PR #1, opened 2017-05-25
GAWB-2077: Initial APIs

MatthewBemis:
# New functionality:

### Endpoints:

GET: `/resourcesTypes` - Returns the list of resource types in the system
POST: `/resources/{resourceType}/{resourceId}` - Creates a new resource of type `resourceType` with the ID `resourceId`
GET: `/resources/{resourceType}/{resourceId}/action/{actionName} - Returns true if the user can perform `actionName` on the resource, returns false if not.

### Sync resourceTypes on boot:

Specify all resource types in config to sync them with OpenAM.
Example config:
```
resourceTypes =
  [
    {
      name = "workflow"
      actions = ["listmembers", "altermembers", "delete", "view", "abort"]
      ownerRoleName = "owner"
      roles = [
        {
          roleName = "owner"
          roleActions = ["listmembers", "altermembers", "delete", "view", "abort"]
        },
        {
          roleName = "reader"
          roleActions = ["view"]
        }
      ]
    }
  ]
```
1) Resource types found in config but not in OpenAM yet will be created.
2) Resource types found in OpenAM but not in config will not be deleted, but a warning message will be displayed on boot.
3) Resource types with actions that differ between OpenAM and config will be synchronized.


src/main/resources/resources.conf
```diff
@@ -0,0 +1,14 @@
+resources {
```
dvoet: I don't think this should be in a separate file, put it in reference.conf for now

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  def createResource(resourceType: ResourceType, resourceId: String, userInfo: UserInfo): Future[Set[OpenAmPolicy]] = {
+    Future.traverse(resourceType.roles) { role =>
+      val roleMembers: Set[SamSubject] = role.roleName match {
+        case resourceType.ownerRoleName => Set(userInfo.userId)
+        case _ => Set.empty
+      }
+      for {
+        group <- directoryDAO.createGroup(SamGroup(SamGroupName(s"${resourceType.name}-${resourceId}-${role.roleName}"), roleMembers))
+        adminUserInfo <- getOpenAmAdminUserInfo
```
MatthewBemis, line 54: i put this here as a possibly temporary solution.

do we want to allow `All Authenticated Users` to manage policies and use their token? or do we want to manage the policies as the admin?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/OpenAmDAO.scala
```diff
@@ -0,0 +1,101 @@
...
+    val usernameHeader = headers.RawHeader("X-OpenAM-Username", username)
+    val passwordHeader = headers.RawHeader("X-OpenAM-Password", password)
+
+    httpRequest[String, AuthenticateResponse](authenticationUrl, Option("{}"), HttpMethods.POST, List(usernameHeader, passwordHeader))
+  }
+
+  def listResourceTypes(userInfo: UserInfo): Future[Set[OpenAmResourceType]] = {
+    val resourceTypesUrl = openAmConfig.url + "/json/resourcetypes?_queryFilter=true"
+
+    httpRequest[String, OpenAmResourceTypeList](resourceTypesUrl, None, HttpMethods.GET, List(authorizationHeader(userInfo))).map(_.result)
```
MatthewBemis: really I want to do something like...

`httpRequest[_, OpenAmResourceTypeList]...`, since a GET isn't marshalling any payload

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/OpenAmDAO.scala
```diff
@@ -0,0 +1,101 @@
...
+  def updateResourceType(updatedResourceType: OpenAmResourceType, userInfo: UserInfo): Future[OpenAmResourceType] = {
+    val resourceTypesUrl = openAmConfig.url + s"/json/resourcetypes/${updatedResourceType.uuid}"
+
+    httpRequest[OpenAmResourceType, OpenAmResourceType](resourceTypesUrl, Option(updatedResourceType), HttpMethods.PUT, List(authorizationHeader(userInfo)))
+  }
+
+  def getDefaultPolicySet(userInfo: UserInfo): Future[OpenAmPolicySet] = {
+    val policySetsUrl = openAmConfig.url + s"/json/applications/iPlanetAMWebAgentService"
+
+    httpRequest[String, OpenAmPolicySet](policySetsUrl, None, HttpMethods.GET, List(authorizationHeader(userInfo)))
```
MatthewBemis: same here, want to have `String` be some sort of wildcard type

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: as admin I think

dvoet (approved):

PR #2, opened 2017-06-01
implemented and tested JndiDirectoryDAO

dvoet:


src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  override def loadUser(userId: SamUserId): Future[Option[SamUser]] = withContext { ctx =>
+    Try {
+      val attributes = ctx.getAttributes(userDn(userId))
+
+      val uid = getAttribute[String](attributes, Attr.uid).getOrElse(throw new WorkbenchException(s"${Attr.uid} attribute missing: $userId"))
+      val emailOption = getAttribute[String](attributes, Attr.email)
+      val firstName = getAttribute[String](attributes, Attr.givenName).getOrElse(throw new WorkbenchException(s"${Attr.givenName} attribute missing: $userId"))
+      val lastName = getAttribute[String](attributes, "sn").getOrElse(throw new WorkbenchException(s"sn attribute missing: $userId"))
```
MatthewBemis: `Attr.sn` instead of `"sn"` in the `getAttribute` call and in the exception

MatthewBemis (approved):
exciting! just one minor comment

PR #22, opened 2017-09-25
Sam Phase II: List user's roles on resource [GAWB-2426]

MatthewBemis:
- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Make sure Swagger is updated if API changes
- [x] **Submitter**: If updating admin endpoints, also update [firecloud-admin-cli](https://github.com/broadinstitute/firecloud-admin-cli)
- [x] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
  * To Demo flag is set
  * Add notes on how to QA
- [x] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [x] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [x] Tell the tech lead (TL) that the PR exists if they wants to look at it
- [x] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Product Owner** sign off
- [x] **Assign to submitter** to finalize


dvoet:
a note on tests - Our service level function signatures are a lot more sane using akka http. So I am thinking that we should have functional level unit tests at the service layer, not the api layer. We should still have api tests but they should concern themselves with ensuring the api exists, has correct return codes and has output in the correct structure (i.e.  json for the right case class).

Thoughts?

MatthewBemis:
@dvoet i agree with that, i had gotten into the bad habit of not testing both. will update...

dvoet (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesSpec.scala
```diff
@@ -55,5 +56,39 @@ class ResourceRoutesSpec extends FlatSpec with Matchers with ScalatestRouteTest
...
+  it should "404 on list resource roles when resource type doesnt exist" in {
+    val resourceType = ResourceType(ResourceTypeName("rt"), Set(ResourceAction("run")), Set(ResourceRole(ResourceRoleName("owner"), Set(ResourceAction("run")))), ResourceRoleName("owner"))
+    val samRoutes = createSamRoutes(Map(resourceType.name -> resourceType))
+
+    Get(s"/api/resource/doesntexist/foo/roles") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.NotFound
+    }
+  }
+
+  it should "200 and return an empty list on list resource roles" in {
```
breilly2: This test is identical to the `Get` part of `it should "200 on list resource roles" in`. Should either do something more interesting or one of the two should be removed.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -21,18 +21,28 @@ class ResourceService(val accessPolicyDAO: AccessPolicyDAO, val directoryDAO: Di
...
+    }
+  }
+
+  def listUserResourceRoles(resourceType: ResourceType, resourceName: ResourceName, userInfo: UserInfo): Future[Set[ResourceRoleName]] = {
+    listResourceAccessPoliciesForUser(resourceType, resourceName, userInfo).map { matchingPolicies =>
+      matchingPolicies.flatMap(_.role)
+    }
+  }
+
+  private def listResourceAccessPoliciesForUser(resourceType: ResourceType, resourceName: ResourceName, userInfo: UserInfo): Future[Set[AccessPolicy]] = {
```
breilly2: This only needs `name` from `resourceType` so I propose that it take a `ResourceTypeName` instead of a full `ResourceType`.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -143,4 +143,49 @@ class ResourceServiceSpec extends FlatSpec with Matchers with TestSupport with B
     //cleanup
     runAndWait(service.deleteResource(resourceType, resourceName))
   }
+
+  it should "listUserResourceRoles when they have at least one role" in {
+    val ownerRoleName = ResourceRoleName("owner")
+    val resourceType = ResourceType(ResourceTypeName(UUID.randomUUID().toString), Set(ResourceAction("a1"), ResourceAction("a2"), ResourceAction("a3")), Set(ResourceRole(ownerRoleName, Set(ResourceAction("a1"), ResourceAction("a2")))), ownerRoleName)
```
breilly2: Are the `ResourceAction`s important to this test case? I don't see them being referenced other than when creating the resource type here.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -143,4 +143,49 @@ class ResourceServiceSpec extends FlatSpec with Matchers with TestSupport with B
...
+      dummyUserInfo
+    ))
+
+    roles shouldEqual Set(ResourceRoleName("owner"))
+
+    runAndWait(service.deleteResource(resourceType, resourceName))
+    runAndWait(service.directoryDAO.deleteUser(dummyUserInfo.userId))
+  }
+
+  it should "listUserResourceRoles when the resource doesn't exist" in {
```
breilly2: `it should "return an empty set from listUserResourceRoles when the resource doesn't exist"`

breilly2 (changes requested):
Only minor changes, but I think they're worth making.

MatthewBemis:
@breilly2 back to you

breilly2 (approved):
👍

PR #39, opened 2017-10-21
[GAWB-2648] Leo call sam to create pet service account at cluster creation time

rtitle:
Changes around pet service account creation. When a pet service account is created, 2 things need to happen:

1. The pet service account email needs to be added as an _attribute_ of the user LDAP record
2. The pet service account needs to be added with its own dn in LDAP, under the ou=pets org unit

Previously, only #1 was happening; now we do both #1 and #2.

Testing notes: https://docs.google.com/document/d/1sVU3KEnJuut-_k58K2yqjAC4zlUXFAfN0Sv35cLxN34/edit?ts=59ea31d7

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala
@@ -28,9 +28,9 @@ trait DirectoryDAO {
   def listFlattenedGroupUsers(groupName: WorkbenchGroupName): Future[Set[WorkbenchUserId]]
   def listAncestorGroups(groupName: WorkbenchGroupName): Future[Set[WorkbenchGroupName]]
 
-  def enableUser(userId: WorkbenchUserId): Future[Unit]
-  def disableUser(userId: WorkbenchUserId): Future[Unit]
-  def isEnabled(userId: WorkbenchUserId): Future[Boolean]
+  def enableIdentity(subject: WorkbenchSubject): Future[Unit]
+  def disableIdentity(subject: WorkbenchSubject): Future[Unit]
+  def isEnabled(subject: WorkbenchSubject): Future[Boolean]
 
   def createPetServiceAccount(petServiceAccount: WorkbenchUserServiceAccount): Future[WorkbenchUserServiceAccount]
   def loadPetServiceAccount(petServiceAccountId: WorkbenchUserServiceAccountId): Future[Option[WorkbenchUserServiceAccount]]
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala
@@ -19,18 +19,18 @@ trait DirectoryDAO {
 
   def loadSubjectFromEmail(email: String): Future[Option[WorkbenchSubject]]
 
-  def createUser(user: WorkbenchUser): Future[WorkbenchUser]
-  def loadUser(userId: WorkbenchUserId): Future[Option[WorkbenchUser]]
-  def loadUsers(userIds: Set[WorkbenchUserId]): Future[Seq[WorkbenchUser]]
-  def deleteUser(userId: WorkbenchUserId): Future[Unit]
+  def createUser(user: WorkbenchPerson): Future[WorkbenchPerson]
+  def loadUser(userId: WorkbenchSubject): Future[Option[WorkbenchPerson]]
+  def loadUsers(userIds: Set[WorkbenchSubject]): Future[Seq[WorkbenchPerson]]
+  def deleteUser(userId: WorkbenchSubject): Future[Unit]
 
-  def listUsersGroups(userId: WorkbenchUserId): Future[Set[WorkbenchGroupName]]
-  def listFlattenedGroupUsers(groupName: WorkbenchGroupName): Future[Set[WorkbenchUserId]]
+  def listUsersGroups(userId: WorkbenchSubject): Future[Set[WorkbenchGroupName]]
+  def listFlattenedGroupUsers(groupName: WorkbenchGroupName): Future[Set[WorkbenchSubject]]
   def listAncestorGroups(groupName: WorkbenchGroupName): Future[Set[WorkbenchGroupName]]
 
-  def enableUser(userId: WorkbenchUserId): Future[Unit]
-  def disableUser(userId: WorkbenchUserId): Future[Unit]
-  def isEnabled(userId: WorkbenchUserId): Future[Boolean]
+  def enableUser(userId: WorkbenchSubject): Future[Unit]
+  def disableUser(userId: WorkbenchSubject): Future[Unit]
+  def isEnabled(userId: WorkbenchSubject): Future[Boolean]
   def getPetServiceAccountForUser(userId: WorkbenchUserId): Future[Option[WorkbenchUserServiceAccountEmail]]
   def addPetServiceAccountToUser(userId: WorkbenchUserId, email: WorkbenchUserServiceAccountEmail): Future[WorkbenchUserServiceAccountEmail]
   def removePetServiceAccountFromUser(userId: WorkbenchUserId): Future[Unit]
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala
@@ -19,19 +19,23 @@ trait DirectoryDAO {
 
   def loadSubjectFromEmail(email: String): Future[Option[WorkbenchSubject]]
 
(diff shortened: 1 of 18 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala
```diff
@@ -19,18 +19,18 @@ trait DirectoryDAO {
 
   def loadSubjectFromEmail(email: String): Future[Option[WorkbenchSubject]]
 
-  def createUser(user: WorkbenchUser): Future[WorkbenchUser]
-  def loadUser(userId: WorkbenchUserId): Future[Option[WorkbenchUser]]
-  def loadUsers(userIds: Set[WorkbenchUserId]): Future[Seq[WorkbenchUser]]
-  def deleteUser(userId: WorkbenchUserId): Future[Unit]
+  def createUser(user: WorkbenchPerson): Future[WorkbenchPerson]
```
rtitle: I changed some method signatures to take the more generic `WorkbenchPerson`, which is a supertrait of `WorkbenchUser` and `WorkbenchUserServiceAccount`. This way these methods can be used on either users or pets. See https://github.com/broadinstitute/workbench-libs/pull/33.

dvoet: I am not sure I like the switch from WorkbenchUser to WorkbenchPerson (or Identity). I know I don't like the switch from WorkbenchUserId to WorkbenchSubject because WorkbenchSubject also includes WorkbenchGroupName which makes no sense here.

The only user related functions that make sense in terms of pets are create, load and delete. The calling code knows and cares which one it is doing so I think there should be different apis. I am ok with sharing code within the implementation.

The functions related to enable and list should remain only in terms of Users.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala
```diff
@@ -19,18 +19,18 @@ trait DirectoryDAO {
...
+  def listUsersGroups(userId: WorkbenchSubject): Future[Set[WorkbenchGroupName]]
+  def listFlattenedGroupUsers(groupName: WorkbenchGroupName): Future[Set[WorkbenchSubject]]
   def listAncestorGroups(groupName: WorkbenchGroupName): Future[Set[WorkbenchGroupName]]
 
-  def enableUser(userId: WorkbenchUserId): Future[Unit]
-  def disableUser(userId: WorkbenchUserId): Future[Unit]
-  def isEnabled(userId: WorkbenchUserId): Future[Boolean]
+  def enableUser(userId: WorkbenchSubject): Future[Unit]
+  def disableUser(userId: WorkbenchSubject): Future[Unit]
+  def isEnabled(userId: WorkbenchSubject): Future[Boolean]
```
dvoet: I am thinking the enable functions here actually belong at the service level. Enable and disable should add or remove both the user and pet. These functions should be renamed. Better yet we should just use the other group functions (but can't yet because the enabled users group has an object class of groupOfNames not groupOfUniqueNames).

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala
rtitle: OK, I can try going the route of having different APIs for users and pets (with maybe a shared private implementation). The `WorkbenchPerson` supertrait might not even be needed in that case. I was actually doing that originally but became annoyed at the code duplication in `DirectoryDAO`. But now I'm annoyed at this weird case class hierarchy. :)

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala
rtitle: Cool, good idea. So 
```
def enableUser(userId: WorkbenchUserId): Future[Unit]
```
would enable both the user and its pet (if one exists)?

dvoet: yes, the same for disable... a user should not be disabled yet have an enabled pet, right?

rtitle: A rogue pet... yeah I can't think of a use case for that. Might as well make enable/disable work atomically for users and pets.

rtitle:
@dvoet this is ready for re-review - it is working in a fiab

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -255,9 +285,21 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
     groups.toSet
   }
 
-  override def enableUser(userId: WorkbenchUserId): Future[Unit] = withContext { ctx =>
+  override def enableUser(userId: WorkbenchUserId): Future[Unit] = {
```
dvoet: I think this function is all biz logic now and belongs in UserService. Make enableIdentity public then when setting up the pet SA just enable that identity.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -91,33 +91,50 @@ class UserService(val directoryDAO: DirectoryDAO, val googleDirectoryDAO: Google
...
+      // add Service Account Actor role to the configured emails so they can assume the identity of the pet service account
+      _ <- Future.traverse(petServiceAccountConfig.serviceAccountUsers) { email =>
+        googleIamDAO.addServiceAccountUserRoleForUser(petServiceAccountConfig.googleProject, petServiceAccount.email, email)
+      }
+      // add the pet service account attribute to the user's LDAP record
+      _ <- directoryDAO.addPetServiceAccountToUser(user.id, petServiceAccount.email)
+      // create an additional LDAP record for the pet service account itself (in a different organizational unit than the user)
+      _ <- directoryDAO.createPetServiceAccount(petServiceAccount)
+      // add the pet service account to the 'enabled users' group
+      _ <- directoryDAO.enableUser(user.id)
```
dvoet: is this an avenue through which a disabled user can inadvertently be enabled? But maybe they can't even call this if they are disabled.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -281,8 +323,20 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
   }
 
   override def disableUser(userId: WorkbenchUserId): Future[Unit] = withContext { ctx =>
```
dvoet: disableUser should also move up to UserService and make disableIdentity public

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
dvoet: I am not certain if this can be called for a disabled user but seems risky. I think we need to only enable the pet SA if the user is enabled.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
rtitle: Gotcha - will do

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
rtitle: I _think_ the user will be blocked at the proxy level if disabled.. but I can change this to enable only the pet as you suggested.

rtitle: Think I found another bug -- `enableUser`/`disableUser` is add/removing the user email from the proxy group -- but NOT removing the pet email. So as it stands the pet email will remain in the proxy group after disabling the user. Fixing that as well.

rtitle:
Pushed fixes in latest commit, retesting

rtitle:
OK, I ran a test with latest code and it works. Ready for review @dvoet

dvoet (approved):

dvoet:
I approved this but there are a couple of follow ons I think we need. First, we discovered that you can't start a VM with a SA from another project. So we need pets per project. Second, since we will have a lot more pets I would like to explore putting the pet ldap entries as a subcontext of their user ldap entry.

PR #41, opened 2017-11-01
Policy is a subject

dvoet:
https://broadinstitute.atlassian.net/browse/GAWB-2785

The problem in the bug is that policies are groups but some of the code was not expecting that. But a policy was not a group in the code but contained a group. But that was also a little weird. So I restructured a bit.

review also https://github.com/broadinstitute/workbench-libs/pull/37

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -7,9 +7,9 @@ object Dependencies {
   val scalaLoggingV = "3.5.0"
   val scalaTestV    = "3.0.1"
 
-  val workbenchUtilV   = "0.2-af345cd"
-  val workbenchModelV  = "0.5-af345cd"
-  val workbenchGoogleV = "0.7-af345cd"
+  val workbenchUtilV   = "0.2-e161b68"
+  val workbenchModelV  = "0.6-e161b68"
+  val workbenchGoogleV = "0.7-e161b68"
 
   val excludeAkkaActor =        ExclusionRule(organization = "com.typesafe.akka", name = "akka-actor_2.12")
   val excludeWorkbenchUtil =    ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-util_2.12")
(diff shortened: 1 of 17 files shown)


breilly2 (approved):
👍 Definitely some good improvements to the model. Just a couple of general thoughts that are probably larger in scope than this PR. I'd have to fiddle with the code to make concrete suggestions so I'm providing this mostly as food for thought.

`ValueObject` doesn't seem to me to be adding a lot of... well... value. It just wraps a `String` and provides a `value` property and some JSON handling. The name also conflicts with my prior notion of a "value object" that could encapsulate multiple properties. I wonder if there's a more specific notion than "value" that we're trying to express here. Maybe some renaming would help but I don't have specific suggestions.

Two things about the `WorkbenchGroup` hierarchy give me pause: the need for a concrete `BasicWorkbenchGroup` and the fact that you seem to cringe every time you have to describe an `AccessPolicy` as being a group. Smells like composition might be better but I don't know how much that would disrupt the model.

Lastly, I'll capture here that we talked a little bit as a group about whether or not the model should be merged into a single artifact (either sam itself of a separate model artifact from sam) instead of being split between sam and workbench-libs. Consolidating would allow us to seal the traits and avoid the catch-all case when matching.

dvoet:
I cringe a little when I say a policy *is a* group instead of *contains a* group because this is a case where an implementation detail is driving our model. In order to implement list functionality efficiently and with minimum ick policies are groups in ldap. We tried modeling policies as if they contained a group but that wasn't working so well which is what this PR fixes.

dvoet:
on the BasicWorkbenchGroup front, that is just scala being mean. We can't extend case classes so we need a trait and a concrete class.

breilly2:
Thanks for the follow-up. You've probably said all of that before but the repetition is helpful. Could you add a code comment near `AccessPolicy` with that explanation of why a policy is a group?

PR #43, opened 2017-11-01
Fix severe SA snafu

helgridly:
There was a bunch of confusion around `WorkbenchUserServiceAccountId`, which sometimes meant the unique "subject id" of the SA, and sometimes the left-part of its email. This clears up that confusion.

I have done this by making the "subject id" the bit that goes into LDAP, which I _think_ was the original intention. If it wasn't, and you'd rather it be the left-part of the email, LMK and I'll fix it.

Requires https://github.com/broadinstitute/workbench-libs/pull/38

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -7,9 +7,9 @@ object Dependencies {
   val scalaLoggingV = "3.5.0"
   val scalaTestV    = "3.0.1"
 
-  val workbenchUtilV   = "0.2-72adc94"
-  val workbenchModelV  = "0.5-82fd71f-SNAP"
-  val workbenchGoogleV = "0.7-82fd71f-SNAP"
+  val workbenchUtilV   = "0.2-af345cd"
+  val workbenchModelV  = "0.5-af345cd"
+  val workbenchGoogleV = "0.7-af345cd"
 
   val excludeAkkaActor =        ExclusionRule(organization = "com.typesafe.akka", name = "akka-actor_2.12")
   val excludeWorkbenchUtil =    ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-util_2.12")
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -8,8 +8,8 @@ object Dependencies {
   val scalaTestV    = "3.0.1"
 
   val workbenchUtilV   = "0.2-72adc94"
-  val workbenchModelV  = "0.4-72adc94"
-  val workbenchGoogleV = "0.5-b4c9b5c"
+  val workbenchModelV  = "0.5-0ba4bca-SNAP"
+  val workbenchGoogleV = "0.7-0ba4bca-SNAP"
 
   val excludeAkkaActor =        ExclusionRule(organization = "com.typesafe.akka", name = "akka-actor_2.12")
   val excludeWorkbenchUtil =    ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-util_2.12")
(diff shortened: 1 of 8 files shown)


helgridly:
> I have done this by making the "subject id" the bit that goes into LDAP, which I think was the original intention. If it wasn't, and you'd rather it be the left-part of the email, LMK and I'll fix it.

Given that tests are failing I think I am wrong here. Am I?

helgridly:
Wait, no. Because LDAP is storing a `WorkbenchSubject`, which is a subject id. I'm so confused.

rtitle:
Uh, it's failing here:
```
[info] - should get a pet service account for a user *** FAILED ***
[info]   None was not equal to Some(WorkbenchUserServiceAccount(1234567890,pet-newuser@my-pet-project.iam.gserviceaccount.com,)) (UserServiceSpec.scala:149)
```
but I don't see the bug (yet). It looks correct..

helgridly:
The problem is I'm trying to store a subject ID in Sam LDAP now. This works fine everywhere except for `loadPetServiceAccount`, which takes a subject ID nobody knows about and can't ascertain because it's Google-generated. And we're no longer storing the account name in Sam. If we try to store the account name in Sam then that fails because the account name isn't a `WorkbenchSubject`. I am beginning to think the answer is we store both in LDAP.

rtitle:
Ah yeah, I see. `MockGoogleIamDAO` does this for the subject ID: 
```
val uniqueId = WorkbenchUserServiceAccountUniqueId(Random.nextLong.toString)
```
and `UserService.createUserPetServiceAccount` doesn't return back the subject ID, just the email.

Maybe `UserService.createUserPetServiceAccount` should return back the full `WorkbenchUserPetServiceAccount` so the caller can see _both_ the subject ID and email which were generated. Then the test could call `loadPetServiceAccount` with the subject ID that was returned by `createUserPetServiceAccount`.

rtitle:
...or just make `MockGoogleIamDAO` deterministic as it was before, so you can verify with tests. :)

I don't think this is a problem outside of unit tests..

helgridly:
We need to store both the subject id and email. We look up using the email, but need a WorkbenchSubject for `dnToSubject`.

helgridly:
Either that or I'm severely misunderstanding what Sam wants to store in a dn.

rtitle:
I think you may be misunderstanding a little (or I am).

The subject is always in the dn (the "primary key"). This was true before this change. The other fields are stored as attributes. For example:

```
ou=people:
  subject: 123   <-- this is in the dn
    email: rtitle@broadinstitute.com  <-- attribute
    pet: pet-123@gserviceaccount.com   <-- attribute

ou=pets:
  subject: 456   <-- this is in the dn
    email: pet-123@gserviceAccount.com  <-- attribute
```

This way a "person" (e.g. rtitle@broadinstitute.org) can easily look up its pet email. And incoming requests _as the pet_ can easily switch back to the user (the user subject ID is embedded in the pet email).

The issue with this test is you're generating random pet subject IDs, and have no way to retrieve them afterwards. I think that can be resolved with better mocks.

dvoet:
that last comment looks right and if there is still confusion let's use faces.

I would like to eventually see either of 2 things 1) the pet attribute of a person stores the dn of the pet, not the email or 2) pets live beneath their user in the directory hierarchy, not in a separate ou.

rtitle:
@dvoet I was going to try your #2, I haven't got it to a PR stage yet.

rtitle (approved):
I threw this on a fiab and it works. :shipit: after fixing namings.

PR #49, opened 2017-11-13
Google group sync

dvoet:
https://broadinstitute.atlassian.net/browse/GAWB-2821

3 big parts
1. copy group synch code from rawls
2. copy group sync monitor from rawls
3. move most of google integrations into google package

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
@@ -233,6 +233,7 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit 
       subject match {
         case _: WorkbenchUserId => WorkbenchUserEmail(emailString)
         case _: WorkbenchGroupIdentity => WorkbenchGroupEmail(emailString)
+        case _: WorkbenchUserServiceAccountSubjectId => WorkbenchUserServiceAccountEmail(emailString)
       }
     }
   }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
@@ -1,5 +1,7 @@
 package org.broadinstitute.dsde.workbench.sam.directory
 
+import java.text.SimpleDateFormat
+import java.util.Date
 import javax.naming._
 import javax.naming.directory._
 
@@ -18,7 +20,6 @@ import org.broadinstitute.dsde.workbench.sam.schema.JndiSchemaDAO.Attr
  * Created by dvoet on 11/5/15.
  */
 class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit executionContext: ExecutionContext) extends DirectoryDAO with DirectorySubjectNameSupport with JndiSupport {
-
   override def createGroup(group: BasicWorkbenchGroup): Future[BasicWorkbenchGroup] = withContext { ctx =>
     try {
       val groupContext = new BaseDirContext {
@@ -36,6 +37,8 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit 
             group.members.foreach(subject => members.add(subjectDn(subject)))
             myAttrs.put(members)
           }
+
+          myAttrs.put(new BasicAttribute(Attr.groupUpdatedTimestamp, formattedDate(new Date())))
 
           myAttrs
         }
@@ -56,10 +59,20 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit 
 
   override def removeGroupMember(groupId: WorkbenchGroupIdentity, removeMember: WorkbenchSubject): Future[Unit] = withContext { ctx =>
     ctx.modifyAttributes(groupDn(groupId), DirContext.REMOVE_ATTRIBUTE, new BasicAttributes(Attr.uniqueMember, subjectDn(removeMember)))
+    updateUpdatedDate(groupId, ctx)
   }
 
   override def addGroupMember(groupId: WorkbenchGroupIdentity, addMember: WorkbenchSubject): Future[Unit] = withContext { ctx =>
     ctx.modifyAttributes(groupDn(groupId), DirContext.ADD_ATTRIBUTE, new BasicAttributes(Attr.uniqueMember, subjectDn(addMember)))
+    updateUpdatedDate(groupId, ctx)
+  }
+
+  private def updateUpdatedDate(groupId: WorkbenchGroupIdentity, ctx: InitialDirContext) = {
+    ctx.modifyAttributes(groupDn(groupId), DirContext.REPLACE_ATTRIBUTE, new BasicAttributes(Attr.groupUpdatedTimestamp, formattedDate(new Date())))
+  }
+
+  override def updateSynchronizedDate(groupId: WorkbenchGroupIdentity): Future[Unit] = withContext { ctx =>
+    ctx.modifyAttributes(groupDn(groupId), DirContext.REPLACE_ATTRIBUTE, new BasicAttributes(Attr.groupSynchronizedTimestamp, formattedDate(new Date()), true))
   }
 
   override def loadGroup(groupName: WorkbenchGroupName): Future[Option[BasicWorkbenchGroup]] = withContext { ctx =>
@@ -210,6 +223,17 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit 
       case Seq() => None
       case Seq(subject) => Option(subject)
       case _ => throw new WorkbenchException(s"Database error: email $email refers to too many subjects: $subjects")
+    }
+  }
+
+  override def loadSubjectEmail(subject: WorkbenchSubject): Future[Option[WorkbenchEmail]] = withContext { ctx =>
+    val subDn = subjectDn(subject)
+    Option(ctx.getAttributes(subDn).get(Attr.email)).map { emailAttr =>
+      val emailString = emailAttr.get.asInstanceOf[String]
+      subject match {
+        case _: WorkbenchUserId => WorkbenchUserEmail(emailString)
+        case _: WorkbenchGroupIdentity => WorkbenchGroupEmail(emailString)
+      }
     }
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
@@ -75,6 +75,11 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit 
     ctx.modifyAttributes(groupDn(groupId), DirContext.REPLACE_ATTRIBUTE, new BasicAttributes(Attr.groupSynchronizedTimestamp, formattedDate(new Date()), true))
   }
 
+  override def getSynchronizedDate(groupId: WorkbenchGroupIdentity): Future[Option[Date]] = withContext { ctx =>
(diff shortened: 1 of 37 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  private def parseMessage(message: PubSubMessage) = {
+    (Try {
+      import org.broadinstitute.dsde.workbench.sam.model.SamJsonSupport.ResourceAndPolicyNameFormat
+      message.contents.parseJson.convertTo[ResourceAndPolicyName]
+    } recover {
+      case _: DeserializationException =>
+        import WorkbenchIdentityJsonSupport.WorkbenchGroupNameFormat
+        message.contents.parseJson.convertTo[WorkbenchGroupName]
```
breilly2, line 151: Is this to handle old-style name-only messages? Based on previous discussions, I thought sam was going to only pick up new-style dn messages and we'd let rawls handle any lingering old-style messages.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -213,6 +231,18 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
     }
   }
 
+  override def loadSubjectEmail(subject: WorkbenchSubject): Future[Option[WorkbenchEmail]] = withContext { ctx =>
+    val subDn = subjectDn(subject)
+    Option(ctx.getAttributes(subDn).get(Attr.email)).map { emailAttr =>
+      val emailString = emailAttr.get.asInstanceOf[String]
+      subject match {
```
breilly2: Should we have a function on `WorkbenchSubject` that takes an email and returns an appropriate `WorkbenchEmail`?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
```diff
@@ -74,8 +74,8 @@ class UserServiceSpec extends FlatSpec with Matchers with TestSupport with Befor
     val mockGoogleDirectoryDAO = service.googleDirectoryDAO.asInstanceOf[MockGoogleDirectoryDAO]
     val groupEmail = WorkbenchGroupEmail(service.toProxyFromUser(defaultUserId.value))
     val allUsersGroupEmail = WorkbenchGroupEmail(service.toGoogleGroupName(UserService.allUsersGroupName.value))
-    mockGoogleDirectoryDAO.groups should contain key (groupEmail)
-    mockGoogleDirectoryDAO.groups(groupEmail) shouldBe Set(defaultUserEmail)
+//    mockGoogleDirectoryDAO.groups should contain key (groupEmail)
```
breilly2, line 76: How about using Mockito to verify that `CloudExtenstions.onUserCreate` was invoked? Somewhere in between not checking anything and leaving commented-out code (that you probably meant to remove anyway... but your hesitation makes me think you want to do _something_ here).

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
dvoet: I was originally thinking the messages would look like a ldap dn but changed my mind when it came time to implement because none of our DAO functions are in terms of dns. The message coming in could be either a serialized ResourceAndPolicyName or WorkbenchGroupName. The only way to figure it out is to try one then the other or make a more complicated json structure that indicates which kind it is.

breilly2: Or send to different topics which is what I was envisioning. I think this had some advantages when it came to rolling out sam and letting it handle any new messages while rawls could continue to drain the queue of old messages. Does publishing to the same queue complicate rollout? At the very least, rawls probably needs a change so it doesn't choke on new-style messages.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -0,0 +1,153 @@
...
+
+      results.groupEmail should equal(target.email)
+      results.items should contain theSameElementsAs (
+        added.map(SyncReportItem("added", _, None)) ++
+          removed.map(SyncReportItem("removed", _, None)) ++
+          Seq(
+            SyncReportItem("added", ge.toProxyFromUser(addError.value), Option(ErrorReport(addException))),
+            SyncReportItem("removed", removeError, Option(ErrorReport(removeException)))))
+
+      added.foreach { email => verify(mockGoogleDirectoryDAO).addMemberToGroup(target.email, WorkbenchUserEmail(email)) }
```
breilly2, line 94: No need to verify behavior that is already stubbed. If the code cares about the value being returned from the stub (defaults to `null` in this case) then the test will fail if the stubbing is wrong. If you weren't stubbing, you would need the `verify` but you (almost?) always don't need both.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -0,0 +1,153 @@
...
+      val defaultUser = WorkbenchUser(defaultUserId, defaultUserEmail)
+
+      // create a user
+      val newUser = service.createUser(defaultUser).futureValue
+      newUser shouldBe UserStatus(UserStatusDetails(defaultUserId, defaultUserEmail), Map("ldap" -> true, "allUsersGroup" -> true, "google" -> true))
+
+      // create a pet service account
+      val emailResponse = googleExtensions.createUserPetServiceAccount(defaultUser).futureValue
+
+      emailResponse.value should endWith("@my-pet-project.iam.gserviceaccount.com")
```
breilly2: I wish the `my-pet-project` part of the setup wasn't so far from the test (tucked away in `reference.conf`). Is there a way to make the test more self-contained, maybe by manually constructing `petServiceAccountConfig` instead of having it read from a config file?

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -0,0 +1,153 @@
...
+      val groupEmail = WorkbenchGroupEmail(googleExtensions.toProxyFromUser(defaultUserId.value))
+      mockGoogleIamDAO.serviceAccounts should contain key (emailResponse)
+      mockGoogleDirectoryDAO.groups should contain key (groupEmail)
+      mockGoogleDirectoryDAO.groups(groupEmail) shouldBe Set(defaultUserEmail, emailResponse)
+
+      // create one again, it should work
+      val petSaResponse2 = googleExtensions.createUserPetServiceAccount(defaultUser).futureValue
+      petSaResponse2 shouldBe emailResponse
+    } finally {
+      runAndWait(schemaDao.clearDatabase())
```
breilly2: If it's worth clearing the database at the end, you might want to also clear it at the beginning just in case something went wrong and with the cleanup during a previous test run.

breilly2 (commented):
Mostly minor things. I do want to consider the same vs. separate topics for the rollout/transition from rawls to sam before approving.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
dvoet: we kept the different queue idea (it is all config based). Rawls submits to and consumes one and sam another.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
dvoet: I think it is required here. I understand that if it is subbed _and_ called probably no need to verify but another passing case would be that it is stubbed _and not_ called. The test is to make sure that we call add on the right members and not on the wrong members. So in this case I don't need to verify unstubbed functions are not called.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
dvoet: I could, it is just a PITA... I think I will change the verification to pull the project from config

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
dvoet: would it be better to just have a standard one way or the other. So far we just clean up at the end. Maybe it is safer to start clean.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
breilly2: Yes, you're right. The test fails without the stubbing but only if it's called. The practice of doing only one of stubbing or verifying assumes that the stubbed return value is vital to some other assertion which isn't the case here.

Since it is pure side effect _and_ the return value doesn't depend on the arguments _and_ you're verifying that the right calls are made, you can slightly simplify the stubbing with blanket declarations:
```
when(mockGoogleDirectoryDAO.addMemberToGroup(any[WorkbenchGroupEmail], any[WorkbenchEmail])).thenReturn(Future.successful(()))
when(mockGoogleDirectoryDAO.removeMemberFromGroup(any[WorkbenchGroupEmail], any[WorkbenchEmail])).thenReturn(Future.successful(()))
```
I like this but not enough to argue passionately about it so it's up to you which style you prefer.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
breilly2: Okay. I was hoping to find a reason to avoid this try-one-then-the-other code but I guess it's just how it has to be.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
breilly2: Could do only pre-cleanup... it depends on what state you want to leave the system. My recent experience is with things like the selenium test for registration where I do want to clean up after the test to leave Google in the expected rest state but also clean up before the test to avoid test failures due to previous failed clean up (which might then only be able to be resolved via manual intervention).

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
breilly2: Ah, that works too. Good idea!

breilly2 (approved):
👍

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
dvoet: thanks, I was trying to figure out how to do that

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
dvoet: there is not an external system to worry about here

dvoet:
we definitely need different pubsub topics

PR #58, opened 2017-11-16
GAWB-2825 make google nested group calls

UrsaStutsman:


src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -97,6 +101,37 @@ class GoogleExtensionSpec extends FlatSpec with Matchers with TestSupport with M
     }
   }
 
+  "GoogleGroup Sync" should "break out of cycle" in {
```
dvoet: "GoogleGroup Sync" should be it

dvoet (approved):

PR #59, opened 2017-11-20
GAWB-2824 Add endpoint to get last group sync date

breilly2:
- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Make sure Swagger is updated if API changes
- [x] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -130,6 +130,10 @@ paths:
           description: Success
           schema:
             $ref: '#/definitions/SyncStatus'
+        204:
+          description: Group is not synchronized to Google
+        404:
+          description: Group not found
       parameters:
         - in: path
           description: Type of resource
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -123,6 +123,33 @@ paths:
       tags:
         - Google
 
+    get:
+      summary: Gets the synchronization state (last synchronized date) for the group.
+      responses:
+        200:
+          description: Success
+          schema:
+            $ref: '#/definitions/SyncStatus'
+      parameters:
+        - in: path
+          description: Type of resource
+          name: resourceTypeName
+          required: true
+          type: string
+        - in: path
+          description: Id of resource
+          name: resourceId
+          required: true
+          type: string
+        - in: path
+          description: Name of policy to synchronize
+          name: policyName
+          required: true
+          type: string
+      operationId: syncStatus
+      tags:
+        - Google
+
   /api/resource/{resourceTypeName}:
     get:
       summary: List resources and policies for this resource for the caller
@@ -669,6 +696,14 @@ definitions:
         type: string
       errorReport:
         $ref: '#/definitions/ErrorReport'
+
+  SyncStatus:
+    description: status of group synchronization
+    required:
+      - lastSyncDate
+    properties:
+      lastSyncDate:
+        type: string
 
   SystemStatus:
     description: status of each aubsystem Sam depends on
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -1,5 +1,7 @@
 package org.broadinstitute.dsde.workbench.sam.google
 
+import java.util.Date
+
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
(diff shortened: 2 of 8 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -165,6 +167,10 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
     } yield ()
   }
 
+  def getSynchronizedDate(groupId: WorkbenchGroupIdentity): Future[Option[Date]] = {
+    directoryDAO.getSynchronizedDate(groupId)
```
breilly2, line 194: Choosing to pass through to DirectoryDAO from here rather than add DirectoryDAO as a dependency to GoogleExtensionRoutes.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
```diff
@@ -103,4 +102,62 @@ class GoogleExtensionRoutesSpec extends FlatSpec with Matchers with ScalatestRou
...
+      val googleExtensions = googleExt
+    }
+
+    //create user
+    Post("/register/user") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.Created
+    }
+
+    Get(s"/api/google/resource/${resourceType.name}/foo/${resourceType.ownerRoleName.value}/sync") ~> samRoutes.route ~> check {
+/* hasn't been created yet so should return not found but doesn't
```
breilly2: Would like to fix this. Also this could be a separate test.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -149,4 +151,18 @@ class GoogleExtensionSpec extends FlatSpec with Matchers with TestSupport with M
     val petSaResponse2 = googleExtensions.createUserPetServiceAccount(defaultUser).futureValue
     petSaResponse2 shouldBe emailResponse
   }
+
+  it should "get a group's last synchronized date" in {
+    val groupName = WorkbenchGroupName("group1")
+
+    val mockDirectoryDAO = mock[DirectoryDAO]
```
breilly2, line 193: It appears we don't currently have a test that actually performs group sync against real Google/LDAP (unless maybe it's in Ursa's PR?). That test would make this more superficial test more meaningful and also provide a skeleton for writing a more meaningful test for this endpoint.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
```diff
@@ -103,4 +102,62 @@ class GoogleExtensionRoutesSpec extends FlatSpec with Matchers with ScalatestRou
...
+      }
+*/
+    }
+
+    Post(s"/api/resource/${resourceType.name}/foo") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.NoContent
+    }
+
+    Get(s"/api/google/resource/${resourceType.name}/foo/${resourceType.ownerRoleName.value}/sync") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.OK
```
dvoet: this is the case where the group exists but has never been sunk, right? I think we should return either No Content or Not Found but not OK with an empty string.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
dvoet: I think it is good to have in this test as it establishes the baseline

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
dvoet: we don't have anything against a real google, that should be in the up coming api integration tests

dvoet (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
breilly2: I don't think it should return 404 Not Found. I was trying to preserve the format of the payload for any case where the group exists (non-404). I can't make a strong argument for this over 204 No Content so I'll go with that.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
breilly2: We can't actually do this check in this test the way it's structured. The sync status is known by DirectoryDAO while existence of the resource is known by some other DAO. Since the mocks aren't linked (and shouldn't be or else they start to be a real alternate implementation) MockDirectoryDAO can't reply one way for a non-existent group and another way for a created-but-not-synced group.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -123,6 +123,33 @@ paths:
       tags:
         - Google
 
+    get:
+      summary: Gets the synchronization state (last synchronized date) for the group.
+      responses:
```
dvoet, line 128: document no content response

dvoet (approved):
minor doc change then 👍

PR #60, opened 2017-11-20
Adding endpoints for add/remove user from policy

UrsaStutsman:


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -127,6 +127,17 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
     }
   }
 
+  def addOrRemoveUserFromPolicy(resourceType: ResourceType, policyName: AccessPolicyName, resource: Resource, email: String, userInfo: UserInfo, add: Boolean = true): Future[AccessPolicy] = {
```
dvoet: Loading in the entire policy and overwriting it can lead to problem if multiple requests come in at the same time. They will clobber each other. I understand that it is convenient to do so because we already have the overwrite code. But it is also easy to just add or remove a member from a policy, in fact the [add/remove]GroupMember functions in JndiDirectoryDAO can be used directly.

Be sure to handle errors when a user being added already exists or a user being removed does not exist. I think both these cases should be treated as OK.

Also in each case call the cloudExtension.onGroupUpdate function.

And a note on the existing implementation: try to avoid the use of flags (the add parameter). 2 separate functions are easier to follow and you can do code reuse within. I would be less critical if it were a private function but only slightly less ;)

Last, write some unit tests.

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -127,6 +127,22 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
     }
   }
 
+  def addUserToPolicy(resourceType: ResourceType, policyName: AccessPolicyName, resource: Resource, email: String, userInfo: UserInfo) = {
```
dvoet: this function (and remove) should take a ResourceAndPolicyName as a parameter.

Also you should be explicit about return types of public functions. And when you do you will find that this is not returning what you think it is.

dvoet (approved):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -272,6 +272,83 @@ paths:
...
+        -Resources
+
+  /api/resource/{resourceTypeName}/{resourceId}/policies/{policyName}/memberEmails/{email}:
+      delete:
+        summary: Remove a user to a policy
+        responses:
+          204:
+            description: Successfully added a user to the policy
+          403:
+            description: You do not have permission to add users to this policy
```
MatthewBemis: "You do not have permission to remove users from this policy"

src/main/resources/swagger/api-docs.yaml
```diff
@@ -272,6 +272,83 @@ paths:
...
+            name: resourceId
+            required: true
+            type: string
+          - in: path
+            description: Name of the policy
+            name: policyName
+            required: true
+            type: string
+          - in: path
+            description: Email of user to be added
```
MatthewBemis: "Email of user to be removed"

src/main/resources/swagger/api-docs.yaml
```diff
@@ -272,6 +272,83 @@ paths:
...
+      operationId: addUserToPolicy
+      tags:
+        -Resources
+
+  /api/resource/{resourceTypeName}/{resourceId}/policies/{policyName}/memberEmails/{email}:
+      delete:
+        summary: Remove a user to a policy
+        responses:
+          204:
+            description: Successfully added a user to the policy
```
MatthewBemis: "Successfully removed a user from the policy"

MatthewBemis (approved):
:+1: pending some small changes to swagger wording leftover from copy/paste

PR #66, opened 2017-12-05
Sam swagger java client api

ansingh7115:
Using Swagger Codegen to generate a Java Client API for Sam that runs on Travis - 

.travis.yml 
- added the artifactory username and password with values encrypted by Travis
- runs the publish script

swagger-client-build.txt
- Template for the build.sbt that we'll be putting in the generated repo
- $organization, $name, $version, $scalaVersions are replaced in the build.sbt template

publish.sh
- uses swagger-codegen to generate new repo, replace build.sbt with templated one, push to artifactory

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


scripts/publish.sh
```diff
@@ -0,0 +1,61 @@
...
+    java -jar swagger-codegen-cli.jar generate -l java --input-spec ./src/main/resources/swagger/api-docs.yaml --output generated
+
+    echo "adding plugins file..."
+    cd generated
+    mkdir project
+    echo "addSbtPlugin(\"com.eed3si9n\" % \"sbt-assembly\" % \"0.14.3\")" > project/plugins.sbt
+    sed -i 's/publishArtifact in (Compile, packageDoc) := false/publishTo := Some(Resolver.file(\"file\", new File(\"publish\")))/g' build.sbt
+
+    echo "compiling Java Client API..."
+    sbt assembly
```
rtitle: This could be 1 sbt command:
```
sbt compile assembly publish
```
Also since compile is a dependency of assembly shouldn't it come first?

ansingh7115: oh - I didn't realize you could just put in one line, thanks
And that's probably true - I can change the order

swagger-api-client/publish.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+set -e
+
+# sbt publish publishes libs to Artifactory for the scala version sbt is running as.
+# sbt +publish publishes libs to Artifactory for all scala versions listed in crossScalaVersions.
+SNAPSHOT=$(( "$TRAVIS_PULL_REQUEST" == "false" && "$TRAVIS_BRANCH" == "develop" ? "SNAP" : "" ))
```
ansingh7115, line 7: How often should this push to artifactory? (Right now it just pushes on every commit for testing purposes.)

swagger-api-client/publish.sh
```diff
@@ -0,0 +1,59 @@
...
+
+echo "getting swagger codegen jar..."
+wget http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/2.2.3/swagger-codegen-cli-2.2.3.jar -O swagger-codegen-cli.jar
+echo "generating Sam Java Client API..."
+java -jar swagger-codegen-cli.jar generate -l java --input-spec ./src/main/resources/swagger/api-docs.yaml --output generated
+
+COMMIT_HASH="${TRAVIS_COMMIT:0:7}"
+VERSION_HASH="1.0-${COMMIT_HASH}${SNAPSHOT}"
```
ansingh7115, line 15: So I want to grab the version number from `project/Version.scala` which sets `val baseModelVersion = "0.1"`
I was told that sbt had a way to do this - I tried googling stuff but can't find examples probably because I'm not googling the right words. What are the right words?

swagger-api-client/publish.sh
helgridly: that's fine IMO. it's what we do for workbench-libs

swagger-api-client/publish.sh
helgridly: You might want inspect:
http://www.scala-sbt.org/1.x/docs/Howto-Inspect-the-Build.html

swagger-api-client/publish.sh
```diff
@@ -0,0 +1,59 @@
...
+ORGANIZATION="org.broadinstitute.dsde"
+APP_NAME="sam"
+PATH_PREFIX="org/broadinstitute/dsde/$APP_NAME"
+SCALA_11_VERSION="2.11.8"
+SCALA_11_VERSION_SHORT="${SCALA_11_VERSION:0:4}"
+SCALA_12_VERSION="2.12.3"
+SCALA_12_VERSION_SHORT="${SCALA_12_VERSION:0:4}"
+SCALA_VERSIONS="List(\"$SCALA_11_VERSION\", \"$SCALA_12_VERSION\")"
```
helgridly, line 23: Blerg. Can you decide what to publish by finding the poms / jars etc and deconstructing the path instead? Defining your list of scala version here and poking it in `crossScalaVersions` feels a bit gross.

swagger-api-client/swagger-client-build.txt
```diff
@@ -0,0 +1,21 @@
+lazy val root = (project in file(".")).
+  settings(
+    organization := "$organization",
+    name := "$name",
```
helgridly, line 4: I think you can get away without templating anything in this file. `organization` and `name` can just be set to broad and sam-client respectively. Things like version can be yanked from `sys.env` like we do for artifactory and snapshot-ness in workbench-libs. Etc.

helgridly (changes requested):
Congrats on getting this done! I think you can do it without templating the build.sbt, and doing so will be much cleaner.

swagger-api-client/publish.sh
ansingh7115: but then...won't it just make a million of them - one for every commit anyone makes on any branch of this repo?

helgridly: I think wb-libs just does it when there's an open PR, but I'm not sure. In either case: who cares :)

rtitle:
Instead of calling swagger-codegen from a bash script, I feel like there should be a way to do this all from the SBT build. I found this SBT plugin: https://github.com/unicredit/sbt-swagger-codegen and example project: https://github.com/unicredit/sbt-swagger-codegen-examples

We could try:
- making a new Sam sub-project containing the Swagger yaml
- adding this plugin to Sam plugins.sbt and enabling it in that sub-project
- making the travis build do something like: `sbt clean swaggerClientCodeGen compile test publish`

Would that work? It might be a lot cleaner than calling `swagger-codegen-cli.jar` and manipulating the output from bash..

helgridly:
I agree this could be sbt-ier. There's also expediency to think of: this is a dependency for Sam auth in Leo, which we need to get done sooner rather than later.

rtitle:
True.. and also I just saw in that plugin documentation: "generates client code using play-ws". Since we don't use Play that's probably not what we want. I can't find an existing SBT plugin that generates Java client code (I found plugins for maven and gradle).

Still, I bet this could be written as an SBT task instead of a bash script, and integrate with the Sam sbt build (and not have to template/manage a sub-sbt project). But maybe that's Future Work. I can poke around a little more..

swagger-api-client/publish.sh
```diff
@@ -0,0 +1,59 @@
...
+set -e
+
+# sbt publish publishes libs to Artifactory for the scala version sbt is running as.
+# sbt +publish publishes libs to Artifactory for all scala versions listed in crossScalaVersions.
+SNAPSHOT=$(( "$TRAVIS_PULL_REQUEST" == "false" && "$TRAVIS_BRANCH" == "develop" ? "SNAP" : "" ))
+
+echo "getting swagger codegen jar..."
+wget http://central.maven.org/maven2/io/swagger/swagger-codegen-cli/2.2.3/swagger-codegen-cli-2.2.3.jar -O swagger-codegen-cli.jar
+echo "generating Sam Java Client API..."
+java -jar swagger-codegen-cli.jar generate -l java --input-spec ./src/main/resources/swagger/api-docs.yaml --output generated
```
rtitle, line 12: Any reason we don't generate scala?

PR #71, opened 2017-12-12
Canshare

dvoet:
https://broadinstitute.atlassian.net/browse/GAWB-2922

enable sharing of policies if a user has the action can_share_{policy name}, the format for specifying the acceptable pattern for action names in config is with a regex

tests:
- add/delete user to/from policy is successful when user has can share for that policy
- add/delete user to/from policy gives when user does not have can share for that policy (existing test slightly tweaked)
- actions may be added to a policy if they match configured regex but not otherwise

.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,8 @@ Group - Create, delete, read, list, add/remove users and groups. Nested groups m
 * read_policies - may read all policies of a resource 
 * alter_policies - may change any policy of a resource
 * delete - may delete a resource
-* read_policy::{policy name} - may add/remove members to/from specified policy of a resource
-* share_policy::{policy name} - may read specified policy of a resource
+* share_policy::{policy name} - may add/remove members to/from specified policy of a resource
+* read_policy::{policy name} - may read specified policy of a resource
 
 ### UML Model
 ![Sam Model](model.png)
--- a/README.md
+++ b/README.md
@@ -51,7 +51,8 @@ Group - Create, delete, read, list, add/remove users and groups. Nested groups m
 * read_policies - may read all policies of a resource 
 * alter_policies - may change any policy of a resource
 * delete - may delete a resource
-* can_share_{policy name} - may add/remove members to/from specified policy of a resource
+* read_policy::{policy name} - may add/remove members to/from specified policy of a resource
+* share_policy::{policy name} - may read specified policy of a resource
 
 ### UML Model
 ![Sam Model](model.png)
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Group - Create, delete, read, list, add/remove users and groups. Nested groups m
 * read_policies - may read all policies of a resource 
 * alter_policies - may change any policy of a resource
 * delete - may delete a resource
-* can_share::{policy name} - may add/remove members to/from specified policy of a resource
+* can_share_{policy name} - may add/remove members to/from specified policy of a resource
 
 ### UML Model
 ![Sam Model](model.png)
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -20,7 +20,7 @@ resourceTypes =
     },
     {
       name = "billing-project"
-      actions = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "launch_notebook", "can_share_.+"]
+      actions = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "launch_notebook", "share_policy::.+", "read_policy::.+"]
       ownerRoleName = "owner"
       roles = [
         {
@@ -29,7 +29,7 @@ resourceTypes =
         },
         {
           roleName = "workspace-creator"
-          roleActions = ["create_workspace", "can_share_can-compute-user"]
+          roleActions = ["create_workspace", "share_policy::can-compute-user", "read_policy::can-compute-user"]
         },
         {
           roleName = "batch-compute-user"
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -5,7 +5,7 @@ resourceTypes =
   [
     {
       name = "workflow-collection"
-      actions = ["delete", "view", "abort", "alter_policies", "read_policies"]
+      actionPatterns = ["delete", "view", "abort", "alter_policies", "read_policies"]
       ownerRoleName = "owner"
       roles = [
         {
@@ -20,7 +20,7 @@ resourceTypes =
     },
     {
       name = "billing-project"
-      actions = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "launch_notebook", "share_policy::.+", "read_policy::.+"]
+      actionPatterns = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "launch_notebook", "share_policy::.+", "read_policy::.+"]
       ownerRoleName = "owner"
       roles = [
         {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -20,7 +20,7 @@ resourceTypes =
     },
     {
       name = "billing-project"
-      actions = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "launch_notebook"]
+      actions = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "launch_notebook", "can_share_.+"]
       ownerRoleName = "owner"
       roles = [
         {
@@ -28,12 +28,8 @@ resourceTypes =
           roleActions = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "launch_notebook"]
         },
         {
-          roleName = "user"
-          roleActions = ["create_workspace"]
-        },
-        {
           roleName = "workspace-creator"
-          roleActions = ["create_workspace"]
+          roleActions = ["create_workspace", "can_share_can-compute-user"]
         },
         {
           roleName = "batch-compute-user"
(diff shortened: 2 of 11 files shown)


src/main/resources/reference.conf
```diff
@@ -20,20 +20,16 @@ resourceTypes =
...
           roleActions = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "launch_notebook"]
         },
-        {
-          roleName = "user"
-          roleActions = ["create_workspace"]
-        },
         {
           roleName = "workspace-creator"
-          roleActions = ["create_workspace"]
+          roleActions = ["create_workspace", "can_share_can-compute-user"]
```
dvoet: I like this because it is an easy way to give all users with workspace-creator role the can share action, no migration required. I don't like it because can-compute-user is a policy name that is defined in rawls.

gdlex4015 (approved):

src/main/resources/reference.conf
breilly2: Can `can-compute-user` be thought of as being defined here and referenced in rawls?

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -46,12 +46,17 @@ object SamResourceActions {
   val readPolicies = ResourceAction("read_policies")
   val alterPolicies = ResourceAction("alter_policies")
   val delete = ResourceAction("delete")
+
+  def canShare(policy: AccessPolicyName) = ResourceAction(s"can_share_${policy.value}")
```
breilly2: Talked about this in person but as a reminder, we should rename this to sound like an action instead of a permission. Maybe `sharePolicy`.

Also, I preferred using a different delimiter instead of `_` both as a visual indication of the pieces being separate (likely a dynamic compound name) and to allow for easier splitting in the future if we need to.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -121,6 +118,14 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
     }
   }
 
+  private def validateActionsAndRoles(resourceType: ResourceType, policyMembership: AccessPolicyMembership) = {
+    val invalidAction = policyMembership.actions.find(a => resourceType.actions.find(_.matches(a)).isEmpty)
```
breilly2: IntelliJ suggests replacing the `find` with:
```
!resourceType.actions.exists(_.matches(a))
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
```diff
@@ -11,12 +11,14 @@ import org.broadinstitute.dsde.workbench.sam.service.ResourceService
...
   val resourceService: ResourceService
 
-  def requireAction(resource: Resource, action: ResourceAction, userInfo: UserInfo): Directive0 = {
+  def requireAction(resource: Resource, action: ResourceAction, userInfo: UserInfo): Directive0 = requireOneOfAction(resource, Set(action), userInfo)
+
+  def requireOneOfAction(resource: Resource, requestedActions: Set[ResourceAction], userInfo: UserInfo): Directive0 = {
     Directives.mapInnerRoute { innerRoute =>
       onSuccess(resourceService.listUserResourceActions(resource, userInfo)) { actions =>
-        if(actions.contains(action)) innerRoute
+        if(!actions.intersect(requestedActions).isEmpty) innerRoute
```
breilly2: IntelliJ suggests:
```
actions.intersect(requestedActions).nonEmpty
```

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -215,6 +215,27 @@ class ResourceServiceSpec extends FlatSpec with Matchers with TestSupport with B
...
+
+    assertResult(actions) {
+      newPolicy.actions
+    }
+
+    val policies = runAndWait(policyDAO.listAccessPolicies(resource))
+
+    assert(policies.contains(newPolicy))
+
+    runAndWait(service.deleteResource(resource, dummyUserInfo))
```
breilly2: This looks like clean-up code that should be in a `finally` block or similar to make sure it runs even if the test fails.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -235,6 +256,18 @@ class ResourceServiceSpec extends FlatSpec with Matchers with TestSupport with B
...
 
+  it should "fail when given an invalid regex action" in {
+    val rt = ResourceType(ResourceTypeName(UUID.randomUUID().toString), Set(ResourceAction("foo-.+-bar")), Set(ResourceRole(ResourceRoleName("owner"), Set(ResourceAction("foo-biz-bar")))), ResourceRoleName("owner"))
+    val resource = Resource(rt.name, ResourceId("my-resource"))
+
+    runAndWait(service.createResourceType(rt))
+    runAndWait(service.createResource(rt, resource.resourceId, dummyUserInfo))
+
+    intercept[WorkbenchExceptionWithErrorReport] {
+      runAndWait(service.overwritePolicy(rt, AccessPolicyName("foo"), resource, AccessPolicyMembership(Set.empty, Set(ResourceAction("foo--bar")), Set.empty)))
```
breilly2, line 268: Good to have this test. I was wondering about cases like this.

src/main/resources/reference.conf
```diff
@@ -20,20 +20,16 @@ resourceTypes =
     },
     {
       name = "billing-project"
-      actions = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "launch_notebook"]
+      actions = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "launch_notebook", "can_share_.+"]
```
breilly2: Now that I see it, I'm not so fond of mixing regular expressions in here. There are now effectively two kinds of `ResourceAction`: those that could be regular expressions (on resource types) and those that aren't (on policies/roles). I think that would be worth formalizing in types. I've spent way too long trying a proof-of-concept of this before suggesting it, but I think it would work out to have something like:
```
sealed trait ResourceAction
case class ResourceActionRequest(value: String) extends ResourceAction with ValueObject
case class ResourceActionDef(value: String) extends ResourceAction with ValueObject {
  private lazy val regex = value.r
  def matches(other: ResourceActionRequest) = regex.pattern.matcher(other.value).matches()
}
```
In my experiment, the only thing I couldn't decide on were what types the actions on roles and policies should be, but I suspect following the types from the match call outward would either work or result in a case where something should be both types.

Anyway, what do you think about trying something like this?

breilly2 (commented):
Mostly minor stuff, except for one change that I feel would be the right thing to do... but I'm willing to change my mind if you have a good argument against it.

src/main/resources/reference.conf
dvoet: I think they can be 2 different things but not extending a base trait. One is a definition of a pattern, the other is the actual action.

dvoet:
@breilly2 I am done with PR changes

README.md
```diff
@@ -51,7 +51,8 @@ Group - Create, delete, read, list, add/remove users and groups. Nested groups m
 * read_policies - may read all policies of a resource 
 * alter_policies - may change any policy of a resource
 * delete - may delete a resource
-* can_share::{policy name} - may add/remove members to/from specified policy of a resource
+* read_policy::{policy name} - may add/remove members to/from specified policy of a resource
+* share_policy::{policy name} - may read specified policy of a resource
```
breilly2: Swapped descriptions?

MatthewBemis (approved):
lgtm

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesSpec.scala
```diff
@@ -86,16 +80,89 @@ class ResourceRoutesSpec extends FlatSpec with Matchers with ScalatestRouteTest
...
 
     Get(s"/api/resource/doesntexist/foo/roles") ~> samRoutes.route ~> check {
       status shouldEqual StatusCodes.NotFound
     }
   }
 
-  "PUT /api/resource/{resourceType}/{resourceId}/policies" should "201 on a new policy being created for a resource" in {
-    val resourceType = ResourceType(ResourceTypeName("rt"), Set(ResourceAction("alter_policies"), ResourceAction("can_compute")), Set(ResourceRole(ResourceRoleName("owner"), Set(ResourceAction("alter_policies")))), ResourceRoleName("owner"))
+  "GET /api/resource/{resourceType}/{resourceId}/policies/{policyName}" should "200 on existing policy of a resource with read_policies" in {
+    val members = AccessPolicyMembership(Set(defaultUserInfo.userEmail), Set(ResourceAction("can_compute")), Set.empty)
```
breilly2: As far as I can tell, all references to "can_compute" and "canCompute" are not significant to these tests. For clarity, I would rather they not be there. To verify my thinking, I went ahead and made those changes and created a PR if that's helpful: https://github.com/broadinstitute/sam/pull/74.

breilly2 (approved):
Minor fix and some test clarity suggestions, then 👍

PR #81, opened 2018-01-12
change pet sa key end points to return credential json file ala google

dvoet:
turns out the private key data returned in the google api call to create a service account key is actually the base64 encoding of what we want to return.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -135,9 +135,7 @@ paths:
       summary: gets a key for the user's pet service account, get_pet_private_key action on cloud-extension/google required
       responses:
         200:
-          description: 'key for a pet service account'
-          schema:
-            $ref: '#/definitions/ServiceAccountKey'
+          description: 'json format key for a pet service account, see https://cloud.google.com/iam/docs/creating-managing-service-account-keys'
         404:
           description: 'user does not exist or caller does not have any access to cloud-extension/google resource'
           schema:
@@ -192,9 +190,7 @@ paths:
       summary: gets a key for the user's pet service account
       responses:
         200:
-          description: 'key for a pet service account'
-          schema:
-            $ref: '#/definitions/ServiceAccountKey'
+          description: 'json format key for a pet service account, see https://cloud.google.com/iam/docs/creating-managing-service-account-keys'
         500:
           description: Internal Error
           schema:
@@ -226,7 +222,7 @@ paths:
           required: true
           type: string
         - in: path
-          description: key ID for the key to remove
+          description: key ID for the key to remove, private_key_id field of json key from get request
           name: keyId
           required: true
           type: string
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
```diff
@@ -71,16 +74,18 @@ trait GoogleExtensionRoutes extends ExtensionRoutes with UserInfoDirectives with
                 }
               }
             } ~
-            get {
-              complete {
-                googleExtensions.createUserPetServiceAccount(WorkbenchUser(userInfo.userId, userInfo.userEmail), GoogleProject(project)).map { petSA =>
-                  StatusCodes.OK -> petSA.serviceAccount.email
+            pathEnd {
```
dvoet, line 79: this is a different, unrelated bug: the missing pathEnd directive meant that the get below and the one under pathPrefix("key") were both in play at the same time.

MatthewBemis (approved):

breilly2 (approved):
👍  Makes sense

src/main/resources/swagger/api-docs.yaml
```diff
@@ -135,9 +135,7 @@ paths:
       summary: gets a key for the user's pet service account, get_pet_private_key action on cloud-extension/google required
       responses:
         200:
-          description: 'key for a pet service account'
-          schema:
-            $ref: '#/definitions/ServiceAccountKey'
+          description: 'json format key for a pet service account, see https://cloud.google.com/iam/docs/creating-managing-service-account-keys'
```
ansingh7115, line 138: Should this have 

> schema:
>              type: string

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
```diff
@@ -60,7 +61,9 @@ trait GoogleExtensionRoutes extends ExtensionRoutes with UserInfoDirectives with
             pathPrefix("key") {
               get {
                 complete {
-                  googleExtensions.getPetServiceAccountKey(WorkbenchUser(userInfo.userId, userInfo.userEmail), GoogleProject(project)).map(x => StatusCodes.OK -> x)
+                  import spray.json._
+                  // parse json to ensure it is json and tells akka http the right content-type
+                  googleExtensions.getPetServiceAccountKey(WorkbenchUser(userInfo.userId, userInfo.userEmail), GoogleProject(project)).map { x => StatusCodes.OK -> x.parseJson }
```
MatthewBemis, line 68: i just realized this, but the other endpoint for getting pet SA keys needs to be parsed into json as well

MatthewBemis (changes requested):

src/main/resources/swagger/api-docs.yaml
dvoet: I don't think so... it is actual json but I don't want to specify what is in it because it is google specific.

ansingh7115: you're right!

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
ansingh7115: +1

MatthewBemis (approved):

PR #93, opened 2018-02-07
GAWB-2933 Store proxy email in LDAP

breilly2:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


UrsaStutsman (approved):
Looks good to me, pending one test fix (in automation tests)

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -334,10 +336,24 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+      case Some(user: WorkbenchUserId) => getUserProxy(user)
+      case _ => Future.successful(None)
+    }
+  }
+
+  def getUserProxy(userId: WorkbenchUserId): Future[Option[WorkbenchEmail]] = {
+    directoryDAO.readUserAttribute[String](userId, Attr.proxyEmail).map(_.map(WorkbenchEmail))
+  }
+
+  private def withProxyEmail[T](userId: WorkbenchUserId)(f: WorkbenchEmail => Future[T]): Future[Option[Future[T]]] = {
```
breilly2: Turns out I think it's going to be better to treat a missing proxy group as an exceptional case to avoid having to deal with futures and options together. Change to this function is coming soon.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -377,6 +377,20 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
     memberDns.map(dnToSubject).contains(subject)
   } recover { case e: NameNotFoundException => false }
 
+  override def addUserAttribute(userId: WorkbenchUserId, attrId: String, value: Any): Future[Unit] = withContext { ctx =>
```
dvoet: I don't like add and readUserAttribute being part of the public interface of DirectoryDAO because it makes it feel like attrId could be anything when in fact there are a fixed number of values it could be.

This could be fixed in 2 ways:
* make attrId an enum-like type
* keep these functions private and have read/addProxyGroup functions

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -377,6 +377,20 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
...
+    val basicAttributes: Attributes = new BasicAttributes(attrId, value, true)
+    ctx.modifyAttributes(userDn(userId), DirContext.ADD_ATTRIBUTE, basicAttributes)
+  }
+
+  override def readUserAttribute[T](userId: WorkbenchUserId, attrId: String): Future[Option[T]] = withContext { ctx =>
+    Try {
+      val attributes = ctx.getAttributes(userDn(userId))
```
dvoet: the second parameter to getAttributes is an array of attribute names to retrieve, use it.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -14,6 +14,7 @@ import org.broadinstitute.dsde.workbench.sam.config.{GoogleServicesConfig, PetSe
 import org.broadinstitute.dsde.workbench.sam.directory.DirectoryDAO
 import org.broadinstitute.dsde.workbench.sam.model._
 import org.broadinstitute.dsde.workbench.sam.openam.AccessPolicyDAO
+import org.broadinstitute.dsde.workbench.sam.schema.JndiSchemaDAO.Attr
```
dvoet: I don't think GoogleExtensions know about Jndi level things

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -334,10 +336,21 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+  }
+
+  def getUserProxy(userId: WorkbenchUserId): Future[Option[WorkbenchEmail]] = {
+    directoryDAO.readUserAttribute[String](userId, Attr.proxyEmail).map(_.map(WorkbenchEmail))
+  }
+
+  private def withProxyEmail[T](userId: WorkbenchUserId)(f: WorkbenchEmail => Future[T]): Future[T] = {
+    getUserProxy(userId) flatMap {
+      case Some(e) => f(e)
+      case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Proxy group does not exist for subject ID: $userId"))
```
dvoet: I think this case can happen both if the user id does not exist or if the proxy address is not present. It seems like the former should be a 404 but the latter is an internal server error. But it seems like we are doing too much error handling... if we get as far as calling this function and the user does not exist, is that an internal server error?

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
breilly2: On the flip side, the concept of a proxy group has so far been encapsulated inside of GoogleExtensions (other than the orch passthrough endpoint). I also had reservations about the general attribute accessors though so I'm happy to change this. I had wished at one point that AttrId was an enum type and I still think that a may be a good idea but I'm not going to tackle that at this time.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
breilly2: Looking at the callers, we can be pretty sure the user exists by the time we get here. Therefore, if the proxy group doesn't exist, that's an unexpected state so I'm good with changing that to an internal server error if that's what you're suggesting.

dvoet (approved):

PR #106, opened 2018-02-27
GAWB-3049: Lock schema during boot

MatthewBemis:
Also see https://github.com/broadinstitute/firecloud-develop/pull/1010

Note: Let Jackie know when this is being merged into dev. There is a dsp-toolbox PR that must go in with it

How this works:

- Add an entry for a schemaLock to LDAP. If it already exists, some other instance must be updating the schema so we will wait for the configured amount of time. 
- If the lock becomes available while we are still waiting, we proceed and update the schema. If it doesn't, we throw an exception. Currently this does not kill the instance (see open questions)
- After we're done updating the schema, we remove the record of the lock so other instances can do their thing

OPEN QUESTIONS:

- Should this operation block everything else? Right now we also initialize things like Google group sync. Liquibase will prevent anything else from happening until the lock is released. Perhaps this is the way we want this mechanism to behave as well.
- I still feel like there is a slight vulnerability here. Say sam101 and sam102 are deployed but then sam102 needs a redeploy for some reason. sam102 may destroy and recreate the schema while somebody is communicating with sam101. This would cause problems, no? A **solution** I can think of is to insert a new entry for each "version" of the schema and not remove it, much like the DATABASECHANGELOG table that Liquibase uses. If a second instance tries to deploy and sees that OpenDJ already has the current schema, it will bypass the operation altogether. The only downside is that it would require us to manually increment a schemaVersion integer somewhere in config or in code whenever we change the schema.

If the answer to both of these questions is "let's do them both" (and I think doing both is the right thing to do), then there is obviously a little bit more work but not much. Just want to solicit feedback before proceeding further.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -10,7 +10,7 @@ import com.typesafe.scalalogging.LazyLogging
 import net.ceedubs.ficus.Ficus._
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.Pem
 import org.broadinstitute.dsde.workbench.google.{HttpGoogleDirectoryDAO, HttpGoogleIamDAO, HttpGooglePubSubDAO, HttpGoogleStorageDAO}
-import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
+import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchException}
 import org.broadinstitute.dsde.workbench.sam.api.{SamRoutes, StandardUserInfoDirectives}
 import org.broadinstitute.dsde.workbench.sam.config._
 import org.broadinstitute.dsde.workbench.sam.directory._
@@ -32,6 +32,7 @@ object Boot extends App with LazyLogging {
 
     val directoryConfig = config.as[DirectoryConfig]("directory")
     val googleServicesConfigOption = config.getAs[GoogleServicesConfig]("googleServices")
+    val schemaLockConfig = config.as[SchemaLockConfig]("schemaLock")
 
     // we need an ActorSystem to host our application in
     implicit val system = ActorSystem("sam")
@@ -40,7 +41,7 @@ object Boot extends App with LazyLogging {
 
     val accessPolicyDAO = new JndiAccessPolicyDAO(directoryConfig)
     val directoryDAO = new JndiDirectoryDAO(directoryConfig)
-    val schemaDAO = new JndiSchemaDAO(directoryConfig)
+    val schemaDAO = new JndiSchemaDAO(directoryConfig, schemaLockConfig)
 
     val resourceTypes = config.as[Map[String, ResourceType]]("resourceTypes").values.toSet
     val resourceTypeMap = resourceTypes.map(rt => rt.name -> rt).toMap
@@ -74,6 +75,9 @@ object Boot extends App with LazyLogging {
 
     for {
       _ <- schemaDAO.init() recover {
+        case e: WorkbenchException =>
+          logger.error("FATAL - could not update schema to latest version. Is the schema lock stuck? See documentation here for more information: [link]")
+          throw e
         case t: Throwable =>
           logger.error("FATAL - could not init ldap schema", t)
           throw t
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/SchemaLockConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/SchemaLockConfig.scala
@@ -1,0 +1,19 @@
+package org.broadinstitute.dsde.workbench.sam.config
+
+/**
+  * Created by mbemis on 2/27/18.
+  */
+
+/**
+  * Schema lock configuration.
+  * @param lockSchemaOnBoot Set to false to disable schema locking altogether
+  * @param recheckTimeInterval The number of seconds to wait before retrying the lock (should be less than or equal to maxTimeToWait)
+  * @param maxTimeToWait The number of seconds to wait before giving up
+  * @param instanceId The id of the instance that is dealing with the lock
+  * @param schemaVersion The version of the schema, to be incremented each time the schema changes
+  */
+case class SchemaLockConfig(lockSchemaOnBoot: Boolean,
+                            recheckTimeInterval: Int,
+                            maxTimeToWait: Int,
+                            instanceId: String,
+                            schemaVersion: Int)
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
@@ -1,17 +1,20 @@
 package org.broadinstitute.dsde.workbench.sam.schema
 
-import javax.naming.{NameAlreadyBoundException, NameNotFoundException}
 import javax.naming.directory._
+import javax.naming.{NameAlreadyBoundException, NameNotFoundException}
 
-import org.broadinstitute.dsde.workbench.sam.config.DirectoryConfig
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.workbench.model.WorkbenchException
+import org.broadinstitute.dsde.workbench.sam.config.{DirectoryConfig, SchemaLockConfig}
+import org.broadinstitute.dsde.workbench.sam.directory.DirectorySubjectNameSupport
+import org.broadinstitute.dsde.workbench.sam.schema.JndiSchemaDAO._
+import org.broadinstitute.dsde.workbench.sam.schema.SchemaStatus._
 import org.broadinstitute.dsde.workbench.sam.util.{BaseDirContext, JndiSupport}
 
-import scala.concurrent.{ExecutionContext, Future}
-import scala.util.Try
-import JndiSchemaDAO._
-import org.broadinstitute.dsde.workbench.sam.directory.DirectorySubjectNameSupport
-
 import scala.collection.JavaConverters._
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
 
 /**
   * Created by mbemis on 10/3/17.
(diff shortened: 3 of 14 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
```diff
@@ -49,15 +49,98 @@ object JndiSchemaDAO {
...
+    }
+  }
+
+  def waitForSchemaLock(remainingAttempts: Int): Future[Unit] = withContext { ctx =>
+    if(remainingAttempts == 0) {
+      logger.error("Could not acquire schema lock...")
+      throw new Exception(s"Operation timed out. Could not acquire schema lock before timeout...")
+    }
+    else {
+      Thread.sleep(schemaLockConfig.recheckTimeInterval * 1000)
```
MatthewBemis: is there a nice way to do this type of thing without Thread.sleep?

breilly2:
I like the schema version idea. Another consequence of not having it is that, during a release, each instance of sam will recreate the schema when only one needs to do it. I imagine something like:

* If "updated to version n" record exists
  * continue
* Else
  * If writing "updating to version n" record succeeds
    * update schema
    * write "updated to version n" record
    * continue
  * Else
    * poll until "updated to version n" record exists
    * continue

I think that addresses both situations.

dvoet:
I agree that both is the right thing to do. But I think this is a good first step and can go out to prod as is. Like I was saying before, tearing down and recreating the schema all the time does not sound like a great long-term approach. So some kind of lock and version is the right approach.

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/SchemaLockConfig.scala
```diff
@@ -0,0 +1,17 @@
+package org.broadinstitute.dsde.workbench.sam.config
+
+/**
+  * Created by mbemis on 2/27/18.
+  */
+
+/**
+  * Google Key Cache configuration.
```
dvoet: wrong comment

src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
```diff
@@ -49,15 +49,98 @@ object JndiSchemaDAO {
...
+      _ <- createSchemaChangeLockSchema()
+      _ <- createOrgUnit(schemaLockOu)
+      _ <- insertSchemaLock()
+    } yield ()
+  }
+
+  def cleanupSchemaLock(): Future[Unit] = {
+    for {
+      _ <- removeSchemaLock()
+      _ <- removeSchemaChangeLockSchema()
```
dvoet: do we need to remove it? That will have its own concurrency issues right: sam101 has the lock, sam102 is waiting, sam101 finishes and removes the schema then sam102 wakes up but can't insert the lock because the schema is gone.

src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
```diff
@@ -49,15 +49,98 @@ object JndiSchemaDAO {
...
+      case Success(attrs) => {
+        val instanceId = attrs.get("instanceId").get.toString
+        logger.info(s"Schema currently locked by sam instance [$instanceId]...")
+        true
+      }
+      case Failure(_) => false
+    }
+  }
+
+  def waitForSchemaLock(remainingAttempts: Int): Future[Unit] = withContext { ctx =>
```
dvoet: I don't think you need a checkSchemaLock function. All this can be a couple lines of code in insertSchemaLock that ends with a recursive call to insertSchemaLock: How about insertSchemaLock takes the number of times to try and interval as parameters. Then attempt insert, catch exception, log message, sleep for interval, call insertSchemaLock decrementing count.

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
MatthewBemis: my initial thought was to remove the schema in case it changed, but that's a bit premature so i'll get rid of that line

src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
dvoet: I think sleep is appropriate here... we want to block

breilly2:
I'm still worried about the 2nd instance dropping and recreating the schema while the 1st continues on and starts serving requests. Even worse, if one instance goes off-the-rails and we only have to restart that one, it will cause downtime in the other unless it can recognize that it doesn't need to drop/recreate the schema.

Not implementing versions now will probably make it work but by brute force only. I would prefer a fix that better isolates the instances. This is tech debt that we can kick down the road but I'm going to assume we'll need to address it, along with zero-downtime rolling deploys, with SamssS. We should update that spike doc accordingly.

src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
```diff
@@ -49,15 +49,67 @@ object JndiSchemaDAO {
...
+      _ <- insertSchemaLock(schemaLockConfig.maxTimeToWait / schemaLockConfig.recheckTimeInterval)
+    } yield ()
+  }
 
   def init(): Future[Unit] = {
     for {
+      _ <- if(schemaLockConfig.lockSchemaOnBoot) initSchemaLock() else Future.successful(())
       _ <- destroySchema()
       _ <- createSchema()
+      _ <- if(schemaLockConfig.lockSchemaOnBoot) removeSchemaLock() else Future.successful(())
```
breilly2: We need to release the lock if `createSchema()` fails.

src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
```diff
@@ -49,15 +49,67 @@ object JndiSchemaDAO {
...
+        case Failure(t: NameAlreadyBoundException) => {
+          logger.info(s"Schema is already locked. Waiting for schema lock...")
+          Thread.sleep(schemaLockConfig.recheckTimeInterval * 1000)
+          insertSchemaLock(remainingAttempts - 1)
+        }
+      }
+    }
+  }
+
+  def removeSchemaLock(): Future[Unit] = withContext { ctx =>
```
breilly2: We should have some documentation on how to manually remove the lock just in case something fails here. Maybe we should also retry this a few times to make it more unlikely that we'll have to manually intervene.

breilly2 (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
breilly2: Importantly, the doc needs to be easy to find by someone doing a release in case something goes wrong.

dvoet:
fine, let's do versions now

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -72,6 +73,9 @@ object Boot extends App with LazyLogging {
 
     for {
       _ <- schemaDAO.init() recover {
+        case e: WorkbenchException =>
```
MatthewBemis: was thinking of making a broad.io link and having it link to some confluence page with instructions

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -72,6 +73,9 @@ object Boot extends App with LazyLogging {
...
+          logger.error("FATAL - could not update schema to latest version. Is the schema lock stuck? See documentation here for more information: [link]")
```
MatthewBemis, line 79: was thinking of making a broad.io link and having it link to some confluence page with instructions

breilly2: It would be great if we could reduce it to a `docker run` command that could be logged with this error.

src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
```diff
@@ -49,40 +52,190 @@ object JndiSchemaDAO {
...
+    logger.info("Acquiring schema lock...")
+
+    for {
+      _ <- createSchemaChangeLockSchema()
+      _ <- createOrgUnit(schemaLockOu)
+      result <- lockSchema()
+    } yield result
+  }
+
+  private def createAndDestroySchema(): Future[SchemaStatus] = {
```
breilly2: Nit: `destroyAndCreateSchema` feels better to me, or something like `recreateSchema`.

MatthewBemis:
jenkins retest

MatthewBemis:
jenkins retest

dvoet (changes requested):
I don't think this addresses the problem. We are still destroying the schema all the time which using versions should allow us to avoid. Let's meet and figure out a design.

src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
```diff
@@ -49,40 +52,190 @@ object JndiSchemaDAO {
...
+    val myAttrs = new BasicAttributes(true)
+    myAttrs.put("completed", true.toString)
+
+    logger.info(s"Schema successfully updated to version [${schemaLockConfig.schemaVersion}]")
+
+    ctx.modifyAttributes(schemaLockDn(schemaLockConfig.schemaVersion), DirContext.REPLACE_ATTRIBUTE, myAttrs)
+  }
+
+  private def readSchemaStatus(): Future[SchemaStatus] = withContext { ctx =>
+    val attributes = Try { ctx.getAttributes(schemaLockDn(schemaLockConfig.schemaVersion)) }
```
breilly2: A note of caution: Since the locks are schema version specific, all sam instances must be configured for the same schema version. Otherwise they will be locking on different objects and not blocking each other.

The only time this would be a problem is if 2 instances are using 2 different versions that both don't exist which is probably extremely unlikely.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
breilly2: This could also be useful to know during development so we can remove version locks so sam will re-run the schema update without having to artificially increment the version.

src/test/resources/reference.conf
```diff
@@ -43,6 +43,14 @@ googleServices {
   }
 }
 
+schemaLock {
+  lockSchemaOnBoot = false
+  recheckTimeInterval = 5
+  maxTimeToWait = 60
+  instanceId = "sam-local"
```
breilly2: Curious to see what value(s) this gets in production.

breilly2 (approved):
👍 Just one nit but otherwise looks good. The lock logic looks solid to me (with the one caveat that I also mentioned).

src/main/scala/org/broadinstitute/dsde/workbench/sam/schema/JndiSchemaDAO.scala
```diff
@@ -49,40 +52,190 @@ object JndiSchemaDAO {
...
 
   def init(): Future[Unit] = {
+    if(schemaLockConfig.lockSchemaOnBoot) { initWithSchemaLock() }
+    else { createAndDestroySchema().map(_ => ()) }
+  }
+
+  private def initWithSchemaLock(): Future[Unit] = {
+    for {
+      schemaLockStatus <- initSchemaLock()
+      schemaUpdateStatus <- schemaLockStatus match {
```
dvoet: what if this step has a failure? Should we remove the lock record?

MatthewBemis: My initial thought was no. For comparison, Liquibase _does_ remove the lock when something fails, but schema updates are incremental and there are a lot more safety checks in place.

In our case, I think if destroyAndCreateSchema fails, it requires a bit more examination because many things could have gone wrong and there are fewer protections in place. I think forcing some additional human intervention is better but maybe that's just flat-out annoying.

I could be easily swayed to the other side of this argument.

src/test/resources/reference.conf
MatthewBemis: thanks for bringing this to my attention because it caused me to look into it further which necessitated some devops work.

see Jackie's commit in https://github.com/broadinstitute/firecloud-develop/pull/1010 for the relevant changes

dvoet (approved):

dvoet:
@MatthewBemis why isn't this in yet?

dvoet:
@MatthewBemis I rebased this and we can push it after tests pass

dvoet:
jenkins retest - firecloud-develop updated

PR #108, opened 2018-02-28
SAM III.5 Feature Branch - GAWB-2526

gpolumbo-broad:
Notes to Reviewers:

The heart of logic for this issue is in `ManagedGroupService.scala`.  All of the methods for this class could have easily been implemented inside `ResourceService.scala`, however, for fear of bloating that class, @UrsaStutsman and I decided to implement a separate class.  

As a result, this required me to change the parameter list of `SamRoutes` constructor, and therefore the `TestSamRoutes` constructor. 

The concern here is:  **_What happens then next time we add a new "special" resource-type?_**  Today it's ManagedGroups, tomorrow Workspaces, etc.  Do we want to be in the habit of bloating the parameter list of `SamRoutes` every time we need to add a resource-type?  

For now, the `ManagedGroupService` is composed with a `ResourceService` and a Map of Resource Types (this map is passed in only because that field is marked `private` in the `ResourceService` class).  This works fine...for now.  But I think we have other options:

1.  Encapsulate the `ManagedGroupService` (and its construction) inside `ResourceService`. 
1.  Create a `ManagedGroupService` factory method in `ResourceService`
1.  Make the `ManagedGroupService` into a trait and mix it into `ResourceService` (not sure how to make this work given that `ManagedGroupService` necessarily relies on methods in `ResourceService` but it looks like there's some wizardry with `type val` that might get us what we need?)
1.  Use Implicits to "inject" the methods from `ManagedGroupService` into `ResourceService` (Admittedly, I've only done a little bit of reading on this and I may be totally off-base with this suggestion)
1.  Keep it as implemented
1.  Other ideas?

My goals are:
1.  Make the construction of `SamRoutes` as easy as possible by not requiring us to expand the parameter list on the constructor every time we add a new special subject type.
1.  Ensure that we can keep logic for managing special resource types in their own classes
1.  Implement a solution that follows team best practices and coding standards

One final question: 
1. `ResourceService.resourceTypes` is `private`.  I'm not sure why.  If we make it public (which seems harmless) then we don't need to pass it into the `ManagedGroupService` constructor. 

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/main/resources/swagger/api-docs.yaml
```diff
@@ -157,6 +157,66 @@ paths:
       tags:
         - Admin
 
+  /api/group/{groupName}:
+    get:
+      summary: Show details of the group
+      responses:
+        200:
+          description: Group found
```
dvoet: need the schema for a 200 response

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -0,0 +1,64 @@
...
+    complete (
+      managedGroupService.loadManagedGroup(ResourceId(groupId)).map {
+        case Some(response) => StatusCodes.OK -> response
+        case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "group not found"))
+      }
+    )
+  }
+
+  private def handlePostGroup(groupId: String, userInfo: UserInfo) = {
+    complete(managedGroupService.createManagedGroup(ResourceId(groupId), userInfo).map(_ => StatusCodes.NoContent))
```
dvoet: this should return status code Created

src/main/resources/swagger/api-docs.yaml
```diff
@@ -157,6 +157,66 @@ paths:
...
+          type: string
+      operationId: getGroup
+      tags:
+        - Group
+    post:
+      summary: Create a new group
+      responses:
+        204:
+          description: Group created
+        500:
```
dvoet: should also return 409 if the group already exists

dvoet: and 400 if the group name is invalid

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  private def handleDeleteGroup(groupId: String, userInfo: UserInfo) = {
+    val resource = Resource(ManagedGroupService.ManagedGroupTypeName, ResourceId(groupId))
+    requireOneOfAction(resource, Set(SamResourceActions.delete), userInfo) {
```
dvoet: use the short cut `requireAction` instead of `requireOneOfAction`

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+  * Created by gpolumbo on 2/21/2018.
+  */
+class ManagedGroupService(resourceService: ResourceService, val resourceTypes: Map[ResourceTypeName, ResourceType]) extends LazyLogging {
+
+  def managedGroupType: ResourceType = resourceTypes.getOrElse(ManagedGroupService.ManagedGroupTypeName, throw new WorkbenchException(s"resource type ${ManagedGroupService.ManagedGroupTypeName.value} not found"))
+  def memberRole = managedGroupType.roles.find(_.roleName == ManagedGroupService.MemberRoleName).getOrElse(throw new WorkbenchException(s"${ManagedGroupService.MemberRoleName} role does not exist in $managedGroupType"))
+
+  // Important Note, see: org/broadinstitute/dsde/workbench/sam/model/SamModel.scala:90
```
dvoet: I guarantee this comment will be out of date

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -0,0 +1,89 @@
...
+  def deleteManagedGroup(groupId: ResourceId) = {
+    resourceService.directoryDAO.deleteGroup(WorkbenchGroupName(groupId.value))
+    resourceService.deleteResource(Resource(managedGroupType.name, groupId))
+  }
+}
+
+object ManagedGroupService {
+  val MemberRoleName = ResourceRoleName("member")
+  val ManagedGroupTypeName = ResourceTypeName("managed-group")
+  val LocalPartRe = "^[A-z0-9_-]+$".r
```
dvoet: vals should start with lower case letter

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -0,0 +1,89 @@
...
+    val errors = validateLocalPartPattern(localPart) ++ validateEmailLength(constructEmail(localPart))
+    if (errors.nonEmpty)
+      throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Cannot create valid email address from group name" , errors.toSeq))
+  }
+
+  private def validateLocalPartPattern(str: String): Option[ErrorReport] = {
+    ManagedGroupService.LocalPartRe.findFirstMatchIn(str) match {
+      case Some(_) => None
+      case None => Option(ErrorReport(s"You have specified a group name that contains characters that are not permitted in an email address. Group name may only contain alphanumeric characters, underscores, and dashes"))
+    }
```
dvoet: regexes can be used in pattern matches so this also works

```
    str match {
      case ManagedGroupService.LocalPartRe() => None
      case _ => Option(ErrorReport(s"You have specified a group name that contains characters that are not permitted in an email address. Group name may only contain alphanumeric characters, underscores, and dashes")) 
    }
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  private def createAggregateGroup(resource: Resource, componentPolicies: Set[AccessPolicy]): Future[BasicWorkbenchGroup] = {
+    val email = generateManagedGroupEmail(resource.resourceId)
+    val workbenchGroupName = WorkbenchGroupName(resource.resourceId.value)
+    val groupMembers: Set[WorkbenchSubject] = componentPolicies.map(_.id)
+    resourceService.directoryDAO.createGroup(BasicWorkbenchGroup(workbenchGroupName, groupMembers, email))
+  }
+
+  private def generateManagedGroupEmail(resourceId: ResourceId): WorkbenchEmail = {
+    val localPart = resourceId.value
```
dvoet: I prefer the term groupName instead of localPart

gpolumbo-broad:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  //  "POST /api/group/{groupName}" should "create a new managed group, the owner group, and the members group, and the All Group and all policies for those groups" in {
+  "POST /api/group/{groupName}" should "create a new managed group with a 204 response code" in {
+    val samRoutes = TestSamRoutes(resourceTypes)
+    assertGroupDoesNotExist(samRoutes)
+    assertCreateGroup(samRoutes)
+  }
+
+  "GET /api/group/{groupName}" should "return a flattened list of users who are in this group" in {
```
dvoet: test name is wrong

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+      status shouldEqual StatusCodes.NoContent
+      responseAs[String].isEmpty shouldEqual true
+    }
+  }
+
+  def assertGetGroup(samRoutes: TestSamRoutes, groupId: String = groupId) = {
+    Get(s"/api/group/$groupId") ~> samRoutes.route ~> check {
+      val expectedResource = Resource(ManagedGroupService.ManagedGroupTypeName, ResourceId(groupId))
+      status shouldEqual StatusCodes.OK
+      responseAs[String] should include (s"${groupId}@${samRoutes.resourceService.emailDomain}")
```
dvoet: In general, route tests should test return codes and input/output formats but not response values. The service level tests should test that output is actually as expected.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupServiceSpec.scala
```diff
@@ -0,0 +1,123 @@
...
+    }
+    runAndWait(managedGroupService.loadManagedGroup(resourceId)).isDefined shouldEqual false
+  }
+
+  it should "fail when the group name has invalid characters" in {
+    val groupName = "Make It Rain!!! $$$$$"
+    intercept[WorkbenchExceptionWithErrorReport] {
+      makeGroup(groupName)
+    }
+    runAndWait(managedGroupService.loadManagedGroup(resourceId)).isDefined shouldEqual false
```
dvoet: drop .isDefined then make it  `shouldEqual None`

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupServiceSpec.scala
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  "ManagedGroupService get" should "return the Managed Group resource" in {
+    makeGroup()
+    runAndWait(managedGroupService.loadManagedGroup(resourceId)).isDefined shouldEqual true
```
dvoet: test that the actual value returned is what is expected:
`runAndWait(managedGroupService.loadManagedGroup(resourceId)) shouldEqual Some(....)`

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupServiceSpec.scala
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  // NOTE: All since we don't have a way to look up policies directly without going through a Resource, this test
```
dvoet: policies are sub contexts of resources so this is not possible give the current ldap structure

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -0,0 +1,64 @@
+package org.broadinstitute.dsde.workbench.sam.api
+
+// package object for implicit ErrorReportSource
```
breilly2: This goes along with line 5? Guessing IntelliJ shifted things around during an auto-import.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -0,0 +1,64 @@
...
+        handlePostGroup(groupId, userInfo)
+      } ~
+      delete {
+        handleDeleteGroup(groupId, userInfo)
+      }
+    }
+  }
+
+  private def handleGetGroup(groupId: String) = {
+    complete (
```
breilly2: Should we protect this so that only members, or even only owners, can read a group?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
breilly2: Also what is the significance in the context of `createManagedGroup`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
gpolumbo-broad: fixed

src/main/resources/swagger/api-docs.yaml
gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: @dvoet my intention was that these should be treated like constants or `static final` values: https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods

If you still want them changed to lowercase, happy to do so, just say so.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: fixed

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
gpolumbo-broad: fixed

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
gpolumbo-broad: fixed

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupServiceSpec.scala
gpolumbo-broad: fixed

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupServiceSpec.scala
gpolumbo-broad: fixed

src/main/resources/swagger/api-docs.yaml
```diff
@@ -157,6 +157,68 @@ paths:
       tags:
         - Admin
 
+  /api/group/{groupName}:
+    get:
+      summary: Show email address of the group
+      responses:
+        200:
```
dvoet, line 182: still need a schema for 200 response

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -0,0 +1,88 @@
...
+  private def validateGroupNameLength(str: String): Option[ErrorReport] = {
+    if (str.length >= maxLength)
+      Option(ErrorReport(s"Email address '$str' is ${str.length} characters in length.  Email address length must be shorter than $maxLength characters"))
+    else
+      None
+  }
+
+  // Per dvoet, when asking for a group, we will just return the group email
+  def loadManagedGroup(groupId: ResourceId): Future[Option[WorkbenchEmail]] = {
+    resourceService.directoryDAO.loadGroup(WorkbenchGroupName(groupId.value)).map(_.map(_.email))
```
dvoet: I think I would prefer this service class taking directoryDAO as a constructor parameter rather than using resourceService's (which probably should be private)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  def deleteManagedGroup(groupId: ResourceId) = {
+    resourceService.directoryDAO.deleteGroup(WorkbenchGroupName(groupId.value))
+    resourceService.deleteResource(Resource(managedGroupType.name, groupId))
+  }
+}
+
+object ManagedGroupService {
```
breilly2: Not sure where that other thread went, but I'll chime in and vote for initial lowercase as well. `val`s are immutable and are far more common than `var`s in scala so there isn't any need to call special attention to these. My brain is already rewired to read the initial uppercase as a type name rather than a val name.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+import akka.http.scaladsl.model.headers.OAuth2BearerToken
+import akka.http.scaladsl.testkit.ScalatestRouteTest
+import org.broadinstitute.dsde.workbench.model.{UserInfo, WorkbenchEmail, WorkbenchUserId}
+import org.broadinstitute.dsde.workbench.sam.{TestSupport, model}
+import org.broadinstitute.dsde.workbench.sam.model._
+import org.broadinstitute.dsde.workbench.sam.service.ManagedGroupService
+import org.scalatest.{BeforeAndAfter, FlatSpec, GivenWhenThen, Matchers}
+
+/**
+  * Created by dvoet on 6/7/17.
```
breilly2: Who created this?

dvoet:
@gpolumbo-broad regarding your design question - I like the ideas behind dependency injection: a component should be told its dependencies, it shouldn't know where to find them. We do that through our constructors. There are certainly other approaches we can use but this is the simplest. I will say bad words about implicits though.

If it is becoming too unwieldy we have created a "Application" object in the past that contains all of our service classes and just pass that around once.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+  private val memberPolicyName = AccessPolicyName(ManagedGroupService.MemberRoleName.value)
+  private val accessPolicyNames = Set(ownerPolicyName, memberPolicyName)
+  private val policyActions: Set[ResourceAction] = accessPolicyNames.flatMap(policyName => Set(SamResourceActions.sharePolicy(policyName), SamResourceActions.readPolicy(policyName)))
+  private val resourceActions = Set(ResourceAction("delete")) union policyActions
+  private val resourceActionPatterns = resourceActions.map(action => ResourceActionPattern(action.value))
+  private val defaultOwnerRole = ResourceRole(ownerRoleName, resourceActions)
+  private val defaultRoles = Set(defaultOwnerRole, ResourceRole(ManagedGroupService.MemberRoleName, Set.empty))
+  private val managedGroupResourceType = ResourceType(ManagedGroupService.ManagedGroupTypeName, resourceActionPatterns, defaultRoles, ownerRoleName)
+  private val resourceTypes = Map(managedGroupResourceType.name -> managedGroupResourceType)
+  private val groupId = "foo"
```
breilly2: I'm usually in favor of inlining these in the tests. Extracting common values implies that usages are somehow linked, that a change in one place necessitates a change in every other place. Since these are used across (independent) tests, this kind of link/interdependency is the opposite of what we want.

To put it another way, a change to the value here should require review of every test that uses it to make sure that the behavior being tested hasn't changed. While slightly more verbose, repeating the instantiation of the value forces that. (Sure, someone could search/replace but the PR would still explicitly declare that it is being changed for multiple test cases and enumerates exactly which test cases are affected.)

In contrast, `ownerRoleName` here is fine because it is a single unique fact about the component being tested rather than data for a test case. There might be some gray area in the values between `ownerRoleName` and `groupId`... I haven't fully analyzed them all... but we can chat in person if you have questions or counter arguments.

breilly2 (changes requested):
Just a few things. I might have a little more feedback but some of what's in this review are the more important points and I didn't want to further delay those suggestions. Let me know if you have questions.

src/main/resources/swagger/api-docs.yaml
gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: Checked with @dvoet and going to make all the values passed to the `ResourceService` constructor private, and will update the `ManagedGroupService` constructor to take all of the required components as parameters in the constructor.

gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
gpolumbo-broad: @breilly2 excellent question.  @dvoet and I talked about this a bit and my understanding was that the decision for now was to leave it as is since we aren't really revealing any information and somebody could "guess" if a group exists simply by trying to POST groups with names they're looking for.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: fixed

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
gpolumbo-broad: fixed

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
gpolumbo-broad: @breilly2 I totally understand where you're coming from and I think it's definitely a tradeoff.  You'll see in some of the tests, I am creating some new instances of the objects I need, whereas in most of the tests, the defaults are just fine.  

That said, the spec files I wrote are now oddballs that do not conform to the coding conventions you've been using.  If the team would prefer that I rewrite these tests in that explicit style, I am happy to do so.

dvoet (approved):

gpolumbo-broad:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -5,24 +5,26 @@ import com.typesafe.scalalogging.LazyLogging
...
+import org.broadinstitute.dsde.workbench.sam.openam.AccessPolicyDAO
 
 import scala.concurrent.Future
 import scala.concurrent.ExecutionContext.Implicits.global
 
 /**
   * Created by gpolumbo on 2/21/2018.
   */
-class ManagedGroupService(resourceService: ResourceService, val resourceTypes: Map[ResourceTypeName, ResourceType]) extends LazyLogging {
+//(private val resourceTypes: Map[ResourceTypeName, ResourceType], private val accessPolicyDAO: AccessPolicyDAO, private val directoryDAO: DirectoryDAO, private val cloudExtensions: CloudExtensions, private val emailDomain: String)
```
breilly2: Nit: removed commented out code

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -37,53 +39,53 @@ class ManagedGroupService(resourceService: ResourceService, val resourceTypes: M
...
+    s"${groupName}@${emailDomain}"
   }
 
-  private def validateEmail(localPart: String) = {
-    val errors = validateLocalPartPattern(localPart) ++ validateEmailLength(constructEmail(localPart))
+  private def validateGroupName(groupName: String) = {
+    val errors = validateGroupNamePattern(groupName) ++ validateGroupNameLength(constructEmail(groupName))
     if (errors.nonEmpty)
-      throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Cannot create valid email address from group name" , errors.toSeq))
+      throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Cannot create valid email address with the group name provided" , errors.toSeq))
```
breilly2: Minor improvement: include the group name in the error message.

breilly2 (approved):
Minor nits then 👍

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: fixed.

gpolumbo-broad:
For GAWB-3260, aside from the intended change, please also note the change to `ManagedGroupService.deleteManagedGroup()`.  I believe that this is the more appropriate implementation.

@dvoet - I have read the CONTRIBUTING.md, and I don't think there's anything additional that I need to do, but I'd appreciate your confirmation.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -54,10 +54,23 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
     }
   }
 
+  // Forcibly delete a group regardless of its membership
+  // If you want to verify that the group is not a member of any other group prior to deletion, use safeDeleteGroup
   override def deleteGroup(groupName: WorkbenchGroupName): Future[Unit] = withContext { ctx =>
```
breilly2: Instead of `safeDeleteGroup`/`deleteGroup`, I'd prefer `deleteGroup`/`forceDeleteGroup`, making "safe" the default behavior. That would make it harder to accidentally discover and call the wrong function when working with a `DirectoryDAO`. I haven't looked at remaining callers of the current `deleteGroup` to know if that would make that code read funny. Thoughts?

gpolumbo-broad: Agreed that this would make more sense.  I contemplated just having the single `deleteGroup` method and rewriting it to always be "safe", but was concerned about breaking code that I might not understand.  Now that I look at it though, up until this PR, the only code calling `DirectoryDAO.deleteGroup()` has been tests.  

Therefore, since this is the first and only production use case for deleting a group so far, I would suggest we just have the one method for `deleteGroup` that does it in the "safe" way, and if we need a "force" version later, we can implement it then. 

I am going to make this change now, which may take a little bit of time to make sure all the tests stay green.

gpolumbo-broad: Fixed

breilly2 (approved):
👍

gpolumbo-broad:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -54,8 +54,13 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
     }
   }
 
-  override def deleteGroup(groupName: WorkbenchGroupName): Future[Unit] = withContext { ctx =>
-    ctx.unbind(groupDn(groupName))
+  override def deleteGroup(groupName: WorkbenchGroupName): Future[Unit] = {
+    listAncestorGroups(groupName).flatMap { ancestors =>
+      if (ancestors.nonEmpty)
```
dvoet: style nit - please always use curly braces, even for 1 line if statements. It makes it easier to read and maintain

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -54,8 +54,13 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
...
+        throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"group ${groupName.value} cannot be deleted because it is a member of at least 1 other group"))
```
dvoet: I don't think it is functionally different but you probably want to return Future.failed instead of throwing the exception.

gpolumbo-broad: Cool, good to know.  FWIW:  https://stackoverflow.com/questions/31194138/difference-between-returning-future-failedexception-and-throwing-an-exception

breilly2: The discussion on the answer is pretty interesting, in particular https://stackoverflow.com/questions/31194138/difference-between-returning-future-failedexception-and-throwing-an-exception#comment50481668_31195611

gpolumbo-broad:
Merged GAWB-2529 (PR: https://github.com/broadinstitute/sam/pull/116) into this branch

MatthewBemis:
:+1: on the group sync stuff

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
```diff
@@ -98,30 +98,41 @@ trait GoogleExtensionRoutes extends ExtensionRoutes with UserInfoDirectives with
...
+                  googleExtensions.getSynchronizedDate(resourceAndPolicyName).map {
+                    case Some(date) => StatusCodes.OK -> Option(GroupSyncResponse(date.toString))
+                    case None => StatusCodes.NoContent -> None
                   }
                 }
               }
             }
           }
+        } ~
+        path("group" / Segment / "sync") { groupName =>
```
dvoet: I thought we decided not to make this end point but to automatically sync the group

PR #118, opened 2018-03-10
GAWB-2526 Automatically sync Managed Group with Google on Create

gpolumbo-broad:
I made a separate PR for this because I have questions that will most likely result in some additional commits. and PR https://github.com/broadinstitute/sam/pull/108 is already quite large.  

In particular, I need answers to a couple of questions, these are also marked as TODO's in the branch.

1.  When creating a Google Group, should we also trigger `GoogleExtensions.onGroupUpdate`? 
2. What should be in the BODY of the response to `POST /api/group/{groupName}`?  Since we're doing the Google Sync as part of the creation, it seems reasonable to me to put the "update report" in the body.

In all cases:


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -160,6 +160,11 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
     } yield ()
   }
 
+  override def onGroupCreate(groupId: WorkbenchGroupIdentity): Future[Map[WorkbenchEmail, Seq[SyncReportItem]]] = {
```
dvoet: I don't think we want a generic onGroupCreate... feels like every group we create should call this which is not true (only managed groups)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -19,12 +19,14 @@ class ManagedGroupService(private val resourceService: ResourceService, private
   def managedGroupType: ResourceType = resourceTypes.getOrElse(ManagedGroupService.managedGroupTypeName, throw new WorkbenchException(s"resource type ${ManagedGroupService.managedGroupTypeName.value} not found"))
   def memberRole = managedGroupType.roles.find(_.roleName == ManagedGroupService.memberRoleName).getOrElse(throw new WorkbenchException(s"${ManagedGroupService.memberRoleName} role does not exist in $managedGroupType"))
 
+  // TODO: What should be returned here?  Just a Resource object?  The report generated from the sync with Google?  Both?
```
dvoet: I don't think we want the google stuff

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -19,12 +19,14 @@ class ManagedGroupService(private val resourceService: ResourceService, private
...
   def createManagedGroup(groupId: ResourceId, userInfo: UserInfo): Future[Resource] = {
     for {
       managedGroup <- resourceService.createResource(managedGroupType, groupId, userInfo)
       _ <- createPolicyForMembers(managedGroup)
       policies <- accessPolicyDAO.listAccessPolicies(managedGroup)
-      _ <- createAggregateGroup(managedGroup, policies)
+      workbenchGroup <- createAggregateGroup(managedGroup, policies)
+      _ <- cloudExtensions.onGroupCreate(workbenchGroup.id)
```
dvoet: I think just call synchronizeGroupMembers directly

dvoet (changes requested):
see comments

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
breilly2: Besides, I don't think we want to wait for the sync to complete before returning.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
breilly2: Probably better to publish a message to the pub/sub topic instead of calling `synchronizeGroupMembers ` directly.

breilly2 (changes requested):
Small change, but may require a little refactoring.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
gpolumbo-broad: Resolved.  Had to add `synchronizeGroupMembers` to the `CloudExtensions` trait.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: I don't disagree, however, I believe implementing in this way would be non-trivial.  If anyone feels strongly that this logic should be written to use Google Pub/Sub, maybe that should be its own standalone issue?

breilly2: I think we have most of the code we need. See `GoogleExtensions.onGroupUpdate`. May just need to extract a method to keep the spirit of the `on[Event]` functions.

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: OK, I definitely have more reading to do to try to understand Akka Actors (I think?) then.  In the short term though, after quickly trying to read through the code it looks to me like `GoogleGroupSyncMonitor.scala` is the thing that manages the `Sub` side of the `Pub/Sub`, and if I'm reading this correctly, it looks like when I create a group in `ManagedGroupService`, I can then just call `GoogleExtensions.onGroupUpdate`, which will ultimately call `GoogleExtensions.synchronizeGroupMembers` for me.  Or, to stick with the pattern used in `ResourceService`, I should call `GoogleExtensions.fireGroupUpdateNotification`.  

I'm happy to keep reading code and verifying my understanding of how this works.  Or if someone else can help me confirm that my interpretation is correct (or not) then we can merge this PR sooner rather than later.  Either way is fine by me.  In the meantime, I need to learn this stuff, so I'll keep reading and experimenting to see if I'm grokking this all correctly.

gpolumbo-broad: Talked with @dvoet to help clarify how this works.  The existing `onGroupUpdate` functionality only updates Groups that already exist.  I added the `publishGroup` method to force the Group creation and use Google Pub/Sub.  I added the method to `CloudExtensions` so that we didn't have to pass a `GooglePubSubDAO` into the `ManagedGroupService` constructor.

breilly2 (approved):
👍 Looks like the changes I was expecting to see.

PR #122, opened 2018-03-20
GAWB-2528 Endpoints for listing/overwriting Admin/Member policies for a Managed Group

gpolumbo-broad:
I went back and forth about whether `JndiDirectoryDao. loadSubjectEmails` should be rewritten to use JNDI directly, or maybe use `loadUsers` and `loadGroups`.  I would take one of these approaches to reduce the number of times we query OpenDJ, which seems like the right way to go, but in the spirit of moving this issue along, I left it as is for now.  I'll change it if anybody wants me to do so.

Please note that this issue is merging back into GAWB-2526. Due to the code freeze, this issue needed to be built off of the code in that branch, so that is also where this is merging back to.

In all cases:


MatthewBemis (commented):
is there a reason to have separate endpoints for the two different policies, as opposed to having the caller pass the policy name into an endpoint like `/api/group/{groupName}/{policyName}`? a lot of the code for the members vs admins path is almost identical

i.e. if you pass the policy name down through the service level, `listPolicyMemberEmails` can be used directly and `listMemberEmails` and `listAdminEmails` can go away

src/main/resources/swagger/api-docs.yaml
```diff
@@ -248,6 +248,108 @@ paths:
...
 
+  /api/group/{groupName}/admins:
+    get:
+      summary: 'Get email addresses for members of the "admin" policy'
+      responses:
+        200:
+          description: 'Email addresses in the "admin" policy'
+          schema:
+            description: 'Email addresses in the "admin" policy'
+            type: string
```
dvoet: this returns an array, right?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -248,6 +248,108 @@ paths:
...
+        - in: path
+          description: Name of group
+          name: groupName
+          required: true
+          type: string
+      operationId: GetGroupAdminEmails
+      tags:
+        - Group
+    put:
+      summary: 'Overwrite email addresses of members of the "admin" policy'
```
dvoet, line 255: In addition to overwrite apis we should also have apis to add/remove individual admins/members.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -248,6 +248,108 @@ paths:
...
+
+  /api/group/{groupName}/members:
+    get:
+      summary: 'Get email addresses for members of the "member" policy'
+      responses:
+        200:
+          description: 'Email addresses in the "member" policy'
+          schema:
+            description: 'Email addresses in the "member" policy'
+            type: string
```
dvoet: array

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -23,37 +23,90 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives {
...
+  private def handleListAdminEmails(managedGroup: Resource, userInfo: UserInfo): Route = {
+    requireAction(managedGroup, SamResourceActions.readPolicy(AccessPolicyName(ManagedGroupService.adminValue)), userInfo) {
+      complete(
+        managedGroupService.listAdminEmails(managedGroup.resourceId).map(StatusCodes.OK -> _)
+      )
+    }
+  }
+
+  private def handleListMemberEmails(managedGroup: Resource, userInfo: UserInfo): Route = {
+    requireAction(managedGroup, SamResourceActions.readPolicy(AccessPolicyName(ManagedGroupService.adminValue)), userInfo) {
```
dvoet: wrong AccessPolicyName

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -242,13 +242,20 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
     }
   }
 
+  // TODO: What do we want this method to do when the subject does not exist? Currently, it throws a NameNotFoundException which is not handled properly
```
dvoet: looks like we should handle it and return None

gpolumbo-broad:
@MatthewBemis I debated about this too.  I ended up doing it this way because my understanding of Managed Groups is that they will only ever have 2 and only 2 policies: Members and Admins.  Doing it the way I implemented, it is very explicit about that fact.  I agree that it makes for some redundancy (especially in how I wrote the tests).  I opted for the redundant code in order to have the explicit endpoints, which I think makes for a slightly better user experience.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -84,10 +84,45 @@ class ManagedGroupService(private val resourceService: ResourceService, private
...
+        resourceService.overwritePolicy(managedGroupType, policyName, resourceAndPolicyName.resource, updatedPolicy)
+      case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Group or policy could not be found: $resourceAndPolicyName"))
+    }
+  }
 }
 
 object ManagedGroupService {
-  val memberRoleName = ResourceRoleName("member")
+  val memberValue = "member"
+  val adminValue = "admin"
```
dvoet: I don't really like have these Strings floating around publicly. I would have a ResourceRoleName and AccessPolicyName val for admin and member and use those throughout. If you want them to use the same private string val, great.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -84,10 +84,45 @@ class ManagedGroupService(private val resourceService: ResourceService, private
...
+  }
+
+  def overwriteMemberEmails(resourceId: ResourceId, emails: Set[WorkbenchEmail]): Future[AccessPolicy] = {
+    overwritePolicyMemberEmails(resourceId, AccessPolicyName(ManagedGroupService.memberValue), emails)
+  }
+
+  private def overwritePolicyMemberEmails(resourceId: ResourceId, policyName: AccessPolicyName, emails: Set[WorkbenchEmail]): Future[AccessPolicy] = {
+    val resourceAndPolicyName = ResourceAndPolicyName(Resource(ManagedGroupService.managedGroupTypeName, resourceId), policyName)
+    accessPolicyDAO.loadPolicy(resourceAndPolicyName).flatMap {
+      case Some(policy) => val updatedPolicy = AccessPolicyMembership(emails, policy.actions, policy.roles)
```
dvoet: policy.copy(memberEmails = emails)

MatthewBemis:
i don't know if the UX value-add warrants the amount of code duplication. i think the main UX value it adds is for users looking at the swagger page, but that can be solved by having a drop down of the available options. this is how we currently do it for billing projects and managed groups: https://firecloud-orchestration.dsde-dev.broadinstitute.org/#!/Groups/addUserToGroup

but regardless of the two-endpoint debate, the code path underneath the routes could still be merged nicely into one

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
dvoet: what happens if any of the incoming emails are invalid or unknown?

dvoet (changes requested):
see comments

src/main/resources/swagger/api-docs.yaml
gpolumbo-broad: fixed

src/main/resources/swagger/api-docs.yaml
gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
gpolumbo-broad: @dvoet, I guess my understanding here was that only Admins should be permitted to see who is a member of the "members" policy.   So the correct behavior is that members of the "member" policy should have `read_policy::member` access.  Should they also have `read_policy::admin`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
gpolumbo-broad: fixed

src/main/resources/swagger/api-docs.yaml
gpolumbo-broad: fixed

gpolumbo-broad:
There are a few lingering TODO's in the code that I would like other's feedback on.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -222,6 +222,164 @@ paths:
...
+      operationId: GetGroupAdminEmails
+      tags:
+        - Group
+    put:
+      summary: 'Overwrite email addresses of members of the "admin" policy'
+      responses:
+        201:
+          description: Admin Policy successfully updated
+        404:
+          description: Group does not exist or you are not a member of the admin policy for the group
```
MatthewBemis: a leftover from merging the two endpoints into one: a bunch of these descriptions specifically reference the "admin" policy

gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -1,59 +1,131 @@
...
+  }
+
+  private def handleDeleteGroup(managedGroup: Resource, userInfo: UserInfo): Route = {
+    requireAction(managedGroup, SamResourceActions.delete, userInfo) {
+      complete(managedGroupService.deleteManagedGroup(managedGroup.resourceId).map(_ => StatusCodes.NoContent))
+    }
+  }
+
+  private def handleListEmails(managedGroup: Resource, accessPolicyName: ManagedGroupPolicyName, userInfo: UserInfo): Route = {
+    requireAction(managedGroup, SamResourceActions.readPolicy(accessPolicyName.asInstanceOf[AccessPolicyName]), userInfo) {
```
dvoet: I don't like the asInstanceOf, why do you need to cast?

you probably want the signature to be `AccessPolicyName with ManagedGroupPolicyName`

and you can tighten that up by declaring a type

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -1,59 +1,131 @@
...
+      managedGroupService.loadManagedGroup(managedGroup.resourceId).map {
         case Some(response) => StatusCodes.OK -> response
         case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "group not found"))
       }
     )
   }
 
-  private def handlePostGroup(groupId: String, userInfo: UserInfo) = {
-    complete(managedGroupService.createManagedGroup(ResourceId(groupId), userInfo).map(_ => StatusCodes.Created))
+  private def handlePostGroup(managedGroup: Resource, userInfo: UserInfo): Route = {
```
dvoet: I would call this handleCreateGroup to be consistent with the others

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -85,10 +86,50 @@ class ManagedGroupService(private val resourceService: ResourceService, private
...
 }
 
 object ManagedGroupService {
-  val memberRoleName = ResourceRoleName("member")
   val managedGroupTypeName = ResourceTypeName("managed-group")
   val groupNameRe = "^[A-z0-9_-]+$".r
+  private val memberValue = "member"
+  private val adminValue = "admin"
+
+  // In lieu of an Enumeration, this trait is being used to ensure that we can only have these policies in a Managed Group
```
dvoet: see this for a good way to implement an enum
https://github.com/broadinstitute/workbench-libs/blob/develop/util/src/main/scala/org/broadinstitute/dsde/workbench/util/health/StatusModel.scala#L16

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -243,9 +243,19 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
...
+      Option(ctx.getAttributes(subDn).get(Attr.email)).map { emailAttr =>
+        WorkbenchEmail(emailAttr.get.asInstanceOf[String])
+      }
+    }.recover {
+      case _: NameNotFoundException => None
+    }.get
+  }
+
+  override def loadSubjectEmails(subjects: Set[WorkbenchSubject]): Future[Set[WorkbenchEmail]] = {
+    Future.traverse(subjects) { subject =>
```
dvoet: look at loadGroups for how to do batched queries... this scales with the number of subjects which probably won't usually be that bad but there will be cases where it is.

dvoet: also throwing an exception when a subject isn't found means that if somehow a group has a reference to a subject that doesn't exist this call will perpetually break. Maybe return a Map[Subject, Option[Email]) and let the caller figure it out? Or just ignore (which is what happens in the batch load).

dvoet (changes requested):
see comments

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
gpolumbo-broad: I tried the case class approach first when I was doing this, and it didn't work because `AccessPolicyName` is already a case class, which means that I can't extend it.  

If you're still not thrilled with this implementation, maybe we can take a few minutes to look at it together

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
gpolumbo-broad: In order to do this all in one query, I think we would need to search the `groupsOu` and `peopleOu` namespaces in a single query and I'm not sure how to do that.  Instead, I simply proxy out the queries to the existing `loadUsers` and `loadGroups` methods.  This is an improvement because we're down to 2 queries instead of 1 per subject, which is better, but admittedly we might be able to do better.  If there's something you can steer me towards to get this done in a single query, please just point me in the right direction.

dvoet (approved):

PR #123, opened 2018-03-20
GAWB-3314: create pet SA in google if it only exists in OpenDJ (also add missing key resiliency)

MatthewBemis:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -186,23 +186,27 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
   }
 
   def createUserPetServiceAccount(user: WorkbenchUser, project: GoogleProject): Future[PetServiceAccount] = {
-    val (petSaID, petSaDisplayName) = toPetSAFromUser(user)
+    val (petSaName, petSaDisplayName) = toPetSAFromUser(user)
 
-    // First find or create the service account in Google, which generates a unique id and email
-    val petSA = googleIamDAO.getOrCreateServiceAccount(project, petSaID, petSaDisplayName)
+    // First find or create the service account in Google, which generates a unique id and email.
+    // We do this first because the pet may have been deleted from Google by an outsider, which would
+    // leave behind a reference in OpenDJ. It is not enough to just see if the pet exists in OpenDJ.
+    val petSA = googleIamDAO.getOrCreateServiceAccount(project, petSaName, petSaDisplayName)
     petSA.flatMap { serviceAccount =>
-      // Set up the service account with the necessary permissions
-      val petServiceAccount = PetServiceAccount(PetServiceAccountId(user.id, project), serviceAccount)
       // If the service account doesn't exist in OpenDJ, create it
       directoryDAO.loadPetServiceAccount(PetServiceAccountId(user.id, project)).flatMap {
         case Some(pet) => Future.successful(pet)
-        case None =>
+        case None => {
+          // Set up the service account with the necessary permissions
+          val petServiceAccount = PetServiceAccount(PetServiceAccountId(user.id, project), serviceAccount)
+
           setUpServiceAccount(petServiceAccount) andThen { case Failure(_) =>
             // If anything fails with setup, clean up any created resources to ensure we don't end up with orphaned pets.
             removePetServiceAccount(petServiceAccount).failed.foreach { e =>
-              logger.warn(s"Error occurred cleaning up pet service account [$petSaID] [$petSaDisplayName]", e)
+              logger.warn(s"Error occurred cleaning up pet service account [$petSaName] [$petSaDisplayName]", e)
             }
           }
+        }
       }
     }
   }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -188,21 +188,22 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
   def createUserPetServiceAccount(user: WorkbenchUser, project: GoogleProject): Future[PetServiceAccount] = {
     val (petSaID, petSaDisplayName) = toPetSAFromUser(user)
 
-    directoryDAO.loadPetServiceAccount(PetServiceAccountId(user.id, project)).flatMap {
-      case Some(pet) => Future.successful(pet)
-      case None =>
-        // First find or create the service account in Google, which generates a unique id and email
-        val petSA = googleIamDAO.getOrCreateServiceAccount(project, petSaID, petSaDisplayName)
-        petSA.flatMap { serviceAccount =>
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -228,7 +229,7 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
       maybePet <- directoryDAO.loadPetServiceAccount(PetServiceAccountId(userId, project))
       result <- maybePet match {
         case Some(pet) => googleKeyCache.removeKey(pet, keyId)
-        case none => Future.successful(())
+        case None => Future.successful(())
```
MatthewBemis, line 236: fortunately this worked before...

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -188,21 +188,22 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
 
-    directoryDAO.loadPetServiceAccount(PetServiceAccountId(user.id, project)).flatMap {
-      case Some(pet) => Future.successful(pet)
-      case None =>
-        // First find or create the service account in Google, which generates a unique id and email
-        val petSA = googleIamDAO.getOrCreateServiceAccount(project, petSaID, petSaDisplayName)
-        petSA.flatMap { serviceAccount =>
-          // Set up the service account with the necessary permissions
-          val petServiceAccount = PetServiceAccount(PetServiceAccountId(user.id, project), serviceAccount)
+    // First find or create the service account in Google, which generates a unique id and email
```
MatthewBemis: gpalloc and other "rogue" beings may delete a pet SA in google, leaving a reference to it behind in Sam.

previously we would check to see if the pet SA existed in OpenDJ, and assume everything was cool in google if it existed. since that's not always going to be the case, i've switched up the ordering to first getOrCreate the pet in google and then create the pet in opendj if needed.

per my local logs, the extra google call appears to add about ~500ms to each call to get an existing key.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
jmthibault79: ha

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -188,21 +188,22 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+    val petSA = googleIamDAO.getOrCreateServiceAccount(project, petSaID, petSaDisplayName)
+    petSA.flatMap { serviceAccount =>
+      // Set up the service account with the necessary permissions
+      val petServiceAccount = PetServiceAccount(PetServiceAccountId(user.id, project), serviceAccount)
```
jmthibault79: This is thrown away if it's in OpenDJ so we should move that to the `None` below

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -188,21 +188,22 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
   def createUserPetServiceAccount(user: WorkbenchUser, project: GoogleProject): Future[PetServiceAccount] = {
     val (petSaID, petSaDisplayName) = toPetSAFromUser(user)
```
jmthibault79: I was wondering what the difference was between`petSaID` and the `PetServiceAccountId` below.  Turns out this is a *name* rather than an ID.  Please rename it as such.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
jmthibault79: Thanks!  Add a comment indicating why we may need to recreate the GooglePet even if it's in OpenDJ, so we don't accidentally "optimize" this in the future

jmthibault79:
Looks good.  I would also like to see an integration test with a rogue pet deletion and confirmation that `getPetServiceAccountKey()` still works afterward.

MatthewBemis:
@jmthibault79 i've added an integration test but i'm concerned now that it could have side effects on other tests on other fiabs since it is using the default project. this would likely be rare, but it's still a possibility.

i've updated it locally to use GPAlloc but there is a permissions issue when deleting the pet SA from google. i'll need to dive in further to see what's wrong. depending on the amount of work required to get it to work with GPAlloc, we may want to merge this with the test ignored and a follow-up to re-enable it with GPAlloc.

jmthibault79:
Thanks!  I wonder if `googleIamDAO` is trying to remove it as the test-runner user and they don't have the right perms

MatthewBemis:
@jmthibault79 in the latest commit i also added some resiliency for the case we originally talked about where a key has been deleted directly from the SA but still exists in the Google bucket. this wasn't related to the errors you were seeing, but i did end up noticing a case where things could go wrong

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -73,11 +73,17 @@ class GoogleKeyCache(val googleIamDAO: GoogleIamDAO, val googleStorageDAO: Googl
...
     val keyRetired = System.currentTimeMillis() - mostRecentKey.timeCreated.toEpochMilli > 86400000L * googleServicesConfig.googleKeyCacheConfig.activeKeyMaxAge
 
-    if(keyRetired) furnishNewKey(pet)
+    val keyPathPattern = """([^\/]+)\/([^\/]+)\/([^\/]+)""".r
+    val keyPathPattern(project, petSaEmail, keyId) = mostRecentKey.value
+
+    //The key may exist in the Google bucket cache, but could have been deleted from the SA directly
+    val keyExistsForSA = serviceAccountKeys.exists(_.id.value.contentEquals(keyId))
+
+    if(keyRetired || !keyExistsForSA) furnishNewKey(pet)
```
MatthewBemis, line 86: also we do not need to worry about deleting the key, we can wait for the lifecycle to pick them up

dvoet: what happens if key is already gone from google when deleted from bucket?

dvoet (approved):
make sure error is handled when deleting a key from google and it does not exist

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
MatthewBemis: it's handled without an issue:

1) pub/sub message is triggered
2) sam tries to delete it, gets a 404 which recovers gracefully in HttpGoogleIamDAO 
3) sam sees this as a success and acks the message

jmthibault79 (approved):
ok to ignore the test for now, to make other tests more reliable

PR #124, opened 2018-03-22
stop looking for PROXY

dvoet:

In all cases:


breilly2 (approved):
I understand this change only in the context of what's going on with tests right now. In the future, we'll need to change this code again when we do the proxy email change. At that time, we'll have to understand this again and know how to handle the implications of the max email length. It's worth adding some comment in here about the reason for `endWith` instead of checking equality. After that, I guess we don't have much choice but to move forward.

I really hate that we have to do this. I think it's time to take a step back and look at our in vivo (full system integration) test granularity to do more in each test, keeping permutation and edge case checking in ex vivo and in vitro (single service and unit) tests.

dvoet:
with luck google will fix it by then

MatthewBemis (approved):

PR #128, opened 2018-03-30
GAWB-3336 SAM /api/groups lists groups for user

gpolumbo-broad:
Found a bug in `MockAccessPolicyDao.listAccessPolicies` where it was taking in a `WorkbenchUserId` but wasn't doing anything with it.  I wrote a spec file for `MockAccessPolicyDao`, which is admittedly weird writing tests for your tests...but I identified that the `listAccessPolicies` method in `MockAccessPolicyDao` as behaving differently than the same method in `JndiAccessPolicyDao`.  Therefore, this new spec file exists for verify the behavior of the `MockAccessPolicyDao`.  

Also note that the new endpoint is `api/groups/` (plural) whereas the other group apis are `api/group/` (singular).  I think its fine, but I'm not sure if anyone else has an opinion,

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -157,6 +157,20 @@ paths:
       tags:
         - Admin
 
+  /api/groups:
+    get:
+      summary: Show all the groups the requesting user belongs to and their policy membership in each group
+      responses:
+        200:
+          description: Managed Group memberships
+        500:
+          description: Internal Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
+      operationId: listGroupMemberships
+      tags:
+        - Group
+
   /api/group/{groupName}:
     get:
       summary: Show email address of the group
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -163,6 +163,10 @@ paths:
       responses:
         200:
           description: Managed Group memberships
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/ResourceAndAccessPolicy'
         500:
           description: Internal Error
           schema:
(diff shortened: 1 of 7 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -157,6 +157,20 @@ paths:
       tags:
         - Admin
 
+  /api/groups:
+    get:
+      summary: Show all the groups the requesting user belongs to and their policy membership in each group
+      responses:
+        200:
+          description: Managed Group memberships
```
MatthewBemis, line 165: what's the json structure returned?

gpolumbo-broad: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -87,6 +87,10 @@ class ManagedGroupService(private val resourceService: ResourceService, private
     } yield ()
   }
 
+  def listGroups(userId: WorkbenchUserId): Future[Set[ResourceIdAndPolicyName]] = {
+    accessPolicyDAO.listAccessPolicies(ManagedGroupService.managedGroupTypeName, userId)
```
dvoet, line 91: if we add other policies they will show up here too... should we be proactive about dropping them?

gpolumbo-broad: Yes, they will show up.  @breilly2 and I just talked about this a bit and I'm of the mind not to be proactive yet.  I think we're going to be addressing this issue as we try to tackle the "Request Access" GAWB-3335 problem.

PR #133, opened 2018-04-09
GAWB-2532 related: return email in list groups endpoint

MatthewBemis:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/JndiAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/JndiAccessPolicyDAO.scala
@@ -169,8 +169,8 @@ class JndiAccessPolicyDAO(protected val directoryConfig: DirectoryConfig)(implic
   }
 
   override def listAccessPoliciesWithEmail(resourceTypeName: ResourceTypeName, userId: WorkbenchUserId): Future[Set[ResourceIdAndPolicyNameWithEmail]] = {
-    listAccessPolicies(resourceTypeName, userId).flatMap { x =>
-      Future.traverse(x) { y => getPolicyEmail(resourceTypeName, ResourceIdAndPolicyName(y.resourceId, y.accessPolicyName)).map(z => ResourceIdAndPolicyNameWithEmail(y.resourceId, y.accessPolicyName, z)) }
+    listAccessPolicies(resourceTypeName, userId).flatMap { ripns =>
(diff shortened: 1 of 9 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/JndiAccessPolicyDAO.scala
```diff
@@ -168,6 +168,19 @@ class JndiAccessPolicyDAO(protected val directoryConfig: DirectoryConfig)(implic
     }
   }
 
+  override def listAccessPoliciesWithEmail(resourceTypeName: ResourceTypeName, userId: WorkbenchUserId): Future[Set[ResourceIdAndPolicyNameWithEmail]] = {
+    listAccessPolicies(resourceTypeName, userId).flatMap { ripns =>
+      Future.traverse(ripns) { ripn => getPolicyEmail(resourceTypeName, ResourceIdAndPolicyName(ripn.resourceId, ripn.accessPolicyName)).map(email => ResourceIdAndPolicyNameWithEmail(ripn.resourceId, ripn.accessPolicyName, email)) }
```
MatthewBemis: one extra query per managed group

alternatively i could re-write the `listAccessPolicies` query but then we don't take advantage of the `memberOf` attribute and the id/policy name being encoded in the dn

src/test/scala/org/broadinstitute/dsde/workbench/sam/openam/MockAccessPolicyDAOSpec.scala
```diff
@@ -86,7 +86,7 @@ class MockAccessPolicyDAOSpec extends FlatSpec with Matchers with TestSupport wi
     runAndWait(mock.userService.createUser(dummyUser))
 
     val expectedGroups = Set(ResourceIdAndPolicyName(ResourceId(groupName), ManagedGroupService.adminPolicyName))
-    runAndWait(jndi.managedGroupService.listGroups(dummyUserInfo.userId)) shouldEqual expectedGroups
-    runAndWait(mock.managedGroupService.listGroups(dummyUserInfo.userId)) shouldEqual expectedGroups
+    runAndWait(jndi.managedGroupService.listGroups(dummyUserInfo.userId)).map(ripn => ResourceIdAndPolicyName(ripn.resourceId, ripn.accessPolicyName)) shouldEqual expectedGroups
```
MatthewBemis, line 89: the email is a uuid thus it's unpredictable. so i'm omitting it from the assertions.

PR #139, opened 2018-04-13
GAWB-3366: Endpoint for getting Pet SA access token

MatthewBemis:
Workbench-libs PR incoming

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
@@ -159,7 +159,7 @@ class SamApiSpec extends FreeSpec with BillingFixtures with Matchers with ScalaF
       val expectedProxyEmail1 = s"${username1}_$userId1@${Config.GCS.appsDomain}"
 */
       val expectedProxyEmail1 = s"$userId1@${Config.GCS.appsDomain}"
-/**/
+/*...*/
       proxyGroup1_1.value should endWith (expectedProxyEmail1)
       proxyGroup1_2.value should endWith (expectedProxyEmail1)
 
@@ -167,7 +167,7 @@ class SamApiSpec extends FreeSpec with BillingFixtures with Matchers with ScalaF
       val expectedProxyEmail2 = s"${username2}_$userId2@${Config.GCS.appsDomain}"
 */
       val expectedProxyEmail2 = s"$userId2@${Config.GCS.appsDomain}"
-/**/
+/*...*/
       proxyGroup2_1.value should endWith (expectedProxyEmail2)
       proxyGroup2_2.value should endWith (expectedProxyEmail2)
     }
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
@@ -159,7 +159,7 @@ class SamApiSpec extends FreeSpec with BillingFixtures with Matchers with ScalaF
       val expectedProxyEmail1 = s"${username1}_$userId1@${Config.GCS.appsDomain}"
 */
       val expectedProxyEmail1 = s"$userId1@${Config.GCS.appsDomain}"
-/*...*/
+// /**/
       proxyGroup1_1.value should endWith (expectedProxyEmail1)
       proxyGroup1_2.value should endWith (expectedProxyEmail1)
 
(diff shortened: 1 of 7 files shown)


gpolumbo-broad (approved):

automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
```diff
@@ -255,6 +255,19 @@ class SamApiSpec extends FreeSpec with BillingFixtures with Matchers with ScalaF
...
+    "should get an access token for a user's own pet service account" in {
+      val user = UserPool.chooseStudent
+
+      withCleanBillingProject(UserPool.chooseProjectOwner, List(user.email)) { project =>
+
+        val token1 = Sam.user.petServiceAccountToken(project)(user.makeAuthToken)
+
+        val token2 = Sam.user.petServiceAccountToken(project)(user.makeAuthToken)
+
+        token1 shouldNot be(token2)
```
rtitle: Does the token necessarily change for subsequent calls to `petServiceAccountToken` from the same user? I could see a service returning the same token for a user until it expires (the Google [javascript API](https://developers.google.com/api-client-library/javascript/reference/referencedocs) does this for example).

A cool test would be to _use_ the token to make a call as the pet, and verify that it succeeds. Would this test be valid?
```
// get my pet's email
val petEmail1 =  Sam.user.petServiceAccountEmail(project)(user.makeAuthToken)
// get my pet's token
val petToken = Sam.user.petServiceAccountToken(project)(user.makeAuthToken)
// get my pet's email using my pet's token
val petEmail2 = Sam.user.petServiceAccountEmail(project)(petToken)
// result should be the same
petEmail2 shouldBe petEmail1
```

automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
```diff
@@ -159,15 +159,15 @@ class SamApiSpec extends FreeSpec with BillingFixtures with Matchers with ScalaF
       val expectedProxyEmail1 = s"${username1}_$userId1@${Config.GCS.appsDomain}"
 */
       val expectedProxyEmail1 = s"$userId1@${Config.GCS.appsDomain}"
-/**/
+// /**/
```
rtitle: commented comment?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -253,6 +257,18 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
 
+  def getPetServiceAccountToken(user: WorkbenchUser, project: GoogleProject): Future[String] = {
+    getPetServiceAccountKey(user, project).flatMap { key =>
+      getAccessTokenUsingJson(key)
+    }
+  }
+
+  def getAccessTokenUsingJson(saKey: String) : Future[String] = Future {
+    val keyStream = new ByteArrayInputStream(saKey.getBytes)
+    val credential = ServiceAccountCredentials.fromStream(keyStream).createScoped(Seq(PlusScopes.USERINFO_EMAIL, PlusScopes.USERINFO_PROFILE, StorageScopes.DEVSTORAGE_READ_ONLY).asJava)
```
rtitle: I'm curious if `DEVSTORAGE_READ_ONLY` scopes are really needed here? Would it make sense to use the same scopes for the pet token as the scopes of the requesting user?

Or alternatively allow specifying scopes in the API request?

automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
MatthewBemis: `/**/` confuses github and makes everything below it appear commented out so i just commented out that comment to negate that effect. i'll change it to something that looks less stupid

rtitle:
Just a few questions above.. overall looks good. I think Leo could use this functionality.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
MatthewBemis: using the same scopes as the requesting user would break saturn UI. they're not allowed to request storage scopes for users, which is why they need to use pets to get around that restriction.

i like the idea of specifying the scopes since it seems like there are several different use-cases with different scopes requirements.

rtitle (approved):

PR #141, opened 2018-04-17
Sam Phase IV Feature Branch

MatthewBemis:
Included in this branch currently:

GAWB-3321
GAWB-3388

----------------

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
@@ -30,7 +30,39 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
     accessPolicyDAO.createResourceType(resourceType.name)
   }
 
-  def createResource(resourceType: ResourceType, resourceId: ResourceId, userInfo: UserInfo): Future[Resource] = {
+  def createResource(resourceType: ResourceType, resourceId: ResourceId, policies: Option[Map[AccessPolicyName, AccessPolicyMembership]] = None, userInfo: UserInfo): Future[Resource] = {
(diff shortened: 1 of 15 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -30,13 +30,29 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+      policiesOpt match {
+        case Some(policies) =>
+          Future.traverse(policies) { case (accessPolicyName, accessPolicyMembership) =>
+            val resourceAndPolicyName = ResourceAndPolicyName(Resource(resourceType.name, resourceId), accessPolicyName)
+            val loadedSubjects = Future.traverse(accessPolicyMembership.memberEmails) { directoryDAO.loadSubjectFromEmail }.map(subjOpts => subjOpts.collect { case Some(subj) => subj })
+            val roles = accessPolicyMembership.roles
+            val actions = accessPolicyMembership.actions
+
+            loadedSubjects.flatMap { subjects =>
+              createPolicy(resourceAndPolicyName, subjects, roles, actions)
```
gpolumbo-broad: https://broadinstitute.atlassian.net/browse/GAWB-2994

We used to not have to worry about validating policies on Resource create because the policies being created we precisely structured and populated.  Now, because we allow users to submit an arbitrary list of policies with whatever data they want in them, I think we need to validate the policies here before creating them.  

At the very least, it seems like line 39 is reintroducing bug GAWB-2994, in that it seems to just silently ignore email addresses that don't already map to a subject in SAM.

dvoet:
jenkins retest

dvoet:
jenkins retest because weird

PR #143, opened 2018-04-19
GAWB-3389 Allowing/disallowing resource ID reuse is configurable

gpolumbo-broad:
ManageGroups are treated special, as is the case with Managed Groups in SAM.  They have their own test, but this test does NOT use configuration defined in `reference.conf`.  Should it?  

In all cases:

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -29,6 +29,7 @@ resourceTypes =
           roleActions = []
         }
       }
+      reuseIds = true
     },
     workflow-collection = {
       actionPatterns = ["delete", "add", "view", "abort", "alter_policies", "read_policies"]
@@ -44,6 +45,7 @@ resourceTypes =
           roleActions = ["view", "add", "delete", "abort"]
         }
       }
+      reuseIds = false
     },
     caas = {
       actionPatterns = ["get_whitelist", "alter_policies", "read_policies"]
@@ -56,6 +58,7 @@ resourceTypes =
         roleActions = ["get_whitelist"]
        }
       }
+      reuseIds = false
     },
     billing-project = {
       actionPatterns = ["create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "list_notebook_cluster", "launch_notebook_cluster", "sync_notebook_cluster", "delete_notebook_cluster", "stop_start_notebook_cluster", "alter_google_role", "share_policy::.+", "read_policy::.+"]
@@ -74,6 +77,7 @@ resourceTypes =
           roleActions = ["launch_notebook_cluster"]
         }
       }
+      reuseIds = false
     },
     notebook-cluster = {
       actionPatterns = ["status", "connect", "sync", "delete", "read_policies", "stop_start"]
@@ -83,6 +87,7 @@ resourceTypes =
           roleActions = ["status", "connect", "sync", "delete", "read_policies", "stop_start"]
         }
       }
+      reuseIds = true
     },
     cloud-extension = {
       actionPatterns = ["get_pet_private_key", "alter_policies", "read_policies"]
@@ -95,5 +100,6 @@ resourceTypes =
           roleActions = ["get_pet_private_key"]
         }
       }
+      reuseIds = false
     }
   }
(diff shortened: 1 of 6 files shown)


src/main/resources/reference.conf
```diff
@@ -74,6 +78,7 @@ resourceTypes =
           roleActions = ["launch_notebook_cluster"]
         }
       }
+      reuseIds = false
```
MatthewBemis: for billing projects we should make sure to override this value to `true` for fiabs so we don't break gpalloc

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupServiceSpec.scala
```diff
@@ -40,7 +40,7 @@ class ManagedGroupServiceSpec extends FlatSpec with Matchers with TestSupport wi
   private val resourceActionPatterns = resourceActions.map(action => ResourceActionPattern(action.value))
   private val defaultOwnerRole = ResourceRole(ManagedGroupService.adminRoleName, resourceActions)
   private val defaultRoles = Set(defaultOwnerRole, ResourceRole(ManagedGroupService.memberRoleName, Set.empty))
-  private val managedGroupResourceType = ResourceType(ManagedGroupService.managedGroupTypeName, resourceActionPatterns, defaultRoles, ManagedGroupService.adminRoleName)
+  private val managedGroupResourceType = ResourceType(ManagedGroupService.managedGroupTypeName, resourceActionPatterns, defaultRoles, ManagedGroupService.adminRoleName, true)
```
MatthewBemis: should this be read from config?

gpolumbo-broad: Yeah, exactly what I was trying to reference in my PR comment.  I would argue yes, it should be pulling from the config.  But this is also a bigger change, which is why I didn't really get into too much detail about it.  

Maybe for this PR we just consider changing the ManagedGroup*Specs and make sure they're pulling from the config file.  

As for the other ResourceTypes, we don't really test them specifically, which I think is a hole in our testing.  That config file is constructed in a careful way, and messing it up could be problematic.  I suppose the automation suite might eventually put that config to test, but I don't know how completely it does so, and that's also pretty far down the process pipeline before we're actually testing.  

On the other hand...the configuration of Resource Types probably doesn't change that often, so is it worth having the testing overhead?  (I would argue that it is worth it)

Managed Groups are the oddballs in the list of Resource Types because they have a bunch of code that's specific to them, which is unusual in SAM.  So modifying the ManagedGroup*Specs to load the resource type from the configs probably makes sense.  

As for the other resource types, if we wanted to test them, I think we would make a specific test suite that simply asserts that we have the policies and actions correctly set up for each resource type.  

@MatthewBemis I think we should face talk about this some more.

src/main/resources/reference.conf
dvoet: I would just set it to true here because we only allow delete in non-prod envs anyway

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -68,15 +68,26 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
     for {
       policiesToDelete <- accessPolicyDAO.listAccessPolicies(resource)
       _ <- Future.traverse(policiesToDelete) {accessPolicyDAO.deletePolicy}
-      _ <- accessPolicyDAO.deleteResource(resource)
+      _ <- attemptDeleteResource(resource)
       _ <- Future.traverse(policiesToDelete) { policy => cloudExtensions.onGroupDelete(policy.email) }
     } yield ()
   }
 
+  private def attemptDeleteResource(resource: Resource): Future[Unit] = {
```
dvoet: I would call this maybeDeleteResource. "attempt" implies to me that we will try and possibly handle failures

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupServiceSpec.scala
dvoet: I agree that they should be read from config

dvoet (approved):

PR #145, opened 2018-04-30
GAWB-3568 ActionPatterns described as objects

gpolumbo-broad:
None of the currently defined actions are Auth Domain constrained yet, as is reflected in `reference.conf`.  All actions will default to `false` in terms of being constrained.  

Please review the descriptions I have provided and provide descriptions for those that are missing.  As a framework for writing these descriptions, I wrote the them so that we could potentially generate text like `s"Users who can perform ${action.name} may ${action.description}".  Feel free to rewrite or let me know how you'd like me to rewrite them.
  
Please also look at `SamResourceActionPatterns` in `SamModel.scala`, I'm not sure if we need to provide descriptions for those ActionPatterns.  

Note also that I created a test in `ResourceServiceSpec.scala` for making sure that the `authDomainConstrained` attribute can be read properly.  It felt a little weird putting it there, but I didn't know where to put it.  

In all cases:

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1,106 +1,246 @@
 akka.http.server.idle-timeout = 180 s
 akka.http.server.request-timeout = 60 s
 
-resourceTypes =
-  {
-    workspace = {
-      actionPatterns = ["delete", "read_policies", "share_policy::owner", "share_policy::writer", "share_policy::reader", "read_policy::owner", "read_policy::project-owner"]
-      ownerRoleName = "owner"
-      roles = {
-        owner = {
-          roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::writer", "share_policy::reader", "read_policy::project-owner"]
-        },
-        writer = {
-          roleActions = ["read_policy::owner", "read_policy::project-owner"]
-        },
-        reader = {
-          roleActions = ["read_policy::owner", "read_policy::project-owner"]
-        }
+resourceTypes = {
+  workspace = {
+    actionPatterns = {
+      delete = {
+        description = "Delete this workspace"
       }
-      reuseIds = false
-    },
-    managed-group = {
-      actionPatterns = ["delete", "read_policies", "share_policy::admin", "share_policy::member", "read_policy::admin", "read_policy::member"]
-      ownerRoleName = "admin"
-      roles = {
-        admin = {
-          roleActions = ["delete", "read_policies", "share_policy::admin", "share_policy::member", "read_policy::admin", "read_policy::member"]
-        },
-        member = {
-          roleActions = []
-        }
+      read_policies = {
+        description = "View all policies and policy details for this workspace"
       }
-      reuseIds = true
-    },
-    workflow-collection = {
-      actionPatterns = ["delete", "add", "view", "abort", "update", "alter_policies", "read_policies"]
-      ownerRoleName = "owner"
-      roles = {
-        owner = {
-          roleActions = ["delete", "add", "view", "abort", "update", "alter_policies", "read_policies"]
-        },
-        reader = {
-          roleActions = ["view"]
-        },
-        writer = {
-          roleActions = ["view", "add", "delete", "abort", "update"]
-        }
+      "share_policy::owner" = {
+        description = "Add/remove members to/from the owner policy for this workspace"
       }
-      reuseIds = false
-    },
-    caas = {
-      actionPatterns = ["get_whitelist", "alter_policies", "read_policies"]
-      ownerRoleName = "owner"
-      roles = {
-       owner = {
-        roleActions = ["alter_policies", "read_policies"]
-       },
-       cromiam = {
-        roleActions = ["get_whitelist"]
-       }
+      "share_policy::writer" = {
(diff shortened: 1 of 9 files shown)


src/main/resources/reference.conf
```diff
@@ -1,106 +1,246 @@
...
+        roleActions = ["launch_batch_compute"]
+      },
+      notebook-user = {
+        roleActions = ["launch_notebook_cluster"]
+      }
+    }
+    reuseIds = true
+  }
+  notebook-cluster = {
+    actionPatterns = {
```
MatthewBemis, line 195: @rtitle given that these are actions that your team cares about, would you (or anyone else from your team) like to do the honors of describing them? there's also some notebook actions under `billing-project`

they would need to be written such that the sentence `s"Users who can perform ${action.name} may ${action.description}".` could be generated

otherwise, we'll do our best to come up with the descriptions

src/main/resources/reference.conf
```diff
@@ -1,106 +1,246 @@
...
-    },
-    caas = {
-      actionPatterns = ["get_whitelist", "alter_policies", "read_policies"]
-      ownerRoleName = "owner"
-      roles = {
-       owner = {
+resourceTypes = {
+  workspace = {
+    actionPatterns = {
+      delete = {
```
MatthewBemis, line 7: this is a total nit but is there a reason why some actions are in quotes and others are not?

gpolumbo-broad: Yes, special characters like `:` and `+`.   So I took the approach of only quoting those keys that needed them, but if you'd prefer to be consistent, I'm cool with that too.

dvoet:
if we want to be able to place the action descriptions in the specified sentence they should not start with capital letters

src/main/resources/reference.conf
```diff
@@ -1,106 +1,246 @@
...
+        description = "Delete this workspace"
+      }
+      read_policies = {
+        description = "View all policies and policy details for this workspace"
+      }
+      "share_policy::owner" = {
+        description = "Add/remove members to/from the owner policy for this workspace"
```
dvoet: how about "may change the membership of the owner policy for this workspace"

I find "add/remove" and "to/from" too distracting

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -58,20 +58,19 @@ object SamResourceActions {
...
 }
 
 case class UserStatusDetails(userSubjectId: WorkbenchUserId, userEmail: WorkbenchEmail) //for backwards compatibility to old API
 case class UserStatus(userInfo: UserStatusDetails, enabled: Map[String, Boolean])
 
-case class ResourceActionPattern(value: String) extends ValueObject {
-  lazy val regex = value.r
-  def matches(other: ResourceAction) = regex.pattern.matcher(other.value).matches()
+case class ResourceActionPattern(value: String, description: String, authDomainConstrained: Boolean) {
+  def matches(other: ResourceAction) = value.r.pattern.matcher(other.value).matches()
```
dvoet, line 64: Is there any benefit to leaving this case class as it was and creating another case class that has the pattern (as an object) plus the 2 new attributes? This came to mind looking at the changes you have for SamResourceActionPatterns. You were forced to put in dummy values which are not pretty.

dvoet: resolved with faces, should move SamResourceActionPatterns to tests as it is not used in production code

dvoet:
this needs to be rebased to develop

src/main/resources/reference.conf
rtitle: Sorry I missed this earlier. Here's an attempt:

status = "view notebook cluster status details and configuration"
connect = "connect to the Jupyter notebook running on the notebook cluster"
sync = "sync files to/from the notebook cluster VMs"
delete = "delete the notebook cluster"
read_policies = "view all policies and policy details for the notebook cluster"
stop_start = "stop and start the notebook cluster VMs"

src/main/resources/reference.conf
```diff
@@ -1,106 +1,246 @@
...
+        description = "List all policies and policy details for this billing-project"
+      }
+      launch_batch_compute = {
+        description = "Launch workflows in this billing-project"
+      }
+      list_notebook_cluster = {
+        description = ""
+      }
+      launch_notebook_cluster = {
+        description = ""
```
rtitle: Launch a new notebook cluster in this billing-project

src/main/resources/reference.conf
```diff
@@ -1,106 +1,246 @@
...
+        description = "Create and delete policies for this billing-project"
+      }
+      read_policies = {
+        description = "List all policies and policy details for this billing-project"
+      }
+      launch_batch_compute = {
+        description = "Launch workflows in this billing-project"
+      }
+      list_notebook_cluster = {
+        description = ""
```
rtitle: List all notebook clusters in this billing-project

gpolumbo-broad:
This branch needed to be rebased off of develop.  I'm going to close this one, and instead, please see https://github.com/broadinstitute/sam/pull/151

PR #146, opened 2018-05-01
fixed list groups when email missing

dvoet:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
@@ -97,8 +97,10 @@ class ManagedGroupService(private val resourceService: ResourceService, private 
       emailLookup <- directoryDAO.batchLoadGroupEmail(ripns.map(ripn => WorkbenchGroupName(ripn.resourceId.value)))
     } yield {
       val emailLookupMap = emailLookup.toMap
-      ripns.map { ripn =>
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -97,8 +97,10 @@ class ManagedGroupService(private val resourceService: ResourceService, private
       emailLookup <- directoryDAO.batchLoadGroupEmail(ripns.map(ripn => WorkbenchGroupName(ripn.resourceId.value)))
     } yield {
       val emailLookupMap = emailLookup.toMap
-      ripns.map { ripn =>
-        ManagedGroupMembershipEntry(ripn.resourceId, ripn.accessPolicyName, emailLookupMap(WorkbenchGroupName(ripn.resourceId.value)))
+      // This will silently ignore any group where the email could not be loaded. This can happen when a
```
gpolumbo-broad, line 100: Are we sure we want to "silently ignore"?  Perhaps we should add just a simple log statement to let somebody know that the data is a little fouled up.

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
dvoet: I don't think this is a big deal nor do I think anything can really be done about it. I think the right thing to do is pretend that it does not exist.

gpolumbo-broad (approved):

gpolumbo-broad:
jenkins retest

PR #147, opened 2018-05-03
Resource validation refactor

gpolumbo-broad:
The goal of this refactor was:

- Get validations happening first, before anything is persisted
- Collect all the validation errors at once into a single `ErrorReport` so that all problems are reported to the requestor at once.
- Make the code eas(y)(ier) to add new validations to
- Reduce or eliminate need to execute some pieces of logic more than once

I'm not thrilled with the `PotentialAccessPolicy`.  The most important thing that it gave us is a place to wrap up the `Map[WorkbenchEmail, Option[WorkbenchSubject]]`, but it is also a package for bundling up into a single object the things we need to create a new `AccessPolicy`.   Anyway, it seems to help for now, so let me know what you think

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
@@ -100,6 +100,7 @@ ldap but it felt unnatural.
 case class AccessPolicy(id: ResourceAndPolicyName, members: Set[WorkbenchSubject], email: WorkbenchEmail, roles: Set[ResourceRoleName], actions: Set[ResourceAction]) extends WorkbenchGroup
 case class AccessPolicyMembership(memberEmails: Set[WorkbenchEmail], actions: Set[ResourceAction], roles: Set[ResourceRoleName])
 case class AccessPolicyResponseEntry(policyName: AccessPolicyName, policy: AccessPolicyMembership, email: WorkbenchEmail)
+case class AccessPolicyCreator(policyName: AccessPolicyName, emailsToSubjects: Map[WorkbenchEmail, Option[WorkbenchSubject]], roles: Set[ResourceRoleName], actions: Set[ResourceAction])
 
 case class BasicWorkbenchGroup(id: WorkbenchGroupName, members: Set[WorkbenchSubject], email: WorkbenchEmail) extends WorkbenchGroup
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
@@ -100,7 +100,6 @@ ldap but it felt unnatural.
 case class AccessPolicy(id: ResourceAndPolicyName, members: Set[WorkbenchSubject], email: WorkbenchEmail, roles: Set[ResourceRoleName], actions: Set[ResourceAction]) extends WorkbenchGroup
 case class AccessPolicyMembership(memberEmails: Set[WorkbenchEmail], actions: Set[ResourceAction], roles: Set[ResourceRoleName])
 case class AccessPolicyResponseEntry(policyName: AccessPolicyName, policy: AccessPolicyMembership, email: WorkbenchEmail)
-case class PotentialAccessPolicy(policyName: AccessPolicyName, emailsToSubjects: Map[WorkbenchEmail, Option[WorkbenchSubject]], roles: Set[ResourceRoleName], actions: Set[ResourceAction])
 
 case class BasicWorkbenchGroup(id: WorkbenchGroupName, members: Set[WorkbenchSubject], email: WorkbenchEmail) extends WorkbenchGroup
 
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -244,5 +299,18 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
     }
   }
 
+  private def makeCreatablePolicies(policies: Map[AccessPolicyName, AccessPolicyMembership]): Future[Set[PotentialAccessPolicy]] = {
+    val eventualPolicyCreators = policies.map {
+      case (accessPolicyName, accessPolicyMembership) => makeCreatablePolicy(accessPolicyName, accessPolicyMembership)
+    }
+    Future.sequence(eventualPolicyCreators.toSet)
```
dvoet: Future.traverse instead of map followed by Future.sequence

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -44,30 +44,53 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
-      }
-    } yield {
-      resource
-    }
+  private def validateCreateResource(resourceType: ResourceType, resourceId: ResourceId, policies: Set[PotentialAccessPolicy], userInfo: UserInfo): Option[Seq[ErrorReport]] = {
+    val ownerPolicyErrors = validateOwnerPolicyExists(resourceType, policies).toSeq
+    val policyErrors = policies.flatMap(policy => validatePolicy(resourceType, policy)).toSeq
 
+    val possibleErrors = ownerPolicyErrors ++ policyErrors
+    if (possibleErrors.nonEmpty) Some(possibleErrors) else None
```
dvoet: general best practice is to never directly create an instance of Some. User Option(possibleErrors) instead of Some(possibleErrors). This protects against accidental Some(null) situations.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -100,6 +100,7 @@ ldap but it felt unnatural.
 case class AccessPolicy(id: ResourceAndPolicyName, members: Set[WorkbenchSubject], email: WorkbenchEmail, roles: Set[ResourceRoleName], actions: Set[ResourceAction]) extends WorkbenchGroup
 case class AccessPolicyMembership(memberEmails: Set[WorkbenchEmail], actions: Set[ResourceAction], roles: Set[ResourceRoleName])
 case class AccessPolicyResponseEntry(policyName: AccessPolicyName, policy: AccessPolicyMembership, email: WorkbenchEmail)
+case class PotentialAccessPolicy(policyName: AccessPolicyName, emailsToSubjects: Map[WorkbenchEmail, Option[WorkbenchSubject]], roles: Set[ResourceRoleName], actions: Set[ResourceAction])
```
dvoet: this seems like something that should be private and internal to ResourceService

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -44,30 +44,53 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+    } yield resource
+  }
 
-      _ <- Future.traverse(policies) { case (accessPolicyName, accessPolicyMembership) =>
-        overwritePolicy(resourceType, accessPolicyName, resource, accessPolicyMembership)
-      }
-    } yield {
-      resource
-    }
+  private def validateCreateResource(resourceType: ResourceType, resourceId: ResourceId, policies: Set[PotentialAccessPolicy], userInfo: UserInfo): Option[Seq[ErrorReport]] = {
```
dvoet: could this return just a Seq instead of Option[Seq]? Note that `case Seq()` matches an empty seq.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -244,5 +299,18 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
     }
   }
 
+  private def makeCreatablePolicies(policies: Map[AccessPolicyName, AccessPolicyMembership]): Future[Set[PotentialAccessPolicy]] = {
```
dvoet: pick a single term.. is it creatable or potential? how about validatable?

dvoet (changes requested):
see comments

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -18,6 +18,9 @@ import scala.util.Success
   * Created by mbemis on 5/22/17.
   */
 class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceType], private val accessPolicyDAO: AccessPolicyDAO, private val directoryDAO: DirectoryDAO, private val cloudExtensions: CloudExtensions, private val emailDomain: String)(implicit val executionContext: ExecutionContext) extends LazyLogging {
+
+  case class PotentialAccessPolicy(policyName: AccessPolicyName, emailsToSubjects: Map[WorkbenchEmail, Option[WorkbenchSubject]], roles: Set[ResourceRoleName], actions: Set[ResourceAction])
```
dvoet: If you feel strongly about the term Potential, fine, but it doesn't really speak to me. Why would a potential policy be different that the real one? I think I prefer Validatable.

MatthewBemis (approved):

gpolumbo-broad:
@dvoet gave thumb in slack

PR #155, opened 2018-05-16
be smarter about cleaning up pet sa keys

dvoet:
remove keys only if service account is accessible
remove keys that don't exist in cache before issuing a new key

In all cases:


src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -528,11 +528,62 @@ class GoogleExtensionSpec(_system: ActorSystem) extends TestKit(_system) with Fl
     val firstKey = runAndWait(googleExtensions.getPetServiceAccountKey(defaultUser, googleProject))
 
     //remove the key we just created
-    runAndWait(googleExtensions.removePetServiceAccountKey(defaultUserId, googleProject, ServiceAccountKeyId(firstKey)))
```
dvoet, line 531: `firstKey` is not a key id so this was incorrect. It only worked because the mock iam dao ignored the key id.

gpolumbo-broad (approved):

PR #156, opened 2018-05-18
Stand alone

gpolumbo-broad:

In all cases:

### DIFF
--- a/docker/stand-alone/.env
+++ b/docker/stand-alone/.env
@@ -3,7 +3,7 @@
 
 # These values MUST be changed in order to be able to authenticate with SAM
 GOOGLE_OAUTH_CLIENT_ID=change_me
-GOOGLE_CLOUD_PROJECT_DOMAIN=change_me
+GOOGLE_CLOUD_PROJECT_ID=change_me
 
 # Variables below can optionally be modified as needed
 SAM_APP_IMAGE=broadinstitute/sam:stand-alone
--- a/docker/stand-alone/.env
+++ b/docker/stand-alone/.env
@@ -6,6 +6,7 @@ GOOGLE_OAUTH_CLIENT_ID=change_me
 GOOGLE_CLOUD_PROJECT_DOMAIN=change_me
 
 # Variables below can optionally be modified as needed
+SAM_APP_IMAGE=broadinstitute/sam:stand-alone
 SAM_HOST=localhost
 EMAIL_DOMAIN=example.foo
 HTTP_PORT=29800
--- a/docker/stand-alone/.env
+++ b/docker/stand-alone/.env
@@ -1,0 +1,12 @@
+# https://docs.docker.com/compose/env-file/
+# https://docs.docker.com/compose/environment-variables/#the-env-file
+
+# These values MUST be changed in order to be able to authenticate with SAM
+GOOGLE_OAUTH_CLIENT_ID=change_me
+GOOGLE_CLOUD_PROJECT_DOMAIN=change_me
+
+# Variables below can optionally be modified as needed
+SAM_HOST=localhost
+EMAIL_DOMAIN=example.foo
+HTTP_PORT=29800
+HTTPS_PORT=29443
--- a/docker/stand-alone/README.md
+++ b/docker/stand-alone/README.md
@@ -21,7 +21,7 @@ order to configure OAuth in SAM, you will need to:
  Google Project.  
     1. Add an "Authorized JavaScript origin" - `https://localhost:29443`
     1. Add an "Authorized redirect URI" - `https://localhost:29443/o2c.html`
-1. In `docker/stand-alone/.env`, set the values for `GOOGLE_OAUTH_CLIENT_ID` and `GOOGLE_CLOUD_PROJECT_DOMAIN`  
+1. In `docker/stand-alone/.env`, set the values for `GOOGLE_OAUTH_CLIENT_ID` and `GOOGLE_CLOUD_PROJECT_ID`  
 
 ### SSL Certificate
 
--- a/docker/stand-alone/README.md
+++ b/docker/stand-alone/README.md
@@ -80,5 +80,5 @@ If you would like to make code changes to SAM and test those changes, do the fol
 1. Run `sbt assembly` - This should build a `.jar` in `target/scala-2.12`
 1. Copy the `.jar` to SAM's root directory: `cp target/scala-2.12/sam-assembly-0.1-xxxxxxx-SNAPSHOT.jar .`
 1. [Re-build](https://docs.docker.com/engine/reference/commandline/build/) the `sam-app` Docker container by running 
-`docker build -t sam/my-tag`.  You can name the tag whatever you want by changing the argument passed to `-t`. 
+`docker build -t sam/my-tag .`  You can name the tag whatever you want by changing the argument passed to `-t`. 
 1. Update `.env` and set `SAM_APP_IMAGE=sam/my-tag` (or whatever name you chose for your tag)
--- a/docker/stand-alone/README.md
+++ b/docker/stand-alone/README.md
@@ -56,9 +56,6 @@ If you have `openssl` installed, you can generate a self-signed certificate usin
 You may then copy the generated `server.crt` file to `ca-bundle.crt` to use your certificate as the certificate chain.
 
 ### Starting the application
-1. Create an external docker network: 
-    
-    ```docker network create sam```
 
 1.  `cd` to the `docker/stand-alone` directory and run:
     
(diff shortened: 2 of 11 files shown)


docker/stand-alone/docker-compose.yaml
```diff
@@ -0,0 +1,34 @@
+version: '2'
+services:
+  sam-app:
+    image: broadinstitute/sam:stand-alone
```
gpolumbo-broad: Before merging, we will need to update this to use the correct image.  (Is sam:dev the right image?)

docker/stand-alone/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+1.  `cd` to the `docker/stand-alone` directory and run:
+    
+    ```docker-compose up```
+1. SAM should be running at: https://sam.example.localhost.org:29443/
+
+### Email Domain
+
+Specify the email domain for your SAM instance in `docker/stand-alone/sam.conf` by changing the 
+`emailDomain = "replace.me.with.your.domain"` field.
```
coverbeck: Shouldn't this step be before "Starting the application"?

Also, it's not clear to me what this value should be? Should it be `sam.example.localhost.org`? If so, could you not default to that? If not, can you provide an example of what it should be?

docker/stand-alone/README.md
```diff
@@ -0,0 +1,68 @@
...
+      googleClientId = "REPLACE_ME_WITH_YOUR_GOOGLE_OAUTH_CLIENT_ID"
+      realm = "REPLACE_ME_WITH_THE_NAME_OF_YOUR_GOOGLE_PROJECT"
+    }
+    ```
+
+### Update /etc/hosts
+
+Add the following to your `/etc/hosts` file:
+
+```127.0.0.1 sam.example.localhost.org```
```
coverbeck: Most people will probably already have `127.0.0.1` in their `/etc/hosts` file, or maybe I'm assuming based on my own hosts file. So I suggest an example with that as well or instead, because you probably don't want to kill your localhost mapping.

```
127.0.0.1 localhost sam.example.localhost.org
```

docker/stand-alone/docker-compose.yaml
coverbeck: Not crucial, although it would be helpful, and it may not even belong here, to give instructions on how to build and run SAM from local sources. I see in the main README.md that there is the `./docker/build.sh jar -d build`, but tying it all together explicitly on how I build SAM locally then run that local Docker image would be helpful.

I believe I will need to do my own PR to add a new resource type to SAM, so I'm anticipating that. :)

coverbeck:
After following the instructions, I got the SAM Swagger UI to come up, but saw two issues, both done via Swagger UI html page:

1. Invoking `/status` returns a 500, with the following response body:

```
{
  "ok": false,
  "systems": {
    "OpenDJ": {
      "ok": false,
      "messages": [
        "could not find group All_Users in opendj"
      ]
    }
  }
}
```

2. Trying to list resource types, `/api/resourceTypes`, returns a 401 with the following response body:

```<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>401 Unauthorized</title>
</head><body>
<h1>Unauthorized</h1>
<p>This server could not verify that you
are authorized to access the document
requested.  Either you supplied the wrong
credentials (e.g., bad password), or your
browser doesn't understand how to supply
the credentials required.</p>
<hr>
<address>Apache Server at sam.example.localhost.org Port 29443</address>
</body></html>
```

I was able to authorize in Swagger before doing that, the red exclamation point had changed to the blue `i`.

MatthewBemis:
@coverbeck you can fix the 401 by hitting the POST /register/user endpoint to get your account registered in sam. we should add a note about this at the end of the setup instructions.

coverbeck:
@MatthewBemis Thanks for the tip. Just tried the POST and a subsequent GET `/register/user` returns a 200 with my email in the response body, so it worked.

However, even after that, I continue to get a 401 on GET `/api/resourceTypes`.

Charles

docker/stand-alone/docker-compose.yaml
gpolumbo-broad: Will update with these instructions

gpolumbo-broad:
@coverbeck we're seeing the same 401 and are working to get it sorted.

MatthewBemis (approved):
👍

gpolumbo-broad:
@coverbeck I believe I have addressed your comments, of course please let us know if there's anything else that doesn't work or could be simplified.

- App now defaults to run on localhost
- All config file settings are now isolated to .env file
- Set more default values so that there's less that you MUST specify to get SAM running
- Instructions added for building custom SAM app and docker image
- Fixed the 401 errors

coverbeck:
Thanks @gpolumbo-broad !

I'll give it a shot and let you know.

docker/stand-alone/.env
```diff
@@ -0,0 +1,13 @@
+# https://docs.docker.com/compose/env-file/
+# https://docs.docker.com/compose/environment-variables/#the-env-file
+
+# These values MUST be changed in order to be able to authenticate with SAM
+GOOGLE_OAUTH_CLIENT_ID=change_me
+GOOGLE_CLOUD_PROJECT_DOMAIN=change_me
```
coverbeck: What is the `GOOGLE_CLOUD_PROJECT_DOMAIN`? Is it the project id in the GCP console? It doesn't seem like it, but I can't tell what it should be.

docker/stand-alone/.env
```diff
@@ -0,0 +1,13 @@
...
+
+# Variables below can optionally be modified as needed
+SAM_APP_IMAGE=broadinstitute/sam:stand-alone
+SAM_HOST=localhost
+EMAIL_DOMAIN=example.foo
```
coverbeck, line 11: What is this used for and what should the value be? Looks like it's optional so it may not matter, but it would be nice to understand.

docker/stand-alone/.env
coverbeck: I tried setting it to my the domain of my organization, ucsc.edu, and it seems to be working.

docker/stand-alone/README.md
```diff
@@ -0,0 +1,84 @@
...
+1. It will take a few minutes to complete the startup process.  When complete, SAM should be accessible at: 
+[https://localhost:29443/](https://localhost:29443/)
+
+## Using SAM
+
+1. Open the SAM Swagger UI at: [https://localhost:29443/](https://localhost:29443/)
+1. Click the `Authorize` button in the upper right of the page to log into SAM
+    1. In the window that pops up, check the 3 checkboxes (`openid`, `email`, and `profile`)
+    1. Click the `Authorize` button at the bottom of the pop-up window
+1. Register your user account in the **Users** section by expanding the `POST /register/user` operation and clicking the
```
coverbeck, line 61: When I did this, I got an error saying the user had already been registered, because I had run this previously against a previous version of the PR.

That said, although maybe this is getting too much into documenting Docker, maybe it would be worth mentioning how to clear the volumes, especially if a user wanted to start from scratch.

docker/stand-alone/.env
dvoet: we use email addresses as alternate identifiers, in production there is a gsuite domain to make them real email addresses. Here it doesn't matter.

docker/stand-alone/README.md
gpolumbo-broad: The easiest way to start over is to run `docker-compose down; docker-compose up`

gpolumbo-broad:
jenkins retest

docker/stand-alone/README.md
coverbeck: I think you need to do `docker-compose down -v` if you don't want users persisted between sessions. But maybe this is getting too much into details and most Docker users will know this already (I just learned about it fairly recently).

docker/stand-alone/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Building SAM
+
+If you would like to make code changes to SAM and test those changes, do the following:
+
+1. `cd` to the SAM root directory
+1. Run `sbt assembly` - This should build a `.jar` in `target/scala-2.12`
+1. Copy the `.jar` to SAM's root directory: `cp target/scala-2.12/sam-assembly-0.1-xxxxxxx-SNAPSHOT.jar .`
+1. [Re-build](https://docs.docker.com/engine/reference/commandline/build/) the `sam-app` Docker container by running 
+`docker build -t sam/my-tag`.  You can name the tag whatever you want by changing the argument passed to `-t`.
```
coverbeck: Should be `docker build -t sam/mytag .` (need a period inside the backtick).

gpolumbo-broad:
Once again, Jenkins retest please

gpolumbo-broad:
reported unit test failure:  https://fc-jenkins.dsp-techops.broadinstitute.org/job/sam-PR-fiab-test/272/

jenkins retest

PR #157, opened 2018-05-22
GAWB-3609: Random pet

MatthewBemis:

In all cases:


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -265,6 +265,45 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+  }
+
+  def getArbitraryPetServiceAccountToken(user: WorkbenchUser, scopes: Set[String]): Future[String] = {
+    getArbitraryPetServiceAccountKey(user).flatMap { key =>
+      getAccessTokenUsingJson(key, scopes)
+    }
+  }
+
+  private def getDefaultServiceAccountForShellProject(user: WorkbenchUser): Future[String] = {
+    val projectName = s"shell-${user.id.value}"
```
dvoet: I think we should have something firecloud specific... was thinking whatever our email domain is... how many characters do we have to play with?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -265,6 +265,45 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+    for {
+      creationOperationId <- googleBillingDAO.createProject(projectName)
+      _ <- pollShellProjectCreation(creationOperationId) //poll until it's created TODO: use the response instead of assuming it worked
+      _ <- googleBillingDAO.addPolicyBindings(projectName, Map("roles/viewer" -> List(s"user:${user.email.value}")))
```
dvoet: why add the user?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -265,6 +265,45 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+  def getArbitraryPetServiceAccountToken(user: WorkbenchUser, scopes: Set[String]): Future[String] = {
+    getArbitraryPetServiceAccountKey(user).flatMap { key =>
+      getAccessTokenUsingJson(key, scopes)
+    }
+  }
+
+  private def getDefaultServiceAccountForShellProject(user: WorkbenchUser): Future[String] = {
+    val projectName = s"shell-${user.id.value}"
+    for {
+      creationOperationId <- googleBillingDAO.createProject(projectName)
```
dvoet: should check if the project already exists and just use it if it does

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
MatthewBemis: subject ID is 21 characters so we have 9 characters to play with

could do something like `fc-$env-111111111111111111111`

`env` would have to be truncated to first 5 chars because staging would be too long in that case

automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
```diff
@@ -282,6 +282,80 @@ class SamApiSpec extends FreeSpec with BillingFixtures with Matchers with ScalaF
...
+
+      withCleanBillingProject(UserPool.chooseProjectOwner, List(user.email)) { project =>
+        // get my pet's email directly (this will also create the pet SA)
+        val petEmailDirect =  Sam.user.petServiceAccountEmail(project)(user.makeAuthToken)
+
+        // get my pet's email thru the arbitrary key endpoint
+        val petEmailArbitrary = getFieldFromJson(Sam.user.arbitraryPetServiceAccountKey()(user.makeAuthToken), "client_email")
+
+        // result should be the same
+        petEmailDirect shouldBe petEmailArbitrary
```
gpolumbo-broad: Are we able to assert this because there's only one SA in the pool?  Hard to tell from just this test, but my concern is that this test may not be covering the need for the SA to be arbitrary, it might be always returning the oldest one or something.  Maybe that's what we want, maybe it's not.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -265,6 +263,49 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
     }
   }
 
+  def getArbitraryPetServiceAccountKey(user: WorkbenchUser): Future[String] = {
+    for {
+      pets <- directoryDAO.getAllPetServiceAccountsForUser(user.id)
+      key <- if(pets.nonEmpty) getPetServiceAccountKey(user, pets.head.id.project) else getDefaultServiceAccountForShellProject(user)
```
gpolumbo-broad, line 270: Is there any reason why always returning `pets.head` could be problematic?  I suspect that this will usually return the same pet for the user, so could this cause any issues?  If this is OK, then perhaps it's a nit, but maybe we shouldn't call the method `getArbitrary...`, because it's not really arbitrary.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
gpolumbo-broad: Is this the first time we're using "shell" projects?  The idea being that we are just creating the shell so that we have some random project in which to create an SA in for the user?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -477,6 +477,43 @@ paths:
       tags:
         - Google
 
+  /api/google/user/petServiceAccount/key:
```
gpolumbo-broad: Versioning of routes

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
MatthewBemis: yes that's correct

automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
MatthewBemis: good point. i'll re-work this test and nest gpalloc calls so i can create another pet in a 2nd project.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
dvoet: 👍

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
dvoet: I don't think it is a problem. It is arbitrary from the caller's perspective. The fact that it may be predictable is an implementation detail.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -265,6 +263,49 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+    getArbitraryPetServiceAccountKey(user).flatMap { key =>
+      getAccessTokenUsingJson(key, scopes)
+    }
+  }
+
+  private def getDefaultServiceAccountForShellProject(user: WorkbenchUser): Future[String] = {
+    val projectName = s"shell-${user.id.value}"
+    for {
+      creationOperationId <- googleBillingDAO.createProject(projectName).map(opId => Option(opId)) recover {
+        case gjre: GoogleJsonResponseException if gjre.getDetails.getCode == StatusCodes.Conflict.intValue => None
```
dvoet: this call does not always fail if the project already exists. Sometimes it still creates a project and the operation fails.

MatthewBemis: that's insane

MatthewBemis:
config changes are merged, so jenkins retest

MatthewBemis:
jenkins retest because sam build failed

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -265,6 +264,52 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+        case t: WorkbenchException => throw t
+        case t: Exception => true
+        case _ => false
+      }
+    }
+
+    retryExponentially(whenCreating)(() => {
+      googleProjectDAO.pollOperation(operationId).map { operation =>
+        if(operation.getDone && Option(operation.getError).exists(_.getCode.intValue() == Code.ALREADY_EXISTS.value())) true
+        if(operation.getDone && Option(operation.getError).isEmpty) true
```
dvoet: need an else here

dvoet (approved):
fix else then 👍

MatthewBemis:
jenkins retest because the current run is going to fail since configs weren't merged yet

gpolumbo-broad (approved):

MatthewBemis:
jenkins retest because i am confused and want more data

MatthewBemis:
jenkins retest with new config changes

PR #158, opened 2018-05-22
added v1 to resources, config, google, user, groups interfaces

dvoet:
new api structure:
/api/config/v1/resourceTypes
/api/google/v1/...
/api/groups/v1/...
/api/resources/v1/...
/register/user/v1/...

note that I did not version admin end points or the status end point

The new *V1* specs are copies of the existing specs with V1 inserted. Any additions to the V1 apis should be tests in the V1 files. When we retire the unversioned apis we can delete the unversioned specs.

Swagger includes only versioned apis.

Last, there is a fair amount of whitespace changes so it is probably worth looking at file compares ignoring whitespace.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
@@ -25,51 +25,55 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
   val managedGroupService: ManagedGroupService
 
   def groupRoutes: server.Route = requireUserInfo { userInfo =>
-    pathPrefix("group" / Segment) { groupId =>
-      val managedGroup = Resource(ManagedGroupService.managedGroupTypeName, ResourceId(groupId))
-
-      pathEndOrSingleSlash {
-        get {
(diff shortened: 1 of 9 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -25,51 +25,55 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
...
-        }
-      } ~
-      pathPrefix("requestAccess") {
-        post {
-          handleRequestAccess(managedGroup, userInfo)
-        }
-      } ~
-      pathPrefix(Segment) { policyName =>
-        val accessPolicyName = ManagedGroupService.getPolicyName(policyName)
+    (pathPrefix("groups" / "v1") | pathPrefix("group")) {
```
gpolumbo-broad, line 28: Are the singular, non-versioned URIs deprecated?  If so, do we know how we want to go about marking them as such in the code, swagger, response payloads, etc.?

gpolumbo-broad (approved):

dvoet:
jenkins retest since google fixed project problems

MatthewBemis (approved):
:+1:

PR #159, opened 2018-05-22
GAWB-3569 Can create specify Auth Domains on Resource creation

gpolumbo-broad:

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -680,7 +680,7 @@ paths:
         204:
           description: Successfully created resource
         400:
-          description: Invalid policies, invalid auth domains, or resource type allows id reuse
+          description: Invalid policies, invalid auth domain, or resource type allows id reuse
           schema:
             $ref: '#/definitions/ErrorReport'
         409:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1126,7 +1126,7 @@ definitions:
         items:
           $ref: '#/definitions/AccessPolicyMembership'
         description: map of initial policies to create
-      authDomains:
+      authDomain:
         type: array
         items:
           type: string
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -662,7 +662,7 @@ paths:
         204:
           description: Successfully created resource
         400:
-          description: Invalid policies or resource type allows id reuse
+          description: Invalid policies, invalid auth domains, or resource type allows id reuse
           schema:
             $ref: '#/definitions/ErrorReport'
         409:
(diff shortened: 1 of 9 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1122,10 +1122,14 @@ definitions:
         type: string
         description: id of the resource to create
       policies:
-        type: object
-        additionalProperties:
+        type: array
+        items:
           $ref: '#/definitions/AccessPolicyMembership'
         description: map of initial policies to create
+      authDomains:
```
MatthewBemis: for here and elsewhere in the PR: authDomains should be singular. there's only ever 1 auth domain for a resource, which may be comprised of any number of managed groups

gpolumbo-broad:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -83,16 +81,45 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+      val errors = constrainableErrors ++ existenceErrors.flatten
+      if (errors.nonEmpty) {
+        Option(ErrorReport("Invalid Auth Domains specified", errors))
+      } else None
+    }
+  }
+
+  private def validateAuthDomainsExist(authDomains: Set[WorkbenchGroupName]): Future[Set[Option[ErrorReport]]] = {
+    Future.traverse(authDomains) { groupName =>
+      directoryDAO.loadGroup(groupName) map {
```
dvoet: this should check against managed groups... I think we should add a "use" action to managed groups that both members and admins have. Then this should check that the user has that action. If the user does not have that action then the error message should say that either the group does not exist or the user does not have access.

dvoet (changes requested):
see comments

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -83,16 +81,44 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+    validateAuthDomainPermissions(authDomain, userInfo).map { permissionsErrors =>
+      val constrainableErrors = validateAuthDomainConstraints(resourceType, authDomain).toSeq
+      val errors = constrainableErrors ++ permissionsErrors.flatten
+      if (errors.nonEmpty) {
+        Option(ErrorReport("Invalid Auth Domain specified", errors))
+      } else None
+    }
+  }
+
+  private def validateAuthDomainPermissions(authDomain: Set[WorkbenchGroupName], userInfo: UserInfo): Future[Set[Option[ErrorReport]]] = Future.sequence(
```
dvoet: User Future.traverse instead of sequence and map

dvoet (approved):

MatthewBemis (approved):

PR #167, opened 2018-06-11
Unboundid

dvoet:
implemented new version of directory and access policy daos using unboundid. I did not touch the schema dao, it still uses jndi. But I did turn off the jndi connection pool (schema dao is the only one to use it now and it just runs at the beginning)

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -1,16 +1,21 @@
 package org.broadinstitute.dsde.workbench.sam
 
 import java.io.File
+import java.net.URI
+import javax.net.SocketFactory
+import javax.net.ssl.SSLContext
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.stream.ActorMaterializer
 import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
+import com.unboundid.ldap.sdk.{LDAPConnection, LDAPConnectionPool}
+import com.unboundid.util.ssl.{SSLUtil, TrustStoreTrustManager}
 import net.ceedubs.ficus.Ficus._
 import org.broadinstitute.dsde.workbench.dataaccess.PubSubNotificationDAO
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.Pem
-import org.broadinstitute.dsde.workbench.google.{HttpGoogleProjectDAO, HttpGoogleDirectoryDAO, HttpGoogleIamDAO, HttpGooglePubSubDAO, HttpGoogleStorageDAO}
+import org.broadinstitute.dsde.workbench.google.{HttpGoogleDirectoryDAO, HttpGoogleIamDAO, HttpGoogleProjectDAO, HttpGooglePubSubDAO, HttpGoogleStorageDAO}
 import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchException}
 import org.broadinstitute.dsde.workbench.sam.api.{SamRoutes, StandardUserInfoDirectives}
 import org.broadinstitute.dsde.workbench.sam.config._
@@ -40,8 +45,16 @@ object Boot extends App with LazyLogging {
     implicit val materializer = ActorMaterializer()
     import scala.concurrent.ExecutionContext.Implicits.global
 
-    val accessPolicyDAO = new JndiAccessPolicyDAO(directoryConfig)
-    val directoryDAO = new JndiDirectoryDAO(directoryConfig)
+    val dirURI = new URI(directoryConfig.directoryUrl)
+    val socketFactory = dirURI.getScheme.toLowerCase match {
+      case "ldaps" => SSLContext.getDefault.getSocketFactory
+      case "ldap" => SocketFactory.getDefault
+      case unsupported => throw new WorkbenchException(s"unsupported directory url scheme: $unsupported")
+    }
+    val ldapConnectionPool = new LDAPConnectionPool(new LDAPConnection(socketFactory, dirURI.getHost, dirURI.getPort, directoryConfig.user, directoryConfig.password), directoryConfig.connectionPoolSize)
+
+    val accessPolicyDAO = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig)
+    val directoryDAO = new LdapDirectoryDAO(ldapConnectionPool, directoryConfig)
     val schemaDAO = new JndiSchemaDAO(directoryConfig, schemaLockConfig)
 
     val resourceTypes = config.as[Map[String, ResourceType]]("resourceTypes").values.toSet
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -46,12 +46,13 @@ object Boot extends App with LazyLogging {
     import scala.concurrent.ExecutionContext.Implicits.global
 
     val dirURI = new URI(directoryConfig.directoryUrl)
-    val socketFactory = dirURI.getScheme.toLowerCase match {
-      case "ldaps" => SSLContext.getDefault.getSocketFactory
-      case "ldap" => SocketFactory.getDefault
+    val (socketFactory, defaultPort) = dirURI.getScheme.toLowerCase match {
+      case "ldaps" => (SSLContext.getDefault.getSocketFactory, 636)
+      case "ldap" => (SocketFactory.getDefault, 389)
       case unsupported => throw new WorkbenchException(s"unsupported directory url scheme: $unsupported")
     }
-    val ldapConnectionPool = new LDAPConnectionPool(new LDAPConnection(socketFactory, dirURI.getHost, dirURI.getPort, directoryConfig.user, directoryConfig.password), directoryConfig.connectionPoolSize)
+    val port = if (dirURI.getPort > 0) dirURI.getPort else defaultPort
+    val ldapConnectionPool = new LDAPConnectionPool(new LDAPConnection(socketFactory, dirURI.getHost, port, directoryConfig.user, directoryConfig.password), directoryConfig.connectionPoolSize)
 
     val accessPolicyDAO = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig)
     val directoryDAO = new LdapDirectoryDAO(ldapConnectionPool, directoryConfig)
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/package.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/package.scala
@@ -88,7 +88,8 @@ package object config {
       config.getString("user"),
       config.getString("password"),
       config.getString("baseDn"),
-      config.getString("enabledUsersGroupDn")
+      config.getString("enabledUsersGroupDn"),
+      config.as[Option[Int]]("connectionPoolSize").getOrElse(20)
     )
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectorySubjectNameSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectorySubjectNameSupport.scala
@@ -1,17 +1,21 @@
 package org.broadinstitute.dsde.workbench.sam.directory
 
+import java.text.SimpleDateFormat
+import java.util.Date
+
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.sam._
 import org.broadinstitute.dsde.workbench.sam.config.DirectoryConfig
 import org.broadinstitute.dsde.workbench.sam.model._
 import org.broadinstitute.dsde.workbench.sam.schema.JndiSchemaDAO.Attr
-import org.broadinstitute.dsde.workbench.sam.util.JndiSupport
+
+import scala.util.matching.Regex
 
 /**
   * Created by dvoet on 6/6/17.
   */
-trait DirectorySubjectNameSupport extends JndiSupport {
+trait DirectorySubjectNameSupport {
   protected val directoryConfig: DirectoryConfig
   val peopleOu = s"ou=people,${directoryConfig.baseDn}"
   val groupsOu = s"ou=groups,${directoryConfig.baseDn}"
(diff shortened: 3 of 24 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectorySubjectNameSupport.scala
```diff
@@ -1,17 +1,21 @@
...
 import org.broadinstitute.dsde.workbench.sam.model._
 import org.broadinstitute.dsde.workbench.sam.schema.JndiSchemaDAO.Attr
-import org.broadinstitute.dsde.workbench.sam.util.JndiSupport
+
+import scala.util.matching.Regex
 
 /**
   * Created by dvoet on 6/6/17.
   */
-trait DirectorySubjectNameSupport extends JndiSupport {
```
dvoet, line 14: the additions to this trait moved from JndiSupport

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -40,8 +45,17 @@ object Boot extends App with LazyLogging {
     implicit val materializer = ActorMaterializer()
     import scala.concurrent.ExecutionContext.Implicits.global
 
-    val accessPolicyDAO = new JndiAccessPolicyDAO(directoryConfig)
-    val directoryDAO = new JndiDirectoryDAO(directoryConfig)
+    val dirURI = new URI(directoryConfig.directoryUrl)
```
dvoet, line 48: jndi supports a ldap url which is what we have in conifg. Unboundid does not so here is a little tap dance to figure it out.

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/package.scala
```diff
@@ -88,7 +88,8 @@ package object config {
       config.getString("user"),
       config.getString("password"),
       config.getString("baseDn"),
-      config.getString("enabledUsersGroupDn")
+      config.getString("enabledUsersGroupDn"),
+      config.as[Option[Int]]("connectionPoolSize").getOrElse(20)
```
dvoet, line 92: default value for backwards compatibility

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -40,8 +45,17 @@ object Boot extends App with LazyLogging {
...
+    val (socketFactory, defaultPort) = dirURI.getScheme.toLowerCase match {
+      case "ldaps" => (SSLContext.getDefault.getSocketFactory, 636)
+      case "ldap" => (SocketFactory.getDefault, 389)
+      case unsupported => throw new WorkbenchException(s"unsupported directory url scheme: $unsupported")
+    }
+    val port = if (dirURI.getPort > 0) dirURI.getPort else defaultPort
+    val ldapConnectionPool = new LDAPConnectionPool(new LDAPConnection(socketFactory, dirURI.getHost, port, directoryConfig.user, directoryConfig.password), directoryConfig.connectionPoolSize)
```
gpolumbo-broad, line 55: Maybe we don't want to do anything now, but going off of @MatthewBemis's observation last week about setting the TTL on the connections, we might consider throwing in a call to https://docs.ldap.com/ldap-sdk/docs/javadoc/com/unboundid/ldap/sdk/LDAPConnectionPool.html#setMaxConnectionAgeMillis() at some point.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectorySubjectNameSupport.scala
```diff
@@ -40,6 +44,20 @@ trait DirectorySubjectNameSupport extends JndiSupport {
...
+    * matchAttributeNames would be Seq("policy", "resourceId") and baseDn would be "resourceType=splat,ou=resources,dc=x,dc=u,dc=com".
+    *
+    * @param matchAttributeNames names of attributes in the leading part of the dn that should match and extract values
+    * @param baseDn the trailing part of the dn that should match but we don't care to extract values
+    * @return pattern with capture groups for each member of matchAttributeNames
+    */
+  protected def dnMatcher(matchAttributeNames: Seq[String], baseDn: String): Regex = {
+    val partStrings = matchAttributeNames.map { attrName => s"$attrName=([^,]+)" }
+    partStrings.mkString("(?i)", ",", s",$baseDn").r
+  }
```
gpolumbo-broad, line 59: Are there tests for this?

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectorySubjectNameSupport.scala
```diff
@@ -62,4 +80,7 @@ trait DirectorySubjectNameSupport extends JndiSupport {
       case _ => throw new WorkbenchException(s"not a group dn [$dn]")
     }
   }
+
+  protected def formattedDate(date: Date) = new SimpleDateFormat("yyyyMMddHHmmss.SSSZ").format(date)
+  protected def parseDate(date: String) = new SimpleDateFormat("yyyyMMddHHmmss.SSSZ").parse(date)
```
gpolumbo-broad: Nit, but extract string.  Maybe add just a couple of basic tests.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -64,16 +64,23 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
...
   }
 
-  override def removeGroupMember(groupId: WorkbenchGroupIdentity, removeMember: WorkbenchSubject): Future[Unit] = withContext { ctx =>
+  override def removeGroupMember(groupId: WorkbenchGroupIdentity, removeMember: WorkbenchSubject): Future[Boolean] = withContext { ctx =>
     ctx.modifyAttributes(groupDn(groupId), DirContext.REMOVE_ATTRIBUTE, new BasicAttributes(Attr.uniqueMember, subjectDn(removeMember)))
     updateUpdatedDate(groupId, ctx)
+    true
   } recover {
-    case _: NoSuchAttributeException => ()  // don't fail if the member is already not in the group
+    case _: NoSuchAttributeException => false
```
gpolumbo-broad, line 72: Is this what we want?  If I am trying to remove `User1` from group `foo` and the user is already not present in the group, maybe that should be OK and just return `true`.  In other words, I'm interpreting the method contract as:  "When I return, the subject specified will no longer be a member of the group and I will return True.  I will throw an exception if the group doesn't exist or some other error occurs while trying to remove the user from the group."  

Alternatively, instead of just returning `false`, should we return an `ErrorReport`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
```diff
@@ -64,16 +64,23 @@ class JndiDirectoryDAO(protected val directoryConfig: DirectoryConfig)(implicit
...
-    updateUpdatedDate(groupId, ctx)
+  override def addGroupMember(groupId: WorkbenchGroupIdentity, addMember: WorkbenchSubject): Future[Boolean] = withContext { ctx =>
+    Try {
+      ctx.modifyAttributes(groupDn(groupId), DirContext.ADD_ATTRIBUTE, new BasicAttributes(Attr.uniqueMember, subjectDn(addMember)))
+      updateUpdatedDate(groupId, ctx)
+    } match {
+      case Success(_) => true
+      case Failure(_: AttributeInUseException) => false
+      case Failure(regrets) => throw regrets
+    }
```
gpolumbo-broad, line 83: This method uses a `Try` block, whereas `removeGroupMember` uses just a `withContext{...}recover{...}`.  Is there a reason why the alternative approaches are used?  Since these methods are a matching pair, I think it makes sense for them to be consistent in implementation.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -286,17 +285,13 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
   }
 
   def addSubjectToPolicy(resourceAndPolicyName: ResourceAndPolicyName, subject: WorkbenchSubject): Future[Unit] = {
-    directoryDAO.addGroupMember(resourceAndPolicyName, subject) recover {
-      case _: AttributeInUseException => // subject is already there
-    } andThen {
+    directoryDAO.addGroupMember(resourceAndPolicyName, subject).map(_ => ()) andThen {
       case Success(_) => fireGroupUpdateNotification(resourceAndPolicyName)
```
gpolumbo-broad, line 289: It is possible that `addGroupMember` returns `Success(false)`, in which case, we probably don't need to call `fireGroupUpdateNotification` because nothing changed.  Furthermore, the fact that the user wasn't added, is not bubbled up in any way here, which means `addSubjectToPolicy` did not successfully perform its task and that's not reported out in any way.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -286,17 +285,13 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
     }
   }
 
   def removeSubjectFromPolicy(resourceAndPolicyName: ResourceAndPolicyName, subject: WorkbenchSubject): Future[Unit] = {
-    directoryDAO.removeGroupMember(resourceAndPolicyName, subject) recover {
-      case _: NoSuchAttributeException => // subject already gone
-    } andThen {
+    directoryDAO.removeGroupMember(resourceAndPolicyName, subject).map(_ => ()) andThen {
       case Success(_) => fireGroupUpdateNotification(resourceAndPolicyName)
```
gpolumbo-broad, line 295: See comment for `addSubjectToPolicy` above.   I'm not sure we want to treat `Success(true)` and `Success(false)` the same way.

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/JndiSupport.scala
```diff
@@ -58,20 +58,6 @@ trait JndiSupport {
     t.get
   }
```
gpolumbo-broad, line 60: I think we should we mark this method and/or class as `deprecated` and identify that it is still permitted to be used by the `JndiSchemaDAO` only.  

Do we have a longer term task to refactor the `JndiSchemaDAO` to use unboundid?  Or are we just going to leave it using JNDI?

src/test/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAOSpec.scala
```diff
@@ -48,6 +53,11 @@ class JndiDirectoryDAOSpec extends FlatSpec with Matchers with TestSupport with
       runAndWait(dao.createGroup(group))
     }
 
+    val conflict = intercept[WorkbenchExceptionWithErrorReport] {
```
gpolumbo-broad, line 56: The name of the spec is `"JndiGroupDirectoryDAO" should "create...` should be renamed to `LdapGroupDirectoryDAO`

gpolumbo-broad (changes requested):
I did not go over `LdapDirectoryDAO` and `LdapPolicyDAO` in terribly close detail.  I scanned them quickly, but otherwise am counting on the tests and the coverage results from Coveralls to verify these classes.  @dvoet if you want me to go over them in closer detail, just let me know.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
dvoet: look in the DirectoryDAO trait for the docs of how this should work. In general though, if at the end of the call the user is not in the group the function is successful. The boolean output signifies whether or not the user as actually removed during the course of the function call. This is a common pattern (https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#remove(java.lang.Object)).

It is not an error so no ErrorReport.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/JndiDirectoryDAO.scala
dvoet: this class is on the chopping block anyway, the change was only made because it made sense to change the return type

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: I am not ready to do anything yet... let's keep with current behavior and see how it goes

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectorySubjectNameSupport.scala
dvoet: this code was just moved, not changed. I don't know if there are specific tests but it is tested.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: at the end of this function the subject is in the policy which is the desired result so everything is ok. I think you misunderstand what Success(false) means, it is not an error, it means the user was already there.

Technically we don't need to fireGroupUpdateNotification on Success(false) but I think that is needless complexity

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: we do, none of this is new functionality

MatthewBemis (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAOSpec.scala
dvoet: fixed

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/JndiSupport.scala
dvoet: deprecated

gpolumbo-broad (approved):

dvoet:
jenkins retest with leo fix merged

dvoet:
jenkins retest

PR #170, opened 2018-06-14
WIP: Dockstore Resource Type

coverbeck:
Add a resource type for Dockstore resources.

I went with the name `tool`, because of Dockstore implements the Tool Registry Service (TRS).  Dockstore has the concept of both tools and workflows, but just like with TRS we can distinguish the two by path, prepending `#workflow/` to the path for workflows.

In other words, the resource id for the workflow `dockstore.org/myworkflow` would be, with the escaped characters,`%23workflow%2Fdockstore.org%2Fmyworkflow`, and the resource id for the tool `dockstore.org/mytool`would be `dockstore.org%2Fmytool`.

We currently only plan to support sharing for workflows in the first iteration, but we may support tools later.

I made this a WIP; looking for feedback, with these questions in particular:

1. Is the name `tool` too generic? Do we want to call it `dockstore-tool`, `dockstore-workflow` or something like that?
2. Should we have separate resource types for Dockstore tools and workflows, or put both Dockstore types under the single SAM resource type, like I currently have it?
3. How do the roles and actions I've defined seem? I've kept it fairly simple, but I don't know if I'm missing anything.

- [  ] I've updated the RC_XXX release ticket with any manual steps required to release this change

In all cases:

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -264,10 +264,10 @@ resourceTypes = {
         description = "delete a tool"
       },
       "read_policies" = {
-        description = "read policies of tool"
+        description = "read tool policies"
       },
       "alter_policies" = {
-        description = "alter policies of tool"
+        description = "alter tool policies"
       },
       "write" = {
         description = "modify tool"
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -258,4 +258,37 @@ resourceTypes = {
     }
     reuseIds = false
   }
+  tool = {
+    actionPatternObjects = {
+      "delete" = {
+        description = "delete a tool"
+      },
+      "read_policies" = {
+        description = "read policies of tool"
+      },
+      "alter_policies" = {
+        description = "alter policies of tool"
+      },
+      "write" = {
+        description = "modify tool"
+      },
+      "read" = {
+        description = "read tool"
+      }
+    }
+    ownerRoleName = "owner"
+    roles = {
+      owner = {
+        roleActions = ["delete", "read_policies", "alter_policies", "write", "read"]
+      },
+      writer = {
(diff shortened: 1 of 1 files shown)


dvoet (approved):

dvoet:
ok to test

geoffjentry:
I would suggest avoiding the word `tool`, it's too heavily overloaded ... to the extent that the relevant GA4GH group has started to avoid it altogether (except where it already existed due to history, e.g. Dockstore)

coverbeck:
@geoffjentry What do you propose? This is all I can think of:

* workflow
* dockstore-tool
* dockstore-workflow
* method (as in the FireCloud methods repository)
* ???

Thanks,

Charles

geoffjentry:
@coverbeck Is there an expectation that this is always going to refer to full workflows? If so I'd use that. If you're also expecting to reference individual tasks I think dockstore-tool is the best of those. Since you added it here, I'd also strongly suggest avoiding method in all contexts.

denis-yuen:
Will likely refer to both full workflows and individual tasks long term. 
I'm ok with most of the names above except for method which would be confusing since it doesn't appear in TRS or Dockstore. 

(We use `Entry` as a base class for tools and workflows, so possibly `dockstore-entry`?)

coverbeck:
OK, I will go with `dockstore-tool`, since we do plan to support individual tasks/tools. Will update PR.

coverbeck:
@dvoet I see the Jenkins Automation Tests failed, but that server looks to be behind a firewall, so I can't see what's going on.

Do I need to worry about that? Are there tests that I can run locally to reproduce. I doubt I broke anything with just a config change (famous last words!), but maybe there are tests against the number of resource types or something like that?

PR #171, opened 2018-06-14
isGroupMember shouldn't care about case of full dn

MatthewBemis:
with this change, isGroupMember will only look at the actual group name instead of the full dn, which can have differing cases due to inconsistencies in past code/migrations.

specific bug described:

dev broke because we'd be looking for membership in `cn=All_Users,ou=groups,dc=dsde-dev,dc=broadinstitute,dc=org`, but the user was in `cn=All_Users,ou=Groups,dc=dsde-dev,dc=broadinstitute,dc=org`. doing string comparison, they're not the same, thus user was seen as not fully activated.

In all cases:


dvoet:
jenkins retest

PR #185, opened 2018-08-03
Gawb 3708 policy email

marctalbott:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -357,10 +357,10 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
     val groupEmail = getSynchronizedEmail(groupId)
 
     for {
-      date <- groupDate
-      email <- groupEmail
+      dateOpt <- groupDate
+      emailOpt <- groupEmail
     } yield {
-      (date, email) match {
+      (dateOpt, emailOpt) match {
         case (Some(date), Some(email)) => Option(GroupSyncResponse(date.toString, email))
         case _ => None
       }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -6,7 +6,6 @@ import java.util.Date
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
-import akka.japi
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.auth.oauth2.ServiceAccountCredentials
 import com.typesafe.scalalogging.LazyLogging
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -352,8 +352,11 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
     } yield ()
   }
 
-  def getSynchronizedState(groupId: WorkbenchGroupIdentity): (Future[Option[Date]], Future[Option[WorkbenchEmail]]) = {
-    (getSynchronizedDate(groupId), getSynchronizedEmail(groupId))
+  def getSynchronizedState(groupId: WorkbenchGroupIdentity): Future[GroupSyncResponse] = {
+    for {
+      date <- getSynchronizedDate(groupId)
+      email <- getSynchronizedEmail(groupId)
+    } yield GroupSyncResponse(date, email)
   }
 
   def getSynchronizedDate(groupId: WorkbenchGroupIdentity): Future[Option[Date]] = {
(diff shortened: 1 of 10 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/sam/directory/MockDirectoryDAO.scala
```diff
@@ -214,6 +214,9 @@ class MockDirectoryDAO(private val groups: mutable.Map[WorkbenchGroupIdentity, W
     Future.successful(groupSynchronzedDates.get(groupId))
   }
 
+  override def getSynchronizedEmail(groupId: WorkbenchGroupIdentity): Future[Option[WorkbenchEmail]] = {
+    Future.successful(for (group <- groups.get(groupId)) yield group.email)
```
gpolumbo-broad: One liner `for` comprehension can be changed to a `.map` https://gist.github.com/loicdescotte/4044169

Should make it slightly more readable imho

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -352,10 +352,29 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+  def getSynchronizedState(groupId: WorkbenchGroupIdentity): Future[Option[GroupSyncResponse]] = {
+    val groupDate = getSynchronizedDate(groupId)
+    val groupEmail = getSynchronizedEmail(groupId)
+
+    for {
+      date <- groupDate
+      email <- groupEmail
+    } yield {
+      if(email.isDefined || date.isDefined)
+        Option(GroupSyncResponse(date.getOrElse("").toString, email.getOrElse(WorkbenchEmail(""))))
```
MatthewBemis: i think in the case of `date` or `email` being `None`, it would be better to omit them by returning `None` instead of an empty string. that way the services that consume this response don't have to have any extra logic to determine whether or not the field is missing.

you could also do this using pattern matching. something like:

```
(dateOpt, emailOpt) match {
  case (None, None) => None
  case (Some(date), email) => Option(GroupSyncResponse(Option(date.toString), email))
  case (None, email) => Option(GroupSyncResponse(None, email))
}
```

MatthewBemis (changes requested):
we'll also need to update swagger. i can show you how/where to do that

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
dvoet: @MatthewBemis's comment got me looking a little more deeply. Changing the end point in this way is a subtle break in backwards compatibility. A response of None in the old way meant that the group was never synced. But in the new way it means that there is neither a sync data nor an email. But I can't think of a good way to introduce email here which makes me think this is the wrong place.

I see it is already in the list access policies call, how about also adding it to the load access policy (singular) call too?

MatthewBemis: the thought here was to contain sync date and policy email to one endpoint that would live under the google path because they are google-related attributes. i originally thought about having it put in the endpoint that you're referring to but it felt slightly out of place to return google sync date in that endpoint.

another alternative i would propose is keeping things mostly the way they are in this PR except changing it to return None in the event of a missing sync date _or_ a missing email. that should ensure we keep backwards compatibility.

code would look something like

```
(dateOpt, emailOpt) match {
  case (Some(date), Some(email) => Option(GroupSyncResponse(date.toString, email))
  case _ => None
}
```

because we create the email attribute at policy creation, in practice we should never have a situation where the email is missing so really it would just be a missing sync date that dictates whether or not we return None, which is equivalent to what it does today.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -352,10 +352,29 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
     } yield ()
   }
 
+  def getSynchronizedState(groupId: WorkbenchGroupIdentity): Future[Option[GroupSyncResponse]] = {
```
gpolumbo-broad, line 355: There should be unit tests for this method.

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -352,10 +352,29 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+    val groupDate = getSynchronizedDate(groupId)
+    val groupEmail = getSynchronizedEmail(groupId)
+
+    for {
+      date <- groupDate
+      email <- groupEmail
```
MatthewBemis: nit: for the sake of readability, i'd rename these to `dateOpt` and `emailOpt`

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -348,6 +348,45 @@ class GoogleExtensionSpec(_system: ActorSystem) extends TestKit(_system) with Fl
     }
   }
 
+  it should "get a group's email" in {
```
MatthewBemis: it's not clear to me what the purpose of this unit test is or how it provides value over the test on line 377

dvoet (approved):

gpolumbo-broad (approved):

PR #189, opened 2018-08-08
GAWB 3720 v2 user endpoints

marctalbott:
Create v2 of userStatus interface which separates diagnostic info from normal status info. Help cut down on expensive Google queries which are often unneeded. 

In all cases:

- [x] Get two thumbsworth of review and PO signoff if necessary
- [x] Verify all tests go green
- [x] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1083,6 +1083,54 @@ paths:
       tags:
         - Users
 
+  /register/user/v2/self:
+    post:
+      summary: create current user in the system using login credentials
+      responses:
+        201:
+          description: 'user successfully created'
+          schema:
+            $ref: '#/definitions/UserStatus'
+        409:
+          description: user already exists
+          schema:
+            $ref: '#/definitions/ErrorReport'
+      operationId: createUserV2
+      tags:
+        - Users
+
+  /register/user/v2/self/info:
+    get:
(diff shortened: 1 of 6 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1392,3 +1440,32 @@ definitions:
       enabled:
         $ref: '#/definitions/Enabled'
         description: the status of the user's account
+
+  UserStatusInfo:
+    description: ''
+    required:
+      - userInfo
```
dvoet: I think we should flatten this

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -27,6 +27,10 @@ object SamJsonSupport {
 
   implicit val UserStatusFormat = jsonFormat2(UserStatus)
 
+  implicit val UserStatusInfoFormat = jsonFormat3(UserStatusInfo)
+
+  implicit val userStatusDiagnosticsFormat = jsonFormat3(UserStatusDiagnostics)
```
dvoet: capital U

dvoet (approved):

MatthewBemis (approved):

PR #190, opened 2018-08-14
add googleSubjectId

Qi77Qi:
This should be merged after https://github.com/broadinstitute/sam/pull/198

Things changed:
* There's some refactoring in involved spec files. With the change, https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala#L15 will be used in tests instead of a mocking object. 
* https://broadinstitute.atlassian.net/wiki/spaces/~952052303/pages/622624856/sam+Support+invite+feature 

TODO: 
- [x] add more unit tests for new code
- [x] once https://github.com/broadinstitute/workbench-libs/pull/157 is released, update sha in dependency
- [x] clean up generators in tests if people are okay with adding scalacheck
- [x] remove CreateWorkbenchUserAPI from workbench-lib

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
@@ -1,46 +1,76 @@
-package org.broadinstitute.dsde.workbench.sam.api
+package org.broadinstitute.dsde.workbench.sam
+package api
 
+import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
-import akka.http.scaladsl.server._
 import akka.http.scaladsl.server.Directives.{headerValueByName, onSuccess}
-import org.broadinstitute.dsde.workbench.model._
-import akka.http.scaladsl.server.Directives.headerValueByName
+import akka.http.scaladsl.server._
+import akka.http.scaladsl.server.directives.OnSuccessMagnet._
 import org.broadinstitute.dsde.workbench.model.google.ServiceAccountSubjectId
-import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchUserId}
+import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, _}
+import org.broadinstitute.dsde.workbench.sam.api.StandardUserInfoDirectives._
+import org.broadinstitute.dsde.workbench.sam.directory.DirectoryDAO
+import org.broadinstitute.dsde.workbench.sam.service.UserService._
 
 import scala.concurrent.{ExecutionContext, Future}
-
+import scala.util.Try
 
 
 trait StandardUserInfoDirectives extends UserInfoDirectives {
   implicit val executionContext: ExecutionContext
 
-  val petSAdomain = "\\S+@\\S+\\.iam\\.gserviceaccount\\.com".r
-
-  private def isPetSA(email: WorkbenchEmail) = {
-    petSAdomain.pattern.matcher(email.value).matches
-  }
-
   def requireUserInfo: Directive1[UserInfo] = (
-    headerValueByName("OIDC_access_token") &
-      headerValueByName("OIDC_CLAIM_user_id") &
-      headerValueByName("OIDC_CLAIM_expires_in") &
-      headerValueByName("OIDC_CLAIM_email")
+    headerValueByName(accessTokenHeader) &
+      headerValueByName(googleSubjectIdHeader) &
+      headerValueByName(expiresInHeader) &
+      headerValueByName(emailHeader)
     ) tflatMap {
-    case (token, userId, expiresIn, email) => {
-      val userInfo = UserInfo(OAuth2BearerToken(token), WorkbenchUserId(userId), WorkbenchEmail(email), expiresIn.toLong)
-      onSuccess(getUserFromPetServiceAccount(userInfo).map {
-        case Some(petOwnerUser) => UserInfo(OAuth2BearerToken(token), petOwnerUser.id, petOwnerUser.email, expiresIn.toLong)
-        case None => userInfo
-      })
-    }
+    case (token, googleSubjectId, expiresIn, email) =>
+      onSuccess(Try(expiresIn.toLong).fold[Future[UserInfo]](
+        t => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"expiresIn $expiresIn can't be converted to Long because  of $t"))),
+        l => getUserInfo(OAuth2BearerToken(token), GoogleSubjectId(googleSubjectId), WorkbenchEmail(email), l, directoryDAO)))
   }
 
-  private def getUserFromPetServiceAccount(userInfo: UserInfo):Future[Option[WorkbenchUser]] = {
-    if (isPetSA(userInfo.userEmail)) {
-      directoryDAO.getUserFromPetServiceAccount(ServiceAccountSubjectId(userInfo.userId.value))
-    } else {
-      Future.successful(None)
+  def requireCreateUser: Directive1[CreateWorkbenchUser] = (
+      headerValueByName(googleSubjectIdHeader) &
+      headerValueByName(emailHeader)
+    ) tflatMap {
+    case (googleSubjectId, email) => {
+      onSuccess(Future.successful(
+        CreateWorkbenchUser(genWorkbenchUserId(System.currentTimeMillis()), GoogleSubjectId(googleSubjectId), WorkbenchEmail(email))))
     }
   }
 }
+
+object StandardUserInfoDirectives{
+  val SAdomain = "\\S+@\\S+\\.iam\\.gserviceaccount\\.com".r
+  val accessTokenHeader = "OIDC_access_token"
+  val expiresInHeader = "OIDC_CLAIM_expires_in"
+  val emailHeader = "OIDC_CLAIM_email"
+  val googleSubjectIdHeader = "OIDC_CLAIM_user_id"
+
+  def isServiceAccount(email: WorkbenchEmail) = {
+    SAdomain.pattern.matcher(email.value).matches
+  }
+
+  def getUserInfo(token: OAuth2BearerToken, googleSubjectId: GoogleSubjectId, email: WorkbenchEmail, expiresIn: Long, directoryDAO: DirectoryDAO)(implicit ec: ExecutionContext): Future[UserInfo] = if (isServiceAccount(email)) {
+    // If it's a PET account, we treat it as its owner
+    directoryDAO.getUserFromPetServiceAccount(ServiceAccountSubjectId(googleSubjectId.value)).flatMap{
+      case Some(pet) => Future.successful(UserInfo(token, pet.id, pet.email, expiresIn.toLong))
+      case None => lookUpByGoogleSubjectId(googleSubjectId, directoryDAO).map(uid => UserInfo(token, uid, email, expiresIn))
+    }
+  } else {
+    lookUpByGoogleSubjectId(googleSubjectId, directoryDAO).map(uid => UserInfo(token, uid, email, expiresIn))
+  }
+
+  private def lookUpByGoogleSubjectId(googleSubjectId: GoogleSubjectId, directoryDAO: DirectoryDAO)(implicit ec: ExecutionContext): Future[WorkbenchUserId] = for{
+    subject <- directoryDAO.loadSubjectFromGoogleSubjectId(googleSubjectId)
+    userInfo <- subject match{
+      case Some(uid: WorkbenchUserId) => Future.successful(uid)
+      case Some(_) => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"subjectId $googleSubjectId is not a WorkbenchUser")))
+      case None => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"google subject Id $googleSubjectId not found in sam")))
+    }
+  }yield userInfo
+}
+
+final case class CreateWorkbenchUser(id: WorkbenchUserId, googleSubjectId: GoogleSubjectId, email: WorkbenchEmail)
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
@@ -1,11 +1,12 @@
-package org.broadinstitute.dsde.workbench.sam.api
+package org.broadinstitute.dsde.workbench.sam
+package api
 
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server
 import akka.http.scaladsl.server.Directives._
-import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchUser, WorkbenchUserId}
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.sam.model.SamJsonSupport._
 import org.broadinstitute.dsde.workbench.sam.service.UserService
 
@@ -20,61 +21,64 @@ trait UserRoutes extends UserInfoDirectives {
 
   def userRoutes: server.Route =
     pathPrefix("user") {
-      requireUserInfo { userInfo =>
-        (pathPrefix("v1") | pathEndOrSingleSlash){
+        (pathPrefix("v1") | pathEndOrSingleSlash) {
           pathEndOrSingleSlash {
             post {
-              complete {
-                userService.createUser(WorkbenchUser(userInfo.userId, userInfo.userEmail)).map(userStatus => StatusCodes.Created -> userStatus)
-              }
-            } ~
-            get {
-              parameter("userDetailsOnly".?) { userDetailsOnly =>
+              requireCreateUser { createUser =>
                 complete {
-                  userService.getUserStatus(userInfo.userId, userDetailsOnly.exists(_.equalsIgnoreCase("true"))).map { statusOption =>
-                    statusOption.map { status =>
-                      StatusCodes.OK -> Option(status)
-                    }.getOrElse(StatusCodes.NotFound -> None)
+                  userService.createUser(createUser).map(userStatus => StatusCodes.Created -> userStatus)
+                }
+              }
+            } ~ requireUserInfo { user =>
+              get {
+                parameter("userDetailsOnly".?) { userDetailsOnly =>
+                  complete {
+                    userService.getUserStatus(user.userId, userDetailsOnly.exists(_.equalsIgnoreCase("true"))).map { statusOption =>
+                      statusOption.map { status =>
+                        StatusCodes.OK -> Option(status)
+                      }.getOrElse(StatusCodes.NotFound -> None)
+                    }
                   }
                 }
               }
             }
           }
-        } ~
-        pathPrefix("v2") {
-          pathPrefix("self") {
-            pathEndOrSingleSlash {
-              post {
-                complete {
-                  userService.createUser(WorkbenchUser(userInfo.userId, userInfo.userEmail)).map(userStatus => StatusCodes.Created -> userStatus)
-                }
-              }
-            } ~
-            path("info") {
-              get {
-                complete {
-                  userService.getUserStatusInfo(userInfo.userId).map { statusOption =>
-                    statusOption.map { status =>
-                      StatusCodes.OK -> Option(status)
-                    }.getOrElse(StatusCodes.NotFound -> None)
+        } ~ pathPrefix("v2") {
+              pathPrefix("self") {
+                pathEndOrSingleSlash {
+                  post {
+                    requireCreateUser { createUser =>
+                      complete {
+                        userService.createUser(createUser).map(userStatus => StatusCodes.Created -> userStatus)
+                      }
+                    }
                   }
+                }~ requireUserInfo {user =>
+                    path("info") {
+                      get {
+                        complete {
+                          userService.getUserStatusInfo(user.userId).map { statusOption =>
+                            statusOption.map { status =>
+                              StatusCodes.OK -> Option(status)
+                            }.getOrElse(StatusCodes.NotFound -> None)
+                          }
+                        }
+                      }
+                    } ~
+                    path("diagnostics") {
+                      get {
+                        complete {
+                          userService.getUserStatusDiagnostics(user.userId).map { statusOption =>
+                            statusOption.map { status =>
+                              StatusCodes.OK -> Option(status)
+                            }.getOrElse(StatusCodes.NotFound -> None)
+                          }
+                        }
+                      }
+                    }
                 }
-              }
-            } ~
-            path("diagnostics") {
-              get {
-                complete {
-                  userService.getUserStatusDiagnostics(userInfo.userId).map { statusOption =>
-                    statusOption.map { status =>
-                      StatusCodes.OK -> Option(status)
-                    }.getOrElse(StatusCodes.NotFound -> None)
-                  }
-                }
-              }
             }
           }
-        }
-      }
     }
 
   def adminUserRoutes: server.Route =
@@ -149,3 +153,4 @@ trait UserRoutes extends UserInfoDirectives {
       }
     }
 }
+
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
@@ -156,20 +156,21 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
     } yield (userEmails ++ groupEmails).toSet
   }
 
-  override def createUser(user: WorkbenchUser): Future[WorkbenchUser] = Future {
-    ldapConnectionPool.add(userDn(user.id),
+  override def createUser(user: WorkbenchUser): Future[WorkbenchUser] = {
+    val attrs = List(
       new Attribute(Attr.email, user.email.value),
       new Attribute(Attr.sn, user.id.value),
       new Attribute(Attr.cn, user.id.value),
       new Attribute(Attr.uid, user.id.value),
-      new Attribute(Attr.googleSubjectId, user.id.value),
       new Attribute("objectclass", Seq("top", "workbenchPerson").asJava)
-    )
-
-    user
-  }.recover {
-    case ldape: LDAPException if ldape.getResultCode == ResultCode.ENTRY_ALREADY_EXISTS =>
-      throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"identity with id ${user.id} already exists"))
+    ) ++ user.googleSubjectId.map(gsid => List(new Attribute(Attr.googleSubjectId, gsid.value))).getOrElse(List.empty)
+    Future {
+      ldapConnectionPool.add(userDn(user.id), attrs: _*)
+      user
+    }.recoverWith {
+      case ldape: LDAPException if ldape.getResultCode == ResultCode.ENTRY_ALREADY_EXISTS =>
+        Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"identity with id ${user.id} already exists")))
+    }
   }
 
   override def loadUser(userId: WorkbenchUserId): Future[Option[WorkbenchUser]] = Future {
@@ -186,7 +187,7 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
     val uid = getAttribute(results, Attr.uid).getOrElse(throw new WorkbenchException(s"${Attr.uid} attribute missing"))
     val email = getAttribute(results, Attr.email).getOrElse(throw new WorkbenchException(s"${Attr.email} attribute missing"))
 
-    WorkbenchUser(WorkbenchUserId(uid), WorkbenchEmail(email))
+    WorkbenchUser(WorkbenchUserId(uid), getAttribute(results, Attr.googleSubjectId).map(GoogleSubjectId), WorkbenchEmail(email))
   }
 
   override def loadUsers(userIds: Set[WorkbenchUserId]): Future[Seq[WorkbenchUser]] = Future {
@@ -264,17 +265,10 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
     result.getOrElse(false)
   }
 
-  override def getUserFromPetServiceAccount(petSA: ServiceAccountSubjectId): Future[Option[WorkbenchUser]] = Future {
-    val matchingUsers = ldapSearchStream(peopleOu, SearchScope.SUBORDINATE_SUBTREE, Filter.createEqualityFilter(Attr.uid, petSA.value)) { entry =>
-      dnToSubject(entry.getDN)
-    }.collect {
-      case pet: PetServiceAccountId => pet.userId
-    }.toSet
-
-    if (matchingUsers.size > 1) {
-      throw new WorkbenchException(s"id $petSA refers to too many subjects: $matchingUsers")
-    } else {
-      matchingUsers.headOption.flatMap(loadUserInternal)
+  override def getUserFromPetServiceAccount(petSA: ServiceAccountSubjectId): Future[Option[WorkbenchUser]] = {
+    loadSubjectFromGoogleSubjectId(GoogleSubjectId(petSA.value)).map{
+      case Some(PetServiceAccountId(userId, _)) => loadUserInternal(userId)
+      case _ => None
     }
   }
 
@@ -346,4 +340,17 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
   override def setManagedGroupAccessInstructions(groupName: WorkbenchGroupName, accessInstructions: String): Future[Unit] = Future {
     ldapConnectionPool.modify(groupDn(groupName), new Modification(ModificationType.REPLACE, Attr.accessInstructions, accessInstructions))
   }
+
+  override def loadSubjectFromGoogleSubjectId(googleSubjectId: GoogleSubjectId): Future[Option[WorkbenchSubject]] =
+    for{
+      entries <- Future{ldapConnectionPool.search(peopleOu, SearchScope.SUB, Filter.createEqualityFilter(Attr.googleSubjectId, googleSubjectId.value)).getSearchEntries.asScala}
+      res <- entries match {
+        case Seq() => Future.successful(None)
+        case Seq(subject) => Future.successful(Try(dnToSubject(subject.getDN)).toOption)
+        case subjects => Future.failed(new WorkbenchException(s"Database error: googleSubjectId $googleSubjectId refers to too many subjects: ${subjects.map(_.getDN)}"))
+      }
+    } yield res
+
+  override def setGoogleSubjectId(userId: WorkbenchUserId, googleSubjectId: GoogleSubjectId): Future[Unit] =
+    Future{ldapConnectionPool.modify(userDn(userId), new Modification(ModificationType.ADD, Attr.googleSubjectId, googleSubjectId.value))}
 }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -17,6 +17,7 @@ import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport.Work
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.model.google._
 import org.broadinstitute.dsde.workbench.sam._
+import org.broadinstitute.dsde.workbench.sam.api.CreateWorkbenchUser
 import org.broadinstitute.dsde.workbench.sam.config.{GoogleServicesConfig, PetServiceAccountConfig}
 import org.broadinstitute.dsde.workbench.sam.directory.DirectoryDAO
 import org.broadinstitute.dsde.workbench.sam.model.SamJsonSupport.ResourceAndPolicyNameFormat
@@ -25,6 +26,7 @@ import org.broadinstitute.dsde.workbench.sam.openam.AccessPolicyDAO
 import org.broadinstitute.dsde.workbench.sam.service.{CloudExtensions, SamApplication}
 import org.broadinstitute.dsde.workbench.util.health.{HealthMonitor, SubsystemStatus, Subsystems}
 import org.broadinstitute.dsde.workbench.util.{FutureSupport, Retry}
+import org.broadinstitute.dsde.workbench.sam.service.UserService._
 import spray.json._
 
 import scala.collection.JavaConverters._
@@ -40,14 +42,14 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
 
   private val maxGroupEmailLength = 64
 
-  private[google] def toProxyFromUser(user: WorkbenchUser): WorkbenchEmail = {
+  private[google] def toProxyFromUser(userId: WorkbenchUserId): WorkbenchEmail = {
 /* Re-enable this code and remove the temporary code below after fixing rawls for GAWB-2933
     val username = user.email.value.split("@").head
     val emailSuffix = s"_${user.id.value}@${googleServicesConfig.appsDomain}"
     val maxUsernameLength = maxGroupEmailLength - emailSuffix.length
     WorkbenchEmail(username.take(maxUsernameLength) + emailSuffix)
 */
-    WorkbenchEmail(s"${googleServicesConfig.resourceNamePrefix.getOrElse("")}PROXY_${user.id.value}@${googleServicesConfig.appsDomain}")
+    WorkbenchEmail(s"${googleServicesConfig.resourceNamePrefix.getOrElse("")}PROXY_${userId.value}@${googleServicesConfig.appsDomain}")
 /**/
   }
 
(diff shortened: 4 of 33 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/package.scala
```diff
@@ -0,0 +1,8 @@
+package org.broadinstitute.dsde.workbench.sam
+
+package object api {
+  val accessTokenHeader = "OIDC_access_token"
+  val expiresInHeader = "OIDC_CLAIM_expires_in"
+  val emailHeader = "OIDC_CLAIM_email"
+  val googleSubjectIdHeader = "OIDC_CLAIM_user_id"
+}
```
dvoet: why put these in the package instead of closer to where they are used?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,70 @@
...
+      headerValueByName(googleSubjectIdHeader) &
+      headerValueByName(emailHeader)
+    ) tflatMap {
+    case (googleSubjectId, email) => {
+      onSuccess(createWorkbenchUserAPI(GoogleSubjectId(googleSubjectId), WorkbenchEmail(email)))
+    }
+  }
+}
+
+object StandardUserInfoDirectives{
```
dvoet: why do we have this new object?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,70 @@
...
   implicit val executionContext: ExecutionContext
 
+  def requireUserInfo: Directive1[UserInfo] = (
+    headerValueByName(accessTokenHeader) &
+      headerValueByName(googleSubjectIdHeader) &
+      headerValueByName(expiresInHeader) &
+      headerValueByName(emailHeader)
+    ) tflatMap {
+    case (token, googleSubjectId, expiresIn, email) =>
+      onSuccess(Try(expiresIn.toLong).fold[Future[UserInfo]](t => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"expiresIn $expiresIn can't be converted to Long"))),
```
dvoet: include t as a cause in the ErrorReport

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,70 @@
...
-    case (token, userId, expiresIn, email) => {
-      val userInfo = UserInfo(OAuth2BearerToken(token), WorkbenchUserId(userId), WorkbenchEmail(email), expiresIn.toLong)
-      onSuccess(getUserFromPetServiceAccount(userInfo).map {
-        case Some(petOwnerUser) => UserInfo(OAuth2BearerToken(token), petOwnerUser.id, petOwnerUser.email, expiresIn.toLong)
-        case None => userInfo
-      })
-    }
+  def createUserInfo(token: OAuth2BearerToken, googleSubjectId: GoogleSubjectId, email: WorkbenchEmail, expiresIn: Long, directoryDAO: DirectoryDAO)(implicit ec: ExecutionContext): Future[UserInfo] = if (isPetSA(email)) {
+    // If it's a PET account, we treat it as its owner
+    directoryDAO.getUserFromPetServiceAccount(ServiceAccountSubjectId(googleSubjectId.value)).flatMap(_.fold[Future[UserInfo]](Future.failed(new WorkbenchException("no owner found for this PET account")))(pet => Future.successful(UserInfo(token, pet.id, pet.email, expiresIn.toLong))))
```
dvoet: I find lines like this difficult to read. First, it is a long line, which by itself is ok, but the important part of what is going on is all the way at the end. I would like to see this broken up a bit to highlight the happy path. I also find the fold in the middle hides the choice of error or not error or even that an Option is in play. Last, the flatMap is unnecessary since the result is either Future.failed or Future.successful which just serves to make the line longer. I would prefer something like

```   
    directoryDAO.getUserFromPetServiceAccount(ServiceAccountSubjectId(googleSubjectId.value)).map {
      case None => throw new WorkbenchException("no owner found for this PET account")
      case Some(pet) => UserInfo(token, pet.id, pet.email, expiresIn.toLong)
    }
```

or even

```
directoryDAO.getUserFromPetServiceAccount(ServiceAccountSubjectId(googleSubjectId.value)).map(_.map(pet => UserInfo(token, pet.id, pet.email, expiresIn.toLong)).getOrElse(throw new WorkbenchException("no owner found for this PET account")))
```
because it puts the failure case at the end

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,70 @@
...
-      headerValueByName("OIDC_CLAIM_email")
-    ) tflatMap {
-    case (token, userId, expiresIn, email) => {
-      val userInfo = UserInfo(OAuth2BearerToken(token), WorkbenchUserId(userId), WorkbenchEmail(email), expiresIn.toLong)
-      onSuccess(getUserFromPetServiceAccount(userInfo).map {
-        case Some(petOwnerUser) => UserInfo(OAuth2BearerToken(token), petOwnerUser.id, petOwnerUser.email, expiresIn.toLong)
-        case None => userInfo
-      })
-    }
+  def createUserInfo(token: OAuth2BearerToken, googleSubjectId: GoogleSubjectId, email: WorkbenchEmail, expiresIn: Long, directoryDAO: DirectoryDAO)(implicit ec: ExecutionContext): Future[UserInfo] = if (isPetSA(email)) {
```
dvoet: I think something got lost here: not every service account is a pet service account. The isPetSA function is probably not correctly named, it should be isServiceAccount. So the case where a SA is not a Pet has been lost.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,70 @@
...
+    // If it's a PET account, we treat it as its owner
+    directoryDAO.getUserFromPetServiceAccount(ServiceAccountSubjectId(googleSubjectId.value)).flatMap(_.fold[Future[UserInfo]](Future.failed(new WorkbenchException("no owner found for this PET account")))(pet => Future.successful(UserInfo(token, pet.id, pet.email, expiresIn.toLong))))
+  } else {
+    for{
+      subject <- directoryDAO.loadSubjectFromGoogleSubjectId(googleSubjectId)
+      userInfo <- subject match{
+        case Some(uid: WorkbenchUserId) => Future.successful(UserInfo(token, uid, email, expiresIn))
+        case Some(_) => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"subjectId $googleSubjectId is not a WorkbenchUser")))
```
dvoet: why would this case ever happen and why would it be a conflict status code? does it make sense for loadSubjectFromGoogleSubjectId to return WorkbenchUserId instead of the generic WorkbenchSubject?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,70 @@
...
+    }yield userInfo
   }
 
-  private def getUserFromPetServiceAccount(userInfo: UserInfo):Future[Option[WorkbenchUser]] = {
-    if (isPetSA(userInfo.userEmail)) {
-      directoryDAO.getUserFromPetServiceAccount(ServiceAccountSubjectId(userInfo.userId.value))
-    } else {
-      Future.successful(None)
-    }
+  def createWorkbenchUserAPI(googleSubjectId: GoogleSubjectId, email: WorkbenchEmail): Future[CreateWorkbenchUserAPI] = if (isPetSA(email)) {
```
dvoet: again, not all service accounts are pets. also I don't think CreateWorkbenchUserAPI belongs in workbench-libs as it is specific to sam functionality

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -151,13 +151,23 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
   }
 
   override def createUser(user: WorkbenchUser): Future[WorkbenchUser] = Future {
-    ldapConnectionPool.add(userDn(user.id),
+    user.googleSubjectId.fold(ldapConnectionPool.add(userDn(user.id),
```
dvoet: too much code duplication here

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -328,4 +338,19 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
...
+
+  override def loadSubjectFromGoogleSubjectId(googleSubjectId: GoogleSubjectId): Future[Option[WorkbenchSubject]] = Future {
+    ldapConnectionPool.search(directoryConfig.baseDn, SearchScope.SUB, Filter.createEqualityFilter(Attr.googleSubjectId, googleSubjectId.value)).getSearchEntries.asScala match {
+      case Seq() => None
+      case Seq(subject) => Option(dnToSubject(subject.getDN))
+      case subjects => throw new WorkbenchException(s"Database error: googleSubjectId $googleSubjectId refers to too many subjects: ${subjects.map(_.getDN)}")
+    }
+  }
+
+  override def updateGoogleSubjectId(user: WorkbenchUser): Future[WorkbenchUser] =
```
dvoet: I am not sure of a case when we actually want to update the subject id. Yes, we need to be able to set it, but once it is set, never change it. So I think this should be called setGoogleSubjectId and the modification type below should not be replace (I want an exception if it already exists). 

Allowing a change means there is a code path that allows someone to hijack another's account.

src/main/scala/org/broadinstitute/dsde/workbench/sam/package.scala
```diff
@@ -1,10 +1,33 @@
...
+
+import org.apache.commons.codec.binary.Hex
+import org.broadinstitute.dsde.workbench.model.{ErrorReportSource, WorkbenchUserId}
 
 /**
   * Created by dvoet on 5/18/17.
   */
 package object sam {
   implicit val errorReportSource = ErrorReportSource("sam")
+
```
dvoet: I think this new code should be in UserService

src/main/scala/org/broadinstitute/dsde/workbench/sam/package.scala
```diff
@@ -1,10 +1,33 @@
...
+    // currentMilli is 13 digits, and it'll be another 200 years before it becomes 14 digits. So we're assuming currentMillis is 13 digits here
+    val bytes = new Array[Byte](4)
+    random.nextBytes(bytes)
+    val r = new String(Hex.encodeHex(bytes))
+    // since googleSubjectId starts with 1, we are replacing 1 with 2 to avoid conflicts with existing uid
+    val front = if(currentMillisString(0) == '1') currentMillisString.replaceFirst("1", "2") else currentMilli
+    front + r
+  }
+
+  def genWorkbenchUserId(currentMilli: Long): WorkbenchUserId = {
```
dvoet: don't abbreviate generate, gen looks too much like get.

src/main/scala/org/broadinstitute/dsde/workbench/sam/package.scala
```diff
@@ -1,10 +1,33 @@
...
   implicit val errorReportSource = ErrorReportSource("sam")
+
+  val random = SecureRandom.getInstance("NativePRNGNonBlocking")
+
+  // Generate a 21 digits unique identifier. First char is fixed 2
+  // CurrentMillis.append(randomString)
+  private[workbench] def genRandom(currentMilli: Long): String = {
+    val currentMillisString = currentMilli.toString
+    // one hexdecimal is 4 bits, one byte can generate 2 hexdecial number, so we only need half the number of bytes, which is 8
+    // currentMilli is 13 digits, and it'll be another 200 years before it becomes 14 digits. So we're assuming currentMillis is 13 digits here
```
dvoet: how short sighted ;)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -14,19 +15,47 @@ import scala.concurrent.{ExecutionContext, Future}
...
 
+  /**
+    * If googleSubjectId exists in ldap, return 409; else if email also exists, we lookup pre-created user record and update
+    * its googleSubjectId field; otherwise, we create a new user
+    *
+    * GoogleSubjectId    Email
+    *      no             no      ---> We've never seen this user before, create a new user
+    *      no             no      ---> Someone invited this user previous and we have a record for this user already. We just need to update GoogleSubjetId field for this user.
+    *      yes            skip    ---> User exists. Do nothing.
+    *      yes            skip    ---> User exists. Do nothing.
```
dvoet: matrix above does not look quite right

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -14,19 +15,47 @@ import scala.concurrent.{ExecutionContext, Future}
...
+    */
+  private def registerUser(user: CreateWorkbenchUserAPI): Future[WorkbenchUser] = for{
+    existingSubject <- directoryDAO.loadSubjectFromGoogleSubjectId(user.googleSubjectId)
+    user <- existingSubject.fold[Future[WorkbenchUser]]{
```
dvoet: the more I see code like this, the more I don't like fold for this situation. I would rather see case statements, they are so much more clear. With fold I can never remember which comes first the Some case or the None plus I need to know that it is an Option in the first place.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
```diff
@@ -1,62 +1,20 @@
-package org.broadinstitute.dsde.workbench.sam.api
+package org.broadinstitute.dsde.workbench.sam
```
dvoet, line 1: I am concerned that this spec is changing so drastically. V1 specs are supposed to remain mostly static (additions are ok) because the v1 api should not change. With large changes to the tests I am concerned behavior is also changing.

Do we need all this change?

dvoet (changes requested):
see comments

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/package.scala
Qi77Qi: They're also used in tests...I don't want to copy paste these strings multiple places

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
Qi77Qi: it's just so I don't have to create an object when I'm testing the functions inside `StandardUserInfoDirectives`...the functions inside `StandardUserInfoDirectives` are also singleton.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
Qi77Qi: The idea is so that we're not throwing exceptions. For `fold`, I can change it to pattern match, but `fold` works more naturally when I'm writing out this code. For your confusion, empty case always comes first, it's the same way when you use `fold` on `Either`, `Validated` etc (failure is the first function)

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
Qi77Qi: the majority of the change comes from the fact that we're creating new userId instead inside register user endpoint instead of using a userId that's passed in. Previous `TestSamRoutes` won't work well because it's tied to a `UserInfo` object, which requires a `userId`.

The other part of the change comes from @gpolumbo-broad 's suggestion about reducing duplicated testing setup code.

src/main/scala/org/broadinstitute/dsde/workbench/sam/package.scala
Qi77Qi: `get` is not something comes into my mind :P

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: I am pushing maintainability of the code over writability. The less arbitrary stuff I need to remember a priori the better. So now I know that the empty case usually comes first, at least in cats. Ask me in a week. Will I be sure?

It takes more work to write code that can be maintained by others, to try to see it through the eyes of someone else, but it is worth it.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/package.scala
dvoet: I definitely agree that they should be vals somewhere, but they are not package level things. They should be in StandardUserInfoDirectives

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
Qi77Qi: does https://github.com/broadinstitute/sam/pull/190/files#diff-569028fd4f15057259e1e7cbd70ac4cfR57 address the case where it's a SA but not PET?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
Qi77Qi: I was mostly following the other functions signatures such as https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectoryDAO.scala#L35 ...It also makes sense to me that a given googleSubjectId might be mapped to a non regular user subject. The case you pointed out is where we received a `googleSubjectId` for non regular user (say `WorkbenchGroup` ). I can change the return code to something else if you have a better one in mind.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
Qi77Qi: I've renamed `isPetSA` function. Correct me if I'm wrong, but the assumption for `createWorkbenchUserAPI` is if the mail is a service account, we treat it as a bad request.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: but the ldap schema does not allow `WorkbenchGroup` to have `googleSubjectId`.  `loadSubjectFromGoogleSubjectId` should never return anything other than a regular user so that function should be responsible for returning the error (and it is probably a 500). That relieves all the callers from having to do this.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,81 @@
...
-      val userInfo = UserInfo(OAuth2BearerToken(token), WorkbenchUserId(userId), WorkbenchEmail(email), expiresIn.toLong)
-      onSuccess(getUserFromPetServiceAccount(userInfo).map {
-        case Some(petOwnerUser) => UserInfo(OAuth2BearerToken(token), petOwnerUser.id, petOwnerUser.email, expiresIn.toLong)
-        case None => userInfo
-      })
-    }
+  def createUserInfo(token: OAuth2BearerToken, googleSubjectId: GoogleSubjectId, email: WorkbenchEmail, expiresIn: Long, directoryDAO: DirectoryDAO)(implicit ec: ExecutionContext): Future[UserInfo] = if (isServiceAccount(email)) {
+    // If it's a PET account, we treat it as its owner
+    directoryDAO.getUserFromPetServiceAccount(ServiceAccountSubjectId(googleSubjectId.value)).flatMap(
+      owner =>
```
dvoet: flatMap right into the case statements below:
```
flatMap {
    case Some(pet) => Future.successful(UserInfo(token, pet.id, pet.email, expiresIn.toLong))
    case None => Future.failed(new WorkbenchException("no owner found for this service account"))
}

File: src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala [hunk 17]
```diff
@@ -1,46 +1,81 @@
...
+        owner match{
+          case Some(pet) => Future.successful(UserInfo(token, pet.id, pet.email, expiresIn.toLong))
+          case None => Future.failed(new WorkbenchException("no owner found for this service account"))
```
dvoet: it is not an error if the service account does not have an owner. If it is ownerless (i.e. not a pet) then it should be treated as a regular user. Refer to the original code.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
Qi77Qi: Okay, I can exclude `WorkbenchGroup` as an improvement....But it won't help this code here right, cuz it can still return a PET

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,81 @@
...
+    }yield userInfo
   }
 
-  private def getUserFromPetServiceAccount(userInfo: UserInfo):Future[Option[WorkbenchUser]] = {
-    if (isPetSA(userInfo.userEmail)) {
-      directoryDAO.getUserFromPetServiceAccount(ServiceAccountSubjectId(userInfo.userId.value))
-    } else {
-      Future.successful(None)
-    }
+  def createWorkbenchUserAPI(googleSubjectId: GoogleSubjectId, email: WorkbenchEmail): Future[CreateWorkbenchUserAPI] = if (isServiceAccount(email)) {
```
dvoet: as a general word on style, I would like to see the entire function in its own block:
```
... = {
  if (isServiceAccount(email)) {
  ....
}
```
instead of
```
= if (isServiceAccount(email)) {
```
in the latter, the if statement gets lost and makes the code harder to read.

Also please put spaces before and after curly braces... That one does not really matter but it bothers me for some reason

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,81 @@
...
+    Future.failed[CreateWorkbenchUserAPI](new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"we can't create user for service account email ${email}")))
```
dvoet: this is also not an error, we allow service accounts to be regular users. again, refer to the original code

dvoet (changes requested):

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,77 @@
...
+  }
+
+  def getUserInfo(token: OAuth2BearerToken, googleSubjectId: GoogleSubjectId, email: WorkbenchEmail, expiresIn: Long, directoryDAO: DirectoryDAO)(implicit ec: ExecutionContext): Future[UserInfo] = if (isServiceAccount(email)) {
+    // If it's a PET account, we treat it as its owner
+    directoryDAO.getUserFromPetServiceAccount(ServiceAccountSubjectId(googleSubjectId.value)).flatMap{
+      case Some(pet) => Future.successful(UserInfo(token, pet.id, pet.email, expiresIn.toLong))
+      case None => lookUpByGoogleSubjectId(googleSubjectId, directoryDAO).map(uid => UserInfo(token, uid, email, expiresIn))
     }
+  } else {
+    lookUpByGoogleSubjectId(googleSubjectId, directoryDAO).map(uid => UserInfo(token, uid, email, expiresIn))
```
gpolumbo-broad, line 63: this line appears to be identical to the `None` case on line 58, can the logic be rewritten to eliminate this duplication?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,77 @@
...
   }
+
+  private def lookUpByGoogleSubjectId(googleSubjectId: GoogleSubjectId, directoryDAO: DirectoryDAO)(implicit ec: ExecutionContext): Future[WorkbenchUserId] = for{
+    subject <- directoryDAO.loadSubjectFromGoogleSubjectId(googleSubjectId)
+    userInfo <- subject match{
+      case Some(uid: WorkbenchUserId) => Future.successful(uid)
+      case Some(_) => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"subjectId $googleSubjectId is not a WorkbenchUser")))
```
gpolumbo-broad, line 70: How likely/possible is this scenario?  Do we need to test for it?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,77 @@
...
+      case None => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"google subject Id $googleSubjectId not found in sam")))
+    }
+  }yield userInfo
```
gpolumbo-broad, line 73: Formatting nit - I think the convention throughout the project is to have a space around all braces.  I see some other places throughout this PR like this where the brace is not padded.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
```diff
@@ -20,61 +21,64 @@ trait UserRoutes extends UserInfoDirectives {
...
-                }
-              }
-            }
-          }
-        } ~
-        pathPrefix("v2") {
-          pathPrefix("self") {
-            pathEndOrSingleSlash {
-              post {
+              requireCreateUser { createUser =>
```
gpolumbo-broad, line 27: naming nit, `createUser` is a verb, when this is actually data describing some new user info, so I think this would read a little clearer if it were called `newUserInfo` or some other noun that accurately describes what this variable is

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -329,4 +323,17 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
...
+      entries <- Future{ldapConnectionPool.search(peopleOu, SearchScope.SUB, Filter.createEqualityFilter(Attr.googleSubjectId, googleSubjectId.value)).getSearchEntries.asScala}
+      res <- entries match {
+        case Seq() => Future.successful(None)
+        case Seq(subject) => Future.successful(Try(dnToSubject(subject.getDN)).toOption)
+        case subjects => Future.failed(new WorkbenchException(s"Database error: googleSubjectId $googleSubjectId refers to too many subjects: ${subjects.map(_.getDN)}"))
+      }
+    } yield res
+
+  override def setGoogleSubjectId(userId: WorkbenchUserId, googleSubjectId: GoogleSubjectId): Future[Unit] =
+    Future{ldapConnectionPool.modify(userDn(userId), new Modification(ModificationType.ADD, Attr.googleSubjectId, googleSubjectId.value))}
```
gpolumbo-broad, line 355: Does this only add `googleSubjectId` to records that do not already have that attribute set?  Can I call this method on a record that already has a `googleSubjectId` set?  If no, then I would suggest renaming the method to `addGoogleSubjectId`.  If this method works in an "add or edit" scenario, then it's fine.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -40,14 +42,14 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
 
   private val maxGroupEmailLength = 64
 
-  private[google] def toProxyFromUser(user: WorkbenchUser): WorkbenchEmail = {
+  private[google] def toProxyFromUser(userId: WorkbenchUserId): WorkbenchEmail = {
```
gpolumbo-broad, line 45: Rename method to be `toProxyFromUserId`

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -217,13 +224,12 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
         // SA does not exist in google, create it and add it to the proxy group
         case None => for {
           sa <- googleIamDAO.createServiceAccount(project, petSaName, petSaDisplayName)
-          _ <- withProxyEmail(user.id) { proxyEmail => googleDirectoryDAO.addMemberToGroup(proxyEmail, sa.email) }
+          r <- withProxyEmail(user.id){proxyEmail => googleDirectoryDAO.addMemberToGroup(proxyEmail, sa.email)}
```
gpolumbo-broad, line 228: Just out of curiosity, does this change have any effect?  I can't imagine that assigning to a variable instead of `_` would make any difference in this context, right?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -115,3 +148,26 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+  // Generate a 21 digits unique identifier. First char is fixed 2
+  // CurrentMillis.append(randomString)
+  private[workbench] def genRandom(currentMilli: Long): String = {
+    val currentMillisString = currentMilli.toString
+    // one hexdecimal is 4 bits, one byte can generate 2 hexdecial number, so we only need half the number of bytes, which is 8
+    // currentMilli is 13 digits, and it'll be another 200 years before it becomes 14 digits. So we're assuming currentMillis is 13 digits here
+    val bytes = new Array[Byte](4)
+    random.nextBytes(bytes)
+    val r = new String(Hex.encodeHex(bytes))
+    // since googleSubjectId starts with 1, we are replacing 1 with 2 to avoid conflicts with existing uid
```
gpolumbo-broad, line 165: I'm still a little uneasy about this.  Just because googleSubjectIds have always started with 1 so far, I don't have faith that they always will.  Since we're returning String anyway, i'd rather see a prefix like "wb-" or something.  Or anything that would be really weird/unlikely for Google to add as a prefix

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
```diff
@@ -1,12 +1,93 @@
...
 }
-object TestSupport extends TestSupport
\ No newline at end of file
+
+trait PropertyBasedTesting extends FlatSpec with PropertyChecks with Configuration with Matchers {
+  implicit override val generatorDrivenConfig: PropertyCheckConfiguration = PropertyCheckConfiguration(minSuccessful = 3)
+}
+
+object TestSupport extends TestSupport{
+  implicit val eqThrowablw: Equality[Throwable] = new Equality[Throwable]{
```
gpolumbo-broad: typo?  `eqThrowablw` -> `eqThrowable`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,77 @@
...
+
+  def createWorkbenchUserAPI(googleSubjectId: GoogleSubjectId, email: WorkbenchEmail): Future[CreateWorkbenchUserAPI] =
+    Future.successful(CreateWorkbenchUserAPI(genWorkbenchUserId(System.currentTimeMillis()), googleSubjectId, email))
 }
+
+final case class CreateWorkbenchUserAPI(id: WorkbenchUserId, googleSubjectId: GoogleSubjectId, email: WorkbenchEmail)
```
gpolumbo-broad: Naming again, this reads as a verb to me, so I'd expect it to be a utility/service class that I can use to create new Workbench Users, when in reality it's a noun/entity/DAO/bean class.  I think this would read better as something like, `NewWorkbenchUser` or something.  

Since this seems to only differ from `WorkbenchUser` in that the `googleSubjectId` is required here (not an `Option`) then maybe this is a `RegisteredWorkbenchUser` or `CompleteWorkbenchUser`?  I guess my point is that I think we can be more expressive in the naming of this class.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectivesSpec.scala
```diff
@@ -0,0 +1,133 @@
...
+    }
+  }
+
+  it should "be able to get a UserInfo object for service account if it is a PET" in {
+    forAll(genServiceAccountSubjectId, genOAuth2BearerToken, genPetEmail){
+      (serviceSubjectId: ServiceAccountSubjectId, token: OAuth2BearerToken, email: WorkbenchEmail) =>
+      val directoryDAO = new MockDirectoryDAO()
+      val gSid = GoogleSubjectId(serviceSubjectId.value)
+      val uid = genWorkbenchUserId(System.currentTimeMillis())
+      directoryDAO.createUser(WorkbenchUser(uid, Some(gSid), email)).futureValue
```
gpolumbo-broad: Looks to me like you're saving the User and Pet with the same subjectId?  I don't think that's right, I think the Pet ends up with its own SubjectId.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectivesSpec.scala
```diff
@@ -0,0 +1,133 @@
...
+      directoryDAO.createPetServiceAccount(PetServiceAccount(PetServiceAccountId(uid, GoogleProject("")), ServiceAccount(serviceSubjectId, email, ServiceAccountDisplayName("")))).futureValue
+      val res = getUserInfo(token, gSid, email, 10L, directoryDAO).futureValue
+      res should be (UserInfo(token, uid, email, 10L))
+    }
+  }
+
+  it should "be able to get a UserInfo object for service account if it is a not PET" in {
```
gpolumbo-broad, line 61: This test is identical to the test on line 35 except here you use a generator for the GoogleSubjectId instead of currentTime

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,77 @@
...
+    subject <- directoryDAO.loadSubjectFromGoogleSubjectId(googleSubjectId)
+    userInfo <- subject match{
+      case Some(uid: WorkbenchUserId) => Future.successful(uid)
+      case Some(_) => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"subjectId $googleSubjectId is not a WorkbenchUser")))
+      case None => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"google subject Id $googleSubjectId not found in sam")))
+    }
+  }yield userInfo
+
+  def createWorkbenchUserAPI(googleSubjectId: GoogleSubjectId, email: WorkbenchEmail): Future[CreateWorkbenchUserAPI] =
+    Future.successful(CreateWorkbenchUserAPI(genWorkbenchUserId(System.currentTimeMillis()), googleSubjectId, email))
```
gpolumbo-broad: Why does this need to return a Future?

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectivesSpec.scala
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+  }
+
+  it should "fail if required headers are missing" in {
+    Get("/") ~> directives.requireUserInfo(x => complete(x.toString)) ~> check {
+      Try(responseAs[String])  === (Success(MissingHeaderRejection(accessTokenHeader)))
+    }
+  }
+}
```
gpolumbo-broad: Should there be positive tests for `directives.requireUserInfo`?

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
Qi77Qi: I don't think we'll be able to. line 58 branch depends on result of `directoryDAO.getUserFromPetServiceAccount` call, while line 61 is executed without needing to call `directoryDAO.getUserFromPetServiceAccount`.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
Qi77Qi: Even if it's unlikely to happen, we can't just ignore this failure case tho. It's covered in https://github.com/broadinstitute/sam/pull/190/files#diff-c5e58369b2988aacd6ad60e4b4cd1a61R111

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
Qi77Qi: yea, I don't pay a lot attention to formatting. Now that we have `scalafmt` now, I can have a separate PR to fix formatting for all files

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
Qi77Qi: I don't think `newUserInfo` makes sense because it's not a new `UserInfo`...it's just an instance of `CreateWorkbenchUserAPI` case class...I'm up for other name suggestions tho

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
Qi77Qi: It's just `add`. If a record already have the attribute, it'll return a failed `Future`.

I think @dvoet suggested `set` since it's used in other code with similar semantics.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectivesSpec.scala
Qi77Qi: `PET` has its own subjectId, but this is owner for the `PET`

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectivesSpec.scala
Qi77Qi: the test above has https://github.com/broadinstitute/sam/pull/190/files/6d08d06361abd77fa8356e61e7c5e5f20c463d14#diff-c5e58369b2988aacd6ad60e4b4cd1a61R54, which we add the PET in ldap, and it's testing https://github.com/broadinstitute/sam/pull/190/files#diff-569028fd4f15057259e1e7cbd70ac4cfR57. But in this test, we didn't add PET in ldap, so it's testing https://github.com/broadinstitute/sam/pull/190/files#diff-569028fd4f15057259e1e7cbd70ac4cfR58


I do realize that those `===` is not working the way I expected...just fixed those

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
gpolumbo-broad: OK, no worries

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
gpolumbo-broad: I guess see my other comment on the naming of `CreateWorkbenchUserAPI` then.  I think if we can agree on what we want the final name for that class to be, it will inform a better variable name here.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
gpolumbo-broad: hmmm...OK, if there's a convention, then I won't try to change it, but I still think it's confusing/misleading.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectivesSpec.scala
Qi77Qi: I tried to add success tests. But it's not easy to test just `Directives` with success case since inside `check` block, it's probably referencing some object that's not exposed, hence I'm getting `NPE`.

All tests that are using `SamRoutes` rely on this code to be in happy path, so I don't feel like spending more time adding successful case for just `requireUserInfo` .

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
dvoet: I think we only care about the ids we have seen so far. We only care that our new ids do not clash with existing ids so we don't need to migrate existing ids to something new. It does not matter if google starts producing ids that clash with ours the day after we start generating our own.

gpolumbo-broad (approved):

Qi77Qi:
jenkins retest

Qi77Qi:
jenkins retest

PR #191, opened 2018-08-16
Update pr template

Qi77Qi:
https://broadinstitute.atlassian.net/browse/GAWB-3744
Feedback is welcome!

Current template is a bit distracting and the information can live in READ.me as apposed to every PR description. Jira ticket link can be helpful for providing some background info.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,16 +1,10 @@
-\<your comments for this PR go here\>
+Ticket: <Link to Jira ticket>
+<Put notes here to help reviewer understand this PR>
 
-Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.
+---
 
-I, the developer opening this PR, do solemnly pinky swear that:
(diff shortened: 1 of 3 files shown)


marctalbott (approved):
I think it looks nice and simple. One possible thing to add would be a link to the section of the readme that holds the old template. This could keep the information easily accessible when opening PRs without sacrificing the simplicity of the new template.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,16 +1,4 @@
-\<your comments for this PR go here\>
+Ticket:
+Notes:
```
gpolumbo-broad: If there's nothing left in this file we should just delete it.

If we still want a checklist, it could be something about who the required reviewers are, and if we want this, it is something we could automate with PullApprove.

gpolumbo-broad (changes requested):

gpolumbo-broad:
I don't think this should be merged until @dvoet has a chance to review it

Qi77Qi:
test

dvoet (approved):

PR #196, opened 2018-08-30
turn on partial unification

Qi77Qi:
turning on partial unification is quite important when using cats as described [here](https://github.com/typelevel/cats#getting-started)

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.workbench.sam.service
 import java.util.UUID
 
 import akka.http.scaladsl.model.StatusCodes
+import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.sam._
@@ -324,9 +325,7 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
   }
 
   private def makeValidatablePolicies(policies: Map[AccessPolicyName, AccessPolicyMembership]): Future[Set[ValidatableAccessPolicy]] = {
-    Future.traverse(policies) {
-      case (accessPolicyName, accessPolicyMembership) => makeCreatablePolicy(accessPolicyName, accessPolicyMembership)
-    }.map(_.toSet)
+    policies.toList.traverse{case (accessPolicyName, accessPolicyMembership) => makeCreatablePolicy(accessPolicyName, accessPolicyMembership)}.map(_.toSet)
   }
 
   private def makeCreatablePolicy(accessPolicyName: AccessPolicyName, accessPolicyMembership: AccessPolicyMembership): Future[ValidatableAccessPolicy] = {
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
@@ -3,7 +3,6 @@ package org.broadinstitute.dsde.workbench.sam.service
 import java.util.UUID
 
 import akka.http.scaladsl.model.StatusCodes
-import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.sam._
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -324,9 +325,7 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
   }
 
   private def makeValidatablePolicies(policies: Map[AccessPolicyName, AccessPolicyMembership]): Future[Set[ValidatableAccessPolicy]] = {
-    Future.traverse(policies) {
-      case (accessPolicyName, accessPolicyMembership) => makeCreatablePolicy(accessPolicyName, accessPolicyMembership)
-    }.map(_.toSet)
+    policies.toList.traverse(policies => makeCreatablePolicy(policies._1, policies._2)).map(_.toSet)
```
dvoet: can you use a case statement so you can name policies._1 and policies._2?

Qi77Qi: updated

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -324,9 +325,7 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
   }
 
   private def makeValidatablePolicies(policies: Map[AccessPolicyName, AccessPolicyMembership]): Future[Set[ValidatableAccessPolicy]] = {
-    Future.traverse(policies) {
-      case (accessPolicyName, accessPolicyMembership) => makeCreatablePolicy(accessPolicyName, accessPolicyMembership)
-    }.map(_.toSet)
+    policies.toList.traverse{case (accessPolicyName, accessPolicyMembership) => makeCreatablePolicy(accessPolicyName, accessPolicyMembership)}.map(_.toSet)
```
dvoet: I checked out this branch and while it compiles, intellij complains about this line. How do I make it ok with it? Also, how can I tell what traverse does? The scaladoc for Future.traverse is clear but this traverse does not have any. It feels to me that this change only serves to make the code more opaque. I.e. unless I already know what traverse does it is much harder for me to figure out what this code does.

Qi77Qi: I'm not sure if there's a way to make IJ recognize `traverse`. But https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Traverse.scala#L36 is the `traverse` being used.

It is not easy to find out what code is being used if you've never come across this pattern, but it's quite common in cats. I can explain better offline.

The reason for this change tho is because `Future.traverse` won't work for `Map` when `partial unification` is on

dvoet: this could be just `Future.traverse(policies.toList)` too

I understand what cats traverse does but I want support from my tools too. I would like IJ not to give me red squiggles when the compiler say it is ok and I would like scaladoc to work.

Qi77Qi: I can change this to `Future.traverse(policies.toList)` if you prefer. But I don't think we should limit ourselves within what IJ can support.

It's a known issue with IJ that it's not compatible with lots of functions cats provides because IJ uses its own presentation compiler(most of them won't be that bad if you're okay with having red squiggles in a few places). 

The benefit of cat's traverse is you can use it on `Option`, `Either` and some other things

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -324,9 +325,7 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
   }
 
   private def makeValidatablePolicies(policies: Map[AccessPolicyName, AccessPolicyMembership]): Future[Set[ValidatableAccessPolicy]] = {
-    Future.traverse(policies) {
-      case (accessPolicyName, accessPolicyMembership) => makeCreatablePolicy(accessPolicyName, accessPolicyMembership)
-    }.map(_.toSet)
+    Future.traverse(policies.toList){case (accessPolicyName, accessPolicyMembership) => makeCreatablePolicy(accessPolicyName, accessPolicyMembership)}.map(_.toSet)
```
gpolumbo-broad: I personally find the multi-line expression easier to read.

gpolumbo-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: I agree. All the mechanical things hide the main functional thing. We should strive to use whitespace and new lines to highlight the stuff that matters most.

dvoet (approved):

Qi77Qi:
jenkins retest

PR #198, opened 2018-08-31
populate googleSubjectId with current uid

Qi77Qi:
After this, we should run migration, so that there won't be users missing googleSubjectId field before we change proxy

Discussion https://broadinstitute.slack.com/archives/CC528F8JX/p1535681551000100

In all cases:


dvoet (approved):

gpolumbo-broad (approved):
I approve it even though it doesn't make sense to me that the `sn`, `cn` also have the same value as the `uid` and `googleSubjectId`.

Qi77Qi:
@gpolumbo-broad , I don't know much about `sn` and `cn`, but I think we should leave them as is for this PR.

The reason for duplicating `googleSubjectId` here is just so we can change proxy to lookup users by `googleSubjectId`. Code change in this PR will be updated in https://github.com/broadinstitute/sam/pull/190, but this PR allows us to change proxy without breaking anything in the mean time. It's `2` in https://broadinstitute.atlassian.net/wiki/spaces/~952052303/pages/622624856/sam+Support+invite+feature

PR #200, opened 2018-09-06
add new endpoint to overwrite policy membership

MatthewBemis:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3788

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -5,6 +5,7 @@ import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server
 import akka.http.scaladsl.server.Directive1
 import akka.http.scaladsl.server.Directives._
+import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
 import org.broadinstitute.dsde.workbench.model.{ErrorReport, WorkbenchEmail, WorkbenchExceptionWithErrorReport}
 import org.broadinstitute.dsde.workbench.sam._
 import org.broadinstitute.dsde.workbench.sam.model.SamJsonSupport._
@@ -113,6 +114,15 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
                       }
                     } ~
                     pathPrefix("memberEmails") {
+                      pathEndOrSingleSlash {
+                        put {
+                          requireAction(resource, SamResourceActions.sharePolicy(AccessPolicyName(policyName)), userInfo) {
+                            entity(as[Set[WorkbenchEmail]]) { membersList =>
+                              complete(resourceService.overwritePolicyMembers(resourceType, AccessPolicyName(policyName), Resource(resourceType.name, ResourceId(resourceId)), membersList).map(_ => StatusCodes.NoContent))
+                            }
+                          }
+                        }
+                      } ~
                       pathPrefix(Segment) { email =>
                         withSubject(WorkbenchEmail(email)) { subject =>
                           pathEndOrSingleSlash {
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -116,7 +116,7 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
                     pathPrefix("memberEmails") {
                       pathEndOrSingleSlash {
                         put {
-                          requireAction(resource, SamResourceActions.sharePolicy(AccessPolicyName(policyName)), userInfo) {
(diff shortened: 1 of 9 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -113,6 +114,15 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
                       }
                     } ~
                     pathPrefix("memberEmails") {
+                      pathEndOrSingleSlash {
+                        put {
+                          requireAction(resource, SamResourceActions.sharePolicy(AccessPolicyName(policyName)), userInfo) {
```
dvoet: also alter policies, see a few lines below:
`                             requireOneOfAction(resource, Set(SamResourceActions.alterPolicies, SamResourceActions.sharePolicy(resourceAndPolicyName.accessPolicyName)), userInfo) {`

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -203,6 +203,21 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+    * @param resourceType
+    * @param policyName
+    * @param resource
+    * @param membersList
+    * @return
+    */
+  def overwritePolicyMembers(resourceType: ResourceType, policyName: AccessPolicyName, resource: Resource, membersList: Set[WorkbenchEmail]): Future[AccessPolicy] = {
+    loadResourcePolicy(ResourceAndPolicyName(resource, policyName)).flatMap {
+      case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "Policy not found"))
+      case Some(accessPolicyMembership) => overwritePolicy(resourceType, policyName, resource, accessPolicyMembership.copy(memberEmails = membersList))
```
dvoet: this is a race condition waiting to happen: thread 1 calls overwritePolicyMembers and loads the policy, thread 2 then overwrites the policy, updating roles (or something other than members), then thread 1 overwrites the policy with the old roles.

It would better if this only updates what it needs to. And no need to load first.

dvoet (changes requested):
see comments

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -203,6 +203,21 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+    * Overwrites an existing policy's membership (keyed by resourceType/resourceId/policyName) if it exists
+    * @param resourceType
+    * @param policyName
+    * @param resource
+    * @param membersList
+    * @return
+    */
+  def overwritePolicyMembers(resourceType: ResourceType, policyName: AccessPolicyName, resource: Resource, membersList: Set[WorkbenchEmail]): Future[AccessPolicy] = {
+    loadResourcePolicy(ResourceAndPolicyName(resource, policyName)).flatMap {
+      case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "Policy not found"))
```
Qi77Qi: just fyi, you can `Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "Policy not found")))` instead of throwing the exception

dvoet: not necessarily disagreeing, but why should throwing an exception be avoided?

marctalbott (approved):
Should the new test in ResourceRoutesV1Spec be in ResourceRoutesSpec as well?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
Qi77Qi: 1. throwing exception with stack trace is expensive
2. throwing exception makes code less compose-able in general.

For this particular code, unless caller for this function wraps it inside a `Try` or `Future`, this will actually crash the app? (if we use `Future.failed(some exception)`, then we don't need to worry about this at all)

dvoet: I would not let the expense sway us. Of course throwing exceptions in the normal code path is bad. Used in exceptional situations the expense does not amount to much.

Thinking about thrown exceptions as a side effect though is a good argument. Side effects are bad. Since we use some java libs too, we will never be without. Also, when in Future or Try blocks it seems to be the only way to signal failure.

In this particular case it doesn't really matter since the throw is within a flatMap of a Future. But since it is a flatMap, it is more correct to use Future.failed

Qi77Qi (approved):

dvoet (approved):

dvoet:
jenkins retest - fiab allocator issue

dvoet:
unrelated sam failure

PR #205, opened 2018-09-11
remove group from cloud first

dvoet:
I have seen test failures in cleanup where we attempt to remove a managed group but sam gets a failure in google land (412 precondition failed in this case). But the code removes the group from ldap first so when the test code retries there is the error that the group does not exist.

This PR fixes that by removing the group from google first. If it fails and the group remains in google then ldap still thinks the group exists. If it fails but the group actually got deleted in google (I have seen that), removing the group from google again on retry is tolerant.

---

**PR checklist**


Qi77Qi (approved):

gpolumbo-broad (approved):

dvoet:
jenkins retest

dvoet:
jenkins retest

dvoet:
failures were known in leo tests and in a cleanup of a NIH linking test, the cleanup succeeded in 3 other tests...

PR #209, opened 2018-09-17
add invite endpoint

Qi77Qi:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3699

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1091,7 +1091,7 @@ paths:
       tags:
         - Resources
 
-  /api/v1/invite/user/?email={inviteeEmail}:
+  /api/v1/invite/user/{inviteeEmail}:
     post:
       summary: create current user in the system using login credentials
       parameters:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1091,6 +1091,28 @@ paths:
       tags:
         - Resources
 
+  /api/v1/invite/user/?email={inviteeEmail}:
+    post:
+      summary: create current user in the system using login credentials
+      parameters:
+        - in: path
+          description: email for the user you want to invite
+          name: inviteeEmail
+          required: true
+          type: string
+      responses:
+        201:
+          description: 'user successfully created'
+          schema:
+            $ref: '#/definitions/UserStatusDetails'
+        409:
+          description: user already exists
+          schema:
+            $ref: '#/definitions/ErrorReport'
+      operationId: inviteUser
+      tags:
+      - Users
+
   /register/user/v1:
     get:
       summary: gets the registration status of the logged in user
@@ -1504,6 +1526,19 @@ definitions:
       enabled:
         type: boolean
         description: true if the user is enabled in ldap
+
+  UserStatusDetails:
+    description: ''
+    required:
+      - userSubjectId
+      - userEmail
+    properties:
+      userSubjectId:
+        type: string
+        description: user id
+      userEmail:
+        type: string
+        description: user email
 
   UserStatusDiagnostics:
     description: ''
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamRoutes.scala
@@ -32,7 +32,7 @@ abstract class SamRoutes(val resourceService: ResourceService, val userService: 
     swaggerRoutes ~
     statusRoutes ~
     pathPrefix("register") { userRoutes } ~
-    pathPrefix("api") { resourceRoutes ~ adminUserRoutes ~ extensionRoutes ~ groupRoutes }
+    pathPrefix("api") { inviteRoute ~ resourceRoutes ~ adminUserRoutes ~ extensionRoutes ~ groupRoutes }
   }
 
   // basis for logRequestResult lifted from http://stackoverflow.com/questions/32475471/how-does-one-log-akka-http-client-requests
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
@@ -27,8 +27,10 @@ trait StandardUserInfoDirectives extends UserInfoDirectives {
     ) tflatMap {
     case (token, googleSubjectId, expiresIn, email) =>
       onSuccess(Try(expiresIn.toLong).fold[Future[UserInfo]](
-        t => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"expiresIn $expiresIn can't be converted to Long because  of $t"))),
-        l => getUserInfo(OAuth2BearerToken(token), GoogleSubjectId(googleSubjectId), WorkbenchEmail(email), l, directoryDAO)))
+        t =>
+          Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"expiresIn $expiresIn can't be converted to Long because  of $t"))),
+        l =>
+          getUserInfo(OAuth2BearerToken(token), GoogleSubjectId(googleSubjectId), WorkbenchEmail(email), l, directoryDAO)))
   }
 
   def requireCreateUser: Directive1[CreateWorkbenchUser] = (
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
@@ -9,6 +9,7 @@ import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.sam.model.SamJsonSupport._
 import org.broadinstitute.dsde.workbench.sam.service.UserService
+import org.broadinstitute.dsde.workbench.sam.service.UserService.genWorkbenchUserId
 
 import scala.concurrent.ExecutionContext
 
@@ -21,7 +22,32 @@ trait UserRoutes extends UserInfoDirectives {
 
   def userRoutes: server.Route =
     pathPrefix("user") {
-        (pathPrefix("v1") | pathEndOrSingleSlash) {
+      (pathPrefix("v1") | pathEndOrSingleSlash) {
+        pathEndOrSingleSlash {
+          post {
+            requireCreateUser { createUser =>
+              complete {
+                userService.createUser(createUser).map(userStatus => StatusCodes.Created -> userStatus)
+              }
+            }
+          } ~ requireUserInfo { user =>
+            get {
+              parameter("userDetailsOnly".?) { userDetailsOnly =>
+                complete {
+                  userService.getUserStatus(user.userId, userDetailsOnly.exists(_.equalsIgnoreCase("true"))).map {
+                    statusOption =>
+                      statusOption
+                        .map { status => StatusCodes.OK -> Option(status)
+                        }
+                        .getOrElse(StatusCodes.NotFound -> None)
+                  }
+                }
+              }
+            }
+          }
+        }
+      } ~ pathPrefix("v2") {
+        pathPrefix("self") {
           pathEndOrSingleSlash {
             post {
               requireCreateUser { createUser =>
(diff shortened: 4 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamRoutes.scala
```diff
@@ -32,7 +32,8 @@ abstract class SamRoutes(val resourceService: ResourceService, val userService:
     swaggerRoutes ~
     statusRoutes ~
     pathPrefix("register") { userRoutes } ~
-    pathPrefix("api") { resourceRoutes ~ adminUserRoutes ~ extensionRoutes ~ groupRoutes }
+    pathPrefix("api") { resourceRoutes ~ adminUserRoutes ~ extensionRoutes ~ groupRoutes } ~
+    inviteRoute
```
dvoet: invite route should be under api (I know it is not obvious, but the api prefix requires the user being authenticated and enabled)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -8,30 +8,36 @@ import javax.naming.NameNotFoundException
...
 class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExtensions)(implicit val executionContext: ExecutionContext) extends LazyLogging {
 
-  def createUser(user: CreateWorkbenchUser): Future[UserStatus] = {
+  def createUser(user: CreateWorkbenchUser): Future[UserStatus] = for {
+      createdUser <- registerUser(user)
+      res <- provisionUser(createdUser)
+    } yield res
+
+  def inviteUser(invitee: InviteUser): Future[UserStatus] = for {
+      createdUser <- directoryDAO.createUser(WorkbenchUser(invitee.inviteeId, None, invitee.inviteeEmail))
```
dvoet: need to validate that the email address is not already in use

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -8,30 +8,36 @@ import javax.naming.NameNotFoundException
...
+      res <- provisionUser(createdUser)
```
dvoet: I don't think we should provision the user when they are invited, we should do that when they register

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -8,30 +8,36 @@ import javax.naming.NameNotFoundException
...
-
 import scala.concurrent.{ExecutionContext, Future}
 
 /**
   * Created by dvoet on 7/14/17.
   */
 class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExtensions)(implicit val executionContext: ExecutionContext) extends LazyLogging {
 
-  def createUser(user: CreateWorkbenchUser): Future[UserStatus] = {
+  def createUser(user: CreateWorkbenchUser): Future[UserStatus] = for {
```
dvoet: this function needs to figure out if the user has been invited. If they have been then set their google subject id and provision, if not, register and provision as you already have.

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
Qi77Qi: that's taken care of in https://github.com/broadinstitute/sam/blob/c1419c7f5203e4868b19f46b5bff58e6f5650b92/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala#L51

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -27,8 +27,10 @@ trait StandardUserInfoDirectives extends UserInfoDirectives {
     ) tflatMap {
     case (token, googleSubjectId, expiresIn, email) =>
       onSuccess(Try(expiresIn.toLong).fold[Future[UserInfo]](
-        t => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"expiresIn $expiresIn can't be converted to Long because  of $t"))),
-        l => getUserInfo(OAuth2BearerToken(token), GoogleSubjectId(googleSubjectId), WorkbenchEmail(email), l, directoryDAO)))
+        t =>
```
dvoet, line 30: nitpick: better names than l and t

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
```diff
@@ -21,64 +22,68 @@ trait UserRoutes extends UserInfoDirectives {
...
+          } ~ requireUserInfo { user =>
+            get {
+              parameter("userDetailsOnly".?) { userDetailsOnly =>
+                complete {
+                  userService.getUserStatus(user.userId, userDetailsOnly.exists(_.equalsIgnoreCase("true"))).map {
+                    statusOption =>
+                      statusOption
+                        .map { status => StatusCodes.OK -> Option(status)
+                        }
+                        .getOrElse(StatusCodes.NotFound -> None)
```
dvoet, line 42: I don't really like this code reformatting. Besides confusing this PR, I find it harder to read.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -8,30 +8,38 @@ import javax.naming.NameNotFoundException
...
+      createdUser <- registerUser(user)
+      res <- provisionUser(createdUser)
+    } yield res
+
+  def inviteUser(invitee: InviteUser): Future[UserStatusDetails] = for {
+      existingSubject <- directoryDAO.loadSubjectFromEmail(invitee.inviteeEmail)
+      createdUser <- existingSubject.fold(directoryDAO.createUser(WorkbenchUser(invitee.inviteeId, None, invitee.inviteeEmail)))(
```
dvoet: please use a case and not fold for Options, so much easier to read

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -8,30 +8,38 @@ import javax.naming.NameNotFoundException
...
+        sub => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"identity ${sub} with email ${invitee.inviteeEmail} already exists"))
```
dvoet: just say that the email already exists, don't include sub in the error message

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
```diff
@@ -89,68 +94,91 @@ trait UserRoutes extends UserInfoDirectives {
...
+          }
+        }
+      }
+    }
+
+  val inviteRoute: server.Route = pathPrefix("invite") {
+    pathPrefix("user") {
+      post {
+        requireUserInfo { userInfo =>
+          parameters('email) { email =>
```
dvoet: should the email parameter just be part of the path? I think that is how we do it in other places.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesSpec.scala
```diff
@@ -38,6 +40,18 @@ class UserRoutesSpec extends UserRoutesSpecHelper {
...
   }
 
+  "POST /api/invite/user/{invitee's email}" should "create user" in{
+    val invitee = genInviteUser.sample.get
+
+    val (user, headers, _, routes) = createTestUser() //create a valid user that can invite someone
+    Post(s"/api/invite/user/?email=${invitee.inviteeEmail}").withHeaders(headers) ~> routes.route ~> check{
+      status shouldEqual StatusCodes.Created
+      val res = responseAs[UserStatusDetails]
+      res.userSubjectId.value.length shouldBe 21
```
dvoet: I don't think we should test for this here. api level tests should test only for status codes and response format

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
```diff
@@ -215,9 +215,43 @@ class UserServiceSpec extends FlatSpec with Matchers with TestSupport with Mocki
...
+    val res = dirDAO.loadUser(user.inviteeId).futureValue
+    res shouldBe Some(WorkbenchUser(user.inviteeId, None, user.inviteeEmail))
+  }
+
+  it should "return conflict when there's an existing subject for a given userId" in{
+    val user = genInviteUser.sample.get
+    val email = genNonPetEmail.sample.get
+    dirDAO.createUser(WorkbenchUser(user.inviteeId, None, email)).futureValue
+    val res = service.inviteUser(user).failed.futureValue
+    res === new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"identity with id ${user.inviteeId} already exists")) shouldBe true
```
dvoet: when testing for exception we usually use the intercept function. It is minor, but the test will Fail instead of Error if the exception is not caught.

dvoet (changes requested):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1091,6 +1091,28 @@ paths:
       tags:
         - Resources
 
+  /api/invite/user/?email={inviteeEmail}:
```
dvoet: there should be a v1 in here

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
Qi77Qi: I didn't look through all code, but I saw query string being used...it's a bit more explicit..

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -8,30 +8,39 @@ import javax.naming.NameNotFoundException
...
+
+  def inviteUser(invitee: InviteUser): Future[UserStatusDetails] = for {
+      existingSubject <- directoryDAO.loadSubjectFromEmail(invitee.inviteeEmail)
+      createdUser <- existingSubject match{
+        case None => directoryDAO.createUser(WorkbenchUser(invitee.inviteeId, None, invitee.inviteeEmail))
+        case Some(__) => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"email ${invitee.inviteeEmail} already exists")))
+      }
+    } yield UserStatusDetails(createdUser.id, createdUser.email)
+
+  protected[service] def provisionUser(user: WorkbenchUser): Future[UserStatus] = {
```
gpolumbo-broad: I'm a little confused by the method naming here.  Just looking at the names, here's what I'd expect them to do:

1.  `provisionUser` - simply inserts the bare minimum for a new user in opendj (email and maybe a name)
1.  `registerUser` - Used to "flesh out" the user record with registration data (the things Firecloud asks for, name, organization, phone number, etc.) and syncs user information with Google as appropriate
1.  `inviteUser` - maybe just a proxy for `provisionUser`, or maybe provides slightly more information beyond provisioning, but certainly less than `registerUser` and does not do anything with Google yet
1.  `createUser` - provisions and registers the user 

It's possible that `provisionUser` and `inviteUser` would be very similar or identical and so maybe only one is needed?

In any case, it's not immediately clear to me what each of these methods do right now.  Furthermore, I found that just reading 1 of them wasn't enough for me to make sense out of any of them.  I had to read each line of all 4 methods to try to determine which method does what and which I might want to use in a particular scenario.    I don't see any issues in the logic, just the api.   This may very well be a nit, but I think if we ever wanted to change the names of these things, now would be the time to do it.

src/test/scala/Generator.scala
```diff
@@ -41,11 +41,16 @@ object Generator {
...
-  val genCreateWorkbenchUserAPI = for{
+  val genCreateWorkbenchUser = for{
     email <- genNonPetEmail
-    userId = genWorkbenchUserId(System.currentTimeMillis())
+    userId <- Gen.posNum[Int].map(x => genWorkbenchUserId(System.currentTimeMillis() + x))
   }yield CreateWorkbenchUser(userId, GoogleSubjectId(userId.value), email)
 
+  val genInviteUser = for{
+    email <- genNonPetEmail
+    userId = genWorkbenchUserId(System.currentTimeMillis())
```
gpolumbo-broad, line 51: Why are you using `Gen.posNum[Int]` on line 46 but not here?

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
```diff
@@ -39,6 +41,17 @@ class UserRoutesV1Spec extends UserRoutesSpecHelper{
     }
   }
 
+  "POST /api/v1/invite/user/{invitee's email}" should "create user" in{
+    val invitee = genInviteUser.sample.get
+
+    val (user, headers, _, routes) = createTestUser() //create a valid user that can invite someone
+    Post(s"/api/v1/invite/user/${invitee.inviteeEmail}").withHeaders(headers) ~> routes.route ~> check{
```
gpolumbo-broad, line 47: From `api-docs.yml` the route is: `/api/v1/invite/user/?email={inviteeEmail}` where the email is passed as a query parameter instead of being part of the path.  Which is the correct syntax?  Are both supported?  If so, there should be another test.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
```diff
@@ -203,21 +204,55 @@ class UserServiceSpec extends FlatSpec with Matchers with TestSupport with Mocki
...
+  "invite user and then create user with same email" should "update googleSubjectId for this user" in {
+    val user = genCreateWorkbenchUser.sample.get
+    service.inviteUser(InviteUser(user.id, user.email)).futureValue
+    val res = dirDAO.loadUser(user.id).futureValue
+    res shouldBe Some(WorkbenchUser(user.id, None, user.email))
+
+    service.createUser(user).futureValue
+    val updated = dirDAO.loadUser(user.id).futureValue
+    updated shouldBe Some(WorkbenchUser(user.id, Some(user.googleSubjectId), user.email))
   }
```
gpolumbo-broad, line 257: Pending my other comments from above regarding `UserService`, there should be tests for `provisionUser`.

gpolumbo-broad (changes requested):

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
Qi77Qi: before invite, there's just `createUser` (what's in `provisionUser` is just what breaks out of `createUser`, because I thought `inviteUser` needs to do all these steps too...I can put it back into `createUser` if it's less confusing now that this code is actually not shared)...`inviteUser` corresponds to `invite` endpoint, and it creates a user entry in ldap, but there's nothing else set up for this user.

src/test/scala/Generator.scala
Qi77Qi: Just to increase more randomness. But I don't think that's needed. I'll remove that

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
Qi77Qi: forgot to update the doc after updating code!

gpolumbo-broad (approved):

PR #211, opened 2018-09-19
set servcie account user

dvoet:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3759
need to set the service account user to make calls to google admin sdk

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -69,7 +69,7 @@ object Boot extends App with LazyLogging {
 
         val googleDirDaos = (googleServicesConfig.adminSdkServiceAccounts match {
           case None => NonEmptyList.one(Pem(WorkbenchEmail(googleServicesConfig.serviceAccountClientId), new File(googleServicesConfig.pemFile), Option(googleServicesConfig.subEmail)))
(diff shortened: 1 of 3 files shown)


gpolumbo-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -69,7 +69,7 @@ object Boot extends App with LazyLogging {
 
         val googleDirDaos = (googleServicesConfig.adminSdkServiceAccounts match {
           case None => NonEmptyList.one(Pem(WorkbenchEmail(googleServicesConfig.serviceAccountClientId), new File(googleServicesConfig.pemFile), Option(googleServicesConfig.subEmail)))
-          case Some(accounts) => accounts.map(account => Json(account.json))
+          case Some(accounts) => accounts.map(account => Json(account.json, Option(googleServicesConfig.subEmail)))
```
Qi77Qi, line 72: if `googleServicesConfig.subEmail` will not be `null`, shall we just use `Some`?

dvoet: it certainly should not be null but we follow the practice of using Option to avoid the possibility of Some(null).

Qi77Qi (approved):

dvoet:
unrelated leo test failure

PR #212, opened 2018-09-19
return auth domains

Qi77Qi:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3573

---

**PR checklist**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -61,3 +61,4 @@ scalacOptions in (Compile, console) --= Seq("-Ywarn-unused:imports", "-Xfatal-wa
 scalacOptions in Test -= "-Ywarn-dead-code" // due to https://github.com/mockito/mockito-scala#notes
 
 addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.2.4")
+addCompilerPlugin("org.scalamacros" %% "paradise" % "2.1.0" cross CrossVersion.full)
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -12,6 +12,8 @@ object Dependencies {
   val workbenchModelV  = "0.12-a19203d"
   val workbenchGoogleV = "0.16-0245949"
   val workbenchNotificationsV = "0.1-f2a0020"
+  val monocleVersion = "1.5.0-cats"
+  val catsEffectVersion = "1.0.0"
 
   val excludeAkkaActor =        ExclusionRule(organization = "com.typesafe.akka", name = "akka-actor_2.12")
   val excludeWorkbenchUtil =    ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-util_2.12")
@@ -40,7 +42,9 @@ object Dependencies {
 
   val googleOAuth2: ModuleID = "com.google.auth" % "google-auth-library-oauth2-http" % "0.9.0"
 
-  val catseffect: ModuleID =              "org.typelevel" %% "cats-effect" % "1.0.0"
+  val catseffect: ModuleID =  "org.typelevel" %% "cats-effect" % catsEffectVersion
+  val monocle: ModuleID = "com.github.julien-truffaut" %%  "monocle-core"  % monocleVersion
+  val mococleMacro: ModuleID = "com.github.julien-truffaut" %%  "monocle-macro" % monocleVersion
 
   val scalaTest: ModuleID =       "org.scalatest" %% "scalatest"    % scalaTestV % "test"
   val mockito: ModuleID =         "org.mockito"    % "mockito-core" % "2.7.22"   % "test"
@@ -80,6 +84,8 @@ object Dependencies {
     googleOAuth2,
 
     catseffect,
+    monocle,
+    mococleMacro,
 
     scalaTest,
     mockito,
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -40,7 +40,7 @@ object Dependencies {
 
   val googleOAuth2: ModuleID = "com.google.auth" % "google-auth-library-oauth2-http" % "0.9.0"
 
-  val cats: ModuleID =              "org.typelevel"       %% "cats-core"                  % "1.3.1"
+  val catseffect: ModuleID =              "org.typelevel" %% "cats-effect" % "1.0.0"
 
   val scalaTest: ModuleID =       "org.scalatest" %% "scalatest"    % scalaTestV % "test"
   val mockito: ModuleID =         "org.mockito"    % "mockito-core" % "2.7.22"   % "test"
@@ -79,7 +79,7 @@ object Dependencies {
 
     googleOAuth2,
 
-    cats,
+    catseffect,
 
     scalaTest,
     mockito,
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -44,7 +44,7 @@ object Dependencies {
 
   val catseffect: ModuleID =  "org.typelevel" %% "cats-effect" % catsEffectVersion
   val monocle: ModuleID = "com.github.julien-truffaut" %%  "monocle-core"  % monocleVersion
-  val mococleMacro: ModuleID = "com.github.julien-truffaut" %%  "monocle-macro" % monocleVersion
+  val monocleMacro: ModuleID = "com.github.julien-truffaut" %%  "monocle-macro" % monocleVersion
 
   val scalaTest: ModuleID =       "org.scalatest" %% "scalatest"    % scalaTestV % "test"
   val mockito: ModuleID =         "org.mockito"    % "mockito-core" % "2.7.22"   % "test"
@@ -85,7 +85,7 @@ object Dependencies {
 
     catseffect,
     monocle,
-    mococleMacro,
+    monocleMacro,
 
     scalaTest,
     mockito,
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1367,6 +1367,14 @@ definitions:
       accessPolicyName:
         type: string
         description: User's policy for the resource
+      authDomains:
+        type: array
+        items:
+          type: string
+      authDomainsUserNotAMemberOf:
+        type: array
+        items:
+          type: string
 
   ResourceRole:
     type: object
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1371,7 +1371,7 @@ definitions:
         type: array
         items:
           type: string
-      authDomainsUserNotAMemberOf:
+      missingAuthDomains:
         type: array
         items:
           type: string
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -33,20 +33,20 @@ import org.broadinstitute.dsde.workbench.util.DelegatePool
 import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.language.postfixOps
+import scala.concurrent.ExecutionContext.Implicits.global
 
 object Boot extends App with LazyLogging {
 
-  private def startup(): Unit = {
+  private def startup(): Future[Unit] = {
     val config = ConfigFactory.load()
-
-    val directoryConfig = config.as[DirectoryConfig]("directory")
-    val googleServicesConfigOption = config.getAs[GoogleServicesConfig]("googleServices")
-    val schemaLockConfig = config.as[SchemaLockConfig]("schemaLock")
 
     // we need an ActorSystem to host our application in
     implicit val system = ActorSystem("sam")
     implicit val materializer = ActorMaterializer()
-    import scala.concurrent.ExecutionContext.Implicits.global
+
+    val directoryConfig = config.as[DirectoryConfig]("directory")
+    val googleServicesConfigOption = config.getAs[GoogleServicesConfig]("googleServices")
+    val schemaLockConfig = config.as[SchemaLockConfig]("schemaLock")
 
     val dirURI = new URI(directoryConfig.directoryUrl)
     val (socketFactory, defaultPort) = dirURI.getScheme.toLowerCase match {
@@ -132,9 +132,11 @@ object Boot extends App with LazyLogging {
 
             _ <- IO.fromFuture(IO(cloudExtention.onBoot(SamApplication(userService, resourceService, statusService))))
 
-            _ <- IO.fromFuture(IO(Http().bindAndHandle(sRoutes.route, "0.0.0.0", 8080))).handleErrorWith{
+            binding <- IO.fromFuture(IO(Http().bindAndHandle(sRoutes.route, "0.0.0.0", 8080))).handleErrorWith{
               case t: Throwable => IO(logger.error("FATAL - failure starting http server", t)) *> IO.raiseError(t)
             }
+            _ <- IO.fromFuture(IO(binding.whenTerminated))
+            _ <- IO(system.terminate())
           } yield ()
       }
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -61,7 +61,7 @@ object Boot extends App with LazyLogging {
 
     val resourceTypes = config.as[Map[String, ResourceType]]("resourceTypes").values.toSet
     val resourceTypeMap = resourceTypes.map(rt => rt.name -> rt).toMap
-    val accessPolicyDAO = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig, resourceTypeMap)
+    val accessPolicyDAO = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig)
 
     val cloudExt = googleServicesConfigOption match {
       case Some(googleServicesConfig) =>
@@ -114,7 +114,7 @@ object Boot extends App with LazyLogging {
           throw t
       }
 
-      _ <- Future.traverse(resourceTypes.map(_.name)) { accessPolicyDAO.createResourceType } recover {
+      _ <- Future.traverse(resourceTypes.map(_.name)) { x => accessPolicyDAO.createResourceType(x).unsafeToFuture() } recover {
         case t: Throwable =>
           logger.error("FATAL - unable to init resource types", t)
           throw t
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -62,9 +62,8 @@ object Boot extends App with LazyLogging {
 
     val resourceTypes = config.as[Map[String, ResourceType]]("resourceTypes").values.toSet
     val resourceTypeMap = resourceTypes.map(rt => rt.name -> rt).toMap
-    def accessPolicyDAO(blockingEc: ExecutionContext) = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig, blockingEc)
 
-    def cloudExt(accessPolicyDAO: AccessPolicyDAO) = googleServicesConfigOption match {
+    def createCloudExt(accessPolicyDAO: AccessPolicyDAO) = googleServicesConfigOption match {
       case Some(googleServicesConfig) =>
         val petServiceAccountConfig = config.as[PetServiceAccountConfig]("petServiceAccount")
 
@@ -88,7 +87,7 @@ object Boot extends App with LazyLogging {
     }
 
 
-    def samRoutes(cloudExtensions: CloudExtensions, accessPolicyDAO: AccessPolicyDAO): (SamRoutes, UserService, ResourceService, StatusService) = {
+    def createSamRoutes(cloudExtensions: CloudExtensions, accessPolicyDAO: AccessPolicyDAO): (SamRoutes, UserService, ResourceService, StatusService) = {
       // TODO - https://broadinstitute.atlassian.net/browse/GAWB-3603
       // This should JUST get the value from "emailDomain", but for now we're keeping the backwards compatibility code to
       // fall back to getting the "googleServices.appsDomain"
@@ -99,14 +98,14 @@ object Boot extends App with LazyLogging {
       val statusService = new StatusService(directoryDAO, cloudExtensions, 10 seconds)
       val managedGroupService = new ManagedGroupService(resourceService, resourceTypeMap, accessPolicyDAO, directoryDAO, cloudExtensions, emailDomain)
 
-      val samR = cloudExtensions match {
+      val samRoutes = cloudExtensions match {
         case googleExt: GoogleExtensions => new SamRoutes(resourceService, userService, statusService, managedGroupService, config.as[SwaggerConfig]("swagger"), directoryDAO) with StandardUserInfoDirectives with GoogleExtensionRoutes {
             val googleExtensions = googleExt
             val cloudExtensions = googleExt
           }
         case _ => new SamRoutes(resourceService, userService, statusService, managedGroupService, config.as[SwaggerConfig]("swagger"), directoryDAO) with StandardUserInfoDirectives with NoExtensionRoutes
       }
-      (samR, userService, resourceService, statusService)
+      (samRoutes, userService, resourceService, statusService)
     }
 
     for {
@@ -122,12 +121,12 @@ object Boot extends App with LazyLogging {
       io = ExecutionContexts.blockingThreadPool.use{
         blockingEc =>
           implicit val cs = IO.contextShift(scala.concurrent.ExecutionContext.Implicits.global)
-          val accessPolicyD = accessPolicyDAO(blockingEc)
-          val cloudExtention = cloudExt(accessPolicyD)
-          val (sRoutes, userService, resourceService, statusService) = samRoutes(cloudExtention, accessPolicyD)
+          val accessPolicyDao = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig, blockingEc)
+          val cloudExtention = createCloudExt(accessPolicyDao)
+          val (sRoutes, userService, resourceService, statusService) = createSamRoutes(cloudExtention, accessPolicyDao)
 
           for{
-            _ <- resourceTypes.toList.map(rt => accessPolicyD.createResourceType(rt.name)).parSequence.handleErrorWith{
+            _ <- resourceTypes.toList.map(rt => accessPolicyDao.createResourceType(rt.name)).parSequence.handleErrorWith{
               case t: Throwable => IO(logger.error("FATAL - failure starting http server", t)) *> IO.raiseError(t)
             }
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -56,12 +56,12 @@ object Boot extends App with LazyLogging {
     val port = if (dirURI.getPort > 0) dirURI.getPort else defaultPort
     val ldapConnectionPool = new LDAPConnectionPool(new LDAPConnection(socketFactory, dirURI.getHost, port, directoryConfig.user, directoryConfig.password), directoryConfig.connectionPoolSize)
 
-    val accessPolicyDAO = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig)
     val directoryDAO = new LdapDirectoryDAO(ldapConnectionPool, directoryConfig)
     val schemaDAO = new JndiSchemaDAO(directoryConfig, schemaLockConfig)
 
     val resourceTypes = config.as[Map[String, ResourceType]]("resourceTypes").values.toSet
     val resourceTypeMap = resourceTypes.map(rt => rt.name -> rt).toMap
+    val accessPolicyDAO = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig, resourceTypeMap)
 
     val cloudExt = googleServicesConfigOption match {
       case Some(googleServicesConfig) =>
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -6,13 +6,14 @@ import java.net.URI
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.stream.ActorMaterializer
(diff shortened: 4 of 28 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -107,15 +112,28 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
...
+    for{
+      resourceId <- getAttribute(entry, Attr.resourceId).toRight(s"${entry.getDN} attribute missing: ${Attr.resourceId}. ")
+      policyName <- getAttribute(entry, Attr.policy).toRight(s"${entry.getDN} attribute missing: ${Attr.policy}. ")
+    } yield ResourceIdAndPolicyName(ResourceId(resourceId), AccessPolicyName(policyName), authDomains, authDomains.filterNot(gn => resourcesUserIsMember.contains(gn)))
+  }
 
-    groupDns.collect {
-      case policyDnPattern(policyName, resourceId) => ResourceIdAndPolicyName(ResourceId(resourceId), AccessPolicyName(policyName))
-    }
+  protected[openam] def listAccessPoliciesIO(resourceTypeName: ResourceTypeName, userId: WorkbenchUserId): IO[Set[ResourceIdAndPolicyName]] = {
```
dvoet: my main concern with this is that it packs too much functionality into the DAO. The DAO should be about how to query, marshal and unmarshal the data. I think the composition of that logic should go into the ResourceService.

build.sbt
```diff
@@ -61,3 +61,4 @@ scalacOptions in (Compile, console) --= Seq("-Ywarn-unused:imports", "-Xfatal-wa
 scalacOptions in Test -= "-Ywarn-dead-code" // due to https://github.com/mockito/mockito-scala#notes
 
 addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.2.4")
+addCompilerPlugin("org.scalamacros" %% "paradise" % "2.1.0" cross CrossVersion.full)
```
dvoet, line 64: what does this do?

project/Dependencies.scala
```diff
@@ -40,7 +42,9 @@ object Dependencies {
 
   val googleOAuth2: ModuleID = "com.google.auth" % "google-auth-library-oauth2-http" % "0.9.0"
 
-  val cats: ModuleID =              "org.typelevel"       %% "cats-core"                  % "1.3.1"
+  val catseffect: ModuleID =  "org.typelevel" %% "cats-effect" % catsEffectVersion
+  val monocle: ModuleID = "com.github.julien-truffaut" %%  "monocle-core"  % monocleVersion
+  val mococleMacro: ModuleID = "com.github.julien-truffaut" %%  "monocle-macro" % monocleVersion
```
dvoet: spelling: mococleMacro

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1367,6 +1367,14 @@ definitions:
       accessPolicyName:
         type: string
         description: User's policy for the resource
+      authDomains:
+        type: array
+        items:
+          type: string
+      authDomainsUserNotAMemberOf:
```
dvoet: how about missingAuthDomains

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -107,15 +112,34 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
...
+      groupDns.collect{case policyDnPattern(policyName, resourceId) => ResourceIdAndPolicyName(ResourceId(resourceId), AccessPolicyName(policyName))}
+    }
+  }
 
-    val policyDnPattern = dnMatcher(List(Attr.policy, Attr.resourceId), resourceTypeDn(resourceTypeName))
+  override def listResourceWithAuthdomains(resourceTypeName: ResourceTypeName, resourceId: Set[ResourceId]): IO[Set[Resource]] = {
+    val filters = resourceId.grouped(batchSize).map(
+      batch =>
+        Filter.createORFilter(
+          batch.map(r => Filter.createANDFilter(Filter.createEqualityFilter(Attr.resourceId, r.value), Filter.createNOTFilter(Filter.createPresenceFilter(Attr.policy)))
```
dvoet: filter on objectclass instead of `Filter.createNOTFilter(Filter.createPresenceFilter(Attr.policy))`

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -107,15 +112,34 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
...
+          ).asJava)).toSeq
+
+    for{
+      stream <- runBlocking(IO(ldapSearchStream(resourcesOu, SearchScope.SUB, filters:_*)(et => unmarshalResourceAuthDomain(et, resourceTypeName))))
```
dvoet: probably also use the resource type dn as the search base and don't descend to lower level.

src/main/scala/org/broadinstitute/dsde/workbench/sam/package.scala
```diff
@@ -1,10 +1,21 @@
...
 /**
   * Created by dvoet on 5/18/17.
   */
 package object sam {
   implicit val errorReportSource = ErrorReportSource("sam")
+
+  private val cachedThreadPool = Executors.newCachedThreadPool()
+  private val BlockingIO   = ExecutionContext.fromExecutor(cachedThreadPool)
+
+  def runBlocking[A](io: IO[A])(implicit cs: ContextShift[IO]): IO[A] = IO.shift(BlockingIO) *> io <* IO.shift
```
dvoet: some comments here:
1) I don't think we should use an implicit for cs.
2) This is not the right place to define a thread pool. If it is here, it is defined statically and can't be overridden or configured. This kind of thing is typically constructed in Boot and passed into the places that need it.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -16,39 +18,42 @@ import scala.collection.JavaConverters._
...
   }
 
-  override def loadResourceAuthDomain(resource: Resource): Future[Set[WorkbenchGroupName]] = {
-    Future(Option(ldapConnectionPool.getEntry(resourceDn(resource)))).flatMap {
-      case None => Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"$resource auth domain not found")))
-      case Some(r) => Future.successful(getAttributes(r, Attr.authDomain).getOrElse(Set.empty).toSet.map(g => WorkbenchGroupName(g)))
+  override def loadResourceAuthDomain(resource: Resource): IO[Set[WorkbenchGroupName]] = {
+    runBlocking((IO(Option(ldapConnectionPool.getEntry(resourceDn(resource)))))).flatMap {
+      case None =>
+        IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"$resource auth domain not found")))
```
dvoet: it is not that the auth domain was not found, it is resource that was not found

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -107,15 +112,34 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
     newPolicy
   }
 
-  override def listAccessPolicies(resourceTypeName: ResourceTypeName, userId: WorkbenchUserId): Future[Set[ResourceIdAndPolicyName]] = Future {
-    val entry = ldapConnectionPool.getEntry(userDn(userId), Attr.memberOf)
-    val groupDns = getAttributes(entry, Attr.memberOf).getOrElse(Set.empty).toSet
+  override def listAccessPolicies(resourceTypeName: ResourceTypeName, userId: WorkbenchUserId): IO[Set[ResourceIdAndPolicyName]] = {
+    val policyDnPattern = dnMatcher(Seq(Attr.policy, Attr.resourceId), resourceTypeDn(resourceTypeName))
```
dvoet: what happens if an exception is thrown on this line as it is not within an IO block?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -107,15 +112,34 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
...
+  override def listResourceWithAuthdomains(resourceTypeName: ResourceTypeName, resourceId: Set[ResourceId]): IO[Set[Resource]] = {
+    val filters = resourceId.grouped(batchSize).map(
+      batch =>
+        Filter.createORFilter(
+          batch.map(r => Filter.createANDFilter(Filter.createEqualityFilter(Attr.resourceId, r.value), Filter.createNOTFilter(Filter.createPresenceFilter(Attr.policy)))
+          ).asJava)).toSeq
+
+    for{
+      stream <- runBlocking(IO(ldapSearchStream(resourcesOu, SearchScope.SUB, filters:_*)(et => unmarshalResourceAuthDomain(et, resourceTypeName))))
+      res <- IO.fromEither(stream.toList.parSequence.leftMap(s => new WorkbenchException(s)))
```
dvoet: I am having trouble figuring out what this line is doing. Why toList and parSequence? One reason I have trouble is because intelliJ does not like parSequence so it is loosing the types, etc.

I am a little concerned because this sucks all the results of the stream into memory

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -134,9 +137,28 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
 
-  def listUserAccessPolicies(resourceType: ResourceType, userInfo: UserInfo): Future[Set[ResourceIdAndPolicyName]] = {
-    accessPolicyDAO.listAccessPolicies(resourceType.name, userInfo.userId)
-  }
+  def listUserAccessPolicies(resourceTypeName: ResourceTypeName, userId: WorkbenchUserId): IO[Set[UserPolicyResponse]] = for{
+      rt <- IO.fromEither(resourceTypes.get(resourceTypeName).toRight(new WorkbenchException(s"missing configration for resourceType ${resourceTypeName}")))
+      isConstrained = rt.isAuthDomainConstrainable
+      ridAndPolcyName <- accessPolicyDAO.listAccessPolicies(resourceTypeName, userId) // List all policies of a given resourceType the user is a member of
+      rids = ridAndPolcyName.map(_.resourceId)
+      resources <- accessPolicyDAO.listResourceWithAuthdomains(resourceTypeName, rids)
```
dvoet: we should only do this if isConstrained

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -134,9 +137,28 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+          else{
+            for{
+              authDomains <- EitherT.fromEither[IO](authDomainMap.get(rnp.resourceId).toRight(s"no auth domain found for ${rnp.resourceId}"))
+              allAuthDomainResourcesUserIsMemberOf <- if(authDomains.nonEmpty) accessPolicyDAO.listAccessPolicies(ManagedGroupService.managedGroupTypeName, userId).attemptT.leftMap(_.getLocalizedMessage) else EitherT.fromEither[IO](Set.empty[ResourceIdAndPolicyName].asRight[String])
+              userNotMemberOf = authDomains.filterNot(x => allAuthDomainResourcesUserIsMemberOf.map(_.resourceId).contains(ResourceId(x.value)))
+            } yield UserPolicyResponse(rnp.resourceId, rnp.accessPolicyName, authDomains, userNotMemberOf)
+          }
+      }
+      r <- results.parSequence.value
+      rr <- IO.fromEither(r.leftMap(s => new WorkbenchException(s)))
```
dvoet: better val names

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -134,9 +137,28 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+            EitherT.fromEither[IO](UserPolicyResponse(rnp.resourceId, rnp.accessPolicyName, Set.empty, Set.empty).asRight[String])
+          else{
+            for{
+              authDomains <- EitherT.fromEither[IO](authDomainMap.get(rnp.resourceId).toRight(s"no auth domain found for ${rnp.resourceId}"))
+              allAuthDomainResourcesUserIsMemberOf <- if(authDomains.nonEmpty) accessPolicyDAO.listAccessPolicies(ManagedGroupService.managedGroupTypeName, userId).attemptT.leftMap(_.getLocalizedMessage) else EitherT.fromEither[IO](Set.empty[ResourceIdAndPolicyName].asRight[String])
+              userNotMemberOf = authDomains.filterNot(x => allAuthDomainResourcesUserIsMemberOf.map(_.resourceId).contains(ResourceId(x.value)))
+            } yield UserPolicyResponse(rnp.resourceId, rnp.accessPolicyName, authDomains, userNotMemberOf)
+          }
+      }
+      r <- results.parSequence.value
```
dvoet: I am having trouble understanding what is executing in parallel

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -134,9 +137,28 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+      resources <- accessPolicyDAO.listResourceWithAuthdomains(resourceTypeName, rids)
+      authDomainMap = resources.map(x => x.resourceId -> x.authDomain).toMap
+      results = ridAndPolcyName.toList.map{
+        rnp =>
+          if(!isConstrained)
+            EitherT.fromEither[IO](UserPolicyResponse(rnp.resourceId, rnp.accessPolicyName, Set.empty, Set.empty).asRight[String])
+          else{
+            for{
+              authDomains <- EitherT.fromEither[IO](authDomainMap.get(rnp.resourceId).toRight(s"no auth domain found for ${rnp.resourceId}"))
+              allAuthDomainResourcesUserIsMemberOf <- if(authDomains.nonEmpty) accessPolicyDAO.listAccessPolicies(ManagedGroupService.managedGroupTypeName, userId).attemptT.leftMap(_.getLocalizedMessage) else EitherT.fromEither[IO](Set.empty[ResourceIdAndPolicyName].asRight[String])
```
dvoet: we should execute this query only once, not for each ridAndPolcyName

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -134,9 +137,28 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+      rids = ridAndPolcyName.map(_.resourceId)
+      resources <- accessPolicyDAO.listResourceWithAuthdomains(resourceTypeName, rids)
+      authDomainMap = resources.map(x => x.resourceId -> x.authDomain).toMap
+      results = ridAndPolcyName.toList.map{
+        rnp =>
+          if(!isConstrained)
+            EitherT.fromEither[IO](UserPolicyResponse(rnp.resourceId, rnp.accessPolicyName, Set.empty, Set.empty).asRight[String])
+          else{
+            for{
+              authDomains <- EitherT.fromEither[IO](authDomainMap.get(rnp.resourceId).toRight(s"no auth domain found for ${rnp.resourceId}"))
```
dvoet: I am not certain, but it seems that this code treats a resource with no auth domain as an error. That is not true. The resource type may allow auth domains but that does not mean all resources of that type have them.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -134,9 +137,28 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+  def listUserAccessPolicies(resourceTypeName: ResourceTypeName, userId: WorkbenchUserId): IO[Set[UserPolicyResponse]] = for{
+      rt <- IO.fromEither(resourceTypes.get(resourceTypeName).toRight(new WorkbenchException(s"missing configration for resourceType ${resourceTypeName}")))
+      isConstrained = rt.isAuthDomainConstrainable
+      ridAndPolcyName <- accessPolicyDAO.listAccessPolicies(resourceTypeName, userId) // List all policies of a given resourceType the user is a member of
+      rids = ridAndPolcyName.map(_.resourceId)
+      resources <- accessPolicyDAO.listResourceWithAuthdomains(resourceTypeName, rids)
+      authDomainMap = resources.map(x => x.resourceId -> x.authDomain).toMap
+      results = ridAndPolcyName.toList.map{
+        rnp =>
+          if(!isConstrained)
```
dvoet: I would flip the if/else: `if(isConstrained) ... else ...` It is easier to read, the `!` is too easy to miss

dvoet (changes requested):
that's enough comments for now ;)

build.sbt
Qi77Qi: it's the macro enabling `@lenses` annotation

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
Qi77Qi: I was thinking building the regex won't throw exception, but maybe it's safer to build it inside IO

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
Qi77Qi: `toList` is needed since there's no `Paralell` for `Set`.  `parSequence` turns `List[Either[String, Resource]]` into `Either[String, List[Resource]` (the resulting `String` is concatenated value in `List[Either[String, Resource]]`). 

```
qi-sam@   val strErr1: Either[String, String] = "error 1. ".asLeft[String]
strErr1: Either[String, String] = Left("error 1. ")

qi-sam@   val strErr2: Either[String, String] = "error 2. ".asLeft[String]
strErr2: Either[String, String] = Left("error 2. ")

qi-sam@   val str3: Either[String, String] = "this is fine".asRight[String]
str3: Either[String, String] = Right("this is fine")

qi-sam@

qi-sam@   val result2: Either[String, List[String]] = List(strErr1, strErr2, str3).parSequence
result2: Either[String, List[String]] = Left("error 1. error 2. ")

```

The API is defined with returning a `Set`, so we're loading the whole stream anyway. If that's an issue, we'll probably need to do something else.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
Qi77Qi: The error message is a bit confusing.

`ridAndPolcyName <- accessPolicyDAO.listAccessPolicies(resourceTypeName, userId)`
```      
resources <- if(isConstrained) accessPolicyDAO.listResourceWithAuthdomains(resourceTypeName, rids) else IO.pure(Set.empty)
      authDomainMap = resources.map(x => x.resourceId -> x.authDomain).toMap
```

this line is saying if there's a resource that exists in `ridAndPolcyName`, but not in `authDomainMap` that's returned by the second query, we treat it as an error. Up for suggestions for the error message

Qi77Qi: just a reminder to myself..log it

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -86,51 +87,60 @@ object Boot extends App with LazyLogging {
...
+      val resourceService = new ResourceService(resourceTypeMap, accessPolicyDAO, directoryDAO, cloudExtensions, emailDomain)
+      val userService = new UserService(directoryDAO, cloudExtensions)
+      val statusService = new StatusService(directoryDAO, cloudExtensions, 10 seconds)
+      val managedGroupService = new ManagedGroupService(resourceService, resourceTypeMap, accessPolicyDAO, directoryDAO, cloudExtensions, emailDomain)
 
-    val samRoutes = cloudExt match {
-      case googleExt: GoogleExtensions => new SamRoutes(resourceService, userService, statusService, managedGroupService, config.as[SwaggerConfig]("swagger"), directoryDAO) with StandardUserInfoDirectives with GoogleExtensionRoutes {
-        val googleExtensions = googleExt
-        val cloudExtensions = googleExt
+      val samR = cloudExtensions match {
```
dvoet: let's keep abbreviations to a minimum please

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -56,14 +57,14 @@ object Boot extends App with LazyLogging {
...
-    val accessPolicyDAO = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig)
     val directoryDAO = new LdapDirectoryDAO(ldapConnectionPool, directoryConfig)
     val schemaDAO = new JndiSchemaDAO(directoryConfig, schemaLockConfig)
 
     val resourceTypes = config.as[Map[String, ResourceType]]("resourceTypes").values.toSet
     val resourceTypeMap = resourceTypes.map(rt => rt.name -> rt).toMap
+    def accessPolicyDAO(blockingEc: ExecutionContext) = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig, blockingEc)
 
-    val cloudExt = googleServicesConfigOption match {
+    def cloudExt(accessPolicyDAO: AccessPolicyDAO) = googleServicesConfigOption match {
```
dvoet: maybe call it createCloudExt and createAccessPolicyDAO above

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -86,51 +87,60 @@ object Boot extends App with LazyLogging {
...
 
-      _ <- Future.traverse(resourceTypes.map(_.name)) { accessPolicyDAO.createResourceType } recover {
-        case t: Throwable =>
-          logger.error("FATAL - unable to init resource types", t)
-          throw t
-      }
+      io = ExecutionContexts.blockingThreadPool.use{
+        blockingEc =>
+          implicit val cs = IO.contextShift(scala.concurrent.ExecutionContext.Implicits.global)
+          val accessPolicyD = accessPolicyDAO(blockingEc)
```
dvoet: abbreviation

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -107,27 +109,47 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
     newPolicy
   }
 
-  override def listAccessPolicies(resourceTypeName: ResourceTypeName, userId: WorkbenchUserId): Future[Set[ResourceIdAndPolicyName]] = Future {
-    val entry = ldapConnectionPool.getEntry(userDn(userId), Attr.memberOf)
-    val groupDns = getAttributes(entry, Attr.memberOf).getOrElse(Set.empty).toSet
+  override def listAccessPolicies(resourceTypeName: ResourceTypeName, userId: WorkbenchUserId): IO[Set[ResourceIdAndPolicyName]] = {
+    for{
+      policyDnPattern <- IO(dnMatcher(Seq(Attr.policy, Attr.resourceId), resourceTypeDn(resourceTypeName)))
```
dvoet, line 114: why put policyDnPattern in an IO instead of in the yield?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -123,38 +126,58 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+      resources <- if(isConstrained) accessPolicyDAO.listResourceWithAuthdomains(resourceTypeName, rids) else IO.pure(Set.empty)
+      authDomainMap = resources.map(x => x.resourceId -> x.authDomain).toMap
+
+      allAuthDomainResourcesUserIsMemberOf <- if(isConstrained) accessPolicyDAO.listAccessPolicies(ManagedGroupService.managedGroupTypeName, userId) else IO.pure(Set.empty[ResourceIdAndPolicyName])
+
+      results = ridAndPolcyName.toList.map{
+        rnp =>
+          if(isConstrained){
+            for{
+              authDomains <- authDomainMap.get(rnp.resourceId).toRight(s"no auth domain found for ${rnp.resourceId}")
```
dvoet: as we discussed, this error condition should be logged at the error level and the resource should be filtered from the response. Thus leaving this api call functional for the user, if not complete.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -123,38 +126,58 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
+    accessPolicyDAO.createPolicy(AccessPolicy(resourceAndPolicyName, members, email, roles, actions)).unsafeToFuture()
   }
 
-  def listUserAccessPolicies(resourceType: ResourceType, userInfo: UserInfo): Future[Set[ResourceIdAndPolicyName]] = {
-    accessPolicyDAO.listAccessPolicies(resourceType.name, userInfo.userId)
-  }
+  def listUserAccessPolicies(resourceTypeName: ResourceTypeName, userId: WorkbenchUserId): IO[Set[UserPolicyResponse]] = for{
+      rt <- IO.fromEither(resourceTypes.get(resourceTypeName).toRight(new WorkbenchException(s"missing configration for resourceType ${resourceTypeName}")))
+      isConstrained = rt.isAuthDomainConstrainable
+      ridAndPolcyName <- accessPolicyDAO.listAccessPolicies(resourceTypeName, userId) // List all policies of a given resourceType the user is a member of
```
marctalbott: Spelling: ridAndPolicyName, just missing an 'i' in policy

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -573,29 +580,71 @@ class ResourceServiceSpec extends FlatSpec with Matchers with TestSupport with B
...
+      _ <- constrainableService.createResource(constrainableResourceType, resource.resourceId, Map(viewPolicyName -> constrainablePolicyMembership), resource.authDomain, dummyUserInfo)
+      _ <- dirDAO.createUser(WorkbenchUser(user.userId, Some(GoogleSubjectId(user.userId.value)), user.userEmail))
+      _ <- constrainableService.createPolicy(policy.id, policy.members + user.userId, policy.roles, policy.actions)
+      r <- constrainableService.listUserAccessPolicies(constrainableResourceType.name, user.userId).unsafeToFuture()
+    } yield r
+
+    val expected = Set(UserPolicyResponse(resource.resourceId, policy.id.accessPolicyName, resource.authDomain, resource.authDomain))
+    res.futureValue(Timeout(Span(10, Seconds))) shouldBe expected
+  }
+
```
dvoet, line 648: add a test for listUserAccessPolicies where there is a resource in a constrainable type but does not have any auth domains

dvoet (changes requested):
last time I think

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -28,7 +31,7 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
     Future.successful(resourceTypes.get(name))
   }
 
-  def createResourceType(resourceType: ResourceType): Future[ResourceTypeName] = {
+  def createResourceType(resourceType: ResourceType): IO[ResourceTypeName] = {
```
marctalbott, line 34: Out of curiosity, why was this return type changed to an `IO` while other functions within `ResourceService` retained their return type and simply had `.unsafeToFuture` tacked on to the end instead? Just an effort to limit the amount of Future -> IO conversions?

marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
Qi77Qi: thanks for review!...yea..it's not consistent right now...I touched those that I need to interact with, but left the rest as is. I'll probably convert more in separate PRs just in case I accidentally change semantics

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
Qi77Qi: I looked at `.r`, and it will indeed throw exception if there's syntax error. So it should be in `IO`. Only pure values go to `yield`

dvoet (approved):

Qi77Qi:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -86,51 +86,60 @@ object Boot extends App with LazyLogging {
...
-          throw t
+          Future.failed(t)
       }
 
-      _ <- Future.traverse(resourceTypes.map(_.name)) { accessPolicyDAO.createResourceType } recover {
-        case t: Throwable =>
-          logger.error("FATAL - unable to init resource types", t)
-          throw t
-      }
+      io = ExecutionContexts.blockingThreadPool.use{
```
dvoet, line 121: I think the tests are failing because `blockingThreadPool` gets released when this block completes but _this block completes after successful start_, not when the app is done.

PR #213, opened 2018-09-21
sync intersection group for constrainable policy when policy or auth domain is updated

marctalbott:
Ticket: [GAWB-3570](https://broadinstitute.atlassian.net/browse/GAWB-3570), [GAWB-3574](https://broadinstitute.atlassian.net/browse/GAWB-3574)
- When policies are updated or created, Sam now looks to see if they are constrainable by checking if any of the policy's actions are constrainable. If the policy is constrainable, then the intersection of the policy's members and the resource's auth domain is calculated and synced with Google
- When managed groups are updated or created, Sam now checks whether there are any policies that are constrained by it and updates those policies
---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -82,7 +82,7 @@ object Boot extends App with LazyLogging {
         val googleKeyCache = new GoogleKeyCache(googleIamDAO, googleStorageDAO, googlePubSubDAO, googleServicesConfig, petServiceAccountConfig)
         val notificationDAO = new PubSubNotificationDAO(googlePubSubDAO, googleServicesConfig.notificationTopic)
 
-        new GoogleExtensions(directoryDAO, accessPolicyDAO, googleDirectoryDAO, googlePubSubDAO, googleIamDAO, googleStorageDAO, googleProjectDAO, googleKeyCache, notificationDAO, googleServicesConfig, petServiceAccountConfig, resourceTypeMap(CloudExtensions.resourceTypeName))
+        new GoogleExtensions(directoryDAO, accessPolicyDAO, googleDirectoryDAO, googlePubSubDAO, googleIamDAO, googleStorageDAO, googleProjectDAO, googleKeyCache, notificationDAO, googleServicesConfig, petServiceAccountConfig, resourceTypeMap)
       case None => NoExtensions
     }
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
@@ -228,14 +228,7 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
 
   override def listIntersectionGroupUsers(groupIds: Set[WorkbenchGroupIdentity]): Future[Set[WorkbenchUserId]] = Future {
     ldapSearchStream(directoryConfig.baseDn, SearchScope.SUB, Filter.createANDFilter(groupIds.map(groupId =>
-      Filter.createEqualityFilter(Attr.memberOf, groupDn(groupId))).asJava))(unmarshalUserOption).collect { case Some(user) => user.id }.toSet
-  }
-
-  private def unmarshalUserOption(results: Entry): Option[WorkbenchUser] = {
-    for {
-      uid <- getAttribute(results, Attr.uid)
-      email <- getAttribute(results, Attr.email)
-    } yield WorkbenchUser(WorkbenchUserId(uid), getAttribute(results, Attr.googleSubjectId).map(GoogleSubjectId), WorkbenchEmail(email))
+      Filter.createEqualityFilter(Attr.memberOf, groupDn(groupId))).asJava))(getAttribute(_, Attr.uid)).flatten.map(WorkbenchUserId).toSet
   }
 
   override def listAncestorGroups(groupId: WorkbenchGroupIdentity): Future[Set[WorkbenchGroupIdentity]] = Future {
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
@@ -226,8 +226,16 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
     groupsOption.getOrElse(Set.empty)
   }
 
-  override def listFlattenedGroupUsers(groupId: WorkbenchGroupIdentity): Future[Set[WorkbenchUserId]] = Future {
-    ldapSearchStream(peopleOu, SearchScope.SUB, Filter.createEqualityFilter(Attr.memberOf, groupDn(groupId)))(unmarshalUser).map(_.id).toSet
+  override def listIntersectionGroupUsers(groupIds: Set[WorkbenchGroupIdentity]): Future[Set[WorkbenchUserId]] = Future {
+    ldapSearchStream(directoryConfig.baseDn, SearchScope.SUB, Filter.createANDFilter(groupIds.map(groupId =>
+      Filter.createEqualityFilter(Attr.memberOf, groupDn(groupId))).asJava))(unmarshalUserOption).collect { case Some(user) => user.id }.toSet
+  }
+
+  private def unmarshalUserOption(results: Entry): Option[WorkbenchUser] = {
+    for {
+      uid <- getAttribute(results, Attr.uid)
+      email <- getAttribute(results, Attr.email)
+    } yield WorkbenchUser(WorkbenchUserId(uid), getAttribute(results, Attr.googleSubjectId).map(GoogleSubjectId), WorkbenchEmail(email))
   }
 
   override def listAncestorGroups(groupId: WorkbenchGroupIdentity): Future[Set[WorkbenchGroupIdentity]] = Future {
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -5,6 +5,8 @@ import java.util.Date
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
+//import cats.effect.IO
+//import cats.implicits._
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.auth.oauth2.ServiceAccountCredentials
@@ -38,7 +40,7 @@ object GoogleExtensions {
   val getPetPrivateKeyAction = ResourceAction("get_pet_private_key")
 }
 
-class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: AccessPolicyDAO, val googleDirectoryDAO: GoogleDirectoryDAO, val googlePubSubDAO: GooglePubSubDAO, val googleIamDAO: GoogleIamDAO, val googleStorageDAO: GoogleStorageDAO, val googleProjectDAO: GoogleProjectDAO, val googleKeyCache: GoogleKeyCache, val notificationDAO: NotificationDAO, val googleServicesConfig: GoogleServicesConfig, val petServiceAccountConfig: PetServiceAccountConfig, extensionResourceType: ResourceType)(implicit val system: ActorSystem, executionContext: ExecutionContext) extends LazyLogging with FutureSupport with CloudExtensions with Retry {
+class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: AccessPolicyDAO, val googleDirectoryDAO: GoogleDirectoryDAO, val googlePubSubDAO: GooglePubSubDAO, val googleIamDAO: GoogleIamDAO, val googleStorageDAO: GoogleStorageDAO, val googleProjectDAO: GoogleProjectDAO, val googleKeyCache: GoogleKeyCache, val notificationDAO: NotificationDAO, val googleServicesConfig: GoogleServicesConfig, val petServiceAccountConfig: PetServiceAccountConfig, val resourceTypes: Map[ResourceTypeName, ResourceType])(implicit val system: ActorSystem, executionContext: ExecutionContext) extends LazyLogging with FutureSupport with CloudExtensions with Retry {
 
   private val maxGroupEmailLength = 64
 
@@ -86,7 +88,7 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
       this
     ))
 
-
+    val extensionResourceType = resourceTypes.getOrElse(CloudExtensions.resourceTypeName, throw new Exception(s"${CloudExtensions.resourceTypeName} resource type not found"))
     val ownerGoogleSubjectId = GoogleSubjectId(googleServicesConfig.serviceAccountClientId)
     for {
       user <- directoryDAO.loadSubjectFromGoogleSubjectId(ownerGoogleSubjectId)
@@ -127,6 +129,17 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
         case (rpn: ResourceAndPolicyName, Some(_)) => rpn.toJson.compactPrint
       }
       _ <- googlePubSubDAO.publishMessages(googleServicesConfig.groupSyncTopic, messagesForIdsWithSyncDates)
+      ancestorGroups <- Future.traverse(groupIdentities) { id => directoryDAO.listAncestorGroups(id) }
+      managedGroupIds = (ancestorGroups.flatten ++ groupIdentities).collect { case ResourceAndPolicyName(Resource(ResourceTypeName("managed-group"), id, _), _) => id }
+      _ <- Future.traverse(managedGroupIds)(onManagedGroupUpdate)
+    } yield ()
+  }
+
+  private def onManagedGroupUpdate(groupId: ResourceId): Future[Unit] = {
+    for {
+      resources <- accessPolicyDAO.listResourcesConstrainedByGroup(WorkbenchGroupName(groupId.value))
+      policies <- Future.traverse(resources) { resource => accessPolicyDAO.listAccessPolicies(resource).unsafeToFuture }
+      _ <- onGroupUpdate(policies.flatten.map(_.id).toList)
     } yield ()
   }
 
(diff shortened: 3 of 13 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -450,6 +440,42 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
     }
   }
 
+  override def synchronizeGroupMembers(groupMembers: Future[Set[WorkbenchUserId]], policy: AccessPolicy): Future[Map[WorkbenchEmail, Seq[SyncReportItem]]] = {
```
dvoet: this should not be a separate function, syncing intersection groups should be part of regular syncing.

dvoet: also having a parameter that is a Future doesn't feel right.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -221,11 +221,41 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
     val workbenchSubjects = policy.emailsToSubjects.values.flatten.toSet
+    val constrainable = isConstrainable(resource, policy.actions)
+
     accessPolicyDAO.loadPolicy(resourceAndPolicyName).flatMap {
       case None => createPolicy(resourceAndPolicyName, workbenchSubjects, generateGroupEmail(), policy.roles, policy.actions)
       case Some(accessPolicy) => accessPolicyDAO.overwritePolicy(AccessPolicy(resourceAndPolicyName, workbenchSubjects, accessPolicy.email, policy.roles, policy.actions))
     } andThen {
-      case Success(policy) => fireGroupUpdateNotification(policy.id)
+      case Success(policy) =>
+        fireGroupUpdateNotification(policy.id)
```
dvoet, line 256: the changes in this class also do not belong here. Calculating the intersection is only something that needs to happen when a group is synced so it belongs in the GoogleExtension. Also, we don't want to call sync when a policy is changed, we want to `fireGroupUpdateNotification` and let that machinery deal with it.

dvoet (changes requested):
in addition to the other comments, one function is missing which should be invoked in `GoogleExtensions.onGroupUpdate`: when a group is updated we need to find all the policies of all the resources that are constrained by the auth domain and fire a sync message for them as well.

I am happy to talk on Monday.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -82,7 +82,7 @@ object Boot extends App with LazyLogging {
         val googleKeyCache = new GoogleKeyCache(googleIamDAO, googleStorageDAO, googlePubSubDAO, googleServicesConfig, petServiceAccountConfig)
         val notificationDAO = new PubSubNotificationDAO(googlePubSubDAO, googleServicesConfig.notificationTopic)
 
-        new GoogleExtensions(directoryDAO, accessPolicyDAO, googleDirectoryDAO, googlePubSubDAO, googleIamDAO, googleStorageDAO, googleProjectDAO, googleKeyCache, notificationDAO, googleServicesConfig, petServiceAccountConfig, resourceTypeMap(CloudExtensions.resourceTypeName))
+        new GoogleExtensions(directoryDAO, accessPolicyDAO, googleDirectoryDAO, googlePubSubDAO, googleIamDAO, googleStorageDAO, googleProjectDAO, googleKeyCache, notificationDAO, googleServicesConfig, petServiceAccountConfig, resourceTypeMap(CloudExtensions.resourceTypeName), resourceTypeMap)
```
Qi77Qi: now that we pass in `resourceTypeMap` for `GoogleExtensions`, do we still need to pass in `resourceTypeMap(CloudExtensions.resourceTypeName)` separately?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -450,6 +471,34 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
 
+  private def isConstrainable(resource: Resource, actions: Set[ResourceAction]): Boolean = {
+    resourceTypes.get(resource.resourceTypeName) match {
+      case Some(resourceType) => {
+        for {
+          actionPattern <- resourceType.actionPatterns
+          action <- actions
+        } yield actionPattern.authDomainConstrainable && actionPattern.matches(action)
+      }.contains(true)
+      case None => false // we really shouldn't get here as this means the resource is invalid, but leaving it in for completeness
```
Qi77Qi: can you log an error message in this case?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -450,6 +471,34 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
   }
 
+  private def isConstrainable(resource: Resource, actions: Set[ResourceAction]): Boolean = {
+    resourceTypes.get(resource.resourceTypeName) match {
+      case Some(resourceType) => {
+        for {
+          actionPattern <- resourceType.actionPatterns
+          action <- actions
+        } yield actionPattern.authDomainConstrainable && actionPattern.matches(action)
+      }.contains(true)
```
Qi77Qi: You don't need to loop through all items to see if there's true. I think you can return true as long as there's one `true`?
maybe something like this will be a little bit more efficient?
```
resourceType.actionPatterns.exists{
          actionPattern =>
            actionPattern.authDomainConstrainable && actions.exists(action => actionPattern.matches(action))
        }
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -450,6 +471,34 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+  private def calculateIntersectionGroup(resource: Resource, policy: AccessPolicy): Future[Set[WorkbenchUserId]] = {
+    for {
+      groups <- accessPolicyDAO.loadResourceAuthDomain(resource)
+      authDomain <- Future.traverse(groups)(groupName =>
+        directoryDAO.listFlattenedGroupUsers(groupName))
+      policyMembers <- Future.traverse(policy.members)(groupName => groupName match {
+        case g: WorkbenchGroupName => directoryDAO.listFlattenedGroupUsers(g)
+        case u: WorkbenchUserId => Future.successful(Set(u))
+        case p: PetServiceAccountId => Future(Set(p.userId))
+        case _ => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, s"Bad policy found"))
```
Qi77Qi: fyi, you can do `Future.failed(new WorkbenchExceptionxxx)` without throwing the exception

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -450,6 +471,34 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+      })
+    } yield {
+      (authDomain + policyMembers.flatten).reduce(_.intersect(_))
```
Qi77Qi: if `authDomain + policyMembers.flatten` is empty, `reduce` will throw exception. Will that be an issue? When it's empty, shall we just return empty set or we really want to propagate the error?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -52,6 +52,18 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
     }
   }
 
+  private def unmarshalResource(results: Entry): Resource = {
+    val resourceTypeName = getAttribute(results, Attr.resourceType).getOrElse(throw new WorkbenchException(s"${Attr.resourceType} attribute missing"))
+    val resourceId = getAttribute(results, Attr.resourceId).getOrElse(throw new WorkbenchException(s"${Attr.resourceId} attribute missing"))
+    val authDomain = getAttributes(results, Attr.authDomain).getOrElse(Set.empty).map(g => WorkbenchGroupName(g)).toSet
+
+    Resource(ResourceTypeName(resourceTypeName), ResourceId(resourceId), authDomain)
```
Qi77Qi: if you want to express the error in return type, you can do something like
```
  private def unmarshalResource(results: Entry): Either[String, Resource] = {
    for{
      resourceTypeName <- getAttribute(results, Attr.resourceType).toRight(s"${Attr.resourceType} attribute missing")
      resourceId <- getAttribute(results, Attr.resourceId).toRight(s"${Attr.resourceId} attribute missing")
    } yield {
      val authDomain =  getAttributes(results, Attr.authDomain).getOrElse(Set.empty).map(g => WorkbenchGroupName(g)).toSet
      Resource(ResourceTypeName(resourceTypeName), ResourceId(resourceId), authDomain)
    }
  }
```

And change caller to
```
  override def listResourcesConstrainedByGroup(groupId: WorkbenchGroupIdentity): Future[Set[Resource]] = for{
    res <- Future(ldapSearchStream(resourcesOu, SearchScope.SUB, Filter.createEqualityFilter(Attr.authDomain, groupId.toString))(unmarshalResource))
    r <- res.parSequence.fold(err => Future.failed(new WorkbenchException(err)), r => Future.successful(r.toSet))
  } yield r
```
(you'll need to import `cats.implicits._` for this. This will return all entries fail to unmarshall.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
marctalbott: Is this possible given that `authDomain + policyMembers.flatten` is a `Set[Set[WorkbenchUserId]]`? My understanding is that if there are no users in either the policy or the auth domain, `authDomain + policyMembers.flatten` will still contain the empty set and therefore not be empty. This would result in `reduce` not throwing an exception, but just returning an empty set, which is the desired behavior in this case as there should not be any users included in the intersection group for an empty policy and auth domain.

Qi77Qi (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -414,6 +426,15 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
 
         group = groupOption.getOrElse(throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"$groupId not found")))
 
+        members <- group match {
+          case accessPolicy: AccessPolicy => if (isConstrainable(accessPolicy.id.resource, accessPolicy.actions)) {
```
dvoet: checking `accessPolicy.actions` alone is not enough, also need to check roles which can be though of as a collections of actions

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -450,6 +471,33 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
     }
   }
 
+  private def isConstrainable(resource: Resource, actions: Set[ResourceAction]): Boolean = {
+    resourceTypes.get(resource.resourceTypeName) match {
+      case Some(resourceType) => resourceType.actionPatterns.exists { actionPattern =>
+        actionPattern.authDomainConstrainable && actions.exists(action => actionPattern.matches(action))
+      }
+      case None =>
+        logger.error(s"Invalid resource type specified. ${resource.resourceTypeName} is not an accepted resource type.")
```
dvoet: Why don't we raise this error?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -450,6 +471,33 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+        false
+    }
+  }
+
+  private def calculateIntersectionGroup(resource: Resource, policy: AccessPolicy): Future[Set[WorkbenchUserId]] = {
+    for {
+      groups <- accessPolicyDAO.loadResourceAuthDomain(resource)
+      authDomain <- Future.traverse(groups)(groupName =>
+        directoryDAO.listFlattenedGroupUsers(groupName))
+      policyMembers <- Future.traverse(policy.members)(groupName => groupName match {
```
dvoet: a policy is a group, just call directoryDAO.listFlattenedGroupUsers(policy.id)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -211,7 +211,7 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
 
   /**
     * Overwrites the policy if it already exists or creates a new policy entry if it does not exist.
-    * Triggers update to Google Group upon successfully updating the policy.
+    * Triggers update to Google Group and syncs intersection group if policy is constrainable upon successfully updating the policy.
```
dvoet: looks like there were not actually any changes to this class, remove this change

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -153,6 +154,98 @@ class GoogleExtensionSpec(_system: ActorSystem) extends TestKit(_system) with Fl
     }
   }
 
+  it should "sync the intersection group if the policy is constrainable" in {
```
dvoet, line 159: should test the cases
only 1 policy role contains a constrainable action
only 1 policy action is a constrainable action
neither roles nor actions on the policy are constrainable but resource type may be constrained

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -127,6 +127,17 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
         case (rpn: ResourceAndPolicyName, Some(_)) => rpn.toJson.compactPrint
       }
       _ <- googlePubSubDAO.publishMessages(googleServicesConfig.groupSyncTopic, messagesForIdsWithSyncDates)
+      managedGroupIds = groupIdentities.collect { case ResourceAndPolicyName(Resource(ResourceTypeName("managed-group"), id, _), _) => id }
```
dvoet: actually, I think we need to go a step further... whatever group is being updated may be a member (directly or indirectly) of a managed group. We need to query ldap for all the managed groups this group is a member of, `listResourcesConstrainedByGroup` on each of those then `onGroupUpdate` on each of those.

dvoet (changes requested):
on the testing front, be sure to test with some nested group structures too... anywhere where we are flattening groups we should test with nested groups to make sure that is working.

also test with nested managed groups as auth domains

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -450,6 +471,33 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+      case Some(resourceType) => resourceType.actionPatterns.exists { actionPattern =>
+        actionPattern.authDomainConstrainable && actions.exists(action => actionPattern.matches(action))
+      }
+      case None =>
+        logger.error(s"Invalid resource type specified. ${resource.resourceTypeName} is not an accepted resource type.")
+        false
+    }
+  }
+
+  private def calculateIntersectionGroup(resource: Resource, policy: AccessPolicy): Future[Set[WorkbenchUserId]] = {
```
dvoet, line 497: looking at some other things, I was reminded that this calculation can be done entirely within 1 ldap call. Something like:
`AND filter(EQUALS filter(memberof=policy dn), EQUALS filter(memberof=group1 dn),EQUALS filter(memberof=group2 dn)...)`

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -226,8 +226,16 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
     groupsOption.getOrElse(Set.empty)
   }
 
-  override def listFlattenedGroupUsers(groupId: WorkbenchGroupIdentity): Future[Set[WorkbenchUserId]] = Future {
-    ldapSearchStream(peopleOu, SearchScope.SUB, Filter.createEqualityFilter(Attr.memberOf, groupDn(groupId)))(unmarshalUser).map(_.id).toSet
+  override def listIntersectionGroupUsers(groupIds: Set[WorkbenchGroupIdentity]): Future[Set[WorkbenchUserId]] = Future {
+    ldapSearchStream(directoryConfig.baseDn, SearchScope.SUB, Filter.createANDFilter(groupIds.map(groupId =>
+      Filter.createEqualityFilter(Attr.memberOf, groupDn(groupId))).asJava))(unmarshalUserOption).collect { case Some(user) => user.id }.toSet
```
dvoet: if all you care about is user.id why do you need unmarshalUserOption? Instead just use 
`getAttribute(_, Attr.uid)`
and then flatten instead of collect

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -127,6 +129,17 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
+      managedGroupIds = (ancestorGroups.flatten ++ groupIdentities).collect { case ResourceAndPolicyName(Resource(ResourceTypeName("managed-group"), id, _), _) => id }
+      _ <- Future.traverse(managedGroupIds)(onManagedGroupUpdate)
+    } yield ()
+  }
+
+  private def onManagedGroupUpdate(groupId: ResourceId): Future[Unit] = {
+    for {
+      resources <- accessPolicyDAO.listResourcesConstrainedByGroup(WorkbenchGroupName(groupId.value))
+      policies <- Future.traverse(resources) { resource => accessPolicyDAO.listAccessPolicies(resource).unsafeToFuture }
+      _ <- onGroupUpdate(policies.flatten.map(_.id).toList)
```
dvoet: I am concerned about an endless recursive loop... let's pair to fix it

dvoet (changes requested):

dvoet (approved):

PR #221, opened 2018-10-05
check auth domain for permission

Qi77Qi:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3571

Not exactly sure about https://github.com/broadinstitute/sam/pull/221/files#diff-1b52ef6306942437bf8b4509e0f37ab2R155, but hoping there's test covers this

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
@@ -27,14 +27,14 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
   def groupRoutes: server.Route = requireUserInfo { userInfo =>
     (pathPrefix("groups" / "v1") | pathPrefix("group")) {
       pathPrefix(Segment) { groupId =>
-        val managedGroup = Resource(ManagedGroupService.managedGroupTypeName, ResourceId(groupId))
+        val managedGroup = FullyQualifiedResourceId(ManagedGroupService.managedGroupTypeName, ResourceId(groupId))
 
         pathEndOrSingleSlash {
           get {
-            handleGetGroup(managedGroup)
+            handleGetGroup(managedGroup.resourceId)
           } ~
           post {
-            handleCreateGroup(managedGroup, userInfo)
+            handleCreateGroup(managedGroup.resourceId, userInfo)
           } ~
           delete {
             handleDeleteGroup(managedGroup, userInfo)
@@ -92,26 +92,26 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
     complete(managedGroupService.listGroups(userInfo.userId).map(StatusCodes.OK -> _))
   }
 
-  private def handleGetGroup(managedGroup: Resource): Route = {
+  private def handleGetGroup(resourceId: ResourceId): Route = {
     complete (
-      managedGroupService.loadManagedGroup(managedGroup.resourceId).map {
+      managedGroupService.loadManagedGroup(resourceId).map {
         case Some(response) => StatusCodes.OK -> response
         case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "group not found"))
       }
     )
   }
 
-  private def handleCreateGroup(managedGroup: Resource, userInfo: UserInfo): Route = {
-    complete(managedGroupService.createManagedGroup(managedGroup.resourceId, userInfo).map(_ => StatusCodes.Created))
+  private def handleCreateGroup(resourceId: ResourceId, userInfo: UserInfo): Route = {
+    complete(managedGroupService.createManagedGroup(resourceId, userInfo).map(_ => StatusCodes.Created))
   }
 
-  private def handleDeleteGroup(managedGroup: Resource, userInfo: UserInfo): Route = {
+  private def handleDeleteGroup(managedGroup: FullyQualifiedResourceId, userInfo: UserInfo): Route = {
     requireAction(managedGroup, SamResourceActions.delete, userInfo.userId) {
       complete(managedGroupService.deleteManagedGroup(managedGroup.resourceId).map(_ => StatusCodes.NoContent))
     }
   }
 
-  private def handleListEmails(managedGroup: Resource, accessPolicyName: ManagedGroupPolicyName, userInfo: UserInfo): Route = {
+  private def handleListEmails(managedGroup: FullyQualifiedResourceId, accessPolicyName: ManagedGroupPolicyName, userInfo: UserInfo): Route = {
     requireAction(managedGroup, SamResourceActions.readPolicy(accessPolicyName), userInfo.userId) {
       complete(
         managedGroupService.listPolicyMemberEmails(managedGroup.resourceId, accessPolicyName).map(StatusCodes.OK -> _)
@@ -119,7 +119,7 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
     }
   }
 
-  private def handleOverwriteEmails(managedGroup: Resource, accessPolicyName: ManagedGroupPolicyName, userInfo: UserInfo): Route = {
+  private def handleOverwriteEmails(managedGroup: FullyQualifiedResourceId, accessPolicyName: ManagedGroupPolicyName, userInfo: UserInfo): Route = {
     requireAction(managedGroup, SamResourceActions.sharePolicy(accessPolicyName), userInfo.userId) {
       entity(as[Set[WorkbenchEmail]]) { members =>
         complete(
@@ -129,7 +129,7 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
     }
   }
 
-  private def handleAddEmailToPolicy(managedGroup: Resource, accessPolicyName: ManagedGroupPolicyName, email: String, userInfo: UserInfo): Route = {
+  private def handleAddEmailToPolicy(managedGroup: FullyQualifiedResourceId, accessPolicyName: ManagedGroupPolicyName, email: String, userInfo: UserInfo): Route = {
     requireAction(managedGroup, SamResourceActions.sharePolicy(accessPolicyName), userInfo.userId) {
       withSubject(WorkbenchEmail(email)) { subject =>
         complete(
@@ -139,7 +139,7 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
     }
   }
 
-  private def handleDeleteEmailFromPolicy(managedGroup: Resource, accessPolicyName: ManagedGroupPolicyName, email: String, userInfo: UserInfo): Route = {
+  private def handleDeleteEmailFromPolicy(managedGroup: FullyQualifiedResourceId, accessPolicyName: ManagedGroupPolicyName, email: String, userInfo: UserInfo): Route = {
     requireAction(managedGroup, SamResourceActions.sharePolicy(accessPolicyName), userInfo.userId) {
       withSubject(WorkbenchEmail(email)) { subject =>
         complete(
@@ -149,7 +149,7 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
     }
   }
 
-  private def handleRequestAccess(managedGroup: Resource, userInfo: UserInfo): Route = {
+  private def handleRequestAccess(managedGroup: FullyQualifiedResourceId, userInfo: UserInfo): Route = {
     requireAction(managedGroup, SamResourceActions.notifyAdmins, userInfo.userId) {
       complete(
         managedGroupService.requestAccess(managedGroup.resourceId, userInfo.userId).map(_ => StatusCodes.NoContent)
@@ -157,7 +157,7 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
     }
   }
 
-  private def handleSetAccessInstructions(managedGroup: Resource, accessInstructions: ManagedGroupAccessInstructions, userInfo: UserInfo): Route = {
+  private def handleSetAccessInstructions(managedGroup: FullyQualifiedResourceId, accessInstructions: ManagedGroupAccessInstructions, userInfo: UserInfo): Route = {
     requireAction(managedGroup, SamResourceActions.setAccessInstructions, userInfo.userId) {
       complete(
         managedGroupService.setAccessInstructions(managedGroup.resourceId, accessInstructions.value).map(_ => StatusCodes.NoContent)
@@ -165,7 +165,7 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
     }
   }
 
-  private def handleGetAccessInstructions(managedGroup: Resource): Route = {
+  private def handleGetAccessInstructions(managedGroup: FullyQualifiedResourceId): Route = {
     complete(
       managedGroupService.getAccessInstructions(managedGroup.resourceId).map {
         case Some(accessInstructions) => StatusCodes.OK -> Option(accessInstructions)
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectorySubjectNameSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectorySubjectNameSupport.scala
@@ -24,22 +24,22 @@ trait DirectorySubjectNameSupport {
   protected def groupDn(groupId: WorkbenchGroupIdentity) = {
     groupId match {
       case WorkbenchGroupName(name) => s"cn=$name,$groupsOu"
-      case rpn: ResourceAndPolicyName => policyDn(rpn)
+      case rpn: FullyQualifiedPolicyId => policyDn(rpn)
       case _ => throw new WorkbenchException(s"unexpected WorkbenchGroupIdentity $groupId")
     }
   }
   protected def userDn(samUserId: WorkbenchUserId) = s"uid=${samUserId.value},$peopleOu"
   protected def petDn(petServiceAccountId: PetServiceAccountId) = s"${Attr.project}=${petServiceAccountId.project.value},${userDn(petServiceAccountId.userId)}"
   protected def resourceTypeDn(resourceTypeName: ResourceTypeName) = s"${Attr.resourceType}=${resourceTypeName.value},$resourcesOu"
-  protected def resourceDn(resource: Resource) = s"${Attr.resourceId}=${resource.resourceId.value},${resourceTypeDn(resource.resourceTypeName)}"
+  protected def resourceDn(resource: FullyQualifiedResourceId) = s"${Attr.resourceId}=${resource.resourceId.value},${resourceTypeDn(resource.resourceTypeName)}"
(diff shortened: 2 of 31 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -84,7 +84,10 @@ object SamResourceActions {
 
 @Lenses case class ResourceTypeName(value: String) extends ValueObject
 
-@Lenses case class Resource(resourceTypeName: ResourceTypeName, resourceId: ResourceId, authDomain: Set[WorkbenchGroupName] = Set.empty)
+@Lenses final case class ResourceIdentity(resourceTypeName: ResourceTypeName, resourceId: ResourceId)
```
marctalbott: Having both `ResourceIdentity` and `ResourceId` seems a bit confusing to me as the names feel too similar. Maybe `ResourceIdAndTypeName`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -139,27 +136,34 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
...
-    }
-    result.getOrElse(Set.empty)
-  }
+    } yield accessPolicies.getOrElse(Set.empty)
 
-  override def listFlattenedPolicyMembers(resourceAndPolicyName: ResourceAndPolicyName): Future[Set[WorkbenchUserId]] = Future {
-    ldapSearchStream(peopleOu, SearchScope.ONE, Filter.createEqualityFilter(Attr.memberOf, policyDn(resourceAndPolicyName)))(unmarshalUser).map(_.id).toSet
-  }
+  override def listFlattenedPolicyMembers(policyId: PolicyIdentity): IO[Set[WorkbenchUserId]] = cs.evalOn(blockingEc)(
+    IO(ldapSearchStream(peopleOu, SearchScope.ONE, Filter.createEqualityFilter(Attr.memberOf, policyDn(policyId)))(unmarshalUser).map(_.id).toSet)
```
marctalbott: I believe this might be a good time to rework this query. @dvoet pointed out the same problem in the (soon to be deleted) function `listFlattenedGroupUsers` in `ldapDirectoryDAO` and the search I was doing to find intersection users in my PR. Check out the bottom row in the [LDAP knowledge share](https://broadinstitute.atlassian.net/wiki/spaces/GAWB/pages/622985222/LDAP). It could speed up the `listFlattenedPolicyMembers` query by a lot

marctalbott, line 191: I believe this might be a good time to rework this query. As @dvoet pointed out in `listFlattenedGroupUsers` in `ldapDirectoryDAO` and in my own PR where I was searching for intersection group users, starting at `peopleOu` makes for a much slower search. Starting the search at the `baseDn` instead and then filtering out non-users will result in a faster search. Check out the bottom entry of the [LDAP knowledge share](https://broadinstitute.atlassian.net/wiki/spaces/GAWB/pages/622985222/LDAP) for more info on timing comparisons.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
Qi77Qi: yea, it is a bit confusing. But I think it's better than `ResourceIdAndTypeName` since that name doesn't indicate well what this field is for. Conceptually, it uniquely identifies a `Resource`, whenever we do equality check or look up for a resource, we need all fields in this case class. https://github.com/broadinstitute/sam/pull/221/files/07b755659f0169b1d1aa8704f0ed08cabbe572a0#diff-a80e785aa2e82ece2775df0a248fde5cR101 is similar idea. I'm not sure if `resourceId` is unique across all resourceTypes, but even if they are, we won't be able to look up a resource just by `resourceId`.

I'm up for other name suggestions tho.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
Qi77Qi: The PR didn't touch this logic other than I changed return type to `IO`. I checked the code and didn't see any tests covering this function, so I'm not inclined to refactor this in this PR since changing the query is much higher risk than changing return type from Future to IO.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -92,34 +92,34 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
     complete(managedGroupService.listGroups(userInfo.userId).map(StatusCodes.OK -> _))
   }
 
-  private def handleGetGroup(managedGroup: Resource): Route = {
+  private def handleGetGroup(resourceId: ResourceId): Route = {
```
dvoet, line 95: can we leave handleGetGroup and handleCreate group with `ResourceIdentity` and the parameter type for consistency?

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectorySubjectNameSupport.scala
```diff
@@ -67,15 +67,15 @@ trait DirectorySubjectNameSupport {
       case groupMatcher(cn) => WorkbenchGroupName(cn)
       case personMatcher(uid) => WorkbenchUserId(uid)
       case petMatcher(petProject, userUid) => PetServiceAccountId(WorkbenchUserId(userUid), GoogleProject(petProject))
-      case policyMatcher(policyName, resourceId, resourceTypeName) => ResourceAndPolicyName(Resource(ResourceTypeName(resourceTypeName), ResourceId(resourceId)), AccessPolicyName(policyName))
+      case policyMatcher(policyName, resourceId, resourceTypeName) => PolicyIdentity(ResourceIdentity(ResourceTypeName(resourceTypeName), ResourceId(resourceId)), AccessPolicyName(policyName)) //TODO: we might need a better type that excludes AuthDomains field
```
dvoet: what does this TODO mean? If it is something that should be done but not as part of this PR make a jira for it.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
dvoet: `FullyQualifiedResourceId` and also `FullyQualifiedPolicyId`

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -161,27 +158,34 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
...
-      val memberOfSubjects = memberOfDns.map(dnToSubject)
-      memberOfSubjects.collect {
-        case ripn@ResourceAndPolicyName(`resource`, _) => unmarshalAccessPolicy(ldapConnectionPool.getEntry(subjectDn(ripn)))
+  ))
+
+  override def listAccessPoliciesForUser(resource: ResourceIdentity, user: WorkbenchUserId): IO[Set[AccessPolicy]] = for {
+      entry <- cs.evalOn(blockingEc)(IO(ldapConnectionPool.getEntry(subjectDn(user), Attr.memberOf)))
+      members <- IO.pure(Option(entry).map(e => getAttributes(e, Attr.memberOf).toList))
+      accessPolicies <- members.traverse{
+        strs =>
```
dvoet: better variable naming please

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -161,27 +158,34 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
...
+          val ripns = strs.mapFilter{
+            str =>
+              for{
+                subject <- Either.catchNonFatal(dnToSubject(str)).toOption
+                ripn <- subject match {
+                  case ripn: PolicyIdentity if ripn.resource.resourceId == resource.resourceId && ripn.resource.resourceTypeName == resource.resourceTypeName => Some(ripn)
+                  case _ => None
+                }
+              } yield ripn
```
dvoet: these are not `ripn`s anymore

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -168,53 +164,60 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
-        resourceType <- resourceTypeOption
-        role <- resourceType.roles.find(_.roleName == resourceRoleName)
-      } yield {
-        role.actions
+  def hasPermission(resource: ResourceIdentity, action: ResourceAction, userInfo: UserInfo): IO[Boolean] = {
+    listUserResourceActions(resource, userInfo).map(_.contains(action))
+  }
+
+  def listUserResourceActions(resource: ResourceIdentity, userInfo: UserInfo): IO[Set[ResourceAction]] = {
+    val resourceType = resourceTypes.get(resource.resourceTypeName)
```
dvoet: unused

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -168,53 +164,60 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
...
-      val maybeActions = for {
-        resourceType <- resourceTypeOption
-        role <- resourceType.roles.find(_.roleName == resourceRoleName)
-      } yield {
-        role.actions
+  def hasPermission(resource: ResourceIdentity, action: ResourceAction, userInfo: UserInfo): IO[Boolean] = {
+    listUserResourceActions(resource, userInfo).map(_.contains(action))
+  }
+
+  def listUserResourceActions(resource: ResourceIdentity, userInfo: UserInfo): IO[Set[ResourceAction]] = {
```
dvoet: 2 comments on this function:
1) functionality: aside from the `authDomainsPoliciesUserIsMemberOf` being not quite right as discussed in person, the other problem is that auth domains should be applied only to specific actions as governed by the resource type (`ResourceActionPattern.authDomainConstrainable`).

2) complexity: the logic here is complicated and layered for comps, maps, flatmaps... oh my. And much of the complication is only required when at least one action is actually constrained by an auth domain. So to simplify, I would try reverting to how it was then add on a conditional filtering step: 
```
if any of the actions are constrainable, load resource auth domains
  if there are auth domains, load user's auth domains
    if the user is missing any auth domains remove all constrainable actions
else (for all ifs) return original list
```

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -697,10 +703,21 @@ class ManagedGroupRoutesSpec extends FlatSpec with Matchers with ScalatestRouteT
...
+    val samRoutes = createResource(resourceTypes, Resource(ManagedGroupService.managedGroupTypeName, ResourceId("foo"), Set.empty))
 
     Post(s"/api/group/$groupId/requestAccess") ~> samRoutes.route ~> check {
       status shouldEqual StatusCodes.NotFound
     }
   }
 }
+
+object ManagedGroupRoutesSpec{
+  def createResource(resourceTypeMap: Map[ResourceTypeName, ResourceType], resource: Resource)(implicit sysmtem: ActorSystem, materializer: Materializer, ec: ExecutionContext): TestSamRoutes ={
```
dvoet: call it `createSamRoutesWithResource`

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/TestSamRoutes.scala
```diff
@@ -22,23 +22,23 @@ import scala.concurrent.ExecutionContext
...
-  def mockDirectoryDao: MockDirectoryDAO = directoryDAO
+  def mockDirectoryDao: DirectoryDAO = directoryDAO
 }
 
 object TestSamRoutes {
   val defaultUserInfo = UserInfo(OAuth2BearerToken("accessToken"), WorkbenchUserId("user1"), WorkbenchEmail("user1@example.com"), 0)
   val defaultGoogleSubjectId = GoogleSubjectId("user1")
 
-  def apply(resourceTypes: Map[ResourceTypeName, ResourceType], userInfo: UserInfo = defaultUserInfo)(implicit system: ActorSystem, materializer: Materializer, executionContext: ExecutionContext) = {
+  def apply(resourceTypes: Map[ResourceTypeName, ResourceType], userInfo: UserInfo = defaultUserInfo, policyAccessDAO: Option[AccessPolicyDAO] = None, policies: Option[mutable.Map[WorkbenchGroupIdentity, WorkbenchGroup]] = None)(implicit system: ActorSystem, materializer: Materializer, executionContext: ExecutionContext) = {
```
dvoet, line 35: why did this change?

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -601,7 +603,11 @@ class ManagedGroupRoutesSpec extends FlatSpec with Matchers with ScalatestRouteT
   }
 
   it should "fail with 404 when the group does not exist" in {
-    val samRoutes = TestSamRoutes(resourceTypes)
+    val groups = TrieMap.empty[WorkbenchGroupIdentity, WorkbenchGroup]
+    val policyDao = new MockAccessPolicyDAO(groups)
+    val samRoutes = TestSamRoutes(resourceTypes, policyAccessDAO = Some(policyDao), policies = Some(groups))
+
+    policyDao.createResource(Resource(ManagedGroupService.managedGroupTypeName, ResourceId("foo"), Set.empty)).unsafeRunSync()
```
dvoet, line 610: why this change?

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -228,8 +233,7 @@ class ManagedGroupRoutesSpec extends FlatSpec with Matchers with ScalatestRouteT
   }
 
   it should "fail with 404 when the group does not exist" in {
-    val samRoutes = TestSamRoutes(resourceTypes)
-
+    val samRoutes = createResource(resourceTypes, Resource(ManagedGroupService.managedGroupTypeName, ResourceId("foo"), Set.empty))
```
dvoet: why all the changes to `createResource`?

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -1,19 +1,25 @@
 package org.broadinstitute.dsde.workbench.sam
 package api
 
+import akka.actor.ActorSystem
```
dvoet, line 4: all the comments I made on ManagedGroupRoutesSpec also apply to ManagedGroupRoutesV1Spec. However I would like to see ManagedGroupRoutesSpec change as little as possible since it is here to ensure we remain backwards compatible.

dvoet (changes requested):
on the test front, I am concerned by all the seemingly unrelated changes. This is a big refactor and I would expect changes only related to that refactor and new functionality. With all the other changes it is hard to tell that the tests themselves did not actually change. And if the tests changed, maybe we are not testing what we think we are anymore.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
Qi77Qi: Giving `handleGetGroup` `ResourceIdentity` is giving it more info than needed though. There're definitely cases where you don't need a `FullyQualifiedResourceId` but `ResourceId` will suffice, and I think it makes sense to have the distinction.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/DirectorySubjectNameSupport.scala
Qi77Qi: I think that was before I decided to use a separate type. I'll remove the comment

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
Qi77Qi: it's because lots of endpoints use [requireAction](https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala#L18), which calls [added new check](https://github.com/broadinstitute/sam/blob/secondAuthDomain/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala#L191)...previously, without the resource which policy belongs to being existent is fine in the test, but due to the new check, we need to add the resource in tests

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
Qi77Qi: I think all changes in these 2 files are due to https://github.com/broadinstitute/sam/pull/221#discussion_r224230593, which can't be avoided since this PR is to add auth domain check, which means we need to load resource for the policy we're checking, but previously in tests, we didn't care about the resource that the policy belongs to.

I don't think it'll break backward compatibility since in real use case, we definitely have resource pre-created before we have policy.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
Qi77Qi: I think all changes in these 2 files are due to #221 (comment), which can't be avoided since this PR is to add auth domain check, which means we need to load resource for the policy we're checking, but previously in tests, we didn't care about the resource that the policy belongs to.

I don't think it'll break backward compatibility since in real use case, we definitely have resource pre-created before we have policy.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/TestSamRoutes.scala
Qi77Qi: it's used https://github.com/broadinstitute/sam/blob/secondAuthDomain/src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala#L718

dvoet (approved):

marctalbott (approved):

Qi77Qi:
jenkins retest

Qi77Qi:
checked with Alex...this jenkins failure should be fine

PR #222, opened 2018-10-05
Public resources

dvoet:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3796
This PR introduces a flag on a policy that declares it public. When public is set to true all users may perform the roles/actions listed by the policy. This means that list resources, list actions, has permission all recognize public policies. Further, when a public policy is synched to google the all users group is place in the google group.

Security: There is a new resource type: resource_type_admin. When sam starts it automatically creates a resource of type resource_type_admin for every other resource type. This controls at the type level who can set resources to public. For example, we want only comms to be able to make workspaces public but we probably want anybody to be able to make methods public.

In order to make a policy public, a user must have either the set_public or set_public::[policy_name] action on the resource_type_admin resource for the given type *and* have either alter_policies or share::[policy_name] on the give resource.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -142,14 +142,14 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
                     } ~
                     pathPrefix("public") {
                       pathEndOrSingleSlash {
-                        requireOneOfAction(resource, Set(SamResourceActions.readPolicies, SamResourceActions.readPolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
-                          get {
+                        get {
+                          requireOneOfAction(resource, Set(SamResourceActions.readPolicies, SamResourceActions.readPolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
                             complete(resourceService.isPublic(resourceAndPolicyName))
                           }
                         } ~
-                        requireOneOfAction(resource, Set(SamResourceActions.alterPolicies, SamResourceActions.sharePolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
-                          requireOneOfAction(FullyQualifiedResourceId(SamResourceTypes.resourceTypeAdminName, ResourceId(resourceType.name.value)), Set(SamResourceActions.setPublic, SamResourceActions.setPublicPolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
-                            put {
+                        put {
+                          requireOneOfAction(resource, Set(SamResourceActions.alterPolicies, SamResourceActions.sharePolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
+                            requireOneOfAction(FullyQualifiedResourceId(SamResourceTypes.resourceTypeAdminName, ResourceId(resourceType.name.value)), Set(SamResourceActions.setPublic, SamResourceActions.setPublicPolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
                               entity(as[Boolean]) { isPublic =>
                                 complete(resourceService.setPublic(resourceAndPolicyName, isPublic).unsafeToFuture().map(_ => StatusCodes.NoContent))
                               }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -151,7 +151,7 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
                           requireOneOfAction(resource, Set(SamResourceActions.alterPolicies, SamResourceActions.sharePolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
                             requireOneOfAction(FullyQualifiedResourceId(SamResourceTypes.resourceTypeAdminName, ResourceId(resourceType.name.value)), Set(SamResourceActions.setPublic, SamResourceActions.setPublicPolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
                               entity(as[Boolean]) { isPublic =>
-                                complete(resourceService.setPublic(resourceAndPolicyName, isPublic).unsafeToFuture().map(_ => StatusCodes.NoContent))
+                                complete(resourceService.setPublic(resourceAndPolicyName, isPublic).map(_ => StatusCodes.NoContent))
                               }
                             }
                           }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -9,6 +9,7 @@ import akka.http.scaladsl.server.Directives._
 import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
 import org.broadinstitute.dsde.workbench.model.{ErrorReport, WorkbenchEmail, WorkbenchExceptionWithErrorReport}
 import org.broadinstitute.dsde.workbench.sam.model.SamJsonSupport._
+import org.broadinstitute.dsde.workbench.sam.model.RootPrimitiveJsonSupport._
 import org.broadinstitute.dsde.workbench.sam.model._
 import org.broadinstitute.dsde.workbench.sam.service.ResourceService
 import spray.json.DefaultJsonProtocol._
@@ -133,6 +134,24 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
                               } ~
                               delete {
                                 complete(resourceService.removeSubjectFromPolicy(resourceAndPolicyName, subject).map(_ => StatusCodes.NoContent))
+                              }
+                            }
+                          }
+                        }
+                      }
+                    } ~
+                    pathPrefix("public") {
+                      pathEndOrSingleSlash {
+                        requireOneOfAction(resource, Set(SamResourceActions.readPolicies, SamResourceActions.readPolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
+                          get {
+                            complete(resourceService.isPublic(resourceAndPolicyName))
+                          }
+                        } ~
+                        requireOneOfAction(resource, Set(SamResourceActions.alterPolicies, SamResourceActions.sharePolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
+                          requireOneOfAction(FullyQualifiedResourceId(SamResourceTypes.resourceTypeAdminName, ResourceId(resourceType.name.value)), Set(SamResourceActions.setPublic, SamResourceActions.setPublicPolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
+                            put {
+                              entity(as[Boolean]) { isPublic =>
+                                complete(resourceService.setPublic(resourceAndPolicyName, isPublic).unsafeToFuture().map(_ => StatusCodes.NoContent))
                               }
                             }
                           }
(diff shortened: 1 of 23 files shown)


marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -409,7 +409,14 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
       for {
         groupOption <- groupId match {
           case basicGroupName: WorkbenchGroupName => directoryDAO.loadGroup(basicGroupName)
-          case rpn: ResourceAndPolicyName => accessPolicyDAO.loadPolicy(rpn)
+          case rpn: ResourceAndPolicyName => accessPolicyDAO.loadPolicy(rpn).map(_.map { loadedPolicy =>
+            if (loadedPolicy.isPublic.contains(true)) {
+              // include all users group when synchronizing a public policy
+              loadedPolicy.copy(members = loadedPolicy.members + allUsersGroupName)
```
Qi77Qi: fyi you can do `AccessPolicy.members.modify(_ + allUsersGroupName)(loadedPolicy)` as well

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -103,7 +118,7 @@ Note that AccessPolicy IS A group, does not have a group. This makes the ldap qu
 and thus resources much easier. We tried modeling with a "has a" relationship in code but a "is a" relationship in
 ldap but it felt unnatural.
  */
-@Lenses case class AccessPolicy(id: ResourceAndPolicyName, members: Set[WorkbenchSubject], email: WorkbenchEmail, roles: Set[ResourceRoleName], actions: Set[ResourceAction]) extends WorkbenchGroup
+@Lenses case class AccessPolicy(id: ResourceAndPolicyName, members: Set[WorkbenchSubject], email: WorkbenchEmail, roles: Set[ResourceRoleName], actions: Set[ResourceAction], isPublic: Option[Boolean] = None) extends WorkbenchGroup
```
Qi77Qi: shall we make `isPublic` as just simple `Boolean`? A policy is either public or private, it doesn't really have a third state right? can we assume a policy is private if user didn't set it public?

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -91,7 +106,7 @@ object SamResourceActions {
 @Lenses case class ResourceId(value: String) extends ValueObject
 
 @Lenses final case class ResourceIdAndPolicyName(resourceId: ResourceId, accessPolicyName: AccessPolicyName)
-@Lenses final case class UserPolicyResponse(resourceId: ResourceId, accessPolicyName: AccessPolicyName, authDomains: Set[WorkbenchGroupName], missingAuthDomains: Set[WorkbenchGroupName])
+@Lenses final case class UserPolicyResponse(resourceId: ResourceId, accessPolicyName: AccessPolicyName, authDomains: Set[WorkbenchGroupName], missingAuthDomains: Set[WorkbenchGroupName], public: Boolean)
```
Qi77Qi: shall we use `isPublic` as well here? just to keep things a bit more consistent

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -139,22 +141,50 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
...
     } yield {
-      val memberOfSubjects = memberOfDns.map(dnToSubject)
-      memberOfSubjects.collect {
-        case ripn@ResourceAndPolicyName(`resource`, _) => unmarshalAccessPolicy(ldapConnectionPool.getEntry(subjectDn(ripn)))
+      entry match {
+        case Some(entry) =>
+          val memberOfDns = getAttributes(entry, Attr.memberOf)
+          val memberOfSubjects = memberOfDns.map(dnToSubject)
+          memberOfSubjects.collect {
+            case ripn@ResourceAndPolicyName(`resource`, _) => unmarshalAccessPolicy(ldapConnectionPool.getEntry(subjectDn(ripn)))
```
Qi77Qi: `ldapConnectionPool.getEntry(subjectDn(ripn))` is not pure. I think you'll need to do something like https://github.com/broadinstitute/sam/pull/221/files#diff-d860b2e8b5a3754c5084cdee4f182cefR143

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
dvoet: the reason it is not `isPublic` here is because it is the external response

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
dvoet: @Qi77Qi I am going to let your PR take care of it ;)

dvoet:
jenkins retest - Error during sbt execution: Error retrieving required libraries

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -138,6 +139,24 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
...
+                        get {
+                          requireOneOfAction(resource, Set(SamResourceActions.readPolicies, SamResourceActions.readPolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
+                            complete(resourceService.isPublic(resourceAndPolicyName))
+                          }
+                        } ~
+                        put {
+                          requireOneOfAction(resource, Set(SamResourceActions.alterPolicies, SamResourceActions.sharePolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
+                            requireOneOfAction(FullyQualifiedResourceId(SamResourceTypes.resourceTypeAdminName, ResourceId(resourceType.name.value)), Set(SamResourceActions.setPublic, SamResourceActions.setPublicPolicy(resourceAndPolicyName.accessPolicyName)), userInfo.userId) {
+                              entity(as[Boolean]) { isPublic =>
+                                complete(resourceService.setPublic(resourceAndPolicyName, isPublic).unsafeToFuture().map(_ => StatusCodes.NoContent))
```
Qi77Qi: I think you don't need to `unsafeToFuture` here anymore

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -197,4 +219,14 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
 
     WorkbenchUser(WorkbenchUserId(uid), getAttribute(entry, Attr.googleSubjectId).map(GoogleSubjectId), WorkbenchEmail(email))
   }
+
+  override def setPolicyIsPublic(policyId: FullyQualifiedPolicyId, public: Boolean): IO[Unit] = {
+    val dateMod = new Modification(ModificationType.REPLACE, Attr.groupUpdatedTimestamp, formattedDate(new Date()))
+    val publicMod = new Modification(ModificationType.REPLACE, Attr.public, public.toString)
+
+    cs.evalOn(blockingEc)(IO(ldapConnectionPool.modify( policyDn(policyId), dateMod, publicMod)).void).recoverWith {
```
Qi77Qi, line 227: very minor, `onError` expects a partial function that returns a `IO[Unit]`, which is a little more constrained than `recoverWith` and maybe slightly better for the use case here

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -335,4 +362,34 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
   }
 
   private def generateGroupEmail() = WorkbenchEmail(s"policy-${UUID.randomUUID}@$emailDomain")
+
+  def isPublic(resourceAndPolicyName: FullyQualifiedPolicyId): IO[Boolean] = {
+    accessPolicyDAO.loadPolicy(resourceAndPolicyName).map {
+      case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "policy not found"))
```
Qi77Qi: `flatMap` and use `IO.raiseError` :P

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -107,28 +108,31 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
...
     val email = WorkbenchEmail(getAttribute(entry, Attr.email).get)
 
     AccessPolicy(
-      FullyQualifiedPolicyId(FullyQualifiedResourceId(resourceTypeName, resourceId), AccessPolicyName(policyName)), members, email, roles, actions)
+      FullyQualifiedPolicyId(FullyQualifiedResourceId(resourceTypeName, resourceId), AccessPolicyName(policyName)), members, email, roles, actions, public.contains(true))
   }
 
   override def overwritePolicyMembers(id: FullyQualifiedPolicyId, memberList: Set[WorkbenchSubject]): IO[Unit] = {
     val memberMod = new Modification(ModificationType.REPLACE, Attr.uniqueMember, memberList.map(subjectDn).toArray:_*)
+    val dateMod = new Modification(ModificationType.REPLACE, Attr.groupUpdatedTimestamp, formattedDate(new Date()))
```
Qi77Qi, line 121: I kinda like `Instant.now()` over `new Date()`.

Qi77Qi (approved):
lgtm...some minor comments

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
dvoet: I think `onError` is different: "Execute a callback on certain errors, then rethrow them." So it rethrows the original error which is not what I want. I think it is meant for side effecting callbacks.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
dvoet: that is not the right type :(

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
Qi77Qi: I don't think that's true
```
qi-workbench-libs@ IO.raiseError[Int](new Exception("1")).onError{case t =>  IO.raiseError(new Exception("2"))}.unsafeRunSync()
java.lang.Exception: 2
  ammonite.$sess.cmd3$$anonfun$1.applyOrElse(cmd3.sc:1)
  ammonite.$sess.cmd3$$anonfun$1.applyOrElse(cmd3.sc:1)
  scala.PartialFunction$AndThen.applyOrElse(PartialFunct
```

Qi77Qi:
```
qi-workbench-libs@ IO.raiseError[Int](new Exception("1")).recoverWith{case t =>  IO(4)}.unsafeRunSync()
res2: Int = 4
```
is what I originally thought would be nice to exclude.

Similar code with `onError` looks like:
```
qi-workbench-libs@ IO.raiseError[Int](new Exception("1")).onError{case t =>  IO(4)}.unsafeRunSync()
java.lang.Exception: 1
  ammonite.$sess.cmd3$.<init>(cmd3.sc:1)
  ammonite.$sess.cmd3$.<clini
``` 

So I was thinking if we really want to recover from error, we should use `recoverWith`, but if we're just converting error type, feels `onError` seems more appropriate.

dvoet: unfortunately it is beyond my patience to wait for another test cycle so I am going to merge but will keep this in mind for the future.

Qi77Qi: sounds good 👍

PR #225, opened 2018-10-09
fix missing auth domains field

Qi77Qi:
Ticket: no ticket

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
@@ -95,7 +95,7 @@ class ManagedGroupService(private val resourceService: ResourceService, private 
       // This will silently ignore any group where the email could not be loaded. This can happen when a
       // managed group is in an inconsistent state (partially created/deleted or created incorrectly).
       // It also includes only admin and member policies
-      ripns.filter(ripn => Seq(ManagedGroupService.adminPolicyName, ManagedGroupService.memberPolicyName).contains(ripn.accessPolicyName)).flatMap { ripn =>
+      ripns.filter(ripn => ManagedGroupService.userMemberShipPolicyNames.contains(ripn.accessPolicyName)).flatMap { ripn =>
         emailLookupMap.get(WorkbenchGroupName(ripn.resourceId.value)).map(email => ManagedGroupMembershipEntry(ripn.resourceId, ripn.accessPolicyName, email))
       }
     }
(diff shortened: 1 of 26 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -169,6 +169,7 @@ object ManagedGroupService {
   val adminPolicyName: ManagedGroupPolicyName = new AccessPolicyName(adminValue) with AllowedManagedGroupPolicyName
   val memberPolicyName: ManagedGroupPolicyName = new AccessPolicyName(memberValue) with AllowedManagedGroupPolicyName
   val adminNotifierPolicyName: ManagedGroupPolicyName = new AccessPolicyName(adminNotifierValue) with AllowedManagedGroupPolicyName
+  val userMemberShipPolicyNames: Set[AccessPolicyName] = Set(adminPolicyName, memberPolicyName)
```
dvoet: membership is one word, not 2: `userMembershipPolicyNames`

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -147,8 +147,9 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
       resources <- if(isConstrained) accessPolicyDAO.listResourceWithAuthdomains(resourceTypeName, rids) else IO.pure(Set.empty)
       authDomainMap = resources.map(x => x.resourceId -> x.authDomain).toMap
 
-      allAuthDomainResourcesUserIsMemberOf <- if(isConstrained) accessPolicyDAO.listAccessPolicies(ManagedGroupService.managedGroupTypeName, userId) else IO.pure(Set.empty[ResourceIdAndPolicyName])
+      allAuthDomain <- if(isConstrained) accessPolicyDAO.listAccessPolicies(ManagedGroupService.managedGroupTypeName, userId) else IO.pure(Set.empty[ResourceIdAndPolicyName])
 
+      allAuthDomainResourcesUserIsMemberOf = allAuthDomain.filter(x => ManagedGroupService.userMemberShipPolicyNames.contains(x.accessPolicyName))
```
dvoet: I think this should call the managed group service to list the groups a user is in instead of duplicating the logic here. If `ManagedGroupService.listGroups` is not quite right as it is, refactor it a little. But I do see the concern that `ManagedGroupService` is already dependent on `ResourceService` and we can't have it going the other way too. Perhaps we should have a `PolicyEvaluatorService` that has `listUserResourceActions` and `hasPermission`.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -664,6 +664,40 @@ class ResourceServiceSpec extends FlatSpec with Matchers with ScalaFutures with
     res.futureValue(Timeout(Span(10, Seconds))) shouldBe expected
   }
 
+  it should "list required authDomains and missing authDomains if user is a member of a universal policy" in {
```
dvoet: what is a universal policy?

dvoet (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
Qi77Qi: I meant it as something like `admin-notifier` which will all users are a member of. any better name suggestions?

dvoet: just say `admin-notifier`

dvoet (approved):

MatthewBemis (approved):

PR #230, opened 2018-10-19
add lock for createPetSerivceAccount

Qi77Qi:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3859
There's no double lock, but I'm thinking maybe we can start with this before making things more complicated and see if most of the race conditions are gone in QA tests

This can't be merged yet since we need a valid project for prod to host firestore

Waiting on https://github.com/broadinstitute/firecloud-automated-testing/pull/76

TODO: 
* figure out tests properly
* PR firecloud-develop to add json cred in config

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -100,51 +100,55 @@ object Boot extends App with LazyLogging {
       val statusService = new StatusService(directoryDAO, cloudExtensions, 10 seconds)
       val managedGroupService = new ManagedGroupService(resourceService, policyEvaluatorService, resourceTypeMap, accessPolicyDAO, directoryDAO, cloudExtensions, emailDomain)
 
-      val samRoutes = cloudExtensions match {
-        case googleExt: GoogleExtensions => new SamRoutes(resourceService, userService, statusService, managedGroupService, config.as[SwaggerConfig]("swagger"), directoryDAO, policyEvaluatorService) with StandardUserInfoDirectives with GoogleExtensionRoutes {
+      val samRoutes = (cloudExtensions, googleFirestoreOps) match {
+        case (googleExt: GoogleExtensions, Some(firestore)) => new SamRoutes(resourceService, userService, statusService, managedGroupService, config.as[SwaggerConfig]("swagger"), directoryDAO, policyEvaluatorService) with StandardUserInfoDirectives with GoogleExtensionRoutes {
```
helgridly: i think from reading it that you mean "you must define both google extensions AND firestore, or neither" -- in which case maybe you want to make the case where only one of them is defined an explicit error, rather than a silent fallback to the no-extensions case?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -100,51 +100,55 @@ object Boot extends App with LazyLogging {
...
-    for {
-      _ <- schemaDAO.init().recoverWith{
-        case e: WorkbenchException =>
-          logger.error("FATAL - could not update schema to latest version. Is the schema lock stuck? See documentation here for more information: [link]")
-          Future.failed(e)
-        case t: Throwable =>
-          logger.error("FATAL - could not init ldap schema", t)
-          Future.failed(t)
-      }
+    val resources: cats.effect.Resource[IO, SamResources] = googleServicesConfigOption match{
```
helgridly: nit: it seems we're pretty inconsistent here, but in general i prefer `match {` to `match{` (and similar). this is probably either a group discussion or a "tell hussein to stop caring" thing though

helgridly: also can you call this var other than `resources`, it's not very descriptive

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -100,51 +100,55 @@ object Boot extends App with LazyLogging {
...
+      samResources =>
+        val accessPolicyDao = new LdapAccessPolicyDAO(ldapConnectionPool, directoryConfig, samResources.blockingExecutionContext)
+        val directoryDAO = new LdapDirectoryDAO(ldapConnectionPool, directoryConfig, samResources.blockingExecutionContext)
+        val cloudExtension = createCloudExt(accessPolicyDao, directoryDAO)
+        val (sRoutes, userService, resourceService, statusService) = createSamRoutes(cloudExtension, accessPolicyDao, directoryDAO, samResources.googleFirestoreOpsInterpreters)
+
+        for{
+          _ <- IO.fromFuture(IO(schemaDAO.init())).onError{
+            case e: WorkbenchException =>
+              IO(logger.error("FATAL - could not update schema to latest version. Is the schema lock stuck? See documentation here for more information: [link]")) *> IO.raiseError(e)
```
helgridly: [link]

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -133,8 +136,8 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
     }.getOrElse(throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"$groupId not found")))
   }
 
-  override def loadSubjectFromEmail(email: WorkbenchEmail): Future[Option[WorkbenchSubject]] = Future {
-    ldapConnectionPool.search(directoryConfig.baseDn, SearchScope.SUB, Filter.createEqualityFilter(Attr.email, email.value)).getSearchEntries.asScala match {
+  override def loadSubjectFromEmail(email: WorkbenchEmail): IO[Option[WorkbenchSubject]] = {
+    cs.evalOn(blockingEc)(IO(ldapConnectionPool.search(directoryConfig.baseDn, SearchScope.SUB, Filter.createEqualityFilter(Attr.email, email.value)).getSearchEntries.asScala)).map{
```
helgridly: checking my understanding: by default, things in this class run on the `executionContext` contained in `cs` (i.e. whatever EC was implicit when this DAO was created), but sometimes for specific functions you want them to run specifically on a blocking EC, and calling `evalOn` is what does that?

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+  }(_ => releaseLock(lockPath))
+
+  private[util] def getLock[A](lockPath: LockPath)(implicit googleFirestoreOps: GoogleFirestoreOps[IO]): IO[Unit] =
+    googleFirestoreOps.transaction { (db, tx) =>
+      val docRef = db.collection(lockPath.collectionName.asString).document(lockPath.document.asString)
+      val res = for {
+        lockStatus <- getLockStatus(tx, docRef)
+        _ <- lockStatus match {
+          case Available => setLock(tx, docRef, lockPath.expiresIn)
+          case Locked => IO.raiseError(new WorkbenchException("can't get lock"))
```
helgridly: put the path in this exception?

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+        }
+      } yield ()
+      res.unsafeRunSync()
+    }
+
+  private[util] def getLockStatus(tx: Transaction, documentReference: DocumentReference): IO[LockStatus] = {
+    IO.async[DocumentSnapshot] { cb =>
+       Try(tx.get(documentReference).get()).fold(t => cb(Left(t)), doc => cb(Right(doc)))
```
helgridly: i'm assuming cb is callback, but i don't know where it's defined or what t and doc are

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+        tx.set(docRef, Map(ISLOCKED -> false).asJava)
+    }
+  }
+
+  private[util] def setLock(tx: Transaction, documentReference: DocumentReference, expiresIn: ExpiresIn): IO[Unit] =
+    for {
+      currentTime <- timer.clock.realTime(EXPIRESINTIMEUNIT)
+      data = Map(
+        EXPIRESIN -> (currentTime + expiresIn.asDuration.toMillis),
+        ISLOCKED -> true
```
helgridly: why _both_ an expiry time _and_ a lock flag? why not just set the expiry to 0 or -1 or null or the current time when you're explicitly unlocking?

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+    } yield ()
+
+  def retry[A](ioa: IO[A], initialDelay: FiniteDuration, maxRetries: Int)
+                         (implicit timer: Timer[IO]): IO[A] = {
+    ioa.handleErrorWith {
+      case error: WorkbenchException =>
+        IO.raiseError(error)
+      case _ =>
+        if (maxRetries > 0)
+          IO.sleep(initialDelay) *> retry(ioa, initialDelay, maxRetries - 1)
```
helgridly: nit: it's not really an _initial_ delay per se, since the same delay is used the first and every time

helgridly (approved):
a few comments but nothing earth-shattering

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: I care about the whitespace too and have given similar feedback in the past. @Qi77Qi can you please add spaces before curly braces?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -226,46 +229,50 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
...
       }
     } yield pet
 
+    val lock = LockPath(CollectionName("createPet"), Document(user.id.value), ExpiresIn(10 seconds))
+
     // In the case of high concurrency, the above code may find that there is no pet SA and then try to create one, getting
     // a 409 Conflict in Google if another call has since created it. This recoverWith will catch that case and return the
     // newly-created pet SA instead of simply failing.
-    pet recoverWith {
+    withLock(lock).use(_ => createPet).unsafeToFuture() recoverWith {
```
dvoet: with the lock I don't think we need the recoverWith anymore (or the above comment)

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+  private[util] val ISLOCKED = "isLocked"
+  private[util] val EXPIRESINTIMEUNIT = TimeUnit.MILLISECONDS
+  implicit val cs = IO.contextShift(ExecutionContext.global)
+  implicit val timer = IO.timer(ExecutionContext.global)
+
+  val executor = new Executor {
+    override def execute(command: Runnable): Unit = ExecutionContext.global.execute(command)
+  }
+
+  def withLock(lockPath: LockPath)(implicit googleFirestoreOps: GoogleFirestoreOps[IO]): Resource[IO, Unit] = Resource.make{
```
dvoet: Sorry, but I will continue to say 👎 to implicits. The whole point of them in this context is to hide that there is some `GoogleFirestoreOps` object that needs to be used. I understand that it can make the calling code look nicer but there are other ways to do that without hiding things. As software engineers, whenever we hide things we make it less understandable and thus harder to maintain and easier to break.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
Qi77Qi: I'll spend a bit more time figuring out how to enable scalafmt on compile...I can't promise I'll remember adding space all the time

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: according to @MatthewBemis there are no docs so we should just remove that.

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.{FiniteDuration, _}
+
+object DistributedLock {
+  private[util] val EXPIRESIN = "expiresIn"
+  private[util] val ISLOCKED = "isLocked"
+  private[util] val EXPIRESINTIMEUNIT = TimeUnit.MILLISECONDS
+  implicit val cs = IO.contextShift(ExecutionContext.global)
+  implicit val timer = IO.timer(ExecutionContext.global)
```
dvoet: I don't like that these are pinned to the global context

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+import scala.util.Try
+import com.google.cloud.firestore.{DocumentReference, DocumentSnapshot, Transaction}
+import org.broadinstitute.dsde.workbench.google.{CollectionName, Document, GoogleFirestoreOps}
+import org.broadinstitute.dsde.workbench.model.WorkbenchException
+
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.{FiniteDuration, _}
+
+object DistributedLock {
```
dvoet: I don't think this should be an object. It should be a class with its dependencies passed in instead of dependencies being implicitly passed to each public function.

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+            expiresIn match{
+              case Some(e) =>
+                timer.clock.realTime(EXPIRESINTIMEUNIT).map{
+                  currentTime =>
+                    if(e < currentTime)
+                      Available
+                    else
+                      Locked
+                }
+              case None => IO.raiseError(new WorkbenchException("Lock expiration time is missing"))
```
dvoet: I would treat this case as expired, otherwise this lock will consistently error

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+        }
+      } yield ()
+      res.unsafeRunSync()
+    }
+
+  private[util] def getLockStatus(tx: Transaction, documentReference: DocumentReference): IO[LockStatus] = {
+    IO.async[DocumentSnapshot] { cb =>
+       Try(tx.get(documentReference).get()).fold(t => cb(Left(t)), doc => cb(Right(doc)))
+      }
+      .flatMap{x =>
```
dvoet: better name than `x`, `documentSnapshot` maybe

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+        if (maxRetries > 0)
+          IO.sleep(initialDelay) *> retry(ioa, initialDelay, maxRetries - 1)
+        else
+          IO.raiseError(new WorkbenchException("can't get lock"))
+    }
+  }
+}
+
+final case class ExpiresIn(asDuration: FiniteDuration) extends AnyVal
+final case class LockPath(collectionName: CollectionName, document: Document, expiresIn: ExpiresIn)
```
dvoet: expiresIn does not seem like it belongs in LockPath

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+    for {
+      currentTime <- timer.clock.realTime(EXPIRESINTIMEUNIT)
+      data = Map(
+        EXPIRESIN -> (currentTime + expiresIn.asDuration.toMillis),
+        ISLOCKED -> true
+      )
+      _ <- IO(tx.set(documentReference, data.asJava))
+    } yield ()
+
+  def retry[A](ioa: IO[A], initialDelay: FiniteDuration, maxRetries: Int)
```
dvoet: better name than `ioa`. This is the thing to retry, right? How about `operation`

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  val executor = new Executor {
+    override def execute(command: Runnable): Unit = ExecutionContext.global.execute(command)
+  }
+
+  def withLock(lockPath: LockPath)(implicit googleFirestoreOps: GoogleFirestoreOps[IO]): Resource[IO, Unit] = Resource.make{
+    retry(getLock(lockPath), 5 seconds, 5)
+  }(_ => releaseLock(lockPath))
+
+  private[util] def getLock[A](lockPath: LockPath)(implicit googleFirestoreOps: GoogleFirestoreOps[IO]): IO[Unit] =
```
dvoet: call this function `acquireLock`

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+  private[util] val EXPIRESINTIMEUNIT = TimeUnit.MILLISECONDS
+  implicit val cs = IO.contextShift(ExecutionContext.global)
+  implicit val timer = IO.timer(ExecutionContext.global)
+
+  val executor = new Executor {
+    override def execute(command: Runnable): Unit = ExecutionContext.global.execute(command)
+  }
+
+  def withLock(lockPath: LockPath)(implicit googleFirestoreOps: GoogleFirestoreOps[IO]): Resource[IO, Unit] = Resource.make{
+    retry(getLock(lockPath), 5 seconds, 5)
```
dvoet: the initial delay and number of retries should be configurable

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
dvoet: it is more like the sleep interval

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+        ISLOCKED -> true
+      )
+      _ <- IO(tx.set(documentReference, data.asJava))
+    } yield ()
+
+  def retry[A](ioa: IO[A], initialDelay: FiniteDuration, maxRetries: Int)
+                         (implicit timer: Timer[IO]): IO[A] = {
+    ioa.handleErrorWith {
+      case error: WorkbenchException =>
+        IO.raiseError(error)
```
dvoet: I am not quite following how the retry will work. retry is called with getLock which throws WorkbenchException if it can't get the lock... wouldn't that be handled by this case and thus not retry?

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
```diff
@@ -0,0 +1,105 @@
...
+    googleFirestoreOps.transaction { (db, tx) =>
+      val docRef = db.collection(lockPath.collectionName.asString).document(lockPath.document.asString)
+      val res = for {
+        lockStatus <- getLockStatus(tx, docRef)
+        _ <- lockStatus match {
+          case Available => setLock(tx, docRef, lockPath.expiresIn)
+          case Locked => IO.raiseError(new WorkbenchException("can't get lock"))
+        }
+      } yield ()
+      res.unsafeRunSync()
```
dvoet: calling `unsafeRunSync` seems odd here. Why doesn't `googleFirestoreOps.transaction` take a function that produces `F[A]` instead of a naked `A`?

dvoet (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLockSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.sam.Generator.genLockPath
+import org.broadinstitute.dsde.workbench.sam.TestSupport
+import org.broadinstitute.dsde.workbench.sam.util.DistributedLock._
+import org.scalatest.{AsyncFlatSpec, Matchers}
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+
+class DistributedLockSpec extends AsyncFlatSpec with Matchers with TestSupport{
```
dvoet: If this requires real google it can't be a unit test

dvoet: perhaps an automation test that makes sure concurrent requests to create a pet work

dvoet: you could probably also mock out google in this case with a TrieMap

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -30,121 +29,174 @@ import org.broadinstitute.dsde.workbench.sam.service._
...
-
-      _ <- io.unsafeToFuture()
-    } yield ()
+    (samRoutes, userService, resourceService, statusService)
   }
-
-  startup()
 }
+
+final case class AppDependencies(googleFirestoreOpsInterpreters: Option[GoogleFirestoreOps[IO]], cloudExtensions: CloudExtensions, directoryDAO: DirectoryDAO, accessPolicyDAO: AccessPolicyDAO)
```
gpolumbo-broad: I'm trying to wrap my head around all these changes.  Are all of these configuration changes required to add the locking functionality?

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -15,7 +17,8 @@ import scala.collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
-class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, protected val directoryConfig: DirectoryConfig)(implicit executionContext: ExecutionContext) extends DirectoryDAO with DirectorySubjectNameSupport with LdapSupport {
+class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, protected val directoryConfig: DirectoryConfig, protected val blockingEc: ExecutionContext)(implicit executionContext: ExecutionContext) extends DirectoryDAO with DirectorySubjectNameSupport with LdapSupport {
```
gpolumbo-broad: How can we be sure that the person constructing a new `LdapDirectoryDAO` is passing `blockingEc` an `ExecutionContext` that actually blocks?  What if they pass in the global EC or something for example?

Can we at least provide some instructions on what sort of of `ExecutionContext` is needed here?  What attributes or behavior must it satisfy and can we check that these things are satisfied in the constructor?

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
Qi77Qi: I think of it as retrying an `IO` of `A`...so I'd prefer `ioa` than `operation`...`ioa` is pretty commonly used through cats-effect doc as well (for instance https://typelevel.org/cats-effect/datatypes/io.html#introduction)...

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLock.scala
Qi77Qi:
```ApiFutures.addCallback(
            db.runTransaction(new Transaction.Function[A]{
              override def updateCallback(transaction: Transaction): A = ops(db, transaction)
            }),
            callBack(cb),
            executor
          )
```
This is the java interface, we have to get an `A` here because of `new Transaction.Function[A]`...`new Transaction.Function[IO[A]]`  doesn't work (java lib probably doesn't understand it).

I think calling `.unsafeRunSync()` here should be okay since it's inside a callback we pass to the java lib, which gets executed asynchronously on java lib's execution context.

src/test/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLockSpec.scala
Qi77Qi: I wanted to test against google api though. cuz I realize a few of my assumptions weren't correct when writing out this test...where should I move this kind of test?

dvoet: Then that is an integration test. There should be 2 levels to the testing, is your code right and are you call google right. The former is the unit test which should have mock google. The latter is an integration test that should be written in /automation... ask @marctalbott :)

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -1,150 +1,223 @@
...
-            val cloudExtensions = googleExt
-          }
-        case _ => new SamRoutes(resourceService, userService, statusService, managedGroupService, config.as[SwaggerConfig]("swagger"), directoryDAO, policyEvaluatorService) with StandardUserInfoDirectives with NoExtensionRoutes
-      }
-      (samRoutes, userService, resourceService, statusService)
+  private[sam] def createAppDependencies(ldapConnectionPool: LDAPConnectionPool, appConfig: AppConfig, resourceTypeMap: Map[ResourceTypeName, ResourceType])(implicit actorSystem: ActorSystem): cats.effect.Resource[IO, AppDependencies] = for {
+    blockingEc <- ExecutionContexts.blockingThreadPool
+    accessPolicyDao = new LdapAccessPolicyDAO(ldapConnectionPool, appConfig.directoryConfig, blockingEc)
+    directoryDAO = new LdapDirectoryDAO(ldapConnectionPool, appConfig.directoryConfig, blockingEc)
+    dep <- appConfig.googleConfig match {
```
dvoet: please let's try to stay away from abbreviations, appDepenencies, or results, not dep. I understand there are some abbreviations that are acceptable (e.g. config on the next line) and there are some places where it is just too verbose. But those should be exceptions. Abbreviations make the code harder to read.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
dvoet: there are not really any constraints but perhaps some doc saying that all ldap calls will be made on the ec and that they block. Also calling it blockingEc is misleading. Perhaps `ldapExecutionContext`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -228,55 +230,50 @@ class GoogleExtensions(val directoryDAO: DirectoryDAO, val accessPolicyDAO: Acce
 
   def createUserPetServiceAccount(user: WorkbenchUser, project: GoogleProject): Future[PetServiceAccount] = {
     val (petSaName, petSaDisplayName) = toPetSAFromUser(user)
-
+    val lockCollectionName = "createUserPetServiceAccount"
+    //TODO: question: a user can only create one PET service?
```
dvoet: there can be 1 pet per project

Qi77Qi:
jenkins retest

Qi77Qi:
jenkins retest

Qi77Qi:
Jenkins retest

Qi77Qi:
Jenkins retest

Qi77Qi:
jenkins retest

dvoet:
why is newrelic-agent-4.6.0.jar commited? Can we pull it from maven instead: https://search.maven.org/artifact/com.newrelic.agent.java/newrelic-agent/4.6.0/jar.

If we must commit it we have some extra hoops to jump through wrt licensing: https://broadinstitute.atlassian.net/wiki/spaces/DSDE/pages/3014826/DSDE+Open+Source+Licensing

automation/README.md
```diff
@@ -9,12 +9,6 @@ See [firecloud-automated-testing](https://github.com/broadinstitute/firecloud-au
 
 ### Set Up
 
-```
```
Qi77Qi, line 12: cc @aweng98

Qi77Qi:
jenkins rebuild

Qi77Qi:
jenkins retest

PR #231, opened 2018-10-24
add endpoint for reading the auth domain of a resource

marctalbott:
Ticket: [GAWB-3925](https://broadinstitute.atlassian.net/browse/GAWB-3925)
- this endpoint takes in a resource and returns the auth domain of that resource

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -840,6 +840,39 @@ paths:
       tags:
         - Resources
 
+  /api/resources/v1/{resourceTypeName}/{resourceId}/authDomain:
+    get:
+      summary: List the groups in the Auth Domain for a resource
+      responses:
+        200:
+          description: List of groups that compose the Auth Domain for this resource. Empty if an Auth Domain has not been set
+          schema:
+            type: array
+            items:
+              type: string
+        403:
+          description: You do not have permission to perform this action on the resource
+        404:
+          description: Resource type or resource does not exist or you are not a member of any policy on the resource
+        500:
+          description: Internal Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
+      parameters:
+        - in: path
+          description: Type of resource
+          name: resourceTypeName
+          required: true
+          type: string
+        - in: path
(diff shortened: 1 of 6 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -840,6 +840,39 @@ paths:
...
+          schema:
+            type: array
+            items:
+              type: string
+        403:
+          description: You do not have permission to perform this action on the resource
+        404:
+          description: Resource type or resource does not exist or you are not a member of any policy on the resource
+        500:
+          description: Internal Error
```
Qi77Qi: I'd prefer `Internal Server Error`

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -85,6 +85,17 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
                   }
                 }
               } ~
+              pathPrefix("authDomain") {
```
Qi77Qi, line 69: I kinda prefer to put this block in an `ResourceRoutes` object since these paths grow pretty nested and I remember spending quite some time figuring out matching braces when working on a separate PR)

Something like
```
object ResourceRoutes{
  val getResourceAuthDomainPath = ...
}
```

src/main/resources/swagger/api-docs.yaml
marctalbott: I can do that, but `Internal Error` is consistent with the other 500 status responses across the Swagger doc

Qi77Qi: maybe we should update all of them? 🤔 (https://tools.ietf.org/html/rfc7231#section-6.6.1)

dvoet (changes requested):
code looks good, also add `read_auth_domain` action to workspace owner, reader and writer roles

dvoet (approved):

Qi77Qi (approved):

PR #232, opened 2018-10-26
list flattened resource users endpoint

marctalbott:
Ticket: [GAWB-3926](https://broadinstitute.atlassian.net/browse/GAWB-3926)

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -253,7 +253,7 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
 
   def getAllResourceUsers(resource: FullyQualifiedResourceId, userInfo: UserInfo): server.Route = {
     get {
-      requireOneOfAction(resource, Set(SamResourceActions.readPolicies), userInfo.userId) {
+      requireAction(resource, SamResourceActions.readPolicies, userInfo.userId) {
         complete(resourceService.listAllFlattenedResourceUsers(resource).map { allUsers =>
           StatusCodes.OK -> allUsers
         })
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -392,4 +392,14 @@ class ResourceService(private val resourceTypes: Map[ResourceTypeName, ResourceT
     _ <- IO.fromFuture(IO(fireGroupUpdateNotification(policyId)))
     } yield ()
   }
+
+  def listAllFlattenedResourceUsers(resourceId: FullyQualifiedResourceId): Future[Set[UserIdInfo]] = {
```
Qi77Qi: do you wanna try change this signature into IO.
Grab me if you need some help :)

Qi77Qi (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -245,4 +250,14 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
       })
     }
   }
+
+  def getAllResourceUsers(resource: FullyQualifiedResourceId, userInfo: UserInfo): server.Route = {
+    get {
+      requireOneOfAction(resource, Set(SamResourceActions.readPolicies), userInfo.userId) {
```
dvoet: it doesn't actually matter but there is a `requireAction` function

dvoet:
would something like this work to list all the users in 1 ldap call:
```
  def listFlattenedResourceMembers(resourceId: FullyQualifiedResourceId): IO[Stream[WorkbenchUser]] = cs.evalOn(blockingEc)(
    IO(ldapSearchStream(peopleOu, SearchScope.ONE, Filter.createSubstringFilter(Attr.memberOf, "", Array.empty[String], resourceDn(resourceId)))(unmarshalUser))
  )
```

marctalbott:
@dvoet it didn't work and I haven't yet figured out how to do it in a single query, but I will look at it again on Monday to see if I can get it down to one ldap call.

dvoet:
I was able to get something on the commandline to work like `'(ismemberof=*resourceId=cclf,resourceType=managed-group,ou=resources,dc=dsde-prod,dc=broadinstitute,dc=org)'` then it is a question of figuring out the right code to make that.

dvoet (approved):

PR #236, opened 2018-10-30
Introduced MockitoSafety to allow mockito mocks to be test instance values

breilly2:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3432

This may or may not be a solution for GAWB-3432. I didn't read the details of the ticket... it was just mentioned as a "flaky test" ticket and this PR helps with flaky test behavior I was experiencing. The flakiness may or not be the same so someone with more knowledge should check before closing the ticket.

---

**PR checklist**


gpolumbo-broad (approved):

Qi77Qi (approved):

breilly2:
jenkins retest (docker hub problems)

breilly2:
@dvoet pointed out that this isn't actually the problem in this case. While introducing `OneInstancePerTest` did seem to help reliability for me, I've since had more success locally so 🤷‍♂️. Even if it does generally help, this isn't the right way to introduce it so I'm closing this PR.

PR #243, opened 2018-11-06
format with scalafmt

Qi77Qi:
Ticket: No ticket
Format all files once. 
scalafmtOnCompile still doesn't work. But you can set up git pre-commit hook if you want.

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -34,144 +34,144 @@ trait ManagedGroupRoutes extends UserInfoDirectives with SecurityDirectives with
...
-          } ~
-          delete {
-            handleDeleteGroup(managedGroup, userInfo)
-          }
-        } ~
-        pathPrefix("requestAccess") {
-          post {
-            handleRequestAccess(managedGroup, userInfo)
-          }
+            post {
```
dvoet: why did this get indented?

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -16,7 +16,14 @@ import javax.net.SocketFactory
 import javax.net.ssl.SSLContext
 import org.broadinstitute.dsde.workbench.dataaccess.PubSubNotificationDAO
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.{Json, Pem}
-import org.broadinstitute.dsde.workbench.google.{GoogleDirectoryDAO, HttpGoogleDirectoryDAO, HttpGoogleIamDAO, HttpGoogleProjectDAO, HttpGooglePubSubDAO, HttpGoogleStorageDAO}
+import org.broadinstitute.dsde.workbench.google.{
+  GoogleDirectoryDAO,
```
marctalbott, line 20: Out of curiosity, what causes each of these to show up on different lines? Just total line length? I think it is easier to read, but was just wondering about the inconsistency between this and line 29 as an example.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -44,143 +43,142 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
...
-                    getActionPermissionForUser(resource, userInfo, action)
-                  }
-                }
-              } ~
-              pathPrefix("authDomain") {
-                pathEndOrSingleSlash {
-                  getResourceAuthDomain(resource, userInfo)
-                }
+                getUserPoliciesForResourceType(resourceType, userInfo) ~
+                  postResource(resourceType, userInfo)
```
marctalbott, line 52: Why the tab for `postResource` here (and with other functions below)? These are just a get and post to the same url, so I'm thinking they should be tabbed in the same amount. Is that just a setting in scalafmt? It seemed to do well with `ManagedGroupRoutes`, but doesn't seem to like `ResourceRoutes` as much. I guess maybe we need to rework the routes files to be more consistent for scalafmt to be able to format them consistently.

marctalbott (approved):
This all looks good to me, I just have a couple of more general questions in addition to the two I already asked:
1. Could you walk me through how to set up the git pre-commit hook/leave some documentation somewhere to help others set it up? That way we could maintain the style consistency that scalafmt would provide without all having to focus our efforts on figuring out how scalafmt would format our code.
2. How easy is it to customize scalafmt? If we decide later on that we don't like how it's formatting something, would it be pretty easy to change that and have another reformatting PR?

Qi77Qi:
will redo this once all other PRs are merged

PR #244, opened 2018-11-06
more Future to IO

Qi77Qi:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3961
* Main file changed is `LdapDirectoryDao`. Unfortunately I forgot to disable scalafmt when switched to this branch working on that file. So diff on this file will be bigger than it should be. Check out https://github.com/broadinstitute/sam/pull/244/files#diff-508e3512e77b252b9224194610088e44L18 for real changes
* while working on this, I notice `/google/petServiceAccount/Segment/Segment` in newrelic response time increased a lot compared what it was before. So I added some metrics around the related ldap call (I inserted metrics in 2 calls. One was for my local testing, but I figured it's nice to have it there anyway, so didn't remove it). It's probably better to have metrics while we're refactoring this code.
---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
@@ -18,20 +18,27 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
 // use ExecutionContexts.blockingThreadPool for blockingEc
-class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, protected val directoryConfig: DirectoryConfig, protected val ecForLdapBlockingIO: ExecutionContext)(implicit executionContext: ExecutionContext) extends DirectoryDAO with DirectorySubjectNameSupport with LdapSupport {
+class LdapDirectoryDAO(
+    protected val ldapConnectionPool: LDAPConnectionPool,
+    protected val directoryConfig: DirectoryConfig,
+    protected val ecForLdapBlockingIO: ExecutionContext)(implicit executionContext: ExecutionContext)
+    extends DirectoryDAO
+    with DirectorySubjectNameSupport
+    with LdapSupport {
   implicit val cs = IO.contextShift(executionContext)
 
   override def createGroup(group: BasicWorkbenchGroup, accessInstructionsOpt: Option[String] = None): IO[BasicWorkbenchGroup] = {
-    val membersAttribute = if (group.members.isEmpty) None else Option(new Attribute(Attr.uniqueMember, group.members.map(subject => subjectDn(subject)).asJava))
+    val membersAttribute =
+      if (group.members.isEmpty) None else Option(new Attribute(Attr.uniqueMember, group.members.map(subject => subjectDn(subject)).asJava))
 
     val accessInstructionsAttr = accessInstructionsOpt.collect {
       case accessInstructions => new Attribute(Attr.accessInstructions, accessInstructions)
     }
 
     val attributes = Seq(
-        new Attribute("objectclass", "top", "workbenchGroup"),
-        new Attribute(Attr.email, group.email.value),
-        new Attribute(Attr.groupUpdatedTimestamp, formattedDate(new Date())),
+      new Attribute("objectclass", "top", "workbenchGroup"),
+      new Attribute(Attr.email, group.email.value),
+      new Attribute(Attr.groupUpdatedTimestamp, formattedDate(new Date())),
     ) ++ membersAttribute ++
       accessInstructionsAttr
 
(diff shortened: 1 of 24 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -42,58 +49,65 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
...
   override def loadGroup(groupName: WorkbenchGroupName): IO[Option[BasicWorkbenchGroup]] = cs.evalOn(ecForLdapBlockingIO) {
-    IO(Option(ldapConnectionPool.getEntry(groupDn(groupName))) map(unmarshalGroup))
-  }
-
-  private def unmarshalGroup(results: Entry): BasicWorkbenchGroup = {
-    val cn = getAttribute(results, Attr.cn).getOrElse(throw new WorkbenchException(s"${Attr.cn} attribute missing: ${results.getDN}"))
-    val email = getAttribute(results, Attr.email).getOrElse(throw new WorkbenchException(s"${Attr.email} attribute missing: ${results.getDN}"))
-    val memberDns = getAttributes(results, Attr.uniqueMember)
+    IO(Option(ldapConnectionPool.getEntry(groupDn(groupName))))
+  }.map(_.flatMap(res => unmarshalGroup(res).toOption))
```
helgridly: this is less readable by a long shot :(

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -42,58 +49,65 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
...
+      memberDns = getAttributes(results, Attr.uniqueMember)
+    } yield BasicWorkbenchGroup(WorkbenchGroupName(cn), memberDns.map(dnToSubject), WorkbenchEmail(email))
 
-  override def loadGroups(groupNames: Set[WorkbenchGroupName]): Future[Seq[BasicWorkbenchGroup]] = Future {
+  override def loadGroups(groupNames: Set[WorkbenchGroupName]): IO[Stream[BasicWorkbenchGroup]] = {
     val filters = groupNames.grouped(batchSize).map(batch => Filter.createORFilter(batch.map(g => Filter.createEqualityFilter(Attr.cn, g.value)).asJava)).toSeq
-    ldapSearchStream(groupsOu, SearchScope.SUB, filters:_*)(unmarshalGroup)
+
+    for {
+      streamOfEither <- cs.evalOn(ecForLdapBlockingIO)(IO(ldapSearchStream(groupsOu, SearchScope.SUB, filters: _*)(unmarshalGroup)))
```
helgridly: :face_with_head_bandage:

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
dvoet: it looks like this quietly drops any errors that happen in unmarshalGroup

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -42,58 +49,65 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
...
-  override def batchLoadGroupEmail(groupNames: Set[WorkbenchGroupName]): Future[Seq[(WorkbenchGroupName, WorkbenchEmail)]] = loadGroups(groupNames).map(_.map(g => g.id -> g.email))
+  override def batchLoadGroupEmail(groupNames: Set[WorkbenchGroupName]): IO[Stream[(WorkbenchGroupName, WorkbenchEmail)]] =
+    loadGroups(groupNames).map(_.map(g => g.id -> g.email))
 
-  override def deleteGroup(groupName: WorkbenchGroupName): Future[Unit] = {
+  override def deleteGroup(groupName: WorkbenchGroupName): Future[Unit] =
     listAncestorGroups(groupName).map { ancestors =>
       if (ancestors.nonEmpty) {
-        throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"group ${groupName.value} cannot be deleted because it is a member of at least 1 other group"))
+        throw new WorkbenchExceptionWithErrorReport(
```
dvoet, line 92: did you mean to get rid of this throw?

dvoet (approved):
2 comments then 👍

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
Qi77Qi: oh, I think this is one of those untouched but accidentally scalafmtted methods...I'll refactor this in a future PR

gpolumbo-broad (approved):

Qi77Qi:
jenkins retest

helgridly (commented):
I'm gonna leave this as a comment for discussion rather than an explicit :-1: , but I find this to be significantly less readable. It feels like we're spending a lot of characters wrangling types in and out of various boxes, and personally I find that with every consecutive call to `map` or `flatMap` I burn a little bit of my ability to understand what's going on. I know it's easy to accidentally drop a `Future` on the floor and miss its exception, but this whole `map flatMap IO.pure cs.evalOn` business feels like boilerplate that gets in the way.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
Qi77Qi: which code you're comparing this with?

Qi77Qi:
@helgridly  If you point to a specific code, I can explain better why the code looks that way better...

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
Qi77Qi: Semantically, cs.evalOn is doing something we didn't have before. We were sharing one global forkandjoin threadpool for everything. VS, with cs.evalOn, we're pushing blocking ldap work to a separate threadpool

dvoet: maybe one good refactor would be to pull all of our ldap calls into a set of utilities that do this evalOn for us. As it stands it litters the code and is easy to forget. There is no reason that `ldapSearchStream(groupsOu, SearchScope.SUB, filters:_*)(unmarshalGroup)` couldn't look the same as before only it returns an IO and runs on the right context.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -41,59 +48,72 @@ class LdapDirectoryDAO(protected val ldapConnectionPool: LDAPConnectionPool, pro
...
-  override def loadGroup(groupName: WorkbenchGroupName): IO[Option[BasicWorkbenchGroup]] = cs.evalOn(ecForLdapBlockingIO) {
-    IO(Option(ldapConnectionPool.getEntry(groupDn(groupName))) map(unmarshalGroup))
+  override def loadGroup(groupName: WorkbenchGroupName): IO[Option[BasicWorkbenchGroup]] = {
+    for {
+      searchResult <- cs.evalOn(ecForLdapBlockingIO)(IO(ldapConnectionPool.getEntry(groupDn(groupName))))
+      res <- Option(searchResult).traverse{
+        entry =>
+          unmarshalGroup(entry).fold(err => IO.raiseError(new WorkbenchException(err)), r => IO.pure(r))
+      }
+    } yield res
```
dvoet: so this particular refactor is not about IO but the desire not to throw exceptions because exceptions are side effects as they cause early exit of the function.

But all this use of Either makes my head hurt. And the fact that I just need to understand that the Left is an error sucks.

So I do understand the reason behind why throw should be avoided but I think this makes the code too hard. If the return type is IO, then fine, return IO.raiseError. But if not, I think we should continue to throw exceptions. It is just so much easier to understand and get right.

dvoet: I am going to riff on this a little bit more:
If we are serious about not throwing exceptions then all of our functions should return an Either and I would argue for Either[ErrorReport, T] just so it is clear the code means an error, not just some String. The idea is that any of our functions should be able to signal an error. If a function does not signal an error today it should not require a bunch of refactoring to do so.  I don't think this will make the code better. 

That's why throwing exceptions is so useful. It allows developers to focus on what is important, trust that when something is in a bad state and we need to get out it just works, anytime, anywhere. Otherwise we need to implement too much plumbing.

Plus we call java code which throws exceptions so we can't get away from it anyway. No matter what we need (and have) a top level exception handler. So let's use it.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -65,18 +65,17 @@ class LdapDirectoryDAO(
...
+
   override def loadGroups(groupNames: Set[WorkbenchGroupName]): IO[Stream[BasicWorkbenchGroup]] = {
     val filters = groupNames.grouped(batchSize).map(batch => Filter.createORFilter(batch.map(g => Filter.createEqualityFilter(Attr.cn, g.value)).asJava)).toSeq
 
-    for {
-      streamOfEither <- cs.evalOn(ecForLdapBlockingIO)(IO(ldapSearchStream(groupsOu, SearchScope.SUB, filters: _*)(unmarshalGroup)))
-      res <- streamOfEither.parSequence.fold(err => IO.raiseError(new WorkbenchException(err)), r => IO.pure(r))
-    } yield res
+    executeLdap(IO(ldapSearchStream(groupsOu, SearchScope.SUB, filters: _*)(unmarshalGroupThrow)))
   }
```
Qi77Qi, line 74: we lose the ability to collect errors this way, but Doug suggested we don't care about collecting here.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -225,9 +225,11 @@ class LdapAccessPolicyDAO(protected val ldapConnectionPool: LDAPConnectionPool,
...
     val publicMod = new Modification(ModificationType.REPLACE, Attr.public, public.toString)
 
-    cs.evalOn(ecForLdapBlockingIO)(IO(ldapConnectionPool.modify( policyDn(policyId), dateMod, publicMod)).void).recoverWith {
+    executeLdap(IO(ldapConnectionPool.modify( policyDn(policyId), dateMod, publicMod)).void).recoverWith {
       case ldape: LDAPException if ldape.getResultCode == ResultCode.NO_SUCH_OBJECT =>
         IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "policy does not exist")))
     }
   }
+
+  private def executeLdap[A](ioa: IO[A]): IO[A] = cs.evalOn(ecForLdapBlockingIO)(ioa)
```
Qi77Qi, line 234: this will only apply to blocking operations with this ldap library, which is probably going to be true for a while

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
Qi77Qi: 1. reason why I made the left type `String` here is because I want to combine these error messages into one easily (if it were Exception, I'd have to write a Semigroup instance for exception to combine them).

2. I think by promoting throwing exceptions, we have to code in a more defensive way and more mentally exhausting to me (have to think about whether a statement will throw all the time and make the decision whether we care to interpret it before propagating it up)

dvoet: I disagree on 2. I find it freeing. We thought about our global exception handler. Now I don't think about exceptions at all because I know they are handled. I only think about exceptions when I know I need to do some special handling which is rare.

Including exceptions in the return type of functions is the equivalent to checked exceptions in java. It forces us to deal with them at every level of the application. So I would argue it is more mental burden.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -11,90 +11,109 @@ import org.broadinstitute.dsde.workbench.sam._
...
 
-  override def loadGroup(groupName: WorkbenchGroupName): IO[Option[BasicWorkbenchGroup]] = cs.evalOn(ecForLdapBlockingIO) {
-    IO(Option(ldapConnectionPool.getEntry(groupDn(groupName))) map(unmarshalGroup))
+  override def loadGroup(groupName: WorkbenchGroupName): IO[Option[BasicWorkbenchGroup]] = {
+    for {
+      searchResult <- executeLdap(IO(ldapConnectionPool.getEntry(groupDn(groupName))))
+      res <- Option(searchResult).traverse{
+        entry =>
+          unmarshalGroup(entry).fold(err => IO.raiseError(new WorkbenchException(err)), r => IO.pure(r))
```
dvoet: use unmarshalGroupThrow

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -11,90 +11,109 @@ import org.broadinstitute.dsde.workbench.sam._
...
 
-  override def loadGroupEmail(groupName: WorkbenchGroupName): Future[Option[WorkbenchEmail]] = Future {
-    Option(ldapConnectionPool.getEntry(groupDn(groupName), Attr.email)) map { results =>
-      WorkbenchEmail(getAttribute(results, Attr.email).getOrElse(throw new WorkbenchException(s"${Attr.email} attribute missing: ${results.getDN}")))
-    }
+  override def loadGroupEmail(groupName: WorkbenchGroupName): IO[Option[WorkbenchEmail]] = {
+    for{
+      searchResult <- executeLdap(IO(ldapConnectionPool.getEntry(groupDn(groupName), Attr.email)))
+      res <- Option(searchResult).traverse{
+        e =>
```
dvoet: better val naming please wrt res and e

dvoet: I also think this could just be
```
    executeLdap(IO(Option(ldapConnectionPool.getEntry(groupDn(groupName), Attr.email))) map { results =>
      WorkbenchEmail(getAttribute(results, Attr.email).getOrElse(throw new WorkbenchException(s"${Attr.email} attribute missing: ${results.getDN}")))
    }
```

Qi77Qi:
jenkins retest

PR #249, opened 2018-11-28
add lock for retrieving service account key

Qi77Qi:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-3992 

This PR only reduces the chance that too many keys will be created at the same time. It doesn't ensure we only create less than 10 keys yet

There shouldn't be significant change in this PR.
Few things needs some attention:
1. new google storage lib list objects with prefix slightly differently when there's `/`
2. new google storage lib returns different exceptions (see https://github.com/broadinstitute/sam/pull/249/files#diff-02552ac0495e804f6baf87d1ab572189R54)
3. With google storage emulator in unit tests, response time increased a bit.

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -21,13 +25,15 @@ import scala.concurrent.{ExecutionContext, Future}
 class GoogleKeyCache(
     val googleIamDAO: GoogleIamDAO,
     val googleStorageDAO: GoogleStorageDAO,
```
dvoet: do we still need the old `googleStorageDAO`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -41,59 +47,65 @@ class GoogleKeyCache(
...
-      case (keyObjects, serviceAccountKeys) => retrieveActiveKey(pet, keyObjects, serviceAccountKeys)
-    }
-  }
+  override def getKey(pet: PetServiceAccount): IO[String] =
+    for {
+      keyObjects <- googleStorageAlg
+        .unsafeListObjectsWithPrefix(googleServicesConfig.googleKeyCacheConfig.bucketName, keyNamePrefix(pet.id.project, pet.serviceAccount.email))
+      keys <- IO.fromFuture(IO(googleIamDAO.listUserManagedServiceAccountKeys(pet.id.project, pet.serviceAccount.email).map(_.toList)))
+      cleanedKeyObjects <- IO.fromFuture(IO(cleanupUnknownKeys(pet, keyObjects, keys)))
+      res <- (cleanedKeyObjects, keys) match {
```
dvoet: avoid abbreviations: `result` instead of `res`

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -36,6 +36,43 @@ import scala.concurrent.duration._
...
+//import org.mockito.Mockito._
+//import org.scalatest.concurrent.ScalaFutures
+//import org.scalatest.mockito.MockitoSugar
+//import org.scalatest.{BeforeAndAfterAll, FlatSpecLike, Matchers, PrivateMethodTester}
+//
+//import scala.concurrent.ExecutionContext.Implicits.global
+//import scala.concurrent.Future
+//import scala.concurrent.duration._
+//import scala.language.postfixOps
+//import scala.util.{Success, Try}
```
dvoet: remove commented imports

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -753,108 +790,103 @@ class GoogleExtensionSpec(_system: ActorSystem) extends TestKit(_system) with Fl
...
-
-    //get a key again, which should return the original cached key created above
-    val secondKey = runAndWait(googleExtensions.getPetServiceAccountKey(defaultUser, googleProject))
-
-    assert(firstKey == secondKey)
+    val firstKey = googleExtensions.getPetServiceAccountKey(defaultUser, googleProject).unsafeRunSync()
+    1 shouldBe(1)
+//    //get a key again, which should return the original cached key created above
+//    val secondKey = googleExtensions.getPetServiceAccountKey(defaultUser, googleProject).unsafeRunSync()
+//    assert(firstKey == secondKey)
```
dvoet: fix this test

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -753,108 +790,103 @@ class GoogleExtensionSpec(_system: ActorSystem) extends TestKit(_system) with Fl
...
+//    val secondKey = googleExtensions.getPetServiceAccountKey(defaultUser, googleProject).unsafeRunSync()
+//
+//    // assert that keys have been removed from service account
+//    assert(removedKeyObjects.forall { removed =>
+//      val existingKeys = runAndWait(googleExtensions.googleIamDAO.listUserManagedServiceAccountKeys(googleProject, petServiceAccount.serviceAccount.email))
+//      !existingKeys.exists(key => removed.value.endsWith(key.id.value))
+//    })
+//
+//    assert(firstKey != secondKey)
+//  }
```
dvoet: put these tests back in

dvoet (changes requested):
I think we can get rid of the the old google storage dao completely. Also tests are unfinished.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
Qi77Qi: isn't `res` common enough to abbreviate?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
Qi77Qi: it's not directly used by `GoogleKeyCache` anymore. But it's referenced in `GoogleKeyCacheMonitorSupervisor` for publishing notifications to pubsub. I guess I can pass the old `googleStorageDAO` directly to `GoogleKeyCacheMonitorSupervisor`.

I didn't want to implement the feature in new google storage because there seems better way to do this now (altho, I haven't looked into it yet)

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -254,7 +254,8 @@ class GoogleExtensions(
     googleDirectoryDAO.deleteGroup(groupEmail)
 
   @deprecated("Use new two-argument version of this function", "Sam Phase 3")
-  def createUserPetServiceAccount(user: WorkbenchUser): Future[PetServiceAccount] = createUserPetServiceAccount(user, petServiceAccountConfig.googleProject)
+  def createUserPetServiceAccount(user: WorkbenchUser): Future[PetServiceAccount] =
```
Qi77Qi, line 248: can we remove this method and the deprecated `deleteUserPetServiceAccount`

marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
dvoet: no, my first thought is that it means `resource`

There are trade offs with abbreviations. They usually make code harder to understand. They can make it easier but only when names get too long. They make the code easier to write but that is pretty much completely offset by autocomplete.

So in the case of `res`, what is gained in exchange for decreased readability?

Qi77Qi: fwiw, I never abbreviated `resource` as `res` and I agree abbreviating `resource` as `res` would be really confusing...I thought `res` or sometimes `ret` is common enough that there's no confusion at all

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -41,59 +51,90 @@ class GoogleKeyCache(
...
+        case (_, Nil) =>
+          furnishNewKey(pet) //mismatch. there were no keys found on the service account, but there may be keys in the bucket
+        case (keyObjects, serviceAccountKeys) =>
+          retrieveActiveKey(pet, keyObjects, serviceAccountKeys)
+      }
+    } yield res
+
+    val lockPath = LockPath(CollectionName(s"${pet.id.project.value}-getKey"), Document(pet.serviceAccount.subjectId.value), 20 seconds)
+    for{
+      (cleanedKeyObjects, keys) <- fetchCleanedKeyObjectsAndKeys(pet)
```
dvoet: There is a race condition with `cleanupUnknownKeys` as it runs outside the lock and could potentially remove keys from google after they are created but not yet in the bucket.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -41,59 +51,75 @@ class GoogleKeyCache(
...
-
-    retrievedKeys.flatMap {
-      case (Nil, _) => furnishNewKey(pet) //mismatch. there were no keys found in the bucket, but there may be keys on the service account
-      case (_, Nil) => furnishNewKey(pet) //mismatch. there were no keys found on the service account, but there may be keys in the bucket
-      case (keyObjects, serviceAccountKeys) => retrieveActiveKey(pet, keyObjects, serviceAccountKeys)
-    }
+  override def getKey(pet: PetServiceAccount): IO[String] = {
+    val fetchKeyObjects = googleStorageAlg
+      .unsafeListObjectsWithPrefix(googleServicesConfig.googleKeyCacheConfig.bucketName, keyNamePrefix(pet.id.project, pet.serviceAccount.email))
+    val fetchKeys = IO.fromFuture(IO(googleIamDAO.listUserManagedServiceAccountKeys(pet.id.project, pet.serviceAccount.email).map(_.toList)))
```
dvoet: I would calls these `keysFromCache` and `keysFromGoogle`

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -41,91 +50,113 @@ class GoogleKeyCache(
...
+      for {
+        (keysFromCache, keysFromIam) <- fetchKeysFromCacheAndIam(pet)
+        cleanedKeyObjects <- IO.fromFuture(IO(cleanupUnknownKeys(pet, keysFromCache, keysFromIam)))
+        key <- retrieveNewKey(keysFromCache, keysFromIam, furnishNewKey(pet))
+      } yield key
     }
+
+    for {
+      (keysFromCache, keysFromIam) <- fetchKeysFromCacheAndIam(pet)
+      knownKeyCachedObjects = keysFromCache.filter { cachedObject =>
```
dvoet: unused

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -41,91 +50,113 @@ class GoogleKeyCache(
...
+        createNewKey //mismatch. there were no keys found in the cache, but there may be keys on the service account
+      case (_, Nil) =>
+        createNewKey //mismatch. there were no keys found on the service account, but there may be keys in the bucket
+      case (keyObjects, serviceAccountKeys) =>
+        val mostRecentKey = keysFromCache.sortBy(_.timeCreated.toEpochMilli).last
+
+        if (isKeyActive(mostRecentKey, keysFromIam)) {
+          googleStorageAlg
+            .unsafeGetObject(googleServicesConfig.googleKeyCacheConfig.bucketName, GcsBlobName(mostRecentKey.value))
+            .flatMap(activeKeyFromCache => activeKeyFromCache.fold(createNewKey)(s => IO.pure(s)))
```
dvoet: no fold please

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -41,91 +50,113 @@ class GoogleKeyCache(
...
-        keyNamePrefix(pet.id.project, pet.serviceAccount.email))
-      keys <- googleIamDAO.listUserManagedServiceAccountKeys(pet.id.project, pet.serviceAccount.email).map(_.toList)
-      cleanedKeyObjects <- cleanupUnknownKeys(pet, keyObjects, keys)
-    } yield (cleanedKeyObjects, keys)
-
-    retrievedKeys.flatMap {
-      case (Nil, _) => furnishNewKey(pet) //mismatch. there were no keys found in the bucket, but there may be keys on the service account
-      case (_, Nil) => furnishNewKey(pet) //mismatch. there were no keys found on the service account, but there may be keys in the bucket
-      case (keyObjects, serviceAccountKeys) => retrieveActiveKey(pet, keyObjects, serviceAccountKeys)
+  override def getKey(pet: PetServiceAccount): IO[String] = {
```
dvoet, line 69: I found some of the logic in here hard to follow so I took a stab at it myself. Let me know what you think or just accept the PR if you like it: https://github.com/broadinstitute/sam/pull/271

dvoet (approved):

PR #252, opened 2018-11-30
Serve Swagger API JSON along with YAML

dinvlad:
Ticket: https://broadinstitute.atlassian.net/projects/GAWB/issues/GAWB-4013

I'd like to add support for fetching Sam Swagger API doc in JSON format.

This will be useful, for example, for calling Sam API from Google Cloud Deployment Manager. The latter can use external APIs, however we must provide a public URL to the Swagger doc in JSON format (YAML is not supported). However, currently only YAML is published in Sam (`/api-docs.yaml`).

The proposed solution is to convert YAML to JSON at build time, and include it as an additional route in Swagger UI (`/api-docs.json`).

---

**PR checklist**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ dist/*
 target/
 lib_managed/
 src_managed/
+src/main/resources/swagger/api-docs.json
 project/boot/
 project/plugins/project/
 docker/stand-alone/server.crt
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -91,9 +91,12 @@ function make_jar()
     echo "building jar..."
     OPENDJ=$(bash ./docker/run-opendj.sh start jenkins | tail -n1)
     echo $OPENDJ
-    
+
     # Get the last commit hash of the model directory and set it as an environment variable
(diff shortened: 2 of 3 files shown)


Qi77Qi (approved):

Qi77Qi:
can you update swagger file as well for the added endpoint https://github.com/broadinstitute/sam/blob/develop/src/main/resources/swagger/api-docs.yaml

dinvlad:
Sure - `/api-docs.yaml` itself is not in that file though, is that expected?

Qi77Qi:
oh, mmm, then I'm not sure if it should actually be in there... maybe we don't want to expose it to user. 
 @dvoet or @MatthewBemis can you guys comment on this?

dvoet (approved):

dvoet:
jenkins retest

dinvlad:
I think we no longer have a need for this. It turns out that the API is not compatible with Deployment Manager, which puts particular requirements on CRUD operations supported by the API. So feel free to close it if you think others won't benefit from it. Thanks

Qi77Qi:
thanks @dinvlad . I think in that case, we can just close this one. We can always reopen this if needed

PR #258, opened 2018-12-06
use ehcache for memberOf lists

dvoet:
This PR adds 2 caches to sam:
- memberOf cache: for each user cache the results of the query that lists which groups the user is a member of. This call is particularly hard on opendj because it is a computed result and it produces a lot of data. Caching this means that when users are added or removed from groups it can take up to the cache expiry for the change to take effect.
- resource cache: cache the results when we load a resource (specifically auth domains). A single one of these calls is not a big deal but in the course of listing resource policies we need to load hundreds or thousands of resources. And they never change.
---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
@@ -24,7 +24,7 @@ class LdapDirectoryDAO(
     protected val ldapConnectionPool: LDAPConnectionPool,
     protected val directoryConfig: DirectoryConfig,
     protected val ecForLdapBlockingIO: ExecutionContext,
-    protected val memberOfCache: Cache[String, Set[String]])(implicit executionContext: ExecutionContext, timer: Timer[IO])
+    protected val memberOfCache: Cache[WorkbenchSubject, Set[String]])(implicit executionContext: ExecutionContext, timer: Timer[IO])
     extends DirectoryDAO
     with DirectorySubjectNameSupport
     with LdapSupport {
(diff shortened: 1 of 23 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -125,15 +127,13 @@ class LdapDirectoryDAO(
         case Left(regrets) => IO.raiseError(regrets)
       }
 
-  override def isGroupMember(groupId: WorkbenchGroupIdentity, member: WorkbenchSubject): Future[Boolean] = Future {
-    val isMember = for {
-      entry <- Option(ldapConnectionPool.getEntry(subjectDn(member), Attr.memberOf))
-      memberOf <- Option(entry.getAttribute(Attr.memberOf))
+  override def isGroupMember(groupId: WorkbenchGroupIdentity, member: WorkbenchSubject): IO[Boolean] = {
+    for {
```
gpolumbo-broad, line 131: Nit?  - `for` comprehension with only one entry in it possibly reads better as a map and flatmap?  (It may very well not)

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
```diff
@@ -51,9 +53,22 @@ trait PropertyBasedTesting extends PropertyChecks with Configuration with Matche
   implicit override val generatorDrivenConfig: PropertyCheckConfiguration = PropertyCheckConfiguration(minSuccessful = 3)
 }
 
-object TestSupport extends TestSupport{
+object TestSupport extends TestSupport {
   private val executor = Executors.newCachedThreadPool()
   val blockingEc = ExecutionContext.fromExecutor(executor)
+  def testCache: Cache[WorkbenchSubject, Set[String]] = {
```
gpolumbo-broad: Is there some way we can test the lookup-and-evict logic?  For example:

- Assert User A is not a member of Group B
- Add User A to Group B
- Assert User A is a member of Group B
- Remove User A from Group B
- Assert User A still shows as member of Group B
- Wait for cache expiry (set it to 5 seconds or something low/reasonable)
- Assert User A is not a member of Group B

gpolumbo-broad (approved):

dvoet:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
dvoet: for the moment I am content that the integration tests fail if this is not right. It is a bit difficult to test because if the cache expiry is too long the tests are slow but if it is too short the test is flakey.

PR #261, opened 2018-12-14
fix stability when large amounts of group syncing happens

dvoet:
Let me tell you a story. Once upon a time integration tests ran on alpha. The sun was out and everyone was happy until the big bad whitelist sync tests came to town. They huffed and they puffed and they blew sam and opendj down. 

First we looked at opendj to see why it was unhappy. The ldap audit logs showed a bunch of slow queries using the isMemberOf attribute to figure out intersection groups. That only happens during google group synching! And there are a lot of calls all involving the TCGA group. Who is calling google group syncing? Hey tests, why are you syncing the whitelist 3 time in quick succession (according to Kibana logs)!?! That is not so great, but still sam and opendj should be able to stand up to that load.

First theory: isMemberOf queries are not that bad when the volume is small. When there are too many it is too cpu intensive and everything slows down. Solution rewrite the queries to do our own recursive lookups since individually they are pretty fast.

Test again: BOOM! no dice. Plus now the out of memory (OOM) killer likes to stomp on sam every now and then. The isMemberOf queries are gone now but now there are BIND requests in the ldap audit logs that are taking as long as 2+ minutes! Why are there so many BIND requests. Those are for authenticating new ldap connections. But we have a ldap connection pool why would we have too many connections?

Next theory: Something is up with our connection pool. After reading the docs a bit, turns out our connection pool has a "create if necessary" feature where if it runs out of connections it will create new one, and it is on by default, how thoughtful. Fortunately there are settings to turn that off and set how long to wait for a connection. OK, turn it off, set the timeout to 30s.

Test again: BOOM! no dice: the OOM killer is having a field day in sam. Opendj seems ok though without any queries above a few hundred ms.

Next theory: sam has too much memory for the VM. Let's try trimming the heap size. Still have problems with 1.5g (down from 2g) and 1g. Hmmmm... what the hell is sam doing!?! Let's look at a thread dump... holy crap why are there >1k threads!?! At rest sam has about 70 threads, when sync starts >1k are added, that would add some non-heap memory pressure! This smells like our ldap execution context which has an unbounded thread pool size.

Next theory: change our unbounded thread pool to a fixed thread pool with the same size as our ldap connection pool.

Test again: now we have a few dice but not all of them. Sam and opendj seem stable as in they are doing things and the things they do are moving along pretty fast. But sam is generally unresponsive to other queries (especially other sync requests) when the sync is being processed in the background. A peak at the thread dump shows that all of threads on sam are busy doing sync stuff leaving no resources for other requests.

Next theory: we need a way to let the background stuff run without crowding everything else out. Let's create a background context, thread pool, ldap pool and run all the background stuff there leaving the main context, etc. for servicing api calls.

Test again: still some problems... the main thread pool is still full of junk but it seems weird. [try a bunch of things, bang head on table, more trying, more banging]. On a whim convert the onGroupUpdate code from Future to IO. Eureka! I am not 100% certain why that works but the code still seems to have the same functionality so I'll take it. 

But the tests still fail on alpha but they cruise along, no timeouts... god damn caching! Now I need to figure that out but that is another story.

new config settings: https://github.com/broadinstitute/firecloud-develop/pull/1474
wait for caching to clear: https://github.com/broadinstitute/firecloud-orchestration/pull/697

---

**PR checklist**

### DIFF
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -60,6 +60,14 @@
         <appender-ref ref="Sentry" />
     </logger>
 
+    <!-- publishing messages can be too verbose logging -->
+    <logger name="org.broadinstitute.dsde.workbench.google.HttpGooglePubSubDAO" level="info" additivity="false">
+        <appender-ref ref="file"/>
+        <appender-ref ref="console"/>
+        <appender-ref ref="SYSLOG"/>
+        <appender-ref ref="Sentry" />
+    </logger>
+
     <root level="warn">
         <appender-ref ref="file"/>
         <appender-ref ref="console"/>
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -11,7 +11,7 @@ import cats.effect.{ExitCode, IO, IOApp}
 import cats.implicits._
 import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
-import com.unboundid.ldap.sdk.{LDAPConnection, LDAPConnectionPool}
+import com.unboundid.ldap.sdk._
 import javax.net.SocketFactory
 import javax.net.ssl.SSLContext
 import org.broadinstitute.dsde.workbench.dataaccess.PubSubNotificationDAO
@@ -89,9 +89,11 @@ object Boot extends IOApp with LazyLogging {
     val port = if (dirURI.getPort > 0) dirURI.getPort else defaultPort
     cats.effect.Resource.make(
       IO {
+        val connectionOptions = new LDAPConnectionOptions
+        connectionOptions.setDisconnectHandler(SamLdapConnectionDisconnectHandler)
         val connectionPool = new LDAPConnectionPool(
-          new LDAPConnection(socketFactory, dirURI.getHost, port, directoryConfig.user, directoryConfig.password),
-          directoryConfig.connectionPoolSize
+          new LDAPConnection(socketFactory, connectionOptions, dirURI.getHost, port, directoryConfig.user, directoryConfig.password),
+          directoryConfig.connectionPoolSize, directoryConfig.connectionPoolSize, SamLdapConnectionPostProcessor
         )
         connectionPool.setCreateIfNecessary(false)
         connectionPool.setMaxWaitTimeMillis(30000)
@@ -267,3 +269,15 @@ final case class AppDependencies(
     directoryDAO: DirectoryDAO,
     accessPolicyDAO: AccessPolicyDAO,
     policyEvaluatorService: PolicyEvaluatorService)
+
+object SamLdapConnectionPostProcessor extends PostConnectProcessor with LazyLogging {
+  override def processPreAuthenticatedConnection(connection: LDAPConnection): Unit = logger.info(s"processPreAuthenticatedConnection -- ${connection.getConnectionID}")
+  override def processPostAuthenticatedConnection(connection: LDAPConnection): Unit = logger.info(s"processPostAuthenticatedConnection -- ${connection.getConnectionID}")
+}
+
+object SamLdapConnectionDisconnectHandler extends DisconnectHandler with LazyLogging {
+  override def handleDisconnect(
+      connection: LDAPConnection, host: String, port: Int, disconnectType: DisconnectType, message: String, cause: Throwable): Unit = {
+    logger.info(s"handleDisconnect -- ${connection.getConnectionID}", cause)
+  }
+}
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -20,9 +20,9 @@ import org.broadinstitute.dsde.workbench.google.util.DistributedLock
 import org.broadinstitute.dsde.workbench.google.{GoogleDirectoryDAO, GoogleFirestoreOpsInterpreters, HttpGoogleDirectoryDAO, HttpGoogleIamDAO, HttpGoogleProjectDAO, HttpGooglePubSubDAO, HttpGoogleStorageDAO}
 import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchException, WorkbenchSubject}
 import org.broadinstitute.dsde.workbench.sam.api.{SamRoutes, StandardUserInfoDirectives}
-import org.broadinstitute.dsde.workbench.sam.config.{AppConfig, DirectoryConfig, GoogleConfig}
+import org.broadinstitute.dsde.workbench.sam.config.{AppConfig, GoogleConfig}
 import org.broadinstitute.dsde.workbench.sam.directory._
-import org.broadinstitute.dsde.workbench.sam.google.{GoogleExtensionRoutes, GoogleExtensions, GoogleKeyCache}
+import org.broadinstitute.dsde.workbench.sam.google._
 import org.broadinstitute.dsde.workbench.sam.model._
 import org.broadinstitute.dsde.workbench.sam.openam._
 import org.broadinstitute.dsde.workbench.sam.schema.JndiSchemaDAO
@@ -68,7 +68,7 @@ object Boot extends IOApp with LazyLogging {
 
         _ <- dependencies.policyEvaluatorService.initPolicy()
 
-        _ <- dependencies.cloudExtensions.onBoot(dependencies.samApplication)
+        _ <- dependencies.cloudExtensionsInitializer.onBoot(dependencies.samApplication)
 
         binding <- IO.fromFuture(IO(Http().bindAndHandle(dependencies.samRoutes.route, "0.0.0.0", 8080))).onError {
           case t: Throwable => IO(logger.error("FATAL - failure starting http server", t)) *> IO.raiseError(t)
@@ -79,8 +79,8 @@ object Boot extends IOApp with LazyLogging {
     }
   }
 
-  private[sam] def createLdapConnectionPool(directoryConfig: DirectoryConfig): cats.effect.Resource[IO, LDAPConnectionPool] = {
-    val dirURI = new URI(directoryConfig.directoryUrl)
+  private[sam] def createLdapConnectionPool(directoryUrl: String, user: String, password: String, connectionPoolSize: Int): cats.effect.Resource[IO, LDAPConnectionPool] = {
+    val dirURI = new URI(directoryUrl)
     val (socketFactory, defaultPort) = dirURI.getScheme.toLowerCase match {
       case "ldaps" => (SSLContext.getDefault.getSocketFactory, 636)
       case "ldap" => (SocketFactory.getDefault, 389)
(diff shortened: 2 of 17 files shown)


Qi77Qi:
is the intention to reduce some burden on opendj?

dvoet:
jenkins retest

src/main/resources/logback.xml
```diff
@@ -60,6 +60,14 @@
         <appender-ref ref="Sentry" />
     </logger>
 
+    <!-- publishing messages can be too verbose logging -->
```
dvoet, line 63: I put this in because it was being pretty verbose, making logging slow down and I thought trigger the OOM killer. It may have been a contributing factor but I don't think causal. Still, it was pretty big and pretty useless so I am leaving this in.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -255,12 +255,40 @@ class LdapDirectoryDAO(
...
-      SearchScope.SUB,
-      Filter.createANDFilter(groupIds.map(groupId => Filter.createEqualityFilter(Attr.memberOf, groupDn(groupId))).asJava)
-    )(getAttribute(_, Attr.uid)).flatten.map(WorkbenchUserId).toSet
+//  override def listIntersectionGroupUsers(groupIds: Set[WorkbenchGroupIdentity]): IO[Set[WorkbenchUserId]] = IO {
+//    ldapSearchStream(
+//      directoryConfig.baseDn,
+//      SearchScope.SUB,
+//      Filter.createANDFilter(groupIds.map(groupId => Filter.createEqualityFilter(Attr.memberOf, groupDn(groupId))).asJava)
+//    )(getAttribute(_, Attr.uid)).flatten.map(WorkbenchUserId).toSet
+//  }
```
dvoet, line 264: this was the big bad isMemberOf query... I am not ready to forget it so I am leaving it in for now

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -449,125 +429,12 @@ class GoogleExtensions(
...
 
-  def getSynchronizedDate(groupId: WorkbenchGroupIdentity): Future[Option[Date]] =
+  def getSynchronizedDate(groupId: WorkbenchGroupIdentity): IO[Option[Date]] =
     directoryDAO.getSynchronizedDate(groupId)
 
-  def getSynchronizedEmail(groupId: WorkbenchGroupIdentity): Future[Option[WorkbenchEmail]] =
+  def getSynchronizedEmail(groupId: WorkbenchGroupIdentity): IO[Option[WorkbenchEmail]] =
     directoryDAO.getSynchronizedEmail(groupId)
 
-  def synchronizeGroupMembers(
```
dvoet, line 458: moved the GoogleGroupSynchronizer

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -160,33 +140,33 @@ class GoogleExtensions(
     googlePubSubDAO.publishMessages(googleServicesConfig.groupSyncTopic, Seq(id.toJson.compactPrint))
 
   override def onGroupUpdate(groupIdentities: Seq[WorkbenchGroupIdentity]): Future[Unit] =
-    onGroupUpdateRecursive(groupIdentities, Seq.empty)
+    onGroupUpdateRecursive(groupIdentities, Seq.empty).unsafeToFuture()
```
dvoet, line 143: the Future to IO changes below

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -647,3 +514,20 @@ class GoogleExtensions(
 
   override val allSubSystems: Set[Subsystems.Subsystem] = Set(Subsystems.GoogleGroups, Subsystems.GooglePubSub, Subsystems.GoogleIam)
 }
+
+case class GoogleExtensionsInitializer(cloudExtensions: GoogleExtensions, googleGroupSynchronizer: GoogleGroupSynchronizer) extends CloudExtensionsInitializer {
```
dvoet, line 518: added this class to limit the impact to Boot.scala and try to keep is cloud agnostic

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
```diff
@@ -0,0 +1,136 @@
...
+import org.broadinstitute.dsde.workbench.sam.directory.DirectoryDAO
+import org.broadinstitute.dsde.workbench.sam.model._
+import org.broadinstitute.dsde.workbench.sam.openam.AccessPolicyDAO
+import org.broadinstitute.dsde.workbench.sam._
+import org.broadinstitute.dsde.workbench.util.FutureSupport
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
+
+class GoogleGroupSynchronizer(directoryDAO: DirectoryDAO,
```
dvoet, line 17: the functionality in this class was pulled from GoogleExtensions. It is separate because all these things now run on the background context, etc. and I didn't want that confusion in GoogleExetensions. Instead Boot figures it all out.

dvoet:
jenkins retest - leo failure looks unrelated

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -79,20 +80,25 @@ object Boot extends IOApp with LazyLogging {
     }
   }
 
-  private[sam] def createLdapConnectionPool(directoryConfig: DirectoryConfig): cats.effect.Resource[IO, LDAPConnectionPool] = {
-    val dirURI = new URI(directoryConfig.directoryUrl)
+  private[sam] def createLdapConnectionPool(directoryUrl: String, user: String, password: String, connectionPoolSize: Int, name: String): cats.effect.Resource[IO, LDAPConnectionPool] = {
```
Qi77Qi, line 83: Why expanding it to individual parameters?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -138,12 +144,20 @@ object Boot extends IOApp with LazyLogging {
...
-      directoryDAO = new LdapDirectoryDAO(ldapConnectionPool, appConfig.directoryConfig, blockingEc, memberOfCache)
+      ldapExecutionContext <- ExecutionContexts.fixedThreadPool[IO](appConfig.directoryConfig.connectionPoolSize)
+      accessPolicyDao = new LdapAccessPolicyDAO(ldapConnectionPool, appConfig.directoryConfig, ldapExecutionContext, memberOfCache, resourceCache)
+      directoryDAO = new LdapDirectoryDAO(ldapConnectionPool, appConfig.directoryConfig, ldapExecutionContext, memberOfCache)
+
+      // This special set of objects are for operations that happen in the background, i.e. not in the immediate service
+      // of an api call. They are meant to partition resources so that background processes can't crowd our api calls.
+      backgroundLdapConnectionPool <- createLdapConnectionPool(appConfig.directoryConfig.directoryUrl, appConfig.directoryConfig.user, appConfig.directoryConfig.password, appConfig.directoryConfig.backgroundConnectionPoolSize, "background")
+      backgroundLdapExecutionContext <- ExecutionContexts.fixedThreadPool[IO](appConfig.directoryConfig.backgroundConnectionPoolSize)
```
Qi77Qi, line 157: I don’t think it’s necessary to create an additional two instances just for running some IOs on a different threadpool. But if we want to get this out ASAP. I guess it should be fine functionally and we can improve in a future PR

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -160,33 +140,33 @@ class GoogleExtensions(
...
       }
       // only sync groups that have already been synchronized
       messagesForIdsWithSyncDates = idsAndSyncDates.collect {
         case (gn: WorkbenchGroupName, Some(_)) => gn.toJson.compactPrint
         case (rpn: FullyQualifiedPolicyId, Some(_)) => rpn.toJson.compactPrint
       }
-      _ <- googlePubSubDAO.publishMessages(googleServicesConfig.groupSyncTopic, messagesForIdsWithSyncDates)
-      ancestorGroups <- Future.traverse(groupIdentities) { id =>
-        directoryDAO.listAncestorGroups(id).unsafeToFuture()
+      _ <- IO.fromFuture(IO(googlePubSubDAO.publishMessages(googleServicesConfig.groupSyncTopic, messagesForIdsWithSyncDates)))
```
Qi77Qi, line 155: If this future runs things in its own threadpool, we may need to add another line to shift back to main ec (otherwise we may end up running things on the pubsub threadpool unintendedly)

marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: so it can be called with different pool size, once for the foreground pool, once for the background pool

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: We can certainly think about alternative approaches but this one is pretty straight forward to understand without confusing business logic.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
dvoet: it is using the global execution context... but IO.fromFuture will shift the context? That doesn't sound right.

Qi77Qi: If it’s using global execution context, then we should be fine (as long as it’s not blocking)

https://github.com/typelevel/cats-effect/issues/444 is what I’m referring to.

I meant to revisit all ‘fromFuture’ calls, but haven’t got to it yet

zarsky-broad:
Soundtrack to this PR: https://www.youtube.com/watch?v=LiGRPYmTFV0

Qi77Qi (approved):
I think we can do some further improvements. But I guess we can have this for now to unblock Sam phase 4

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
dvoet: this is a discussion for another time but after reading that issue I feel like all these execution contexts and context shifts are more trouble than they are worth. I long for the days of single thread per api call. So easy to understand, so easy to debug.

I agree we should get rid of fromFuture but to be complete we need to update our google daos too.

PR #273, opened 2019-01-16
create proxy groups for invites

dvoet:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-4056
The gist is that a proxy group is not created when an invite is created. The intention was to create it when the user comes to the site. But this then requires a bunch of special case logic:
* exclude invites from google group sync (which we neglected)
* trigger group sync when an invite registers
* be careful in the ordering of proxy group creation and group syncing

It just seems easier, (less code and less cognitive load) to always create a proxy group.

We also need to back populate all existing invitations.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -216,10 +216,10 @@ object UserService {
   def genWorkbenchUserId(currentMilli: Long): WorkbenchUserId =
     WorkbenchUserId(genRandom(currentMilli))
 
-  def validateEmailAddress(email: WorkbenchEmail) = {
+  def validateEmailAddress(email: WorkbenchEmail): IO[Unit] = {
     email.value match {
       case UserService.emailRegex() => IO.unit
-      case _ => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "invalid email address")))
+      case _ => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"invalid email address [${email.value}]")))
     }
   }
 }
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -198,4 +215,11 @@ object UserService {
 
   def genWorkbenchUserId(currentMilli: Long): WorkbenchUserId =
     WorkbenchUserId(genRandom(currentMilli))
+
+  def validateEmailAddress(email: WorkbenchEmail) = {
```
gpolumbo-broad: Tests for this?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -198,4 +215,11 @@ object UserService {
...
+    email.value match {
+      case UserService.emailRegex() => IO.unit
+      case _ => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "invalid email address")))
```
Qi77Qi: shall we put the actual email address in the message as well just for debugging purpose?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -198,4 +215,11 @@ object UserService {
 
   def genWorkbenchUserId(currentMilli: Long): WorkbenchUserId =
     WorkbenchUserId(genRandom(currentMilli))
+
+  def validateEmailAddress(email: WorkbenchEmail) = {
+    email.value match {
```
gpolumbo-broad, line 220: If the user entered some leading/trailing whitespace, can we just lop that off for them rather than throwing an error?

gpolumbo-broad (changes requested):

Qi77Qi (approved):

gpolumbo-broad (approved):

PR #275, opened 2019-02-01
Use KMS in Sam

marctalbott:
Ticket: [GAWB-4037](https://broadinstitute.atlassian.net/browse/GAWB-4037)
- This creates a KMS key and keyring and grants access to that key to the All_Users group

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/KmsConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/KmsConfig.scala
@@ -1,5 +1,8 @@
 package org.broadinstitute.dsde.workbench.sam.config
 
+import org.broadinstitute.dsde.workbench.google.{KeyId, KeyRingId, Location}
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
 /**
   * created by mtalbott 1/25/18
   *
@@ -10,8 +13,10 @@ package org.broadinstitute.dsde.workbench.sam.config
   */
 
 final case class GoogleKmsConfig(
-                                  project: String,
-                                  location: String,
-                                  keyRingId: String,
-                                  keyId: String
+                                  project: GoogleProject,
+                                  location: Location,
+                                  keyRingId: KeyRingId,
+                                  keyId: KeyId,
+                                  rotationPeriod: Int
+
                           )
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/KmsConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/KmsConfig.scala
@@ -1,0 +1,17 @@
+package org.broadinstitute.dsde.workbench.sam.config
+
+/**
+  * created by mtalbott 1/25/18
+  *
+  * @param project google project for the key
+  * @param location location for the key
+  * @param keyRingId name of the key ring that contains the key
+  * @param keyId name of the key
+  */
+
+final case class GoogleKmsConfig(
+                                  project: String,
+                                  location: String,
+                                  keyRingId: String,
+                                  keyId: String
+                          )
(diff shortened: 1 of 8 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -134,6 +128,25 @@ class GoogleExtensions(
...
+      _ <- googleKms.createKeyRing(GoogleProject(googleServicesConfig.googleKms.project),
+        Location(googleServicesConfig.googleKms.location),
+        KeyRingId(googleServicesConfig.googleKms.keyRingId)) handleErrorWith { case _: AlreadyExistsException => IO.unit }
+
+      _ <- googleKms.createKey(GoogleProject(googleServicesConfig.googleKms.project),
+        Location(googleServicesConfig.googleKms.location),
+        KeyRingId(googleServicesConfig.googleKms.keyRingId),
+        KeyId(googleServicesConfig.googleKms.keyId),
+        Option(Timestamp.newBuilder().setSeconds(System.currentTimeMillis() / 1000 + 180.days.toSeconds).build()),
+        Option(Duration.newBuilder().setSeconds(180.days.toSeconds).build())
```
dvoet: make 180 days configurable

dvoet (changes requested):
1 config change then 👍

src/test/resources/reference.conf
```diff
@@ -55,6 +55,13 @@ googleServices {
       workerCount = 1
     }
   }
+
+  kms {
+    project = "broad-dsde-dev"
+    location = "global"
+    keyRingId = "mtalbott"
+    keyId = "mtalbott"
```
Qi77Qi: shall we change `keyRingId` and `keyId` to something else that's not specific to you

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/KmsConfig.scala
```diff
@@ -0,0 +1,17 @@
...
+  * @param location location for the key
+  * @param keyRingId name of the key ring that contains the key
+  * @param keyId name of the key
+  */
+
+final case class GoogleKmsConfig(
+                                  project: String,
+                                  location: String,
+                                  keyRingId: String,
+                                  keyId: String
```
Qi77Qi: seems like there're custom types for `project`, `location`, `keyRingId` etc, why not use those? that way, you won't need to wrap them [here](https://github.com/broadinstitute/sam/pull/275/files#diff-ab230e3b899f0b14c89b226c3c761c0eR138). It's also more type safe

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/KmsConfig.scala
```diff
@@ -0,0 +1,22 @@
...
+  * @param keyRingId name of the key ring that contains the key
+  * @param keyId name of the key
+  */
+
+final case class GoogleKmsConfig(
+                                  project: GoogleProject,
+                                  location: Location,
+                                  keyRingId: KeyRingId,
+                                  keyId: KeyId,
+                                  rotationPeriod: Int
```
Qi77Qi: what is `Int`? if this is duration, we should probably use `FiniteDuration`

Qi77Qi (approved):
one minor comment

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/KmsConfig.scala
marctalbott: It's the number of days. I call `.days.toSeconds` on it. I suppose we could use a Duration, but the number of days as an int was what I thought of first

Qi77Qi: so lots config libraries support reading `FiniteDuration` directly. x days probably can be `x day` in config directly...see https://github.com/iheartradio/ficus#examples, there's an example for `minutes`, but I think it should definitely have `day` too

```
  // something such as "15 minutes" can be converted to a FiniteDuration
  val retryInterval: FiniteDuration = config.as[FiniteDuration]("retryInterval")
```

The benefit of this is we don't need to guess what unit this `Int` is anymore. You can also just do `toSeconds` at call site

Qi77Qi: did a bit more digging, I see https://github.com/iheartradio/ficus/blob/a6db5d65f3c3237a27aee592661678962a2aca09/src/test/scala/net/ceedubs/ficus/readers/DurationReadersSpec.scala#L34, I think you should be able to do `180 days` in config file and read it as a `FiniteDuration`

dvoet (approved):

PR #276, opened 2019-02-05
GAWB-4079: Add get_cost action on workflow-collection resource

ansingh7115:
Ticket: https://broadinstitute.atlassian.net/browse/GAWB-4079

---

**PR checklist**


src/main/resources/reference.conf
```diff
@@ -187,13 +190,13 @@ resourceTypes = {
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["delete", "add", "view", "abort", "update", "alter_policies", "read_policies"]
+        roleActions = ["delete", "add", "view", "abort", "get_cost", "update", "alter_policies", "read_policies"]
       }
       reader = {
         roleActions = ["view"]
```
dvoet: should reader should have get_cost too?

src/main/resources/reference.conf
```diff
@@ -174,6 +174,9 @@ resourceTypes = {
       abort = {
         description = "abort any workflow in this workflow-collection"
       }
+      get_cost = {
+        description = "get the cost of a any workflow in this workflow-collection"
```
dvoet: drop `a`

src/main/resources/reference.conf
ansingh7115: I'm not sure yet - I was thinking no and then we add it if people want it. My logic was - if someone doesn't own the workspace or have the ability to run anything in it - why do they care about the costs?

ansingh7115: is it complicated to remove an action from a role? If not, I don't mind readers having get_cost

dvoet (approved):

PR #283, opened 2019-03-13
AP-145 only constrained auth domain needs sync

Qi77Qi:
Ticket: https://broadworkbench.atlassian.net/browse/AP-145

Please check test change to see if the logic is what we want.

---

**PR checklist**


dvoet (changes requested):
👍 on the code. There is also an automation test that makes sure that public policies are synced correctly but it does not operate on a resource type that is constrainable (which is why this problem was missed). Can you add a test that works on workspaces?

dvoet:
jenkins retest - looks like docker error at end of sam build

dvoet:
jenkins retest - known gcloud issue

dvoet:
getting the right automated test is harder than we have time for. It requires a population step in fiab. We will get that done in a follow on PR. In the mean time I manually did the population step and ran the new test code and it passed so I am merging.

dvoet (approved):

PR #294, opened 2019-04-19
Sam GroupApiSpec test

aweng98:
Jira [QA-493](https://broadworkbench.atlassian.net/browse/QA-493)
depend on [workbench-lib PR 221](https://github.com/broadinstitute/workbench-libs/pull/221)

### DIFF
--- a/automation/project/Dependencies.scala
+++ b/automation/project/Dependencies.scala
@@ -11,7 +11,7 @@ object Dependencies {
 
   val workbenchGoogleV = "0.18-8328aae"
   val workbenchGoogle2V = "0.1-8328aae"
-  val workbenchServiceTestV = "0.16-ebc1ddc"
+  val workbenchServiceTestV = "0.16-6542421-SNAP"
 
   val workbenchModel: ModuleID = "org.broadinstitute.dsde.workbench" %% "workbench-model" % workbenchModelV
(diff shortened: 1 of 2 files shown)


aweng98:
jenkins retest

dvoet (commented):
why does this test do rawls stuff... sam tests should know nothing of rawls

ansingh7115:
tbh I think these tests belong in rawls.

aweng98:
@dvoet Because mimicking UI actions or/and performing verifications. For example, `Rawls.billing.removeUserFromBillingProject`, I don't know another way doing it without touching Rawls stuff.

aweng98:
@ansingh7115 If it's a Rawls api tests, I would still need Sam stuff in tests. I thought about it before, either in Sam or Rawls, tests kind needs little of both.

ansingh7115:
Sorry, accidentally deleted my comment. -_-
Original comment was these tests should be in rawls.
@aweng98  if it needs both, put it in rawls. Rawls has a dependency on sam, so it's appropriate to call sam from a rawls test. It's not appropriate to call rawls from a sam test because Sam does _not_ have a dependency on rawls.

aweng98:
@asingh7115 ah, I'll redo it in Rawls. thank you.

PR #295, opened 2019-05-05
log when GoogleGroupSyncMonitorActor is being terminated

Qi77Qi:
I tried searching `akka\:\/\/sam\/user\/` in kibana, and there were lots of actors existed when ldap exception happened.

Memory usage looks fine in new relic, so code is probably okay and we are terminating actors properly after starting new one. This PR is just to add some logging so that we know for sure terminating actors happened.

I do wonder if throwing exception and causing actor to die and spawning new ones is necessary though. Is it enough to log errors (which triggers sentry alert) in this case and GoogleGroupSyncMonitorActor will move on to process other messages, or there is a reason we do actually want GoogleGroupSyncMonitorActor to die first when error happens?

---

**PR checklist**


dvoet (approved):

dvoet:
Terminating the actor is cleanest

geoffjentry:
"Let it crash!" (TM) :)

Qi77Qi:
Going to merge this since it's just adding simple log message

PR #297, opened 2019-05-14
CA-257: Access policies should no longer be added to managed groups

ansingh7115:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -67,7 +67,7 @@ class GoogleExtensions(
   private val maxGroupEmailLength = 64
 
   private[google] def toProxyFromUser(userId: WorkbenchUserId): WorkbenchEmail =
-    /* Re-enable this code and remove the temporary code below after fixing rawls for GAWB-2933
+  /* Re-enable this code and remove the temporary code below after fixing rawls for GAWB-2933
     val username = user.email.value.split("@").head
     val emailSuffix = s"_${user.id.value}@${googleServicesConfig.appsDomain}"
     val maxUsernameLength = maxGroupEmailLength - emailSuffix.length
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -367,10 +367,17 @@ class ResourceService(
         throw t
     }
 
-  def addSubjectToPolicy(policyIdentity: FullyQualifiedPolicyId, subject: WorkbenchSubject): Future[Unit] =
-    directoryDAO.addGroupMember(policyIdentity, subject).unsafeToFuture().map(_ => ()) andThen {
-      case Success(_) => fireGroupUpdateNotification(policyIdentity)
+  def addSubjectToPolicy(policyIdentity: FullyQualifiedPolicyId, subject: WorkbenchSubject): Future[Unit] = {
+    subject match {
+      case subject: FullyQualifiedPolicyId if policyIdentity.resource.resourceTypeName.equals(ManagedGroupService.managedGroupTypeName) =>
+        throw new WorkbenchExceptionWithErrorReport(ErrorReport(s"Access policies cannot be added to managed groups."))
```
dvoet: this should use a 400 bad request error code (as it stands it will be a 500 error)

dvoet: also it is nicer if you use `Future.failed` instead of `throw`

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -171,18 +171,18 @@ class GoogleExtensions(
...
+
+      ancestorManagedGroups <- managedGroupNames traverse { id =>
         directoryDAO.listAncestorGroups(id)
       }
-      managedGroupIds = (ancestorGroups.flatten ++ groupIdentities).filterNot(visitedGroups.contains).collect {
+      managedGroupIds = (ancestorManagedGroups.flatten ++ groupIdentities).filterNot(visitedGroups.contains).collect {
         case FullyQualifiedPolicyId(FullyQualifiedResourceId(ManagedGroupService.managedGroupTypeName, id), ManagedGroupService.adminPolicyName | ManagedGroupService.memberPolicyName) => id
       }
-      _ <- managedGroupIds.traverse(id => onManagedGroupUpdate(id, visitedGroups ++ groupIdentities ++ ancestorGroups.flatten))
+      _ <- managedGroupIds.traverse(id => onManagedGroupUpdate(id, visitedGroups ++ groupIdentities ++ ancestorManagedGroups.flatten))
```
dvoet: I think the changes in this function do what we want but I would like to clean it up a little more. The main thing is that the recursion is unnecessary. But removing that means a bigger restructuring.

I feel like there are 2 steps, 1) figure out what messages to send and 2) send them. The first step has 2 branches: if it is a policy just create a single message for the policy but if it is a managed group, find any related policies via auth domains of the group or an ancestor, create a message for each policy and a message for the managed group. I think that code will end up being more straight forward.

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
ansingh7115: Ah - I didn't realize we were getting rid of the recursion - I'll change it!

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -171,18 +171,18 @@ class GoogleExtensions(
...
-      messagesForIdsWithSyncDates = idsAndSyncDates.collect {
-        case (gn: WorkbenchGroupName, Some(_)) => gn.toJson.compactPrint
-        case (rpn: FullyQualifiedPolicyId, Some(_)) => rpn.toJson.compactPrint
-      }
-      _ <- IO.fromFuture(IO(googlePubSubDAO.publishMessages(googleServicesConfig.groupSyncTopic, messagesForIdsWithSyncDates)))
-      ancestorGroups <- groupIdentities.toList.traverse { id =>
+      managedGroupNames = idsAndSyncDates.collect { case (groupName: WorkbenchGroupName, Some(_)) => groupName }
+      accessPolicyIds   = idsAndSyncDates.collect { case (accessPolicyId: FullyQualifiedPolicyId, Some(_)) => accessPolicyId }
+
+      _ <- IO.fromFuture(IO(googlePubSubDAO.publishMessages(googleServicesConfig.groupSyncTopic, managedGroupNames.map(_.toJson.compactPrint) ++ accessPolicyIds.map(_.toJson.compactPrint))))
```
gpolumbo-broad: This is doing an awful lot in one line, could it be broken down?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -367,10 +367,17 @@ class ResourceService(
...
+      case _ =>    {
+        directoryDAO.addGroupMember(policyIdentity, subject).unsafeToFuture().map(_ => ()) andThen {
```
gpolumbo-broad, line 375: Is using `andThen` safe here?  I think I would rather this thing only return when and if the directoryDAO update completed successfully and the `fireGroupUpdateNotification` completed successfully.

dvoet: let's leave the `andThen` eradication for a later PR... I want to make sure this has good enough performance first

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -162,37 +162,48 @@ class GoogleExtensions(
...
-      _ <- IO.fromFuture(IO(googlePubSubDAO.publishMessages(googleServicesConfig.groupSyncTopic, messagesForIdsWithSyncDates)))
-      ancestorGroups <- groupIdentities.toList.traverse { id =>
-        directoryDAO.listAncestorGroups(id)
+
+      // separate out the access policies and managed groups
+      managedGroupNames = previouslySyncedIds.collect { case groupName: WorkbenchGroupName => groupName }
+      accessPolicyIds = previouslySyncedIds.collect { case accessPolicyId: FullyQualifiedPolicyId => accessPolicyId }
+
+      // get all the ancestor groups of each managed group
+      ancestorGroupsOfManagerGroups: List[Set[WorkbenchGroupIdentity]] <- managedGroupNames traverse { id =>
```
dvoet: ancestorGroupsOfManagedGroups not ancestorGroupsOfManagerGroups (managed not manager)

ansingh7115:
Got to fix the tests

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -162,37 +162,56 @@ class GoogleExtensions(
...
       }
-      managedGroupIds = (ancestorGroups.flatten ++ groupIdentities).filterNot(visitedGroups.contains).collect {
+
+      // publish all the messages
+      _ <- IO.fromFuture(IO(publishMessages(messages.flatten)))
+
+    } yield ()
+  }.unsafeToFuture()
+
+  private def getGroupPublishMessages(groupName: WorkbenchGroupName): IO[List[String]] = {
```
gpolumbo-broad: Nit - maybe rename this to `makeGroupPublishMessages` or `makeMessagesForGroupSync` or `constructMessagesForGroupSync` or something like that.  Until I read the method, I thought you were getting (retrieving) messages from the pub/sub queue or something.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -162,37 +162,56 @@ class GoogleExtensions(
...
-        case (rpn: FullyQualifiedPolicyId, Some(_)) => rpn.toJson.compactPrint
+      // only sync groups that have been synchronized in the past
+      previouslySyncedIds <- groupIdentities.toList.traverseFilter { id =>
+        directoryDAO.getSynchronizedDate(id).map(dateOption => dateOption.map(_ => id))
       }
-      _ <- IO.fromFuture(IO(googlePubSubDAO.publishMessages(googleServicesConfig.groupSyncTopic, messagesForIdsWithSyncDates)))
-      ancestorGroups <- groupIdentities.toList.traverse { id =>
-        directoryDAO.listAncestorGroups(id)
+
+      // get all the messages for the previously synced groups
```
gpolumbo-broad: nit - "Make/collect messages for all groups that need to be synced"

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -162,37 +162,56 @@ class GoogleExtensions(
   override def publishGroup(id: WorkbenchGroupName): Future[Unit] =
     googlePubSubDAO.publishMessages(googleServicesConfig.groupSyncTopic, Seq(id.toJson.compactPrint))
 
-  override def onGroupUpdate(groupIdentities: Seq[WorkbenchGroupIdentity]): Future[Unit] =
-    onGroupUpdateRecursive(groupIdentities, Seq.empty).unsafeToFuture()
-
-  private def onGroupUpdateRecursive(groupIdentities: Seq[WorkbenchGroupIdentity], visitedGroups: Seq[WorkbenchGroupIdentity]): IO[Unit] =
+  override def onGroupUpdate(groupIdentities: Seq[WorkbenchGroupIdentity]): Future[Unit] = {
```
gpolumbo-broad, line 184: Could add a comment to document the how/why we update what we do when we update a group?  You and Doug spent a fair amount of time talking about why we do the group sync and how/why we need to grab ancestors of Group Y but not of Group X, etc.  Furthermore, all of this update stuff, IIRC, is because Google needs to store the intersection groups.  Sam doesn't care about the intersection groups (I think?) because that's basically what LDAP does for us, is calculate intersections of groups.  You and Doug understand how this logic works right now, and understand WHY we need to do it this way.  In 6 months, if someone other than you or Doug needs to modify this, will they understand why we're doing it this way so as not to break it?

gpolumbo-broad (changes requested):
I left just a couple of nits, but I think the documentation one is the one I'm most concerned about.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
ansingh7115: good point!

gpolumbo-broad: Awesome!

gpolumbo-broad (approved):
Good docs are good!

PR #310, opened 2019-07-02
implementation of shadow runner

dvoet:
Ticket: <Link to Jira ticket>

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowCaster.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  protected def reportResult[T](functionName: String, realResult: T, realTime: Long, shadowResult: T, shadowTime: Long) = {
+    val eq = realResult != shadowResult
+    val timeDiff = shadowTime - realTime
+
+    logger.debug(s"$functionName, equal: $eq, time increase $timeDiff")
+  }
+
+  private def measure[A](fa: IO[A]): IO[(A, Long)] = {
+    val clock = Clock.create[IO]
```
Qi77Qi: you don't want to need `Clock` every time you call measure. There's a global one exist in `IOApp` that you can just require implicitly (either as function implicit parameter, or treat it the same way you did with ec, something like `implicit val timer: Timer[IO]`)

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowCaster.scala
```diff
@@ -0,0 +1,39 @@
...
+
+    val realFuture = measure(real).unsafeToFuture()
+    val shadowFuture = measure(shadow).unsafeToFuture()
+
+    Future.reduceLeft[(T, Long), (T, Long)](List(realFuture, shadowFuture)) { case ((realResult, realTime), (shadowResult, shadowTime)) =>
+      reportResult(functionName, realResult, realTime, shadowResult, shadowTime)
+      (realResult, realTime)
+    }
+
+    IO.fromFuture(IO(realFuture)).map(_._1)
```
Qi77Qi: Some syntactic improvement, you can do
```
    (realFuture, shadowFuture).mapN {case ((realResult, realTime), (shadowResult, shadowTime)) =>
      reportResult(functionName, realResult, realTime, shadowResult, shadowTime)
      (realResult, realTime)
    }
```

I realize yesterday observe won't do exactly what you want.
The following snippet is close to what you have, but not exactly(in this code, reportResult returns IO[Unit]). The difference is real and shadow aren't run in parallel. Shadow runs after real asynchronously, and report stats.

```
    for {
      start <- timer.clock.realTime(MILLISECONDS)
      r <- real
      realEnd <- timer.clock.realTime(MILLISECONDS)
      _ <- shadow.runAsync {
        cb =>
          cb match {
            case Left(error) => IO(logger.error("shadow failed"))
            case Right(t) =>
              for {
                shadowEnd <- timer.clock.realTime(MILLISECONDS)
                _ <- reportResult(functionName, r, realEnd - start, t, shadowEnd - realEnd) //shadow runs after real
              } yield ()
          }
      }.toIO
    } yield r
```
I'll let you know if I think of something better

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowCaster.scala
```diff
@@ -0,0 +1,39 @@
+package org.broadinstitute.dsde.workbench.sam.util
+import cats.effect._
+import com.typesafe.scalalogging.LazyLogging
+
+import scala.concurrent.duration.MILLISECONDS
+import scala.concurrent.{ExecutionContext, Future}
+
+trait ShadowCaster extends LazyLogging {
```
helgridly: I find the term "caster" to be really distracting, like it's a D&D character or something

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowCaster.scala
```diff
@@ -0,0 +1,39 @@
...
+  implicit val executionContext: ExecutionContext
+
+  protected def runWithShadow[T](functionName: String, real: IO[T], shadow: IO[T]): IO[T] = {
+
+    val realFuture = measure(real).unsafeToFuture()
+    val shadowFuture = measure(shadow).unsafeToFuture()
+
+    Future.reduceLeft[(T, Long), (T, Long)](List(realFuture, shadowFuture)) { case ((realResult, realTime), (shadowResult, shadowTime)) =>
```
helgridly: i find this hard to read, my brain goes "real shadow result real real time shadow time result". how about a `TimedResult(result: T, time: Long)` case class to pass around fewer parameters

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowCaster.scala
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    IO.fromFuture(IO(realFuture)).map(_._1)
+  }
+
+  protected def reportResult[T](functionName: String, realResult: T, realTime: Long, shadowResult: T, shadowTime: Long) = {
+    val eq = realResult != shadowResult
+    val timeDiff = shadowTime - realTime
+
+    logger.debug(s"$functionName, equal: $eq, time increase $timeDiff")
```
helgridly: is it useful to say _how_ they're not equal here, for easier debugging? scalatest writes fairly good output for `shouldEqual`, maybe you can steal it

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowCaster.scala
gpolumbo-broad: I agree that extracting the tuples would make this easier to read

src/test/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowCasterSpec.scala
```diff
@@ -0,0 +1,45 @@
...
+      val real = IO.pure(10)
+      val shadow = IO.timer(executionContext).sleep(1 second).map(_ => 15)
+
+      val resultPromise: Promise[(String, Any, Long, Any, Long)] = Promise()
+
+      def runTest(): IO[Int] = {
+        runWithShadow("runTest", real, shadow)
+      }
+
+      override protected def reportResult[T](functionName: String, realResult: T, realTime: Long, shadowResult: T, shadowTime: Long): Unit = {
```
gpolumbo-broad: I think the report is critical functionality for the ShadowCaster.  I suggest not overriding this and testing it, possibly with more complex objects than `Int`.

dvoet: I need to override it in this test to test the other functionality... when we write a real report function we can test that specifically

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowCaster.scala
dvoet: we will spend more time figuring out how to report

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowCaster.scala
dvoet: interesting... I had been fixating on running in parallel but I guess we don't need to

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowRunner.scala
```diff
@@ -0,0 +1,44 @@
...
+trait ShadowRunner {
+  implicit val executionContext: ExecutionContext
+  val clock: Clock[IO]
+
+  protected def runWithShadow[T](functionName: String, real: IO[T], shadow: IO[T]): IO[T] = {
+    for {
+      realTimedResult <- measure(real)
+      _ <- measure(shadow).runAsync {
+        case Left(regrets) =>
+          reportShadowFailure(functionName, regrets)
```
helgridly: don't you want to cross-ref this with the real, in case that fails too

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowRunner.scala
```diff
@@ -0,0 +1,44 @@
...
+        case Right(shadowTimedResult) =>
+          reportResult(functionName, realTimedResult, shadowTimedResult)
+          IO.unit
+      }.toIO
+    } yield {
+      realTimedResult.result
+    }
+  }
+
+  protected def reportShadowFailure(functionName: String, regrets: Throwable): Unit = {
```
Qi77Qi: I would make these `IO[Unit]` just so that you don't need to worry about when exactly this eager function will be evaluated.

Qi77Qi (approved):
small comment...overall looks good to me

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/ShadowRunner.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  private def measure[A](fa: IO[A]): IO[TimedResult[A]] = {
+    for {
+      start  <- clock.monotonic(MILLISECONDS)
+      result <- fa
+      finish <- clock.monotonic(MILLISECONDS)
+    } yield TimedResult(result, finish - start)
+  }
+}
+
```
Qi77Qi, line 79: I would make `time` a FiniteDuration

marctalbott (approved):

PR #311, opened 2019-07-03
first effort at setting up scalike tables

marctalbott:
Ticket: [CA-327](https://broadworkbench.atlassian.net/browse/CA-327)
This PR includes the case classes we should need to model the new Postgres schema as well as the Scalike binders to turn JDBC things into our case classes and vice versa.

Much of this is inspired by Anu's work with Scalike in Hamm. It is currently entirely untested because I wasn't sure how to test these tables without any queries to use them, so if anyone has any ideas that'd be much appreciated. Input on naming would also be super useful. I tried to keep a relatively consistent naming scheme across the different tables, but let me know if there's a better/shorter/clearer way to name all these case classes.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/db/WorkbenchParameterBinderFactory.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/db/WorkbenchParameterBinderFactory.scala
@@ -1,0 +1,15 @@
+package org.broadinstitute.dsde.workbench.sam.db
+
+import org.broadinstitute.dsde.workbench.sam.model.ValueObject
+import scalikejdbc.ParameterBinderFactory
+
+// These aren't used yet, but we will want these for the queries in the DAOs
+object WorkbenchParameterBinderFactory {
+  implicit val databaseIdPbf: ParameterBinderFactory[DatabaseId] = ParameterBinderFactory[DatabaseId] {
+    value => (stmt, idx) => stmt.setLong(idx, value.value)
+  }
+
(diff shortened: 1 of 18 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/db/WorkbenchParameterBinderFactory.scala
```diff
@@ -0,0 +1,15 @@
+package org.broadinstitute.dsde.workbench.sam.db
+
+import org.broadinstitute.dsde.workbench.sam.model.ValueObject
+import scalikejdbc.ParameterBinderFactory
+
+// These aren't used yet, but we will want these for the queries in the DAOs
```
dvoet: remove this comment, generally add comments like this in the PR so we don't accidentally leave it in the code

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/WorkbenchParameterBinderFactory.scala
```diff
@@ -0,0 +1,14 @@
+package org.broadinstitute.dsde.workbench.sam.db
+
+import org.broadinstitute.dsde.workbench.sam.model.ValueObject
+import scalikejdbc.ParameterBinderFactory
+
+object WorkbenchParameterBinderFactory {
```
marctalbott, line 6: These aren't used yet, but they will be used in queries by the DAOs

marctalbott:
Closing this in favor of https://github.com/broadinstitute/sam/pull/314 which has better versions of what's done in this PR

PR #312, opened 2019-07-12
implement NewRelicShadowResultReporter

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/CA-328
detects real vs shadow differences, logs and reports to new relic

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapDirectoryDAO.scala
```diff
@@ -8,11 +8,12 @@ import cats.implicits._
 import com.unboundid.ldap.sdk._
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.model.google._
+import org.broadinstitute.dsde.workbench.newrelic.NewRelicMetrics
```
dvoet, line 11: changes in this file are switching to `NewRelicMetrics` in workbench-libs

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
```diff
@@ -0,0 +1,144 @@
...
+        MatchResult(false, Seq(s"real contained status code [${realThrowable.errorReport.statusCode}], shadow threw [${shadowThrowable.getClass.getName}] message [${shadowThrowable.getMessage}]"))
+
+      case (Left(_), Left(_)) =>
+        // both just plain old failures, that they both failed is enough
+        MatchResult(true, Seq.empty)
+
+      case (Right(realTraversable: Traversable[_]), Right(shadowTraversable: Traversable[_])) =>
+        // both traversables, we don't care about order, we care that each item in real matches an item in shadow, that
+        // there are the same number of occurrences, and both real and shadow have the same number of distinct items
+        val real = realTraversable.groupBy(x => x).map { case (key, values) => (key, values.size) }
```
helgridly: comment this: map of distinct element -> count of occurrences

helgridly: comment this: map of distinct element -> count of occurrences

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
```diff
@@ -0,0 +1,144 @@
...
+        val shadow = shadowTraversable.groupBy(x => x).map { case (key, values) => (key, values.size) }
+        val sizeMatch = createMatchResult(real.size, shadow.size, s"unequal distinct item count, [${real.mkString(",")}] vs [${shadow.mkString(",")}]")
+        val itemsMatch = real.keySet.map { realValue =>
+          shadow.keySet.find(shadowValue => resultsMatch(Right(realValue), Right(shadowValue)).matches) match {
```
helgridly: why recurse just to check the equality of the two occurrence counts? can't you just call `createMatchResult`

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
```diff
@@ -0,0 +1,144 @@
...
+      case (Left(_), Left(_)) =>
+        // both just plain old failures, that they both failed is enough
+        MatchResult(true, Seq.empty)
+
+      case (Right(realTraversable: Traversable[_]), Right(shadowTraversable: Traversable[_])) =>
+        // both traversables, we don't care about order, we care that each item in real matches an item in shadow, that
+        // there are the same number of occurrences, and both real and shadow have the same number of distinct items
+        val real = realTraversable.groupBy(x => x).map { case (key, values) => (key, values.size) }
+        val shadow = shadowTraversable.groupBy(x => x).map { case (key, values) => (key, values.size) }
+        val sizeMatch = createMatchResult(real.size, shadow.size, s"unequal distinct item count, [${real.mkString(",")}] vs [${shadow.mkString(",")}]")
```
helgridly: i think that comparing the two keysets instead would make it clearer what you're doing

src/test/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporterSpec.scala
```diff
@@ -0,0 +1,226 @@
...
+    when(newRelicMetrics.gauge(any[String], any[Float])).thenReturn(IO.unit)
+    when(newRelicMetrics.incrementCounterIO(any[String], any[Int])).thenReturn(IO.unit)
+    new NewRelicShadowResultReporter("test", newRelicMetrics)
+  }
+
+  // match cases
+  "NewRelicShadowResultReporter" should "match complex case class structure" in {
+    val reporter = createResultReporter
+    val probe = TestCaseClass("asdfasdf", Seq(TestInnerCaseClass(MyValueObject("qqq"), 88), TestInnerCaseClass(MyValueObject("ppp"), 99)))
+    val result = reporter.resultsMatch(Right(probe), Right(probe))
```
helgridly: do one of these as `probe.copy()` to make sure you're not just testing memory location (and below also)

helgridly (changes requested):
mostly nits

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
dvoet: that's what the comment above is supposed to be, not clear enough?

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
dvoet: unfortunately comparing keysets is not quite right because the keys are case classes that may have embedded collections and order needs to be ignored.

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
dvoet: plain equality does not take into account ignoring of order

dvoet: I tried to make this clear in the scaladoc, let me know if it needs more

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
helgridly: i spent a while getting confused, hence asking for the comment. i think it'd be clearer if the comparisons were commented inline rather than in a blob at the top

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
helgridly: wait, so you're calling the items inside the keyset -- keys -- "value"?

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
helgridly: :( i think i'd find this easier to read if you broke out your definition of equality into a separate function rather than piecing it together inline

helgridly: i realise that seems to conflict with my comment 3 lines above but i will leave it as a statement of how confused i am reading this code

helgridly: doesn't your groupby above also need to use the same notion of equality that you're talking about here?

helgridly: oh -- if you don't care about ordering, then sorting the two traversables and comparing them pairwise should be all you need to do

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
```diff
@@ -0,0 +1,144 @@
...
+      case (Right(realValue), Right(shadowValue)) => createMatchResult(realValue, shadowValue, "values unequal")
+
+      case (_, _) =>
+        // either real or shadow failed but not both
+        MatchResult(false, Seq(s"real and shadow are not comparable: real [$realResult], shadow [$shadowResult]"))
+    }
+  }
+
+  // checkName is passed by-name to avoid unnecessary toString and string concatenation when items match
+  private def createMatchResult[T](real: T, shadow: T, checkName: => String): MatchResult = {
```
gpolumbo-broad, line 160: nit - I don't like the parameter name `checkName`.  I would prefer `errorMessage` or `errorMessagePrefix` or something like that.  I didn't really know what `checkName` meant until I went and saw places where you actually call this method.

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
dvoet: well, yes... they are the actual values...

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporter.scala
dvoet: can't sort because there I don't know the ordering

src/test/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporterSpec.scala
```diff
@@ -0,0 +1,227 @@
...
+    when(newRelicMetrics.gauge(any[String], any[Float])).thenReturn(IO.unit)
+    when(newRelicMetrics.incrementCounterIO(any[String], any[Int])).thenReturn(IO.unit)
+    new NewRelicShadowResultReporter("test", newRelicMetrics)
+  }
+
+  // match cases
+  "NewRelicShadowResultReporter" should "match complex case class structure" in {
+    val reporter = createResultReporter
+    val probe = TestCaseClass("asdfasdf", Seq(TestInnerCaseClass(MyValueObject("qqq"), 88), TestInnerCaseClass(MyValueObject("ppp"), 99)))
+    // copy probe to make sure not just testing reference equality
```
gpolumbo-broad, line 26: This comment makes a good point, do we care about this in other tests too then?

src/test/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporterSpec.scala
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  it should "detect mismatch in case class in complex case class structure" in {
+    val reporter = createResultReporter
+    val real = TestCaseClass("asdfasdf", Seq(TestInnerCaseClass(MyValueObject("qqq"), 88), TestInnerCaseClass(MyValueObject("ppp"), 99)), Some(TestInnerCaseClass(MyValueObject("oooo"), 77)))
+    val shadow = TestCaseClass("asdfasdf", Seq(TestInnerCaseClass(MyValueObject("qqq"), 88), TestInnerCaseClass(MyValueObject("ppp"), 99)), Some(TestInnerCaseClass(MyValueObject("ooo"), 77)))
+    val result = reporter.resultsMatch(Right(real), Right(shadow))
+    withClue(result.mismatchReasons) {
+      result.matches should be (false)
+      result.mismatchReasons should contain theSameElementsAs Seq("values unequal: real [oooo], shadow [ooo]")
```
gpolumbo-broad, line 110: I think it's fine for most of the "mismatch" tests to just assert `should contain`.  However, I think the format and content of the FULL mismatch message is also important.  I would like to see one test of a complex mismatch where the assertion is a `String should equal String`.

src/test/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporterSpec.scala
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  it should "detect mismatch in primitive" in {
+    val reporter = createResultReporter
+    val real = 1
+    val shadow = 2
+    val result = reporter.resultsMatch(Right(real), Right(shadow))
+    withClue(result.mismatchReasons) {
+      result.matches should be (false)
+      result.mismatchReasons should contain theSameElementsAs Seq("values unequal: real [1], shadow [2]")
```
gpolumbo-broad, line 121: I know the intent of the square brackets as a delimiter here.  However, in debugging this issue if we see it IRL, my first instinct is that the comparison was between two `Array[Int]`.  Could we reformat the messaging slightly and do away with the delimiters entirely?  Otherwise, I'm not sure what else to use that might read easier.

helgridly (approved):

gpolumbo-broad (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporterSpec.scala
dvoet: in other tests there isn't a copy function... I think doing it once is good enough

src/test/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporterSpec.scala
dvoet: the square brackets are not really delimiters, they are more to make it clear when there are empty strings... we need something

gpolumbo-broad (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/util/NewRelicShadowResultReporterSpec.scala
dvoet: I added a new test and visually verified that it logs what I want. The full message is important but not really testable... it is mostly important that a human can understand it and it has all pertinent information. If you would like to review what is logged that would be great.

PR #314, opened 2019-07-16
scalike tables, first two queries, start of tests

marctalbott:
Ticket: [CA-308](https://broadworkbench.atlassian.net/browse/CA-308) and [CA-327](https://broadworkbench.atlassian.net/browse/CA-327)
- I made this branch from https://github.com/broadinstitute/sam/pull/311. After writing some queries and the tests, we've made some adjustments to the table classes, so I'm going to close that PR and this PR will close out CA-327. This is only part of the work required for CA-308 as there are still many more queries to write. However, I want to get this merged into the feature branch as this PR is feeling kind of large already.

---

**PR checklist**


dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/ResourceActionPatternTable.scala
```diff
@@ -0,0 +1,26 @@
...
+  import SamTypeBinders._
+  def apply(e: ResultName[ResourceActionPatternRecord])(rs: WrappedResultSet): ResourceActionPatternRecord = ResourceActionPatternRecord(
+    rs.get(e.id),
+    rs.get(e.resourceTypeId),
+    rs.get(e.actionPattern)
+  )
+}
+
+object ResourceActionPatternTableBinders {
+
```
ansingh7115: Why is this empty?

marctalbott: We used to have the type binders for each table in the same file as the table itself and I forgot to delete this one

gpolumbo-broad:
General questions that are probably more directed for @dvoet:

1. All the table names are prefixed with `SAM_`.  I advocate that we don't do this.  I don't foresee us using the same schema for another app or set of tables. 
2. If we have a column that is a foreign key, I advocate that we name that column `{FOREIGN_TABLE_NAME}_id`.  For example, in `SAM_RESOURCE_POLICY_ACTION`, the column `action_id` is a foreign key to `SAM_RESOURCE_ACTION`.  I suggest we rename that column to be `resource_action_id`.  This makes navigating the schema easier without needing to describe tables or look at a schema diagram.

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/PolicyTable.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import org.broadinstitute.dsde.workbench.sam.db.{DatabaseKey, SamTypeBinders}
+import org.broadinstitute.dsde.workbench.sam.model.AccessPolicyName
+import scalikejdbc._
+
+final case class PolicyPK(value: Long) extends DatabaseKey
+final case class PolicyRecord(id: PolicyPK,
+                              resourceId: ResourcePK,
+                              groupId: GroupPK,
+                              name: AccessPolicyName)
```
gpolumbo-broad, line 11: Great!  Glad to see there is a name here.  I was looking at the schema diagram and spent several minutes debating in my head whether Policies should have their own name or whether they inherit their name from the Group.  I added the row to the table in the Lucidchart schema diagram

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/ResourceActionPatternTable.scala
```diff
@@ -0,0 +1,22 @@
...
+import scalikejdbc._
+
+final case class ResourceActionPatternPK(value: Long) extends DatabaseKey
+final case class ResourceActionPatternName(value: String) extends ValueObject
+final case class ResourceActionPatternRecord(id: ResourceActionPatternPK,
+                                             resourceTypeId: ResourceTypePK,
+                                             actionPattern: ResourceActionPatternName)
+
+object ResourceActionPatternTable extends SQLSyntaxSupport[ResourceActionPatternRecord] {
+  override def tableName: String = "SAM_ACTION_PATTERN"
```
gpolumbo-broad: Name mismatch.  The schema calls this `SAM_RESOURCE_ACTION_PATTERN`.  I suggest we call it `SAM_RESOURCE_TYPE_ACTION_PATTERN` to make it clear by the name what thing these action patterns are for, `RESOURCE_TYPE`

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -0,0 +1,216 @@
...
+          .list().apply()
+      }
+    } yield {
+      if (results.isEmpty) {
+        None
+      } else {
+        val email = results.head._1
+        val members: Set[WorkbenchSubject] = results.collect {
+          case (_, Some (userId), None) => userId
+          case (_, None, Some (subGroupName)) => subGroupName
```
gpolumbo-broad: nit - space after `Some`

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/ResourceActionPatternTable.scala
marctalbott: Ah, the table name in `initial_schema.xml` was `SAM_ACTION_PATTERN`, so I'll change both to fit the Lucidchart schema

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/PetServiceAccountTable.scala
```diff
@@ -0,0 +1,23 @@
+package org.broadinstitute.dsde.workbench.sam.db.tables
+
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.model.{GoogleSubjectId, WorkbenchEmail, WorkbenchUserId}
+import org.broadinstitute.dsde.workbench.sam.db.SamTypeBinders
+import scalikejdbc._
+
+final case class PetServiceAccountRecord(userId: WorkbenchUserId,
```
ansingh7115, line 8: userId a primary key according to the changeset definition. Should it be of type `UserIdPK`

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/ResourceRoleTable.scala
```diff
@@ -0,0 +1,21 @@
+package org.broadinstitute.dsde.workbench.sam.db.tables
+
+import org.broadinstitute.dsde.workbench.sam.db.{DatabaseKey, SamTypeBinders}
+import org.broadinstitute.dsde.workbench.sam.model.ResourceRoleName
+import scalikejdbc._
+
+final case class ResourceRolePK(value: Long) extends DatabaseKey
+final case class ResourceRoleRecord(id: ResourceRolePK,
+                                    resourceTypeId: ResourceTypePK,
+                                    role: ResourceRoleName)
```
ansingh7115, line 10: For unique constraints, we did something like this in Hamm: https://github.com/DataBiosphere/hamm/blob/master/core/src/main/scala/org/broadinstitute/dsp/workbench/hamm/db/JobTable.scala#L91

Not saying we need to do that here - but it _is_ difficult to know if there are any unique keys on a table just from looking at the class definitions the way it is. But there might be better ways of doing it and we could add that later.

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/PetServiceAccountTable.scala
marctalbott: It is a primary key here and in the user table. However, we are using the pre-existing `WorkbenchUserId` case class since it can be passed around easier rather than using a new case class like `UserIdPK`. The `userId` is also a bit different in that it is a string and not an auto-incrementing sequence like the other `DatabaseKey` extending classes that we are using in other tables.

ansingh7115 (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/ResourceRoleTable.scala
marctalbott: After some discussion, we're going to hold off on implementing unique keys like Hamm did for now. There is agreement (at least between me and Anu) that the record case classes don't do a good job of representing some of the underlying constraints that are described in the Liquibase changeset. The unique key case classes that Hamm uses would improve that representation, but they would not have any other purpose at this point in time which is why we won't be implementing them right now. That said, we might want to consider ways to improve how well the case classes represent the underlying tables.

PR #315, opened 2019-07-19
implemented isGroupMember and tests

dvoet:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
@@ -8,7 +8,7 @@ import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.model.google.ServiceAccountSubjectId
 import org.broadinstitute.dsde.workbench.sam.db._
 import org.broadinstitute.dsde.workbench.sam.db.tables._
-import org.broadinstitute.dsde.workbench.sam.model.{BasicWorkbenchGroup, FullyQualifiedPolicyId}
+import org.broadinstitute.dsde.workbench.sam.model.{BasicWorkbenchGroup, FullyQualifiedPolicyId, FullyQualifiedResourceId}
 import org.broadinstitute.dsde.workbench.sam.util.DatabaseSupport
 import org.broadinstitute.dsde.workbench.sam.errorReportSource
 import scalikejdbc._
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -154,7 +154,67 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
+      // in the recursive query below, UNION, as opposed to UNION ALL, should break out of cycles because it removes duplicates
+      val query = sql"""WITH RECURSIVE sub_group(parent_group_id, member_group_id, member_user_id) AS (
+        $topGroupQuery
+        UNION
+        SELECT ${gm.groupId}, ${gm.memberGroupId}, ${gm.memberUserId}
+        FROM sub_group sg, ${GroupMemberTable.as(gm)}
+        WHERE ${gm.groupId} = sg.member_group_id
+      )
+      SELECT count(1)
+        FROM sub_group sg WHERE $memberClause"""
```
gpolumbo-broad, line 313: There are a couple of places where you alias the table, `sub_group sg`, and a bunch of places where you reference columns on `sg`.  I'm wondering if you can externalize these strings.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -154,7 +154,67 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
   override def removeGroupMember(groupId: WorkbenchGroupIdentity, removeMember: WorkbenchSubject): IO[Boolean] = ???
 
-  override def isGroupMember(groupId: WorkbenchGroupIdentity, member: WorkbenchSubject): IO[Boolean] = ???
+  override def isGroupMember(groupId: WorkbenchGroupIdentity, member: WorkbenchSubject): IO[Boolean] = {
+    val gm = GroupMemberTable.syntax("gm")
+
+    val memberClause: SQLSyntax = member match {
+      case WorkbenchGroupName(groupName) =>
+        val g = GroupTable.syntax("g")
+        sqls"sg.member_group_id = (select ${g.id} from ${GroupTable.as(g)} where ${g.name} = $groupName)"
```
gpolumbo-broad: We just merged @marctalbott's branch.  We have a query for this that we can/should reuse with minor tweaking:

https://github.com/broadinstitute/sam/blob/4e05af7571fd061b9d93dcb2f90131ecb56f5fbf/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala#L248

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -154,7 +154,67 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
+        val rt = ResourceTypeTable.syntax("rt")
+        val r = ResourceTable.syntax("r")
+        val p = PolicyTable.syntax("p")
+        sqls"""sg.member_group_id = (select ${p.groupId}
+              from ${ResourceTypeTable.as(rt)}
+              join ${ResourceTable.as(r)} on ${rt.id} = ${r.resourceTypeId}
+              join ${PolicyTable.as(p)} on ${r.id} = ${p.resourceId}
+              where ${rt.resourceTypeName} = $resourceTypeName)
+              and ${r.name} = $resourceId
+              and ${p.name} = $accessPolicyName)"""
```
gpolumbo-broad: Sub query helper method: 

https://github.com/broadinstitute/sam/blob/4e05af7571fd061b9d93dcb2f90131ecb56f5fbf/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala#L254

Noticed in that implementation that we didn't do the join on Resource Type too that you did here, so that will need to be added.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -154,7 +154,67 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
+
+      case WorkbenchUserId(userId) => sqls"sg.member_user_id = $userId"
```
gpolumbo-broad: Consider extracting `sqls"sg.member_user_id = "`?  I like it the way it is for readability, but for maintainability it might be better to extract it, especially if we end up just calling the sub query helper methods.  Either way is fine with me, but my preference would be to extract it, especially since it is a table alias/column that we are defining ourselves without scalike's help.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -154,7 +154,67 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
+        sqls"""select ${p.groupId}, ${gm.memberGroupId}, ${gm.memberUserId}
+              from ${ResourceTypeTable.as(rt)}
+              join ${ResourceTable.as(r)} on ${rt.id} = ${r.resourceTypeId}
+              join ${PolicyTable.as(p)} on ${r.id} = ${p.resourceId}
+              join ${GroupMemberTable.as(gm)} on ${p.groupId} = ${gm.groupId}
+              where ${rt.resourceTypeName} = $resourceTypeName)
+              and ${r.name} = $resourceId
+              and ${p.name} = $accessPolicyName"""
+
+    }
```
gpolumbo-broad: Similar to:

https://github.com/broadinstitute/sam/blob/4e05af7571fd061b9d93dcb2f90131ecb56f5fbf/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala#L240

Could we write a method like, `def groupMembersQuery(groupId: WorkbenchGroupIdentity): SQLSyntax = ???`

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -154,7 +154,67 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
+
+    runInTransaction { implicit session =>
+      // https://www.postgresql.org/docs/9.6/queries-with.html
+      // in the recursive query below, UNION, as opposed to UNION ALL, should break out of cycles because it removes duplicates
+      val query = sql"""WITH RECURSIVE sub_group(parent_group_id, member_group_id, member_user_id) AS (
```
gpolumbo-broad: Might want to extract all of this up to constructing `query` into its own method, something like `traceGroupMembership`.  I still want to read up on `WITH RECURSIVE`, but if I'm reading this right, we'll end up with a row of data that represents the path of group memberships between some parent group and the member.  Might be useful to be able to look at that in some cases?  

Then your `isMemberOf` method is really just the `sql("(${traceGroupMembership(groupId, member)}) SELECT count(1)...")`

Or ignore me....I just love extracting methods!

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -154,7 +154,67 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
     */
   override def removeGroupMember(groupId: WorkbenchGroupIdentity, removeMember: WorkbenchSubject): IO[Boolean] = ???
 
-  override def isGroupMember(groupId: WorkbenchGroupIdentity, member: WorkbenchSubject): IO[Boolean] = ???
+  override def isGroupMember(groupId: WorkbenchGroupIdentity, member: WorkbenchSubject): IO[Boolean] = {
+    val gm = GroupMemberTable.syntax("gm")
+
+    val memberClause: SQLSyntax = member match {
```
gpolumbo-broad, line 291: I think you can refactor this to use sub-query helper method we just merged:  

https://github.com/broadinstitute/sam/blob/4e05af7571fd061b9d93dcb2f90131ecb56f5fbf/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala#L240

Or I made comments to replace the individual case statement blocks with calls to the specific helper methods, but probably works better just to reuse `workbenchGroupIdentityToGroupPK`

gpolumbo-broad (changes requested):
A lot of my comments are based on latest code that @marctalbott just merged that he and I have been pairing on.  I think it can help simplify your stuff here.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
dvoet: these are strictly internal to the query, declared in the query and used in the query, I don't think there is any value in externalizing

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
dvoet: I expect we will end up extracting a bunch but we need to dig into the other queries to see how to do it best

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -241,24 +241,28 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
     groupId match {
       case group: WorkbenchGroupName => groupPKQueryForGroup(group)
       case policy: FullyQualifiedPolicyId => groupPKQueryForPolicy(policy)
-      case _ => throw new WorkbenchException(s"unexpected WorkbenchGroupIdentity $groupId")
```
gpolumbo-broad, line 244: Why remove this exception?

gpolumbo-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
dvoet: because the cases are already exhaustive

PR #320, opened 2019-07-24
Pet queries

marctalbott:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/initial_schema.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/initial_schema.xml
@@ -63,6 +63,9 @@
             <column name="google_subject_id" type="VARCHAR(30)">
                 <constraints unique="true" nullable="false"/>
             </column>
+            <column name="display_name" type="VARCHAR(100)">
+                <constraints unique="false" nullable="false"/>
+            </column>
         </createTable>
     </changeSet>
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
@@ -5,7 +5,7 @@ import java.util.Date
 
 import cats.effect.{ContextShift, IO}
 import org.broadinstitute.dsde.workbench.model._
-import org.broadinstitute.dsde.workbench.model.google.{GoogleProject, ServiceAccount, ServiceAccountDisplayName, ServiceAccountSubjectId}
+import org.broadinstitute.dsde.workbench.model.google.{ServiceAccount, ServiceAccountSubjectId}
 import org.broadinstitute.dsde.workbench.sam.db._
 import org.broadinstitute.dsde.workbench.sam.db.tables._
 import org.broadinstitute.dsde.workbench.sam.model.{BasicWorkbenchGroup, FullyQualifiedPolicyId}
@@ -346,25 +346,20 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
   override def loadUser(userId: WorkbenchUserId): IO[Option[WorkbenchUser]] = {
     runInTransaction { implicit session =>
       val userTable = UserTable.syntax
-      val userColumn = UserTable.column
 
-      import SamTypeBinders._
-      val loadUserQuery = samsql"select ${userTable.id}, ${userTable.email}, ${userTable.googleSubjectId} from ${UserTable.table} where ${userTable.id} = ${userId}"
-      loadUserQuery.map(rs => WorkbenchUser(rs.get[WorkbenchUserId](userColumn.id), rs.stringOpt(userColumn.googleSubjectId).map(GoogleSubjectId), rs.get[WorkbenchEmail](userColumn.email)))
-        .single().apply()
+      val loadUserQuery = samsql"select ${userTable.resultAll} from ${UserTable as userTable} where ${userTable.id} = ${userId}"
+      loadUserQuery.map(UserTable(userTable))
+        .single().apply().map(unmarshalUserRecord)
     }
   }
 
   override def loadUsers(userIds: Set[WorkbenchUserId]): IO[Stream[WorkbenchUser]] = {
     runInTransaction { implicit session =>
       val userTable = UserTable.syntax
-      val userColumn = UserTable.column
 
-      import SamTypeBinders._
-
-      val loadUsersQuery = samsql"select ${userTable.id}, ${userTable.email}, ${userTable.googleSubjectId} from ${UserTable.table} where ${userTable.id} in (${userIds})"
-      loadUsersQuery.map(rs => WorkbenchUser(rs.get[WorkbenchUserId](userColumn.id), rs.stringOpt(userColumn.googleSubjectId).map(GoogleSubjectId), rs.get[WorkbenchEmail](userColumn.email)))
-        .list().apply().toStream
+      val loadUsersQuery = samsql"select ${userTable.resultAll} from ${UserTable as userTable} where ${userTable.id} in (${userIds})"
+      loadUsersQuery.map(UserTable(userTable))
+        .list().apply().map(unmarshalUserRecord).toStream
     }
   }
 
@@ -476,18 +471,17 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
   override def isEnabled(subject: WorkbenchSubject): IO[Boolean] = ???
 
   override def getUserFromPetServiceAccount(petSA: ServiceAccountSubjectId): IO[Option[WorkbenchUser]] = {
-    import SamTypeBinders._
-
     runInTransaction { implicit session =>
       val petServiceAccountTable = PetServiceAccountTable.syntax
-
       val userTable = UserTable.syntax
-      val userColumn = UserTable.column
 
-      val loadUserQuery = samsql"""select ${userTable.id}, ${userTable.email}, ${userTable.googleSubjectId} from ${UserTable.table} left join ${PetServiceAccountTable.table} on ${petServiceAccountTable.userId} = ${userTable.id} where ${petServiceAccountTable.googleSubjectId} = ${petSA}""" //TODO harmonize the subjectId types
+      val loadUserQuery = samsql"""select ${userTable.resultAll}
+                from ${UserTable as userTable}
+                join ${PetServiceAccountTable as petServiceAccountTable} on ${petServiceAccountTable.userId} = ${userTable.id}
+                where ${petServiceAccountTable.googleSubjectId} = ${petSA}""" //TODO harmonize the subjectId types
 
-      loadUserQuery.map(rs => WorkbenchUser(rs.get[WorkbenchUserId](userColumn.id), rs.stringOpt(userColumn.googleSubjectId).map(GoogleSubjectId), rs.get[WorkbenchEmail](userColumn.email)))
-        .single().apply()
+      val userRecordOpt: Option[UserRecord] = loadUserQuery.map(UserTable(userTable)).single().apply()
+      userRecordOpt.map(unmarshalUserRecord)
     }
   }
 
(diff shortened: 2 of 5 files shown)


src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/initial_schema.xml
```diff
@@ -63,6 +63,9 @@
             <column name="google_subject_id" type="VARCHAR(30)">
                 <constraints unique="true" nullable="false"/>
             </column>
+            <column name="display_name" type="VARCHAR(100)">
```
MatthewBemis, line 66: varchar(100) strategically chosen because that's the google-defined limit for display names

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -475,26 +475,91 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
+  override def getUserFromPetServiceAccount(petSA: ServiceAccountSubjectId): IO[Option[WorkbenchUser]] = {
+    import SamTypeBinders._
+
+    runInTransaction { implicit session =>
+      val petServiceAccountTable = PetServiceAccountTable.syntax
+
+      val userTable = UserTable.syntax
+      val userColumn = UserTable.column
+
+      val loadUserQuery = samsql"""select ${userTable.id}, ${userTable.email}, ${userTable.googleSubjectId} from ${UserTable.table} left join ${PetServiceAccountTable.table} on ${petServiceAccountTable.userId} = ${userTable.id} where ${petServiceAccountTable.googleSubjectId} = ${GoogleSubjectId(petSA.value)}""" //TODO harmonize the subjectId types
```
marctalbott: note on this todo: We could replace `GoogleSubjectId(petSA.value)` with just `petSA` and everything seems to still work (generated query is identical). Not sure if that would really address the underlying disharmony between `GoogleSubjectId` and `ServiceAccountSubjectId` that this comment is getting at, but it does make this particular query look a little nicer

src/test/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAOSpec.scala
```diff
@@ -240,8 +241,6 @@ class PostgresDirectoryDAOSpec extends FreeSpec with Matchers with BeforeAndAfte
 
         dao.loadUser(user.id).unsafeRunSync() shouldEqual Option(user)
       }
-
-      "not delete a user's pet SA when the user is deleted" is pending
```
marctalbott, line 244: we realized that this test is unnecessary because a user cannot be deleted if it still has any pets since it would throw a foreign key exception

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -475,26 +475,91 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
+  override def getUserFromPetServiceAccount(petSA: ServiceAccountSubjectId): IO[Option[WorkbenchUser]] = {
+    import SamTypeBinders._
+
+    runInTransaction { implicit session =>
+      val petServiceAccountTable = PetServiceAccountTable.syntax
+
+      val userTable = UserTable.syntax
+      val userColumn = UserTable.column
+
+      val loadUserQuery = samsql"""select ${userTable.id}, ${userTable.email}, ${userTable.googleSubjectId} from ${UserTable.table} left join ${PetServiceAccountTable.table} on ${petServiceAccountTable.userId} = ${userTable.id} where ${petServiceAccountTable.googleSubjectId} = ${petSA}""" //TODO harmonize the subjectId types
```
dvoet: should not be using `PetServiceAccountTable.table` or `UserTable.column`, I think those are useful in insert, update and delete but not select. 

Instead
```
select ${userTable.id}, ${userTable.email}, ${userTable.googleSubjectId} 
from ${UserTable as userTable} left join ${PetServiceAccountTable as petServiceAccountTable} 
on ${petServiceAccountTable.userId} = ${userTable.id} 
where ${petServiceAccountTable.googleSubjectId} = ${GoogleSubjectId(petSA.value)}
```
and why is this a left join?

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -475,26 +475,91 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
+
+      loadUserQuery.map(rs => WorkbenchUser(rs.get[WorkbenchUserId](userColumn.id), rs.stringOpt(userColumn.googleSubjectId).map(GoogleSubjectId), rs.get[WorkbenchEmail](userColumn.email)))
```
dvoet: this should be loading a `UserRecord`

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -475,26 +475,91 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
       petServiceAccount
     }
   }
 
-  override def loadPetServiceAccount(petServiceAccountId: PetServiceAccountId): IO[Option[PetServiceAccount]] = ???
+  override def loadPetServiceAccount(petServiceAccountId: PetServiceAccountId): IO[Option[PetServiceAccount]] = {
+    import SamTypeBinders._
+    runInTransaction { implicit session =>
+      val petServiceAccountTable = PetServiceAccountTable.syntax
+      val petServiceAccountColumn = PetServiceAccountTable.column
```
dvoet: don't use `.column`

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -475,26 +475,91 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
+
+      val loadPetQuery = samsql"""select ${petServiceAccountTable.userId}, ${petServiceAccountTable.project}, ${petServiceAccountTable.googleSubjectId}, ${petServiceAccountTable.email}, ${petServiceAccountTable.displayName} from ${PetServiceAccountTable.table} where ${petServiceAccountTable.userId} = ${petServiceAccountId.userId} and ${petServiceAccountTable.project} = ${petServiceAccountId.project}"""
+
+      loadPetQuery.map(rs => PetServiceAccount(
```
dvoet: should be loading a `PetServiceAccountRecord`

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -475,26 +475,91 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
+      )).list().apply()
+    }
+  }
+
+  override def updatePetServiceAccount(petServiceAccount: PetServiceAccount): IO[PetServiceAccount] = {
+    runInTransaction { implicit session =>
+      val petServiceAccountColumn = PetServiceAccountTable.column
+      val updatePetQuery = samsql"""update ${PetServiceAccountTable.table} set
+        ${petServiceAccountColumn.userId} = ${petServiceAccount.id.userId},
+        ${petServiceAccountColumn.project} = ${petServiceAccount.id.project},
```
dvoet: don't update `userId` or `project` as that is part of the primary key

dvoet (changes requested):
comments about using `.column` and `.table` function apply to more than just the places I commented. Also we should map our result sets to `*Record` objects when loading whole rows then convert those the return type we care about.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -346,25 +346,20 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
   override def loadUser(userId: WorkbenchUserId): IO[Option[WorkbenchUser]] = {
     runInTransaction { implicit session =>
       val userTable = UserTable.syntax
-      val userColumn = UserTable.column
 
-      import SamTypeBinders._
-      val loadUserQuery = samsql"select ${userTable.id}, ${userTable.email}, ${userTable.googleSubjectId} from ${UserTable.table} where ${userTable.id} = ${userId}"
-      loadUserQuery.map(rs => WorkbenchUser(rs.get[WorkbenchUserId](userColumn.id), rs.stringOpt(userColumn.googleSubjectId).map(GoogleSubjectId), rs.get[WorkbenchEmail](userColumn.email)))
-        .single().apply()
+      val loadUserQuery = samsql"select ${userTable.resultAll} from ${UserTable as userTable} where ${userTable.id} = ${userId}"
```
marctalbott, line 350: Using `.resultAll` here and elsewhere so that the ResultSet will have the correct scalike-generated column aliases. As an example, here's a query that is generated using the `.resultAll`: 

```sql
select SAM_USER.id as i_on_SAM_USER, SAM_USER.email as e_on_SAM_USER, SAM_USER.google_subject_id as gsi_on_SAM_USER from SAM_USER where SAM_USER.id = testUser;
```
Without the `.resultAll` and the resulting SQL it generates, `loadUserQuery.map(UserTable(userTable))` would fail because it wouldn't be able to find the column `i_on_SAM_USER` which is what Scalike is looking for when we use `UserTable.apply`.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -346,25 +346,20 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
-      val userColumn = UserTable.column
-
-      import SamTypeBinders._
 
-      val loadUsersQuery = samsql"select ${userTable.id}, ${userTable.email}, ${userTable.googleSubjectId} from ${UserTable.table} where ${userTable.id} in (${userIds})"
-      loadUsersQuery.map(rs => WorkbenchUser(rs.get[WorkbenchUserId](userColumn.id), rs.stringOpt(userColumn.googleSubjectId).map(GoogleSubjectId), rs.get[WorkbenchEmail](userColumn.email)))
-        .list().apply().toStream
+      val loadUsersQuery = samsql"select ${userTable.resultAll} from ${UserTable as userTable} where ${userTable.id} in (${userIds})"
+      loadUsersQuery.map(UserTable(userTable))
+        .list().apply().map(unmarshalUserRecord).toStream
```
dvoet, line 362: the `.toStream` here is worrisome... the whole point of this returning a stream is that it does not load everything into memory.

I don't think it needs to be fixed in this PR but maybe make a JIRA to figure out how to stream results from postgres. [This](http://scalikejdbc.org/documentation/operations.html#foreach-operation) looks interesting plus the section below on Setting JDBC fetchSize.

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -475,26 +470,91 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
-  override def getUserFromPetServiceAccount(petSA: ServiceAccountSubjectId): IO[Option[WorkbenchUser]] = ???
+  override def getUserFromPetServiceAccount(petSA: ServiceAccountSubjectId): IO[Option[WorkbenchUser]] = {
+    runInTransaction { implicit session =>
+      val petServiceAccountTable = PetServiceAccountTable.syntax
+      val userTable = UserTable.syntax
+
+      val loadUserQuery = samsql"""select ${userTable.resultAll}
+                from ${UserTable as userTable}
+                join ${PetServiceAccountTable as petServiceAccountTable} on ${petServiceAccountTable.userId} = ${userTable.id}
+                where ${petServiceAccountTable.googleSubjectId} = ${petSA}""" //TODO harmonize the subjectId types
```
dvoet: fix the TODO

marctalbott: Resolved by changing `PetServiceAccountRecord` to have a `ServiceAccountSubjectId` instead of a `GoogleSubjectId` for the `googleSubjectId` field which is what `PetServiceAccount`s have

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
marctalbott: Agreed, https://broadworkbench.atlassian.net/browse/CA-428

dvoet (approved):

PR #324, opened 2019-07-26
AccessPolicyDAO createResource and createResourceType

ansingh7115:
Ticket: <Link to Jira ticket>

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/ResourceTable.scala
```diff
@@ -18,4 +19,11 @@ object ResourceTable extends SQLSyntaxSupport[ResourceRecord] {
     rs.get(e.name),
     rs.get(e.resourceTypeId)
   )
+
+  def pkQuery(resourceId: ResourceId, resourceTypeName: ResourceTypeName, resourceTableAlias: String = "r"): SQLSyntax = {
+    val r = ResourceTable.syntax(resourceTableAlias)
+    samsqls"""select ${r.id}
+              from ${ResourceTable as r}
+              where ${r.name} = ${resourceId} and ${r.resourceTypeId} = (${ResourceTypeTable.pkQuery(resourceTypeName)})"""
```
dvoet, line 27: I am a little concerned about the subquery here because this is probably itself a subquery. Does postgres handle multiple levels of subqueries well?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -0,0 +1,100 @@
...
+import scalikejdbc._
+
+import scala.concurrent.ExecutionContext
+
+class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
+                              protected val ecForDatabaseIO: ExecutionContext)(implicit executionContext: ExecutionContext) extends AccessPolicyDAO with DatabaseSupport with LazyLogging {
+
+  implicit val cs: ContextShift[IO] = IO.contextShift(executionContext)
+
+  def createResourceType(resourceTypeName: ResourceTypeName): IO[ResourceTypeName] = {
```
dvoet: we need to figure out how to insert all the other resource type information (action patterns, roles)

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -0,0 +1,100 @@
...
+    val resourceTypeTableColumn = ResourceTypeTable.column
+    val insertResourceTypeQuery = samsql"""insert into ${ResourceTypeTable.table} (${resourceTypeTableColumn.name}) values (${resourceTypeName.value})"""
+
+    ResourceTypePK(insertResourceTypeQuery.updateAndReturnGeneratedKey().apply())
+  }
+
+  // 1. Maybe check if groups given in authdomains parameter exist, not sure if we need to do that
+  // 2. Create Resource
+  // 3. Create the entries in the join table for the auth domains
+  // Do we need to care or do something special if the ResourceType doesn't exist?
```
dvoet: no

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  private def insertResourceType(resourceTypeName: ResourceTypeName)(implicit session: DBSession): ResourceTypePK = {
+    val resourceTypeTableColumn = ResourceTypeTable.column
+    val insertResourceTypeQuery = samsql"""insert into ${ResourceTypeTable.table} (${resourceTypeTableColumn.name}) values (${resourceTypeName.value})"""
+
+    ResourceTypePK(insertResourceTypeQuery.updateAndReturnGeneratedKey().apply())
+  }
+
+  // 1. Maybe check if groups given in authdomains parameter exist, not sure if we need to do that
```
dvoet: don't need to

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -0,0 +1,100 @@
...
+
+      if (resource.authDomain.nonEmpty) {
+        insertAuthDomainsForResource(resourcePK, resource.authDomain)
+      }
+
+      resource
+    }.recoverWith {
+      case sqlException: PSQLException if sqlException.getSQLState == PSQLStateExtensions.UNIQUE_VIOLATION => {
+        logger.debug(s"Uniqueness violation", sqlException)
+        IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, "A resource of this type and name already exists")))
```
dvoet: include the sqlException in the error report

src/test/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAOSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+        dao.createResourceType(resourceType).unsafeRunSync()
+        dao.createResource(resource).unsafeRunSync() shouldEqual resource
+      }
+
+      "returns a WorkbenchExceptionWithErrorReport when a resource with the same name and type already exists" in {
+        dao.createResourceType(resourceType).unsafeRunSync()
+        dao.createResource(resource).unsafeRunSync()
+
+        intercept[WorkbenchExceptionWithErrorReport] {
+          dao.createResource(resource).unsafeRunSync()
```
dvoet, line 49: check status code

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/ResourceTable.scala
gpolumbo-broad: totally - reading that PostgreSQL perf article you sent last week, we should try to do all this with `JOIN` if we can, rather than sub-queries.  I was thinking about this as I wrote it and sort of wanted to get this working first and if performance was bad, then optimize.  But happy to take a stab at trying to find a modular way to do this with JOINs too

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
ansingh7115: But the `LdapAccessPolicyDAO` `createResourceType` only creates the resourceType - it does not do any adding of actions and roles, etc. `createResourceType` is also only given a `ResourceTypeName`.

ansingh7115: You mean the resource types that is stored in memory? 
- We could add a parameter to the class for the resourcetypes/actions/patterns/roles and initially just pass them through there. 
- Or create an object in Boot.scala
- Or change the `createResourceType` constructor to take in all that extra stuff and then the the LDAPAccessPolicyDAO `createResourceType` would just not use the the extra information (make it an Optional paramter perhaps)

Idk, lots of ideas

dvoet: that's the figure out part :)

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/ResourceTable.scala
dvoet: I am fine if you want to jira it for now... I satisfied myself that a single PK subquery is fine, I just don't know about another level.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -0,0 +1,98 @@
...
+
+      if (resource.authDomain.nonEmpty) {
+        insertAuthDomainsForResource(resourcePK, resource.authDomain)
+      }
+
+      resource
+    }.recoverWith {
+      case sqlException: PSQLException => {
+        logger.debug(s"createResource psql exception on resource $resource", sqlException)
+        IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"Create resource failed with the following database exception: ${sqlException.getMessage}")))
```
dvoet: `ErrorReport` can take an exception in its constructor, use that instead of embedding the exception message.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -0,0 +1,158 @@
...
+      val resourceRolePK = ResourceRolePK(insertRoleQuery.updateAndReturnGeneratedKey().apply())
+
+        val insertRoleActionQuery =
+          samsql"""insert into ${RoleActionTable.table}
+                   select ${r.id}, ${a.id}
+                   from ${ResourceRoleTable as r}, ${ResourceActionTable as a}
+                   where ${r.resourceTypeId} = ${resourceTypePK}
+                   and   ${a.resourceTypeId} = ${resourceTypePK}
+                   and   ${r.role}   = ${role.roleName}
+                   and   ${a.action} IN (${role.actions.mkString(",")})
```
dvoet: I don't think you need `mkString`

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -0,0 +1,158 @@
...
+                              protected val ecForDatabaseIO: ExecutionContext)(implicit executionContext: ExecutionContext) extends AccessPolicyDAO with DatabaseSupport with LazyLogging {
+
+  implicit val cs: ContextShift[IO] = IO.contextShift(executionContext)
+
+  // Create resourceType should do the following:
+  // 1. Create the resource type
+  // 2. Add the roles
+  // 3. Add the actions
+  // 4. Add patterns
+  def createResourceType(resourceType: ResourceType): IO[ResourceType] = {
```
dvoet, line 31: A problem to address is that this function needs to do the right thing if the resource type already exists. The "right thing" is definitely not to raise an exception (as the current code will do). The "right thing" probably includes making the action patterns be correct, adding any new roles, adding any new actions, updating role-actions. Perhaps remove roles that no longer exist if they are not in use. Essentially everything that needs to happen to make the database state look like what is in config. And keep in mind that if more than 1 sam starts at the same time there will be more than one client trying to do this to be resilient to concurrency issues.

If it makes sense to merge this PR and then work on the above, that is fine.

dvoet (approved):

PR #332, opened 2019-07-31
implement loadResourceAuthDomain

dvoet:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
@@ -148,6 +148,8 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
       val rt = ResourceTypeTable.syntax("rt")
       val g = GroupTable.syntax("g")
 
+      // left joins below so we can detect the difference between a resource does not exist vs.
+      // a resource exists but does not have any auth domains
       val query = samsql"""select ${g.result.name}
               from ${ResourceTable as r}
               join ${ResourceTypeTable as rt} on ${r.resourceTypeId} = ${rt.id}
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.workbench.sam.openam
 
 import akka.http.scaladsl.model.StatusCodes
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -138,7 +140,44 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+      val r = ResourceTable.syntax("r")
+      val rt = ResourceTypeTable.syntax("rt")
+      val g = GroupTable.syntax("g")
+
+      val query = samsql"""select ${g.result.name}
+              from ${ResourceTable as r}
+              join ${ResourceTypeTable as rt} on ${r.resourceTypeId} = ${rt.id}
+              left outer join ${AuthDomainTable as ad} on ${r.id} = ${ad.resourceId}
+              left outer join ${GroupTable as g} on  ${ad.groupId} = ${g.id}
+              where ${r.name} = ${resource.resourceId} and ${rt.name} = ${resource.resourceTypeName}"""
```
dvoet, line 158: I didn't use a pkQuery here and just wrote out the joins because that what made sense to me

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -138,7 +140,44 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+      val ad = AuthDomainTable.syntax("ad")
+      val r = ResourceTable.syntax("r")
+      val rt = ResourceTypeTable.syntax("rt")
+      val g = GroupTable.syntax("g")
+
+      val query = samsql"""select ${g.result.name}
+              from ${ResourceTable as r}
+              join ${ResourceTypeTable as rt} on ${r.resourceTypeId} = ${rt.id}
+              left outer join ${AuthDomainTable as ad} on ${r.id} = ${ad.resourceId}
+              left outer join ${GroupTable as g} on  ${ad.groupId} = ${g.id}
```
gpolumbo-broad, line 157: It took me a little while to figure out why you were doing the outer joins here.  I think the query would work fine with regular inner joins, but then if there were no results, we wouldn't be able to tell if it was because the resource didn't exist or if it was because the resource is just not constrained.  Is that right?  If so, can we put a brief comment here that calls this out, because I don't think it's immediately obvious.  Good job finding a way to get it all done in one query though.

gpolumbo-broad (approved):

gpolumbo-broad:
This branch's tests are failing because of something we're fixing in the target branch.  OK to merge this.

PR #336, opened 2019-08-02
Create resource type

gpolumbo-broad:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
@@ -69,6 +69,7 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
     }
 
     if (roleActionValues.nonEmpty) {
+      // on conflict do nothing
       val insertQuery = samsql"insert into ${RoleActionTable.table}(${RoleActionTable.column.resourceRoleId}, ${RoleActionTable.column.resourceActionId}) values ${roleActionValues}"
       insertQuery.update().apply()
     } else {
@@ -102,6 +103,7 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
 
   private def insertRoles(roles: Set[ResourceRole], resourceTypePK: ResourceTypePK)(implicit session: DBSession): Int = {
     val roleValues = roles.map(role => samsqls"(${resourceTypePK}, ${role.roleName})")
+    // on conflict do nothing
     val insertRolesQuery =
       samsql"""insert into ${ResourceRoleTable.table}(${ResourceRoleTable.column.resourceTypeId}, ${ResourceRoleTable.column.role})
                  values ${roleValues}"""
@@ -117,6 +119,7 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
       samsqls"(${resourceTypePK}, ${action})"
     }
 
+    // on conflict do nothing
     val insertActionQuery = samsql"""insert into ${ResourceActionTable.table}(${ResourceActionTable.column.resourceTypeId}, ${ResourceActionTable.column.action}) values ${uniqueActionValues}"""
 
     insertActionQuery.update().apply()
@@ -127,6 +130,7 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
     val actionPatternValues = actionPatterns map { actionPattern =>
       samsqls"(${resourceTypePK}, ${actionPattern.value}, ${actionPattern.description}, ${actionPattern.authDomainConstrainable})"
     }
+    // Upsert.  possibly throw an error if trying to change isAuthDomainConstrainable from TRUE to FALSE
     val actionPatternQuery =
       samsql"""insert into ${ResourceActionPatternTable.table}
               (${resourceActionPatternTableColumn.resourceTypeId},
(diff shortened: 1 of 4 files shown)


gpolumbo-broad:
@dvoet - This works, that was objective #1.  However, take a look at:

https://github.com/broadinstitute/sam/pull/336/files#diff-fa47ac5a3ac54133b082c1abbdaad58eR57

I don't think there's a way to do the `INSERT INTO SAM_ROLE_ACTION` with one query or without N*2 subqueries.  So you'll see I have it querying the database to get all the `ID`s for the Roles and Actions.  Not sure if you can think of a fancier way to do it.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -23,76 +23,153 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
-  // 1. Create the resource type
-  // 2. Add the roles
-  // 3. Add the actions
-  // 4. Add patterns
+  // This method obtains an EXCLUSIVE lock on the ResourceType table because if we boot multiple Sam instances at once,
+  // they will all try to (re)create ResourceTypes at the same time and could collide. The lock is automatically
+  // released at the end of the transaction.
   def createResourceType(resourceType: ResourceType): IO[ResourceType] = {
+    // Check that actions match action patterns
+    validateRoleActions(resourceType)
```
dvoet: I don't think validation belongs here. We currently don't do it at all so this would be a behavior change. If we want to add it we should put it at the service level.

dvoet: also be careful with stuff like this as it executes outside the IO... the right way would be to make validateRoleActions return an IO then flatmap with the runInTransaction call.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -23,76 +23,153 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+      samsql"""insert into ${ResourceRoleTable.table}(${ResourceRoleTable.column.resourceTypeId}, ${ResourceRoleTable.column.role})
+                  values ${roleValues}
+               on conflict do nothing"""
+
+    insertRolesQuery.update().apply()
   }
 
+  // Note: Each ResourceAction that you are saving here should match the regex pattern of at least 1
+  // ResourceActionPattern defined for the specified ResourceType.  This method DOES NOT perform that validation for
+  // you.  It is up to the caller to make sure the actions being saved are valid.
```
dvoet: remove this comment, it refers to a higher level construct that may or may not apply

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -23,76 +23,153 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+               on conflict do nothing"""
+
+    insertActionQuery.update().apply()
+  }
+
+  // Performs an UPSERT when a record already exists for this exact ActionPattern for this ResourceType.  Query will
+  // overwrite the `description` and `isAuthDomainConstrainable` columns if the ResourceTypeActionPattern already
+  // exists.
+  // Question:  Should we permit the ability to change `isAuthDomainConstrainable` from TRUE to FALSE?  This may have
+  // consequences to how we enforce auth domains.
```
dvoet: yes to the question, make the comment scaladoc, not plain comment

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -23,76 +23,153 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
-                   select ${r.id}, ${a.id}
-                   from ${ResourceRoleTable as r}, ${ResourceActionTable as a}
-                   where ${r.resourceTypeId} = ${resourceTypePK}
-                   and   ${a.resourceTypeId} = ${resourceTypePK}
-                   and   ${r.role}   = ${role.roleName}
-                   and   ${a.action} IN (${role.actions})
-            """
-        insertRoleActionQuery.update().apply()
-      }
+    if (roleActionValues.nonEmpty) {
```
dvoet, line 58: The filter on line 65 seems to do the same thing as this if condition. One or the other will do but don't need both. I prefer to have it here.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -23,76 +23,153 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+      insertQuery.update().apply()
+    } else {
+      0
+    }
+  }
+
+  private def selectActionsForResourceType(resourceTypePK: ResourceTypePK)(implicit session: DBSession): List[ResourceActionRecord] = {
+    val actionsQuery =
+      samsql"""select *
+               from ${ResourceActionTable.table}
```
dvoet: use a `ResourceActionTable.syntax` instead of `.table` and `.column`:
```val rat = ResourceActionTable.syntax("rat")```
then don't do `select *`, do `select ${rat.result.*}`

applies to other selects below. The rule is that `.table` and `.column` are used in inserts, updates and deletes and `.syntax` used in selects

dvoet (changes requested):

dvoet (approved):

PR #337, opened 2019-08-02
[WIP] CA-308/CA-310 Compare Postgres and Ldap DirectoryDAO results

marctalbott:
Ticket: [CA-308](https://broadworkbench.atlassian.net/browse/CA-308) and/or [CA-310](https://broadworkbench.atlassian.net/browse/CA-310)
Inspired by the ['Shared Tests' section in the Scalatest FreeSpec documentation](http://doc.scalatest.org/1.5/org/scalatest/FreeSpec.html), this PR combines many of the tests in `PostgresDirectoryDAOSpec` and `LdapDirectoryDAOSpec` into a single trait called `DirectoryDAOBehaviors` (some tests like the Ldap caching tests are specific to the implementation and are therefore not in the shared trait). `PostgresDirectoryDAOSpec` and `LdapDirectoryDAOSpec` can then say `behave like directoryDAO(dao)` (where `dao` is a Postgres or Ldap `DirectoryDAO` respectively) and all the tests in `DirectoryDAOBehaviors` will be run with the given `dao`. 

This is only a partial solution to comparing the results of the two DAO implementations because the service level tests that currently use the `LdapDirectoryDAO` should also be tested with a `PostgresDirectoryDAO`. A lot of the service level tests that do this are in `GoogleExtensionSpec` and instantiate the `LdapDirectoryDAO` in each test which makes testing with the Postgres DAO a bit more annoying IMO and is why I haven't done it yet. 

This is also a work in progress because there are still inconsistencies with how the DAOs behave as evidenced by the fact that some tests in `DirectoryDAOBehaviors` fail for both `PostgresDirectoryDAO` and `LdapDirectoryDAO` (not the same tests). 

I'm opening this PR because I'm unsure that this is testing the right things or that this is even useful at all and would like some feedback.

---

**PR checklist**


dvoet:
I am ambivalent about this change. On the up side it helps to ensure that we test the same things across both DAOs. On the down side it make a lot of changes to the ldap dao test and makes it hard to verify that we are not inadvertently making changes in the wrong places.

marctalbott:
After discussing with Doug, holding off on this work for now in favor of `AccessPolicyDAO` work. To paraphrase his words, he wasn't convinced that this was the best way to do it and had his concerns as he mentioned above, so going to keep working on the `AccessPolicyDAO` and we can revisit this later.

PR #342, opened 2019-08-21
create, load, delete policies

marctalbott:
Ticket: [CA-309](https://broadworkbench.atlassian.net/browse/CA-309)
Queries in this PR: `createPolicy`, `loadPolicy`, `deletePolicy`

- also of note, we ran into an issue where we wanted to share code (specifically queries) between the `AccessPolicyDAO` and the `DirectoryDAO`. The only place where code was shared between the two DAOs was in `DatabaseSupport`, but we didn't feel that that was a good place to share queries and that it should be reserved for utility functions like `runInTransaction` or (in the future) streaming functions. We considered putting the code in the table classes, but that didn't feel like a good fit either as the code we wanted to share spanned multiple tables. We considered a couple other options, but felt like the best option was to create a new DAO that both the `AccessPolicyDAO` and the `DirectoryDAO` would use. Since the code we shared was related to groups, we called this DAO the `GroupDAO`. Our hope is that in the future, we can further break down the existing DAOs into smaller DAOs. For example, we might be able to break the `DirectoryDAO` into a `GroupDAO`, `UserDAO`, and `PetServiceAccountDAO`. We are open to alternative means of sharing code between DAOs, but this felt like the best choice.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/db/dao/PostgresGroupDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/db/dao/PostgresGroupDAO.scala
@@ -1,0 +1,50 @@
+package org.broadinstitute.dsde.workbench.sam.db.dao
+import cats.effect.{ContextShift, IO}
+import org.broadinstitute.dsde.workbench.model.{WorkbenchException, WorkbenchGroupIdentity, WorkbenchSubject, WorkbenchUserId}
+import org.broadinstitute.dsde.workbench.sam.db.{DbReference, SamTypeBinders}
+import org.broadinstitute.dsde.workbench.sam.db.tables.{GroupMemberTable, GroupPK, GroupTable}
+import org.broadinstitute.dsde.workbench.sam.util.DatabaseSupport
+import scalikejdbc.{DBSession, SQLSyntax}
+import org.broadinstitute.dsde.workbench.sam.db.SamParameterBinderFactory.SqlInterpolationWithSamBinders
+
+import scala.concurrent.ExecutionContext
+
+class PostgresGroupDAO(protected val dbRef: DbReference,
+               protected val ecForDatabaseIO: ExecutionContext)(implicit executionContext: ExecutionContext) extends DatabaseSupport {
+  implicit val cs: ContextShift[IO] = IO.contextShift(executionContext)
+
+  def insertGroupMembers(groupId: GroupPK, members: Set[WorkbenchSubject])(implicit session: DBSession): Int = {
+    if (members.isEmpty) {
+      0
+    } else {
+      val memberUsers: List[SQLSyntax] = members.collect {
+        case userId: WorkbenchUserId => samsqls"(${groupId}, ${Option(userId)}, ${None})"
+      }.toList
+
+      val memberGroupPKQueries = members.collect {
+        case id: WorkbenchGroupIdentity => samsqls"(${workbenchGroupIdentityToGroupPK(id)})"
+      }
+
+      import SamTypeBinders._
+      val memberGroupPKs: List[GroupPK] = if (memberGroupPKQueries.nonEmpty) {
+        val g = GroupTable.syntax("g")
+        samsql"select ${g.result.id} from ${GroupTable as g} where ${g.id} in (${memberGroupPKQueries})"
+          .map(rs => rs.get[GroupPK](g.resultName.id)).list().apply()
+      } else {
+        List.empty
+      }
+
+      val memberGroups: List[SQLSyntax] = memberGroupPKs.map { groupPK =>
+        samsqls"(${groupId}, ${None}, ${Option(groupPK)})"
+      }
+
+      if (memberGroups.size != memberGroupPKQueries.size) {
+        throw new WorkbenchException(s"Some member groups not found.")
+      } else {
+        val gm = GroupMemberTable.column
+        samsql"insert into ${GroupMemberTable.table} (${gm.groupId}, ${gm.memberUserId}, ${gm.memberGroupId}) values ${memberUsers ++ memberGroups}"
+          .update().apply()
+      }
+    }
+  }
+}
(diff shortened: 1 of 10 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -343,11 +310,11 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
                                        p: QuerySQLSyntaxProvider[SQLSyntaxSupport[PolicyRecord], PolicyRecord],
```
gpolumbo-broad, line 310: Parameter not used, do we need it?

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/dao/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,77 @@
...
+class PostgresGroupDAO(protected val dbRef: DbReference,
+                       protected val ecForDatabaseIO: ExecutionContext)(implicit executionContext: ExecutionContext) extends DatabaseSupport {
+  implicit val cs: ContextShift[IO] = IO.contextShift(executionContext)
+
+  def insertGroupMembers(groupId: GroupPK, members: Set[WorkbenchSubject])(implicit session: DBSession): Int = {
+    if (members.isEmpty) {
+      0
+    } else {
+      val memberUsers: List[SQLSyntax] = members.collect {
+        case userId: WorkbenchUserId => samsqls"(${groupId}, ${Option(userId)}, ${None})"
```
gpolumbo-broad: Remove Options

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/dao/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,77 @@
...
+      }.toList
+
+      val memberGroupPKQueries = members.collect {
+        case id: WorkbenchGroupIdentity => samsqls"(${workbenchGroupIdentityToGroupPK(id)})"
+      }
+
+      import SamTypeBinders._
+      val memberGroupPKs: List[GroupPK] = if (memberGroupPKQueries.nonEmpty) {
```
gpolumbo-broad, line 31: Add a TODO here to check to see if there's any way we can do this to find the IDs without needed to run N subqueries.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -286,9 +290,126 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+      samsqls"""select concat(${r.id}, '_', ${policy.id.accessPolicyName}) from ${ResourceTable as r}
+               join ${ResourceTypeTable as rt} on ${r.resourceTypeId} = ${rt.id}
+               where ${r.name} = ${policy.id.resource.resourceId}
+               and ${rt.name} = ${policy.id.resource.resourceTypeName}"""
+
+    GroupPK(samsql"""insert into ${GroupTable.table} (${gCol.name}, ${gCol.email}, ${gCol.updatedDate})
+               values ((${policyGroupName}), ${policy.email}, ${Instant.now()})""".updateAndReturnGeneratedKey().apply())
+  }
+
+  private def loadResourcePK(resource: FullyQualifiedResourceId): SQLSyntax = {
```
gpolumbo-broad: Consider moving this query into `ResourceTable`.  There's already the `pkQuery` method in there.  I think these two methods are doing the same thing, they're just parameter overloaded.  

In addition, I prefer the `JOIN` implementation here over the subquery implementation in the `ResourceTable.pkQuery` method

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -286,9 +290,126 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+    val r = ResourceTable.syntax("r")
+    val rt = ResourceTypeTable.syntax("rt")
+    samsqls"select ${r.id} from ${ResourceTable as r} join ${ResourceTypeTable as rt} on ${r.resourceTypeId} = ${rt.id} where ${r.name} = ${resource.resourceId} and ${rt.name} = ${resource.resourceTypeName}"
+  }
+
+  override def deletePolicy(policy: FullyQualifiedPolicyId): IO[Unit] = {
```
gpolumbo-broad, line 348: This method counts on a cascading deleting when we delete the Group, can we just add a comment to that point.  I know it's kind of obvious, but it's sort of not too.  It also makes me want to go double check the schema definition to REALLY make sure that cascading deletes is on.  If we call it out in a comment, then I feel happier that it's done right

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -286,9 +290,126 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+    runInTransaction { implicit session =>
+      samsql"""delete from ${GroupTable as g}
+        using ${PolicyTable as p}
+        where ${g.id} = ${p.groupId}
+        and ${p.name} = ${policy.accessPolicyName}
+        and ${p.resourceId} = (${loadResourcePK(policy.resource)})""".update().apply()
+    }
+  }
+
+  override def loadPolicy(resourceAndPolicyName: FullyQualifiedPolicyId): IO[Option[AccessPolicy]] = {
```
gpolumbo-broad, line 361: I don't think you should change anything for this PR.

This is awesome that you got this to work with a single query, it's probably pretty performant, but I do find it challenging to grok.  I think it would become easier if we broke it down into subqueries, maybe one for policy info, one for members, one for roles, and one for actions.  If we had the time, I'd be curious to see the perf difference to see if we could balance readability with perf.

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
marctalbott: we don't, but in the next PR I'm going to open that finishes off the pending directory dao tests we will need it, so I'm going to leave it for now since it would just get added right back in

PR #351, opened 2019-08-30
create actions if don't exist

dvoet:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
@@ -319,7 +319,7 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
         import SamTypeBinders._
         val r = ResourceTable.syntax("r")
         val p = PolicyTable.syntax("p")
-        val resourceTypePK = samsql"select ${r.result.resourceTypeId} from ${PolicyTable as p} join ${ResourceTable as r} on ${p.resourceId} = ${r.id}"
+        val resourceTypePK = samsql"select ${r.result.resourceTypeId} from ${PolicyTable as p} join ${ResourceTable as r} on ${p.resourceId} = ${r.id} where ${p.id} = ${policyId}"
           .map(rs => rs.get[ResourceTypePK](r.resultName.resourceTypeId)).single().apply().getOrElse(throw new WorkbenchException(s"could not find resource type id for policy id $policyId"))
         insertActions(actions, resourceTypePK)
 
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -307,9 +307,28 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+      val inserted = samsql"$insertQuery".update().apply()
+
+      if (inserted != actions.size) {
+        // in this case some actions that we want to insert did not exist in ResourceActionTable
+        // add them now and rerun the insert ignoring conflicts
+        // this case should happen rarely
+        import SamTypeBinders._
+        val r = ResourceTable.syntax("r")
+        val p = PolicyTable.syntax("p")
+        val resourceTypePK = samsql"select ${r.result.resourceTypeId} from ${PolicyTable as p} join ${ResourceTable as r} on ${p.resourceId} = ${r.id}"
```
marctalbott: Add a where clause to select the `resourceTypePK` for just this policy: 
```suggestion
        val resourceTypePK = samsql"select ${r.result.resourceTypeId} from ${PolicyTable as p} join ${ResourceTable as r} on ${p.resourceId} = ${r.id} where ${p.id} = ${policyId}"
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -409,7 +428,7 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
       samsql"""delete from ${PolicyRoleTable as pr}
               using ${PolicyTable as p}
               where ${p.name} = ${id.accessPolicyName}
-              and ${p.resourceId} = (${ResourceTable.loadResourcePK(id.resource)}) returning ${p.id}""".updateAndReturnGeneratedKey().apply().toLong)
+              and ${p.resourceId} = (${ResourceTable.loadResourcePK(id.resource)}) returning ${p.id}""".updateAndReturnGeneratedKey().apply())
```
marctalbott, line 431: We only added the `.toLong` because Matt had been seeing some weird errors, but if this works then we should also change [line 409 of `overwritePolicyMembersInternal`](https://github.com/broadinstitute/sam/pull/351/files#diff-fa47ac5a3ac54133b082c1abbdaad58eR409) because we do the same thing there.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
dvoet: good catch

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
dvoet: code compiles, tests pass so looks ok to me

marctalbott (approved):

PR #355, opened 2019-09-05
change (fix?) overwriting Resource Types

marctalbott:
Ticket: [CA-309](https://broadworkbench.atlassian.net/browse/CA-309)
This PR changes how using `createResourceType` to overwrite resource types works. `createResourceType` can be used to update the action patterns, roles, and those roles' actions for a resource type. 

Action patterns can be added and removed at will. There's nothing special here, we just delete all of the existing action patterns for the resource type and then insert all of the given action patterns for a full overwrite.

Roles cannot be deleted. If they are removed in the overwrite, they will be designated as deprecated. Deprecated roles still behave as they did before deprecation, but they cannot be added to any new policies and overwriting an existing policy with that role will remove the role from the policy.

Actions can be added and removed from roles at will. We delete all actions from non-deprecated roles and then re-add them all. Records in the ResourceAction and PolicyAction tables are unaffected.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
@@ -68,10 +68,7 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
         .update().apply()
     } else {
       samsql"""delete from ${RoleActionTable as ra}
-                 using ${ResourceRoleTable as rr}
-                 where ${ra.resourceRoleId} = ${rr.id}
-                 and ${ra.resourceRoleId} in (${resourceTypeRoles.map(_.id)})
-                 and ${rr.role} not in (${roles.map(_.roleName)})"""
+                 where (${ra.resourceRoleId}, ${ra.resourceActionId}) not in (${roleActionValues})"""
         .update().apply()
 
       val insertQuery =
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
@@ -37,7 +37,7 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
 
       insertActionPatterns(resourceType.actionPatterns, resourceTypePK)
       insertRoles(resourceType.roles, resourceTypePK)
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -59,10 +59,15 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
     }
 
     if (roleActionValues.nonEmpty) {
+      val ra = RoleActionTable.syntax("ra")
+
+      samsql"""delete from ${RoleActionTable as ra}
+              where ${ra.resourceRoleId} in (${resourceTypeRoles.map(_.id)})"""
+        .update().apply()
+
```
dvoet, line 73: I think I prefer leaving the `on conflict do nothing` below and having the delete statement above only delete actions that are not in the list (hint, use a `not in` clause). That way the usual behavior on startup is effectively not database changes.

I would follow the same pattern in `overwriteActionPatterns`. I think the leaving `overwriteRoles` the way you currently have it is ok. Something about removing rows and re-adding them on every restart of sam makes me shiver and in these cases it is pretty easy not to do so.

It is probably also easy to deprecate only what is needed in `overwriteRoles` too and be sure to undeprecate what was re-added but I have less strong emotions about that.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -305,10 +321,19 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
     if (actions.nonEmpty) {
       val insertQuery = samsqls"""insert into ${PolicyActionTable.table} (${paCol.resourcePolicyId}, ${paCol.resourceActionId})
-            select ${policyId}, ${ra.result.id} from ${ResourceActionTable as ra} where ${ra.action} in (${actions})"""
+            select ${policyId}, ${ra.result.id}
+            from ${ResourceActionTable as ra}
+            join ${ResourceTypeTable as rt} on ${ra.resourceTypeId} = ${rt.id}
+            join ${ResourceTable as r} on ${r.resourceTypeId} = ${rt.id}
+            join ${PolicyTable as p} on ${p.resourceId} = ${r.id}
+            where ${ra.action} in (${actions})
+            and ${p.id} = ${policyId}"""
```
dvoet, line 359: good catch

dvoet (changes requested):

dvoet (approved):

PR #362, opened 2019-10-08
CA-548 release connection at end of ldapSearchStream, don't close

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/CA-548
during profiling I saw that ldap threads were being created and killed in great numbers despite connection pooling. I discovered that `close` on a connection really closes it even if it is part of a pool. Generally we let unboundid handle getting and release connections from the pool but there is this one place where we do it ourselves.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
@@ -36,27 +36,32 @@ trait LdapSupport extends DirectorySubjectNameSupport {
   protected def ldapSearchStream[T](baseDn: String, searchScope: SearchScope, filters: Filter*)(unmarshaller: Entry => T): Stream[T] =
     filters.flatMap { filter =>
       val search = new SearchRequest(baseDn, searchScope, filter)
-      val entrySource = new LDAPEntrySource(ldapConnectionPool.getConnection, search, true)
-      ldapEntrySourceStream(entrySource)(unmarshaller)
+      val connection = ldapConnectionPool.getConnection
+      val entrySource = new LDAPEntrySource(connection, search, false)
+      ldapEntrySourceStream(entrySource, connection)(unmarshaller)
     }.toStream
 
   // this is the magic recursive stream generator
-  private def ldapEntrySourceStream[T](entrySource: LDAPEntrySource)(unmarshaller: Entry => T): Stream[T] =
+  private def ldapEntrySourceStream[T](entrySource: LDAPEntrySource, connection: LDAPConnection)(unmarshaller: Entry => T): Stream[T] =
     Try(Option(entrySource.nextEntry)) match {
       case Success(None) =>
         // reached the last element, Stream.empty terminates the stream
+        ldapConnectionPool.releaseConnection(connection)
         Stream.empty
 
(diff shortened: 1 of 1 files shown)


marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
```diff
@@ -36,27 +36,32 @@ trait LdapSupport extends DirectorySubjectNameSupport {
   protected def ldapSearchStream[T](baseDn: String, searchScope: SearchScope, filters: Filter*)(unmarshaller: Entry => T): Stream[T] =
     filters.flatMap { filter =>
       val search = new SearchRequest(baseDn, searchScope, filter)
-      val entrySource = new LDAPEntrySource(ldapConnectionPool.getConnection, search, true)
-      ldapEntrySourceStream(entrySource)(unmarshaller)
+      val connection = ldapConnectionPool.getConnection
+      val entrySource = new LDAPEntrySource(connection, search, false)
```
dvoet: notice that the last param switched from true to false

davidangb (approved):
:+1:

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
helgridly: nit: can you put the bool variable name in the call so it's clear what it does

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
```diff
@@ -36,27 +36,32 @@ trait LdapSupport extends DirectorySubjectNameSupport {
...
+      ldapEntrySourceStream(entrySource, connection)(unmarshaller)
     }.toStream
 
   // this is the magic recursive stream generator
-  private def ldapEntrySourceStream[T](entrySource: LDAPEntrySource)(unmarshaller: Entry => T): Stream[T] =
+  private def ldapEntrySourceStream[T](entrySource: LDAPEntrySource, connection: LDAPConnection)(unmarshaller: Entry => T): Stream[T] =
```
helgridly, line 48: so you create a single connection in `ldapSearchStream` and pass it through here but where does it actually get used to do work? all i see is it getting released. or is that related to the mystery bool?

helgridly (commented):
honestly i don't understand what's happening here

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
dvoet: line 40 is where it actually does the query. The rest of it is reading off the `LDAPEntrySource` until there are no more entries. In the old code the connection was automatically closed by `LDAPEntrySource`. The new code tells `LDAPEntrySource` not to do that then releases the connection when there are no more entries.

helgridly: thanks!

helgridly (approved):

dvoet:
jenkins retest

PR #363, opened 2019-10-08
Exactly one member per group member record

marctalbott:
Ticket: [CA-549](https://broadworkbench.atlassian.net/browse/CA-549)
We found that we had a whole lot of entries in the `SAM_GROUP_MEMBER` table that had null values for both `member_user_id` and `member_group_id`. This was occurring because the migration script was migrating policies too late, but that will be addressed in a future PR. It never should have been possible for this to slip through, so this PR adds a constraint to the `SAM_GROUP_MEMBER` table to prevent it.

---

**PR checklist**


dvoet (approved):

dvoet:
add a test please

gpolumbo-broad (approved):

PR #364, opened 2019-10-09
added capability for creating a non-default resource to return inform…

kcibul:
Ticket: https://broadworkbench.atlassian.net/browse/AS-55

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -732,6 +732,8 @@ paths:
     post:
       summary: Create a new resource, cannot be used when resource type allows both authorization domains and id reuse
       responses:
+        201:
+          description: Successfully created resource, returning resource information
         204:
           description: Successfully created resource
         400:
@@ -1512,6 +1514,9 @@ definitions:
         type: array
         items:
           type: string
+      returnResource:
+        type: boolean
+        description: return created resource information with 201 response code
 
   Enabled:
     description: the status of the user's account
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1516,6 +1516,7 @@ definitions:
           type: string
       returnResource:
         type: boolean
+        default: false
         description: return created resource information with 201 response code
 
   Enabled:
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -10,7 +10,7 @@ import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
 import org.broadinstitute.dsde.workbench.sam.model.SamJsonSupport._
 import org.broadinstitute.dsde.workbench.sam.model.RootPrimitiveJsonSupport._
-import org.broadinstitute.dsde.workbench.sam.model._
+import org.broadinstitute.dsde.workbench.sam.model.{CreateResourcePolicyResponse, CreateResourceResponse, _}
 import org.broadinstitute.dsde.workbench.sam.service.ResourceService
 import spray.json.DefaultJsonProtocol._
 import spray.json.JsBoolean
@@ -124,14 +124,23 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
 
   def postResource(resourceType: ResourceType, userInfo: UserInfo): server.Route =
     post {
-      entity(as[CreateResourceRequest]) { createResourceRequest =>
-        if (resourceType.reuseIds && resourceType.isAuthDomainConstrainable) {
-          throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "this api may not be used for resource types that allow both authorization domains and id reuse"))
-        }
-        complete(
-          resourceService
-            .createResource(resourceType, createResourceRequest.resourceId, createResourceRequest.policies, createResourceRequest.authDomain, userInfo.userId)
-            .map(_ => StatusCodes.NoContent))
+        entity(as[CreateResourceRequest]) { createResourceRequest =>
+          if (resourceType.reuseIds && resourceType.isAuthDomainConstrainable) {
+            throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "this api may not be used for resource types that allow both authorization domains and id reuse"))
+          }
+          complete(
+            createResourceRequest.returnResource match {
+              case true =>
+                StatusCodes.Created -> resourceService
+                  .createResource(resourceType, createResourceRequest.resourceId, createResourceRequest.policies, createResourceRequest.authDomain, userInfo.userId)
+                  .map { r => CreateResourceResponse(r.resourceTypeName, r.resourceId, r.authDomain, r.accessPolicies.map(ap => CreateResourcePolicyResponse(ap.id, ap.email))) }
+              case _ =>
+                resourceService
+                  .createResource (resourceType, createResourceRequest.resourceId, createResourceRequest.policies, createResourceRequest.authDomain, userInfo.userId)
+                  .map { _ => StatusCodes.NoContent }
+            }
+          )
+
       }
     }
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -15,8 +15,9 @@ import org.broadinstitute.dsde.workbench.sam.service.ResourceService
 import spray.json.DefaultJsonProtocol._
 import spray.json.JsBoolean
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 import ImplicitConversions.ioOnSuccessMagnet
+import akka.http.scaladsl.marshalling.ToResponseMarshallable
 
 /**
   * Created by mbemis on 5/22/17.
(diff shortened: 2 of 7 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1512,6 +1514,9 @@ definitions:
         type: array
         items:
           type: string
+      returnResource:
+        type: boolean
+        description: return created resource information with 201 response code
```
davidangb, line 1520: add `required` and `default` keys, assuming this is optional

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -124,14 +124,23 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
...
+                StatusCodes.Created -> resourceService
+                  .createResource(resourceType, createResourceRequest.resourceId, createResourceRequest.policies, createResourceRequest.authDomain, userInfo.userId)
+                  .map { r => CreateResourceResponse(r.resourceTypeName, r.resourceId, r.authDomain, r.accessPolicies.map(ap => CreateResourcePolicyResponse(ap.id, ap.email))) }
+              case _ =>
+                resourceService
+                  .createResource (resourceType, createResourceRequest.resourceId, createResourceRequest.policies, createResourceRequest.authDomain, userInfo.userId)
+                  .map { _ => StatusCodes.NoContent }
+            }
+          )
+
```
davidangb, line 142: instead of copy/pasting the `createResource(...)` call in both cases, could you do that once? Something like:
```
resourceService.createResource(...).map{ r =>
  if (createResourceRequest.returnResource)
    StatusCodes.Created -> CreateResourceResponse(r. ...)
  else
    StatusCodes.NoContent
}
```

note also for personal style I'm using `if` instead of `match`ing on a boolean; your call on style here.

( I did not test this suggestion; I'm coding right in github )

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -129,7 +134,8 @@ object SamResourceTypes {
 @Lenses final case class CreateResourceRequest(
     resourceId: ResourceId,
     policies: Map[AccessPolicyName, AccessPolicyMembership],
-    authDomain: Set[WorkbenchGroupName])
+    authDomain: Set[WorkbenchGroupName],
+    returnResource: Boolean = false)
```
MatthewBemis: i saw a little bit of discussion on slack about this-- so akka-http doesn't support query params on a post?

just asking because this feels kind of weird but if it's necessary then i'm okay with it

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -100,9 +103,11 @@ object SamResourceTypes {
 @Lenses final case class ResourceTypeName(value: String) extends ValueObject
 
 @Lenses final case class FullyQualifiedResourceId(resourceTypeName: ResourceTypeName, resourceId: ResourceId)
-@Lenses final case class Resource(resourceTypeName: ResourceTypeName, resourceId: ResourceId, authDomain: Set[WorkbenchGroupName]) {
+@Lenses final case class Resource(resourceTypeName: ResourceTypeName, resourceId: ResourceId, authDomain: Set[WorkbenchGroupName], accessPolicies: Set[AccessPolicy] = Set.empty) {
   val fullyQualifiedId = FullyQualifiedResourceId(resourceTypeName, resourceId)
 }
+@Lenses final case class CreateResourceResponse(resourceTypeName: ResourceTypeName, resourceId: ResourceId, authDomain: Set[WorkbenchGroupName], accessPolicies: Set[CreateResourcePolicyResponse] = Set())
```
davidangb: can you get away without setting the default for `accessPolicies`? Looks like `CreateResourceResponse` is only used once, and that usage specifies a value.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV1Spec.scala
```diff
@@ -111,6 +111,26 @@ class ResourceRoutesV1Spec extends FlatSpec with Matchers with ScalatestRouteTes
...
+    }
+  }
+
+  "POST /api/resources/v1/{resourceType} with createResource = false" should "204 create resource with content" in {
+    val resourceType = ResourceType(ResourceTypeName("rt"), Set(ResourceActionPattern("run", "", false)), Set(ResourceRole(ResourceRoleName("owner"), Set(ResourceAction("run")))), ResourceRoleName("owner"))
+    val samRoutes = TestSamRoutes(Map(resourceType.name -> resourceType))
+
+    val createResourceRequest = CreateResourceRequest(ResourceId("foo"), Map(AccessPolicyName("goober") -> AccessPolicyMembership(Set(defaultUserInfo.userEmail), Set(ResourceAction("run")), Set(resourceType.ownerRoleName))), Set.empty, false)
+    Post(s"/api/resources/v1/${resourceType.name}", createResourceRequest) ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.NoContent
```
davidangb, line 140: could you also verify the response payload has what you expect it to have?

davidangb (commented):
comments inline!

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
MatthewBemis: also should this be an Option[Boolean]?  the automated tests seem unhappy with this.

```
org.broadinstitute.dsde.workbench.service.RestException: {
  "statusCode": 400,
  "source": "rawls",
  "timestamp": 1570641340939,
  "causes": [],
  "stackTrace": [],
  "message": "Sam call to HttpMethod(POST) /api/resources/v1/workspace failed with error HttpEntity.Strict(text/plain; charset=UTF-8,The request content was malformed:\nObject is missing required member 'returnResource')"
}
```

src/main/resources/swagger/api-docs.yaml
kcibul: see above for required, added `default`

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
kcibul: good catch

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV1Spec.scala
kcibul: there should be no payload here (NoContext response codes aren't allowed to have a payload)

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
kcibul: yeah -- I couldn't get it to work and banged on it for like an hour.  Basically it seems to support getting the parameter, but not it's value (it was always empty string). 

Moving to option though

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
kcibul: tried -- couldn't get it to go (implicit marshaller related?) . If you have some time to help make it work, I'm game

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV1Spec.scala
davidangb: ack sorry! I meant to put this comment on the test above, the `StatusCodes.Created` one

MatthewBemis (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
davidangb: this compiles for me, haven't run tests on it:
```
          val resourceMaker: Future[ToResponseMarshallable] = resourceService
            .createResource(resourceType, createResourceRequest.resourceId, createResourceRequest.policies, createResourceRequest.authDomain, userInfo.userId)
              .map { r =>
                if (createResourceRequest.returnResource.contains(true)) {
                  StatusCodes.Created -> CreateResourceResponse(r.resourceTypeName, r.resourceId, r.authDomain, r.accessPolicies.map(ap => CreateResourcePolicyResponse(ap.id, ap.email)))
                }  else {
                  StatusCodes.NoContent
                }
          }

          complete(resourceMaker)
```

davidangb: note I'm using  `createResourceRequest.returnResource.contains(true)` ... I'm not convinced `returnResource` needs to be an `Option`, so you might do away with the `contains` too.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -135,7 +135,7 @@ object SamResourceTypes {
     resourceId: ResourceId,
     policies: Map[AccessPolicyName, AccessPolicyMembership],
     authDomain: Set[WorkbenchGroupName],
-    returnResource: Boolean = false)
+    returnResource: Option[Boolean] = Some(false))
```
davidangb, line 138: I don't think this needs to be an `Option`, as long as you're specifying the default via `returnResource: Boolean = false`

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
davidangb: ahhh sorry - ignore my other comments about `Option`, I see now why you made that change.

MatthewBemis:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
kcibul: updated -- I think it needs to be an option for the akka-http binding/parsing to work properly, Matt saw some errors along these lines earlier

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV1Spec.scala
kcibul: ah ok -- added some content checking there

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
kcibul: 👍

davidangb (approved):
:+1: once tests pass

PR #367, opened 2019-10-11
Performance informed changes

dvoet:
1. Convert a bunch of things from `Future` to `IO`
1. Remove unnecessary passing of execution contexts and/or consolidate on passing just context shift
1. Add some sleeps in `initResourceTypes` until we can remove the shadow
1. Move all shadow execution to its own context

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -7,7 +7,7 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.stream.ActorMaterializer
 import cats.data.NonEmptyList
-import cats.effect.{ExitCode, IO, IOApp}
+import cats.effect.{ContextShift, ExitCode, IO, IOApp}
 import cats.implicits._
 import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
@@ -22,7 +22,7 @@ import org.broadinstitute.dsde.workbench.google2.{GoogleFirestoreInterpreter, Go
 import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchException, WorkbenchSubject}
 import org.broadinstitute.dsde.workbench.newrelic.NewRelicMetrics
 import org.broadinstitute.dsde.workbench.sam.api.{SamRoutes, StandardUserInfoDirectives}
-import org.broadinstitute.dsde.workbench.sam.config.DataStores.{OpenDJ, Postgres}
+import org.broadinstitute.dsde.workbench.sam.config.DataStores.{DataStore, OpenDJ, Postgres}
 import org.broadinstitute.dsde.workbench.sam.config.{AppConfig, DataStoreConfig, GoogleConfig}
 import org.broadinstitute.dsde.workbench.sam.db.DbReference
 import org.broadinstitute.dsde.workbench.sam.directory._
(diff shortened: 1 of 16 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -54,17 +55,24 @@ class ResourceService(
           // make sure resource type admin is added first because the rest depends on it
           createdAdminType <- createResourceType(resourceTypeAdmin)
 
-          result <- resourceTypes.values.filterNot(_.name == SamResourceTypes.resourceTypeAdminName).toList.parTraverse { rt =>
+          // sleep added so shadow dao can catch up, remove when removing opendj
```
gpolumbo-broad, line 58: Might I suggest we add a link to the Jira issue:  https://broadworkbench.atlassian.net/browse/CA-526

That way when we get around to resolving that issue, we'll leave a comment on it saying to grep the code for all references to CA-526.  We can let @marctalbott know that this is the way we should tag code we know we want to remove.

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
```diff
@@ -49,6 +49,8 @@ import scala.concurrent.{Await, Awaitable, ExecutionContext}
   */
 trait TestSupport{
   def runAndWait[T](f: Awaitable[T]): T = Await.result(f, Duration.Inf)
+  def runAndWait[T](f: IO[T]): T = f.unsafeRunSync()
```
gpolumbo-broad, line 52: nice!  we should make sure we remember to use this when writing tests in the future

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
```diff
@@ -176,7 +177,7 @@ class PolicyEvaluatorServiceSpec extends AsyncFlatSpec with Matchers with TestSu
       r shouldBe false
     }
 
-    res.unsafeToFuture()
+    res.unsafeRunSync()
```
gpolumbo-broad, line 180: Can you use your new `runAndWait` here?  and in each of the other tests that are still using `unsafeRunSync`?

gpolumbo-broad (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
dvoet: well... this was more so I didn't have to change a bunch of code (and intellij doesn't support inlining this)

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -225,8 +225,8 @@ object Boot extends IOApp with LazyLogging {
                                  postgresExecutionContext: ExecutionContext,
                                  memberOfCache: Cache[WorkbenchSubject, Set[String]],
                                  newRelicMetrics: NewRelicMetrics): DirectoryDAO = {
-    val ldapDirectoryDAO = new LdapDirectoryDAO(ldapConnectionPool, appConfig.directoryConfig, ldapExecutionContext, memberOfCache)
-    val postgresDirectoryDAO = new PostgresDirectoryDAO(dbReference, postgresExecutionContext)
+    val ldapDirectoryDAO = new LdapDirectoryDAO(ldapConnectionPool, appConfig.directoryConfig, ldapExecutionContext, memberOfCache)(chooseRightContextShift(appConfig.dataStoreConfig, OpenDJ), implicitly)
```
marctalbott, line 228: Why is `implicitly` only used here and not when instantiating the other DAOs? Or at least the other ldap DAO?

marctalbott (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
dvoet: I think calling `unsafeRunSync` directly is better

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: because there is a second implicit parameter for this one that is not required for the others

gpolumbo-broad (approved):

PR #371, opened 2019-10-25
CA-571 short circuit for hasPermission

kcibul:
Ticket: CA-571 

Performance improvement for hasPermission, based on short-circuiting for most common case with a cheaper check

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
@@ -37,7 +37,7 @@ class PolicyEvaluatorService(
   def hasPermission(resource: FullyQualifiedResourceId, action: ResourceAction, userId: WorkbenchUserId): IO[Boolean] = {
     // first attempt the cheap, short circuit and fallback to the full check if it returns false
     for {
-      attempt1 <- hasPermissionShortCircuit(resource, action, userId)
+      attempt1 <- hasPermissionShallowCheck(resource, action, userId)
       attempt2 <- if (attempt1) IO.pure(attempt1) else hasPermissionFullCheck(resource, action, userId)
     } yield {
       attempt2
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -35,6 +35,16 @@ class PolicyEvaluatorService(
   }
 
   def hasPermission(resource: FullyQualifiedResourceId, action: ResourceAction, userId: WorkbenchUserId): IO[Boolean] = {
+    // first attempt the cheap, short circuit and fallback to the full check if it returns false
+    for {
```
devonbush, line 39: nitpicky, but I don't think 'short circuit' is quite the right metaphor, and seeing 'short circuit' in an auth check method name is unsettling.  
maybe hasPermissionShallowCheck?

kcibul: yeah -- i looked for a better name.  I liked short circuit because the procedure here *is* a short-circuited boolean OR evaluation (e.g A || B returns after evaluating A if it is true and continues on to evaluate B if A was false).  The fact that A is shallow doesn't matter here.

However, I like the idea of changing the name of the `hasPermissionShortCircuit` to `hasPermissionShallowCheck` or something since the short circuiting is actually above.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -48,6 +58,38 @@ class PolicyEvaluatorService(
...
+    * calling isMemberOf for a user), this method leverages that to do a fast check for presence
+    */
+  def hasPermissionShortCircuit(resource: FullyQualifiedResourceId, action: ResourceAction, userId: WorkbenchUserId): IO[Boolean] = {
+
+    // if this is auth domain constrainable (or is unknown), we can not use this approach
+    resourceTypes.get(resource.resourceTypeName) match {
+      case Some(x) if !x.isAuthDomainConstrainable =>
+      // nothing
+      case _ =>
+        return IO { false }
```
Qi77Qi: 1. `return` should never be used in scala, and it's never needed
2. There're a few places where the code wrapping things into IO context, and using `map`, `flatMap` just to get inside the context again. It'll be easier to read if you just use `for comprehension` instead.

this function can be something like
```
    for {
      res <- resourceTypes.get(resource.resourceTypeName).traverse {
        rt =>
          if (!rt.isAuthDomainConstrainable) {
            // compute the list of roles that contain the action requested
            val roleNamesWithAction = rt.roles.filter(_.actions.contains(action)).map(_.roleName)

            for {
              // get the policies for this resource and filter to ones that contain the roles with the action, or the action directly
              policies <- accessPolicyDAO.listAccessPolicies(resource).map(_.toList.filter( p => p.roles.intersect(roleNamesWithAction).nonEmpty || p.actions.contains( action ) ))
              members = policies.flatMap(_.members)
            } yield members.contains(userId)
          } else IO.pure(false)
      }
    } yield res.getOrElse(false)
```

`.traverse` is very useful, what it does here is "do something if the `Option` is defined; otherwise do nothing"...

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
```diff
@@ -206,6 +206,104 @@ class PolicyEvaluatorServiceSpec extends FlatSpec with Matchers with TestSupport
     res.unsafeRunSync()
   }
 
+
+  "hasPermission" should "return false if given action is not allowed for a user using the short circuit" in {
+    val user = genUserInfo.sample.get
```
Qi77Qi, line 211: this is fine as is...but you can also try using `forAll`, so that the test will generate a few samples instead of just one, something like https://github.com/broadinstitute/sam/blob/842caafa451d7100407a5023ed57c668d5378c2a/src/test/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectivesSpec.scala#L32

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
kcibul: much nicer -- thanks for the feedback

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
kcibul: I was just following the convention of the rest of the tests, but this is good to know about, thanks

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -48,6 +58,36 @@ class PolicyEvaluatorService(
     }
   }
 
+  /**
+    * Check if the user has the action on the resource.  A true result means they do, and can be trusted.  A false result only means that the short
+    * circuit approach did not find that to be true, but a more exhaustive check might.
```
dvoet: replace short circuit in comment with better name

kcibul: replaced with 'shallow check' here, and another reference to 'short circuit' as well

dvoet (approved):

kcibul:
jenkins retest

PR #375, opened 2019-11-06
AS-118, AS-125 - Sam Performance

kcibul:
Ticket: AS-118,  AS-125
---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -298,7 +298,7 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
   def getUserResourceRoles(resource: FullyQualifiedResourceId, userInfo: UserInfo): server.Route =
     get {
       traceRequest { span =>
-        complete(resourceService.listUserResourceRoles(resource, userInfo).map { roles =>
+        complete(resourceService.listUserResourceRoles(resource, userInfo, span).map { roles =>
           StatusCodes.OK -> roles
         })
       }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -199,7 +199,7 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
   def getActionPermissionForUser(resource: FullyQualifiedResourceId, userInfo: UserInfo, action: String): server.Route =
     get {
       traceRequest { span =>
-        complete(policyEvaluatorService.hasPermission(resource, ResourceAction(action), userInfo.userId).map { hasPermission =>
+        complete(policyEvaluatorService.hasPermission(resource, ResourceAction(action), userInfo.userId, span).map { hasPermission =>
           StatusCodes.OK -> JsBoolean(hasPermission)
         })
       }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
@@ -44,7 +44,12 @@ trait SecurityDirectives {
       }
     }
 
-  private def hasPermissionOneOf(resource: FullyQualifiedResourceId, actions: Set[ResourceAction], userId: WorkbenchUserId): IO[Boolean] =
-    actions.toList.existsM(policyEvaluatorService.hasPermission(resource, _, userId))
-  
+  private def hasPermissionOneOf(resource: FullyQualifiedResourceId, actions: Iterable[ResourceAction], userId: WorkbenchUserId): IO[Boolean] =
+    //  first quickly check if we have permission using the shallow check across all actions, then try the full check
+    for {
+      attempt1 <- actions.toList.existsM(policyEvaluatorService.hasPermissionShallowCheck(resource, _, userId))
+      attempt2 <- if (attempt1) IO.pure(attempt1) else actions.toList.existsM(policyEvaluatorService.hasPermissionFullCheck(resource, _, userId))
+    } yield {
+      attempt2
+    }
 }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
@@ -6,15 +6,46 @@ import akka.http.scaladsl.server.{Directive0, Directives}
 import org.broadinstitute.dsde.workbench.model.{ErrorReport, WorkbenchExceptionWithErrorReport, WorkbenchUserId}
 import org.broadinstitute.dsde.workbench.sam._
 import org.broadinstitute.dsde.workbench.sam.model.{FullyQualifiedResourceId, ResourceAction}
-import org.broadinstitute.dsde.workbench.sam.service.PolicyEvaluatorService
+import org.broadinstitute.dsde.workbench.sam.service.{PolicyEvaluatorService, ResourceService}
 import ImplicitConversions.ioOnSuccessMagnet
 import cats.effect.IO
 
 trait SecurityDirectives {
   def policyEvaluatorService: PolicyEvaluatorService
+  def resourceService: ResourceService
 
-  def requireAction(resource: FullyQualifiedResourceId, action: ResourceAction, userId: WorkbenchUserId): Directive0 =
-    requireOneOfAction(resource, Set(action), userId)
+  def requireAction(resource: FullyQualifiedResourceId, action: ResourceAction, userId: WorkbenchUserId): Directive0 = {
+    val requestedActions = Set(action)
+
+    Directives.mapInnerRoute { innerRoute =>
+      onSuccess(policyEvaluatorService.hasPermission(resource, action, userId)) { hasPermission =>
+        if (hasPermission) {
+          innerRoute
+        } else {
+
+          // in the case where we don't have the required action, we need to figure out if we should return
+          // a Not Found (you have no access) vs a Forbidden (you have access, just not the right kind)
+          onSuccess(policyEvaluatorService.listResourceAccessPoliciesForUser(resource, userId)) { policies =>
+
+            if (policies.isEmpty) {
+              Directives.failWith(
+                new WorkbenchExceptionWithErrorReport(
+                  ErrorReport(StatusCodes.NotFound, s"Resource ${resource.resourceTypeName.value}/${resource.resourceId.value} not found")))
+            } else {
+              Directives.failWith(
+                new WorkbenchExceptionWithErrorReport(ErrorReport(
+                  StatusCodes.Forbidden,
+                  s"You may not perform any of ${requestedActions.mkString("[", ", ", "]").toString.toUpperCase} on ${resource.resourceTypeName.value}/${resource.resourceId.value}"
+                )))
+
+
+            }
+          }
+        }
+      }
+    }
+  }
+
 
   def requireOneOfAction(resource: FullyQualifiedResourceId, requestedActions: Set[ResourceAction], userId: WorkbenchUserId): Directive0 =
     Directives.mapInnerRoute { innerRoute =>
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
@@ -6,14 +6,13 @@ import akka.http.scaladsl.server.{Directive0, Directives}
 import org.broadinstitute.dsde.workbench.model.{ErrorReport, WorkbenchExceptionWithErrorReport, WorkbenchUserId}
 import org.broadinstitute.dsde.workbench.sam._
 import org.broadinstitute.dsde.workbench.sam.model.{FullyQualifiedResourceId, ResourceAction}
-import org.broadinstitute.dsde.workbench.sam.service.{PolicyEvaluatorService, ResourceService}
+import org.broadinstitute.dsde.workbench.sam.service.PolicyEvaluatorService
 import ImplicitConversions.ioOnSuccessMagnet
 import cats.implicits._
 import cats.effect.IO
 
 trait SecurityDirectives {
   def policyEvaluatorService: PolicyEvaluatorService
-  def resourceService: ResourceService
 
   def requireAction(resource: FullyQualifiedResourceId, action: ResourceAction, userId: WorkbenchUserId): Directive0 =
     requireOneOfAction(resource, Set(action), userId)
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
@@ -8,17 +8,19 @@ import org.broadinstitute.dsde.workbench.sam._
 import org.broadinstitute.dsde.workbench.sam.model.{FullyQualifiedResourceId, ResourceAction}
 import org.broadinstitute.dsde.workbench.sam.service.{PolicyEvaluatorService, ResourceService}
 import ImplicitConversions.ioOnSuccessMagnet
+import cats.implicits._
 import cats.effect.IO
 
 trait SecurityDirectives {
   def policyEvaluatorService: PolicyEvaluatorService
   def resourceService: ResourceService
 
-  def requireAction(resource: FullyQualifiedResourceId, action: ResourceAction, userId: WorkbenchUserId): Directive0 = {
-    val requestedActions = Set(action)
+  def requireAction(resource: FullyQualifiedResourceId, action: ResourceAction, userId: WorkbenchUserId): Directive0 =
+    requireOneOfAction(resource, Set(action), userId)
 
+  def requireOneOfAction(resource: FullyQualifiedResourceId, requestedActions: Set[ResourceAction], userId: WorkbenchUserId): Directive0 =
     Directives.mapInnerRoute { innerRoute =>
-      onSuccess(policyEvaluatorService.hasPermission(resource, action, userId)) { hasPermission =>
+      onSuccess(hasPermissionOneOf(resource, requestedActions, userId)) { hasPermission =>
         if (hasPermission) {
           innerRoute
         } else {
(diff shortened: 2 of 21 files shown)


kcibul:
jenkins retest

araina2018:
@kcibul  - i tried looking up a JIRA ticket with INVALID CREDENTIALS but can't find one. Will check with Gary.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
```diff
@@ -18,30 +19,32 @@ trait SecurityDirectives {
...
-      actionsAttempt1 <- policyEvaluatorService.listUserResourceActions(resource, userId, force = false)
-      actionsAttempt2 <- if (hasAccess(requestedActions, actionsAttempt1)) IO.pure(actionsAttempt1)
-      else policyEvaluatorService.listUserResourceActions(resource, userId, force = true)
-    } yield actionsAttempt2
-
-  private def hasAccess(requestedActions: Set[ResourceAction], actions: Set[ResourceAction]): Boolean =
-    actions.intersect(requestedActions).nonEmpty
+  private def hasPermissionOneOf(resource: FullyQualifiedResourceId, actions: Set[ResourceAction], userId: WorkbenchUserId): IO[Boolean] =
+    actions.toList.existsM(policyEvaluatorService.hasPermission(resource, _, userId))
+
```
davidangb: for consideration ... would it be more optimal to try 
```
actions.toList.existsM(policyEvaluatorService.hasPermissionShallowCheck(resource, _, userId))
```
first, and iif that is false, then try
```
policyEvaluatorService.listUserResourceActions
```
and look in the list for any of the requested actions?

Additionally, consider changing these directives to accept an `Iterable[ResourceAction]` instead of `Set[ResourceAction]`, which would allow callers of these directives to optimize by listing the most-likely action first in the iteration order.

kcibul: great points, will incorporate

davidangb: ah, and if you do  the `listUserResourceActions` as part of the permission check, you can reuse its result for figuring out 403 vs 404

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -247,7 +247,60 @@ class LdapAccessPolicyDAO(
...
+  // for each policy, see if we are a member
+  //    -> check if we are directly a member, if so... yes
+  //    -> if there are groups... check if we are a member of that group
+  //        --> etc...
+    for {
+      allPolicies <- listAccessPolicies(resource).map(_.toSet)
+    } yield {
+      // KCIBUL: help me!!!
+      allPolicies.filter( x => isUserMemberOfPolicy(x, userId).unsafeRunSync() )
+    }
```
davidangb: does this do what you need?
```
    for {
      allPolicies <- listAccessPolicies(resource)
      myPolicies <- allPolicies.filterA(p => isUserMemberOfPolicy(p, userId))
    } yield {
      allPolicies.toSet
    }
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -247,7 +247,60 @@ class LdapAccessPolicyDAO(
...
       ))
 
-  override def listAccessPoliciesForUser(resource: FullyQualifiedResourceId, user: WorkbenchUserId): IO[Set[AccessPolicy]] =
+
+  private def isUserMemberOfPolicy(policy: AccessPolicy, userId: WorkbenchUserId): IO[Boolean] = {
+    if (policy.members.contains(userId)) {
+      IO.pure(true)
+    } else {
+      // KCIBUL -- what
+      policy.members.collect{case x:WorkbenchGroupName => x}.toList.existsM(isUserMemberOfGroup(_, userId))
```
davidangb: re: your "what" comment ... this code looks fine to me!

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
kcibul: yep, thanks! (but yielding myPolicies)

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
kcibul: ah that was pre my discovery of collect :D

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -156,8 +157,10 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
   }
 
   def getUserPoliciesForResourceType(resourceType: ResourceType, userInfo: UserInfo): server.Route =
-    get {
-      complete(policyEvaluatorService.listUserAccessPolicies(resourceType.name, userInfo.userId))
+    traceRequest { span =>
```
davidangb: `span` is never used, change to `_`

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -217,21 +222,25 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
     }
 
   def getResourcePolicies(resource: FullyQualifiedResourceId, userInfo: UserInfo): server.Route =
-    get {
-      requireAction(resource, SamResourceActions.readPolicies, userInfo.userId) {
-        complete(resourceService.listResourcePolicies(resource).map { response =>
-          StatusCodes.OK -> response.toSet
-        })
+    traceRequest { span =>
```
davidangb: `span` is never used, change to `_`

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -217,21 +222,25 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
...
     }
 
   def getPolicy(policyId: FullyQualifiedPolicyId, userInfo: UserInfo): server.Route =
-    get {
-      requireOneOfAction(policyId.resource, Set(SamResourceActions.readPolicies, SamResourceActions.readPolicy(policyId.accessPolicyName)), userInfo.userId) {
-        complete(resourceService.loadResourcePolicy(policyId).map {
-          case Some(response) => StatusCodes.OK -> response
-          case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "policy not found"))
-        })
+    traceRequest { span =>
```
davidangb: `span` => `_`

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
```diff
@@ -18,30 +19,37 @@ trait SecurityDirectives {
...
     for {
-      actionsAttempt1 <- policyEvaluatorService.listUserResourceActions(resource, userId, force = false)
-      actionsAttempt2 <- if (hasAccess(requestedActions, actionsAttempt1)) IO.pure(actionsAttempt1)
-      else policyEvaluatorService.listUserResourceActions(resource, userId, force = true)
-    } yield actionsAttempt2
-
-  private def hasAccess(requestedActions: Set[ResourceAction], actions: Set[ResourceAction]): Boolean =
-    actions.intersect(requestedActions).nonEmpty
+      attempt1 <- actions.toList.existsM(policyEvaluatorService.hasPermissionShallowCheck(resource, _, userId))
+      attempt2 <- if (attempt1) IO.pure(attempt1) else actions.toList.existsM(policyEvaluatorService.hasPermissionFullCheck(resource, _, userId))
```
davidangb, line 51: `hasPermissionFullCheck` will call `listUserResourceActions` for *each* of the `actions`, at least until it finds a `true` result. This is inefficient. It would be better to call `listUserResourceActions` once, and look for any of the requested `actions` in the list.

This "look for any action in the list" logic might be better encapsulated in `PolicyEvaluatorService` than here?

davidangb: in fact this is a regression from the previous implementation, which called `listUserResourceActions` only once. (in both the previous impl and in `hasPermissionFullCheck` I realize we call `listUserResourceActions` first with force=false and then with force=true)

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
davidangb: heh, totally right - sorry for the bad pseudocode

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
```diff
@@ -103,5 +107,37 @@ trait LdapSupport extends DirectorySubjectNameSupport {
...
+    executeLdap(IO(ldapSearchStream(groupsOu, SearchScope.SUB, filters: _*)(unmarshalGroupThrow)))
+  }
+
+  def isUserMemberOfGroup(groupName: WorkbenchGroupName, userId: WorkbenchUserId): IO[Boolean] = {
+    val members = ldapLoadGroup(groupName).map(_.get.members)
+    val isDirectMember = members.map( _.contains(userId))
+    val isGroupMember = members.flatMap( _.collect{case x:WorkbenchGroupName => x}.toList.existsM(isUserMemberOfGroup(_, userId)))
+
+    List(isDirectMember, isGroupMember).sequence.map( _.contains(true) )
+  }
```
davidangb, line 137: this also needs a short-circuit - we don't want to query the subgroups if the user is a direct member. Something like ... 
```
  def isUserMemberOfGroup(groupName: WorkbenchGroupName, userId: WorkbenchUserId): IO[Boolean] =
    for {
      group <- ldapLoadGroup(groupName)
      members = group.map(_.members).getOrElse(Set.empty[WorkbenchSubject])
      isDirectMember = members.contains(userId)
      isGroupMember <- if (isDirectMember) IO.pure(false) else members.collect{case x:WorkbenchGroupName => x}.toList.existsM(isUserMemberOfGroup(_, userId))
    } yield {
      isDirectMember || isGroupMember
    }
```
(though I don't like the hardcoded `IO.pure(false)` for isGroupMember

davidangb (commented):
more comments inline

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
dvoet: Agreed about the regression. In the case where the user does not have access this will do more ismemberof calls than before. In reality this looks like there are at most 2 actions ever required and almost all of those are in the policy altering apis. So this may be an acceptable regression.

In the case where the user does have permission but not shallowly assigned, caching should ensure there is only 1 ismemberof call.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
```diff
@@ -247,28 +247,22 @@ class LdapAccessPolicyDAO(
...
+      allPolicies <- listAccessPolicies(resource)
+      userPolicies <- allPolicies.filterA(p => isUserMemberOfPolicy(p, userId))
+    } yield userPolicies.toSet
+
+
+  private def isUserMemberOfPolicy(policy: AccessPolicy, userId: WorkbenchUserId): IO[Boolean] = {
+    if (policy.members.contains(userId)) {
+      IO.pure(true)
+    } else {
+      policy.members.collect{case x:WorkbenchGroupName => x}.toList.existsM(ldapIsUserMemberOfGroup(_, userId))
```
dvoet: `members.collect{case x:WorkbenchGroupName => x}.toList.existsM(ldapIsUserMemberOfGroup(_, userId))` is repeated in `ldapIsUserMemberOfGroup`. Can you pull that into a function named like `isSubGroupMember`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
```diff
@@ -103,5 +107,40 @@ trait LdapSupport extends DirectorySubjectNameSupport {
...
+    res.value
+  }
+
+  def ldapLoadGroups(groupNames: Set[WorkbenchGroupName]): IO[Stream[BasicWorkbenchGroup]] = {
+    val filters = groupNames.grouped(batchSize).map(batch => Filter.createORFilter(batch.map(g => Filter.createEqualityFilter(Attr.cn, g.value)).asJava)).toSeq
+
+    executeLdap(IO(ldapSearchStream(groupsOu, SearchScope.SUB, filters: _*)(unmarshalGroupThrow)))
+  }
+
+  def ldapIsUserMemberOfGroup(groupName: WorkbenchGroupName, userId: WorkbenchUserId): IO[Boolean] = {
```
dvoet: add `Recursive` suffix to name

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -71,23 +75,49 @@ class PolicyEvaluatorService(
...
-          // we can only use this approach if the resource type isn't auth-domain constrainable
-          if (!rt.isAuthDomainConstrainable) {
-
-            // compute the list of roles that contain the action requested
-            val roleNamesWithAction = rt.roles.filter(_.actions.contains(action)).map(_.roleName)
+          val roleNamesWithAction = rt.roles.filter(_.actions.contains(action)).map(_.roleName)
+          val hasAction = directMemberHasActionOnResource(resource, roleNamesWithAction, action, userId)
 
+          // if the resource type OR action are NOT auth domain constrainable... just return
+          if (!rt.isAuthDomainConstrainable || !rt.actionPatterns.exists( ap => ap.authDomainConstrainable & ap.matches(action)) ) {
```
dvoet: is there a reason there is only one `&` in `ap.authDomainConstrainable & ap.matches(action)`

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -71,23 +75,49 @@ class PolicyEvaluatorService(
...
+                  }
+              }
+            } yield res2
+          }
       }
     } yield res.getOrElse(false)
 
   }
 
+  def isUserMemberOfGroup(groupName: WorkbenchGroupName, userId: WorkbenchUserId): IO[Boolean] = {
```
dvoet: this seems duplicate of code at the ldap layer, can we just rewrite `isGroupMember` to be a recursive call?

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
```diff
@@ -103,5 +107,49 @@ trait LdapSupport extends DirectorySubjectNameSupport {
...
+    executeLdap(IO(ldapSearchStream(groupsOu, SearchScope.SUB, filters: _*)(unmarshalGroupThrow)))
+  }
+
+//  override def isGroupMember(groupId: WorkbenchGroupIdentity, member: WorkbenchSubject): IO[Boolean] =
+//    for {
+//      memberOf <- ldapLoadMemberOf(member)
+//    } yield {
+//      val memberships = memberOf.map(_.toLowerCase) //toLowerCase because the dn can have varying capitalization
+//      memberships.contains(groupDn(groupId).toLowerCase)
+//    }
```
davidangb: delete commented-out code

davidangb (approved):
:+1: assuming tests pass.

I'd still like to resolve the multiple queries in the negative case of `hasPermissionOneOf` but I'm fine doing that as a second PR so we can get this one in.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
```diff
@@ -154,6 +155,49 @@ class PolicyEvaluatorServiceSpec extends FlatSpec with Matchers with TestSupport
     }
   }
 
+  "hasPermission" should "return true if given action is granted through membership in another policy" in {
```
kcibul, line 158: @dvoet can you review this test please?  thanks

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/LdapAccessPolicyDAO.scala
kcibul: this moved, but I made the change in the new home LdapSupport

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
kcibul: this no longer exists, it's just 'isGroupMember'

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
kcibul: not a good one ;) . fixed to be &&

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
kcibul: yep, great suggestion.  really cleaned things up

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/LdapSupport.scala
kcibul: done

kcibul:
jenkins retest

kcibul:
jenkins retest

kcibul:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -71,21 +75,39 @@ class PolicyEvaluatorService(
...
-          } else IO.pure(false)
+              // first, check if the resource in question HAS an auth domain (likely not) and if so continue [TERRA]
+              authDomainsResult <- accessPolicyDAO.loadResourceAuthDomain(resource)
+              res2 <- authDomainsResult match {
+                case LoadResourceAuthDomainResult.NotConstrained | LoadResourceAuthDomainResult.ResourceNotFound =>
+                  hasAction
+                case LoadResourceAuthDomainResult.Constrained(authDomains) =>
+                  // if there is more than one group, just fallback
+                  if (authDomains.size > 1) {
+                    IO.pure(false)
```
dvoet, line 94: since isGroupMember is better, I don't think you need to fallback in this case:
```authDomains.traverse(directoryDAO.isGroupMember(_, userId)).map(_.forall(identity))```

File: src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala [hunk 16]
```diff
@@ -154,6 +155,49 @@ class PolicyEvaluatorServiceSpec extends FlatSpec with Matchers with TestSupport
...
+    val policy2WithUser = AccessPolicy.members.set(Set(policy.id))(policy2ExtraAction)
+
+    val policy2 = SamLenses.resourceIdentityAccessPolicy.set(resource2.fullyQualifiedId)(policy2WithUser)
+
+    val res = for{
+      _ <- setup()
+      _ <- policyDAO.createResourceType(managedGroupResourceType)
+      _ <- dirDAO.createUser(WorkbenchUser(user.userId, Some(GoogleSubjectId(user.userId.value)), user.userEmail))
+      _ <- resource.authDomain.toList.parTraverse(a => managedGroupService.createManagedGroup(ResourceId(a.value), dummyUserInfo))
+      _ <- resource2.authDomain.toList.parTraverse(a => managedGroupService.createManagedGroup(ResourceId(a.value), dummyUserInfo))
```
dvoet: Just a regular traverse is fine

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
```diff
@@ -154,6 +155,49 @@ class PolicyEvaluatorServiceSpec extends FlatSpec with Matchers with TestSupport
...
+
+    val samplePolicy = genPolicy.sample.get
+    val policyWithUser = AccessPolicy.members.set(samplePolicy.members + user.userId)(samplePolicy)
+    val policy = SamLenses.resourceIdentityAccessPolicy.set(resource.fullyQualifiedId)(policyWithUser)
+
+    val resource2 = genResource.sample.get.copy(resourceTypeName = defaultResourceType.name)
+
+    val samplePolicy2 = genPolicy.sample.get
+    val policy2ExtraAction = AccessPolicy.actions.set(samplePolicy.actions + action)(samplePolicy2)
+    val policy2WithUser = AccessPolicy.members.set(Set(policy.id))(policy2ExtraAction)
```
dvoet: rename `policy2WithUser` to `policy2WithNestedPolicy`

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
dvoet: LGTM

dvoet (approved):
a few comments but looks good

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
kcibul: after talking...  going to save this for a future PR, we're at the point where we could make the Shallow check... full, but still fast.  But also by that point all the LDAP stuff will be not long for this world

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -96,6 +96,7 @@ class PolicyEvaluatorService(
                     directoryDAO.isGroupMember(authDomains.head, userId)
                   }
               }
+            res2 <- if (authConstraintOk) hasAction else IO.pure(false)
```
davidangb, line 99: nit: indent; makes reading the for-comp easier

davidangb (approved):
one nit inline, :+1: if tests pass

PR #376, opened 2019-11-06
pr feedback for #375

davidangb:
pr feedback for #375

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
@@ -42,15 +42,15 @@ class PolicyEvaluatorService(
     // first attempt the shallow check and fallback to the full check if it returns false
     for {
       attempt1 <- hasPermissionShallowCheck(resource, action, userId)
-      attempt2 <- if (attempt1) IO.pure(attempt1) else traceIOWithParent("fullCheck", parentSpan)(_ => hasPermissionFullCheck(resource, action, userId))
+      attempt2 <- if (attempt1) IO.pure(attempt1) else traceIOWithParent("fullCheck", parentSpan)(_ => hasPermissionFullCheck(resource, List(action), userId))
     } yield {
       attempt2
     }
   }
 
-  def hasPermissionFullCheck(resource: FullyQualifiedResourceId, action: ResourceAction, userId: WorkbenchUserId, parentSpan: Span = null): IO[Boolean] = {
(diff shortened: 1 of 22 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -42,15 +42,15 @@ class PolicyEvaluatorService(
...
       attempt1 <- hasPermissionShallowCheck(resource, action, userId)
-      attempt2 <- if (attempt1) IO.pure(attempt1) else traceIOWithParent("fullCheck", parentSpan)(_ => hasPermissionFullCheck(resource, action, userId))
+      attempt2 <- if (attempt1) IO.pure(attempt1) else traceIOWithParent("fullCheck", parentSpan)(_ => hasPermissionFullCheck(resource, List(action), userId))
     } yield {
       attempt2
     }
   }
 
-  def hasPermissionFullCheck(resource: FullyQualifiedResourceId, action: ResourceAction, userId: WorkbenchUserId, parentSpan: Span = null): IO[Boolean] = {
+  def hasPermissionFullCheck(resource: FullyQualifiedResourceId, actions: Iterable[ResourceAction], userId: WorkbenchUserId, parentSpan: Span = null): IO[Boolean] = {
```
davidangb, line 51: you could also just have this accept a `Set`, there's no real need to genericize this to `Iterable`

davidangb:
manual run of swatomation-pipeline at https://fc-jenkins.dsp-techops.broadinstitute.org/view/Testing/view/Pipelines/job/swatomation-pipeline/26826/ ... since this PR isn't against `develop` it won't automatically kick off tests

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
kcibul: I don't know about this one.  It makes the semantic less intuitive (do you mean all of the actions?  one of the actions?) and breaks the symmetry of full/shallow check signature.  It's also only taken advantage of in one place, and callers to the full check now have to wrap their action in a collection of some sort.

kcibul: to be fair though it IS more efficient, but only in the cases where you have multiple actions that fail the shallow check and then want to test them with the full check.  My reasoning is that the underlying benefit from that isn't worth it against the comments above.

davidangb: yeah, I'm definitely optimizing for efficiency. I'd be totally fine changing the method name (`hasAnyPermissionFullCheck`), or changing the signature of the shallow check to allow multiple permissions. I recognize the optimization is in the less-frequent case but I'd hate to have suboptimal performance for the sake of code structure.

PR #381, opened 2019-11-07
[CA-577] Remove references to auth domains when deleting a resource

marctalbott:
Ticket: [CA-577](https://broadworkbench.atlassian.net/browse/CA-577)
From the ticket description

> When a resource that does not allow id reuse is deleted we leave its recording in the db with no policies. References to auth domain groups are not removed which prevents those groups from being deleted in postgres. This causes automation test failures during cleanup. 

This PR changes resource deletion to clean up references to auth domain groups on resources that don't allow id reuse when they are deleted.

---

**PR checklist**

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
@@ -17,7 +17,7 @@ import org.broadinstitute.dsde.workbench.sam.directory.{DirectoryDAO, LdapDirect
 import org.broadinstitute.dsde.workbench.sam.model._
 import org.broadinstitute.dsde.workbench.sam.openam.{AccessPolicyDAO, LdapAccessPolicyDAO, PostgresAccessPolicyDAO}
 import org.broadinstitute.dsde.workbench.sam.schema.JndiSchemaDAO
-import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.concurrent.{ScalaFutures, Eventually}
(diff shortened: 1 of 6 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -703,6 +703,19 @@ class ResourceServiceSpec extends FlatSpec with Matchers with ScalaFutures with
...
+    val managedGroupName = "fooGroup"
+    constrainableService.createResourceType(resourceType).unsafeRunSync()
+    constrainableService.createResourceType(managedGroupResourceType).unsafeRunSync()
+    managedGroupService.createManagedGroup(ResourceId(managedGroupName), dummyUserInfo).unsafeRunSync()
+    val resource = constrainableService.createResource(resourceType, ResourceId(UUID.randomUUID().toString), Map(AccessPolicyName(constrainableReaderRoleName.value) -> constrainablePolicyMembership), Set(WorkbenchGroupName(managedGroupName)), dummyUserInfo.userId).unsafeRunSync()
+
+    runAndWait(constrainableService.deleteResource(resource.fullyQualifiedId))
+    runAndWait(managedGroupService.deleteManagedGroup(ResourceId(managedGroupName)))
+    managedGroupService.loadManagedGroup(ResourceId(managedGroupName)).unsafeRunSync() shouldBe None
+  }
```
dvoet, line 724: just being a little paranoid here, but can you add to your test another resource with an auth domain and verify that it remains unchanged? The failure case I am thinking about on the postgres side is that the query is wrong and deletes all auth domains.

marctalbott:
Got verbal approval from Doug to merge this

PR #401, opened 2020-01-13
CA-526, CA-638, CA-639 out with ldap (almost)

dvoet:
Ticket: 
https://broadworkbench.atlassian.net/browse/CA-526
https://broadworkbench.atlassian.net/browse/CA-638
https://broadworkbench.atlassian.net/browse/CA-639

remove all the ldap code we can, remove the shadow code, keep ldap registration in the right places, fix Boot.scala to do the right things, fix tests

please also review https://github.com/broadinstitute/firecloud-develop/pull/1976

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapRegistrationDAO.scala
```diff
@@ -0,0 +1,135 @@
...
+  override def createUser(user: WorkbenchUser): IO[WorkbenchUser] = {
+    val attrs = List(
+      new Attribute(Attr.email, user.email.value),
+      new Attribute(Attr.sn, user.id.value),
+      new Attribute(Attr.cn, user.id.value),
+      new Attribute(Attr.uid, user.id.value),
+      new Attribute("objectclass", Seq("top", "workbenchPerson").asJava)
+    ) ++ user.googleSubjectId.map(gsid => List(new Attribute(Attr.googleSubjectId, gsid.value))).getOrElse(List.empty)
+
+    executeLdap(IO(ldapConnectionPool.add(userDn(user.id), attrs: _*))).adaptError {
```
gpolumbo-broad, line 52: How come we just call `executeLdap` directly here instead of wrapping in `retryLdapBusyWithBackoff`?  Looks like we only use the retry in `enableIdentity`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
```diff
@@ -176,8 +169,7 @@ object AppConfig {
     val emailDomain = config.as[Option[String]]("emailDomain").getOrElse(config.getString("googleServices.appsDomain"))
     val resourceTypes = config.as[Map[String, ResourceType]]("resourceTypes").values.toSet
     val liquibaseConfig = config.as[LiquibaseConfig]("liquibase")
-    val dataStoreConfig = config.as[DataStoreConfig]("dataStore")
 
-    AppConfig(emailDomain, directoryConfig, schemaLockConfig, distributedLockConfig, swaggerConfig, googleConfigOption, resourceTypes, liquibaseConfig, dataStoreConfig)
+    AppConfig(emailDomain, directoryConfig, schemaLockConfig, distributedLockConfig, swaggerConfig, googleConfigOption, resourceTypes, liquibaseConfig)
```
marctalbott, line 173: Can we remove all of the `CacheConfig` stuff too?

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/RegistrationDAO.scala
```diff
@@ -0,0 +1,14 @@
...
+trait RegistrationDAO {
+    def createUser(user: WorkbenchUser): IO[WorkbenchUser]
+    def loadUser(userId: WorkbenchUserId) : IO[Option[WorkbenchUser]]
+    def deleteUser(userId: WorkbenchUserId): IO[Unit]
+    def enableIdentity(subject: WorkbenchSubject): IO[Unit]
+    def disableIdentity(subject: WorkbenchSubject): IO[Unit]
+    def isEnabled(subject: WorkbenchSubject): IO[Boolean]
+    def createPetServiceAccount(petServiceAccount: PetServiceAccount): IO[PetServiceAccount]
+    def deletePetServiceAccount(petServiceAccountId: PetServiceAccountId): IO[Unit]
+  }
```
gpolumbo-broad: Can we add a comment to this class and/or `LdapRegistrationDAO` briefly explaining something like:  

> This class is here because even though Postgres is the source of record, Apache proxies still query LDAP directly to determine if incoming requests are for a User in the Enabled Users group.  This means that whenever we create, delete, enable, or disable users in Postgres, we must ensure that the change is reflected in LDAP.  Once we move away from a solution that requires that the Apache proxies query this group, we can remove the `RegistrationDAO`.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -334,7 +335,9 @@ class GoogleExtensions(
         case (None, _) =>
           for {
             p <- directoryDAO.createPetServiceAccount(PetServiceAccount(PetServiceAccountId(user.id, project), serviceAccount))
+            _ <- registrationDAO.createPetServiceAccount(PetServiceAccount(PetServiceAccountId(user.id, project), serviceAccount))
             _ <- directoryDAO.enableIdentity(p.id)
+            _ <- registrationDAO.enableIdentity(p.id)
```
gpolumbo-broad, line 340: Could we just reuse/call the `enablePetServiceAccount` method here?

marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -98,7 +99,7 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
           googleStatus <- cloudExtensions.getUserStatus(user)
           allUsersGroup <- cloudExtensions.getOrCreateAllUsersGroup(directoryDAO)
           allUsersStatus <- directoryDAO.isGroupMember(allUsersGroup.id, user.id).unsafeToFuture() recover { case e: NameNotFoundException => false }
-          ldapStatus <- directoryDAO.isEnabled(user.id).unsafeToFuture()
+          ldapStatus <- registrationDAO.isEnabled(user.id).unsafeToFuture()
```
gpolumbo-broad, line 102: I think this one should still get the `isEnabled` information from the `directoryDAO`.  Same in the other `get` methods below.  Basically, anywhere in the Sam code where we are checking `isEnabled` I think we should be getting this information from Postgres.   Which I guess means the only place we are using `isEnabled` is in the tests.

dvoet:
jenkins retest with new firecloud-develop stuff

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
```diff
@@ -116,6 +90,7 @@ object TestSupport extends TestSupport {
   def genSamDependencies(resourceTypes: Map[ResourceTypeName, ResourceType] = Map.empty, googIamDAO: Option[GoogleIamDAO] = None, googleServicesConfig: GoogleServicesConfig = googleServicesConfig, cloudExtensions: Option[CloudExtensions] = None, googleDirectoryDAO: Option[GoogleDirectoryDAO] = None, policyAccessDAO: Option[AccessPolicyDAO] = None)(implicit system: ActorSystem) = {
     val googleDirectoryDAO = new MockGoogleDirectoryDAO()
     val directoryDAO = new MockDirectoryDAO()
+    val registrationDAO = new MockDirectoryDAO()
```
gpolumbo-broad, line 93: Should `MockDirectoryDAO` also extend `RegistrationDAO`?  I know _technically_ this works, but I think we should be explicit about the contract.

src/test/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapRegistrationDAOSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+      dao.createUser(user).unsafeRunSync()
+    }
+
+    assertResult(Some(user)) {
+      dao.loadUser(user.id).unsafeRunSync()
+    }
+
+    dao.deleteUser(user.id).unsafeRunSync()
+
+    assertResult(None) {
```
gpolumbo-broad, line 56: Should we also assert that the pet was deleted along with the User?

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapRegistrationDAO.scala
dvoet: none of this is new code, that's just how it was. The retry was added after the fact to account for problems at higher scale of user registration

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
dvoet: I just want to call enable in ldap in every place we call enable for postgres

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
dvoet: at the moment I disagree... this is used for diagnostics only and it is not useful to check the thing that does not really matter

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
dvoet: but DirectoryDAO extends RegistrationDAO... why do we need more?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
dvoet: also looking at it more closely `enablePetServiceAccount` adds the pet to the proxy group as well which is handled differently in the pet creation process

gpolumbo-broad (approved):

dvoet:
jenkins retest now that google is better

dvoet:
jenkins retest - bond startup error

dvoet:
jenkins retest - looks like unrelated rawls failure

PR #404, opened 2020-02-04
[CA-675] Properly update pet service accounts in OpenDJ

marctalbott:
Ticket: [CA-675](https://broadworkbench.atlassian.net/browse/CA-675)
I'm not sure when this started, but Leo automation tests running against Alpha began failing with numerous requests from pet service accounts failing as Unauthorized. I found that those pet service accounts had different google subject ids in Postgres and OpenDJ. I also found that the google subject id in Postgres matched up with the google subject ids I was seeing in the Kibana logs. Since the google subject id in OpenDJ was different than the google subject id that was coming in with the request, the proxy returned 401s. 
This PR resolves prevents OpenDJ and Postgres from getting out of sync when updating pet service accounts by adding `updatePetServiceAccount` to the `RegistrationDAO`.

---

**PR checklist**

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
@@ -278,7 +278,7 @@ class GoogleExtensionSpec(_system: ActorSystem) extends TestKit(_system) with Fl
   }
 
   "GoogleExtension" should "get a pet service account for a user" in {
-    val (dirDAO: DirectoryDAO, mockGoogleIamDAO: MockGoogleIamDAO, mockGoogleDirectoryDAO: MockGoogleDirectoryDAO, googleExtensions: GoogleExtensions, service: UserService, defaultUserId: WorkbenchUserId, defaultUserEmail: WorkbenchEmail, defaultUserProxyEmail: WorkbenchEmail, createDefaultUser: CreateWorkbenchUser) = initPetTest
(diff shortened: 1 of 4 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -387,6 +388,7 @@ class GoogleExtensionSpec(_system: ActorSystem) extends TestKit(_system) with Fl
 
     val petServiceAccount2 = googleExtensions.createUserPetServiceAccount(defaultUser, googleProject).unsafeRunSync()
     petServiceAccount.serviceAccount shouldNot equal(petServiceAccount2.serviceAccount)
+    regDAO.loadPetServiceAccount(petServiceAccount.id).unsafeRunSync() shouldBe Some(petServiceAccount2)
```
kyuksel, line 391: Is this test failing without your changes?

dvoet (approved):

gpolumbo-broad (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
marctalbott: The test as it was before was succeeding in dev sam. I added this assertion to the test before I made the fix in `GoogleExtensions` and saw that the assertion was failing because the `RegistrationDAO` (LDAP) had the wrong value for the pet service account's google subject id. After I made the fix, this test passed again.

PR #407, opened 2020-02-24
[CA-540] Step 1/3Add delete_billing_project to allow a billing project owner to delete…

yonghaoy:
Add delete_billing_project to allow a billing project owner to delete. This will be used in CA-540 and the ralws change will come after this PR get checked in. 
Ticket: <https://broadworkbench.atlassian.net/browse/CA-540>

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -269,7 +269,7 @@ resourceTypes = {
         description = "add this billing-project to a service-perimeter"
       }
       delete = {
-        description = "delete this billing-project"
+        description = ""
       }
     }
     ownerRoleName = "owner"
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -275,7 +275,7 @@ resourceTypes = {
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["view_status", "create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "list_notebook_cluster", "launch_notebook_cluster", "sync_notebook_cluster", "delete_notebook_cluster", "alter_google_role", "delete_billing_project"]
+        roleActions = ["view_status", "create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "list_notebook_cluster", "launch_notebook_cluster", "sync_notebook_cluster", "delete_notebook_cluster", "alter_google_role", "add_to_service_perimeter", "delete_billing_project"]
       }
       workspace-creator = {
         roleActions = ["view_status", "create_workspace", "share_policy::can-compute-user", "read_policy::can-compute-user"]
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -268,11 +268,17 @@ resourceTypes = {
       add_to_service_perimeter = {
         description = "add this billing-project to a service-perimeter"
       }
+      delete_service_perimeter = {
```
gpolumbo-broad: It looks to me like in the ticket we decided not to permit the removal of a project from a perimeter.  

If I am wrong, can we make sure the ticket is updated accordingly?  Also, if I am wrong and we _do_ need this permission, then I would rather it be called `remove_from_perimeter` or `delete_project_from_perimeter` or something like that.

gpolumbo-broad (changes requested):

src/main/resources/reference.conf
yonghaoy: Ahh, sorry, I forgot when I added delete_service_perimeter but that is not intended. Removed.

Thanks

dvoet (changes requested):
this is incorrect, the action should just be called `delete`. Only the `delete` action will allow the resource to be deleted from sam.

dvoet (approved):

gpolumbo-broad (approved):

dvoet:
I think the tests are failing because the delete action is specified twice leading to a db error
```
insert into SAM_RESOURCE_TYPE_ACTION_PATTERN (resource_type_id, action_pattern, description, is_auth_domain_constrainable) values (16, 'view_status', 'view status of a billing-project', false), (16, 'stop_start_notebook_cluster', 'stop and start notebook clusters in this billing-project', false), (16, 'launch_notebook_cluster', 'launch a new notebook cluster in this billing-project', false), (16, 'alter_google_role', 'modify the role of users on the google project (only a fixed set of roles are permitted)', false), (16, 'add_to_service_perimeter', 'add this billing-project to a service-perimeter', false), (16, 'delete', '', false), (16, 'read_policy"::.+"', 'view the membership of the specified policy', false), (16, 'create_workspace', 'create new workspaces in this billing-project', false), (16, 'launch_batch_compute', 'launch workflows in this billing-project', false), (16, 'delete_notebook_cluster', '', false), (16, 'list_notebook_cluster', 'list all notebook clusters in this billing-project', false), (16, 'delete', 'delete this billing-project', false), (16, 'sync_notebook_cluster', '', false), (16, 'share_policy"::.+"', 'modify the membership of the specified policy', false), (16, 'alter_policies', 'create and delete policies for this billing-project', false), (16, 'read_policies', 'list all policies and policy details for this billing-project', false) on conflict (resource_type_id, action_pattern) do update set description = EXCLUDED.description, is_auth_domain_constrainable = EXCLUDED.is_auth_domain_constrainable

[ERROR] [17:53:39.315] [scala-execution-context-global-72] o.b.dsde.workbench.sam.Boot$ - FATAL - failure starting http server
org.postgresql.util.PSQLException: ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time
```

probably should change the code to deduplicate

PR #412, opened 2020-03-04
CA-682 Integration with Identity Concentrator

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/CA-682

There are a few things going on in this PR
- The obvious: calling out to identity concentrator to get google subject id and email of user when a jwt is used to authenticate
- Introduction of http4s and circe to make http calls because akka http does not use IO which led to some library incompatibilities
- upgrade some libraries to fix incompatibilities
- change header handling to prevent the case where valid jwt is used and OIDC headers are also specified and thus used to impersonate another user
- more extensive direct testing of StandardUserInfoDirectives and removed use of headers in all other tests

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -117,32 +122,78 @@ object Boot extends IOApp with LazyLogging {
...
-            val googleKmsInterpreter = GoogleKmsInterpreter[IO](googleKmsClient, blockingEc)
-            val resourceTypeMap = appConfig.resourceTypes.map(rt => rt.name -> rt).toMap
-            val cloudExtension = createGoogleCloudExt(foregroundAccessPolicyDAO, foregroundDirectoryDAO, registrationDAO, config, resourceTypeMap, lock, newGoogleStorage, googleKmsInterpreter)
-            val googleGroupSynchronizer = new GoogleGroupSynchronizer(backgroundDirectoryDAO, backgroundAccessPolicyDAO, cloudExtension.googleDirectoryDAO, cloudExtension, resourceTypeMap)(backgroundLdapExecutionContext)
-            val cloudExtensionsInitializer = new GoogleExtensionsInitializer(cloudExtension, googleGroupSynchronizer)
-            createAppDepenciesWithSamRoutes(appConfig, cloudExtensionsInitializer, foregroundAccessPolicyDAO, foregroundDirectoryDAO, registrationDAO)
-          }
-        case None => cats.effect.Resource.pure[IO, AppDependencies](createAppDepenciesWithSamRoutes(appConfig, NoExtensionsInitializer, foregroundAccessPolicyDAO, foregroundDirectoryDAO, registrationDAO))
-      }
-    } yield appDependencies
```
dvoet, line 145: extracted this code to a function down below which returns `cloudExtensionsInitializer` and move the call to `createAppDepenciesWithSamRoutes` to the yield

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -13,56 +13,92 @@ import org.broadinstitute.dsde.workbench.sam.api.StandardUserInfoDirectives._
...
-        optionalHeaderValueByName(googleSubjectIdHeader) &
-        optionalHeaderValueByName(expiresInHeader) &
-        optionalHeaderValueByName(emailHeader) &
-        headerValueByName(authorizationHeader)
+    (optionalHeaderValueByName(accessTokenHeader) &
+      optionalHeaderValueByName(googleSubjectIdHeader) &
+      optionalHeaderValueByName(expiresInHeader) &
+      optionalHeaderValueByName(emailHeader) &
+      headerValueByName(authorizationHeader) &
+      provide(identityConcentratorService)
```
dvoet: using `provide` here so the existence or not of `identityConcentratorService` can fit into the pattern matching below

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -92,47 +128,54 @@ object StandardUserInfoDirectives {
...
   }
 
-  private def getBearerTokenFromAuthHeader(authorizationHeader: String): IO[OAuth2BearerToken] = {
-    val bearerPrefix = "bearer "
-    // need to be case insensitive when looking for the 'bearer' prefix so can't use stripPrefix
-    if (authorizationHeader.toLowerCase.startsWith(bearerPrefix)) {
-      IO.pure(OAuth2BearerToken(authorizationHeader.drop(bearerPrefix.size)))
-    } else {
-      IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Unauthorized, "could not parse authorization header, not a bearer token")))
-    }
```
dvoet, line 116: all code related to parsing the auth header to a JsObject when it is a JWT moved to `JwtAuthorizationHeader` below

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -52,10 +52,10 @@ class GoogleKeyCache(
       ))
 
     googleStorageAlg
-      .createBucket(googleServicesConfig.serviceAccountClientProject, googleServicesConfig.googleKeyCacheConfig.bucketName, List.empty)
+      .insertBucket(googleServicesConfig.serviceAccountClientProject, googleServicesConfig.googleKeyCacheConfig.bucketName)
```
dvoet, line 55: changes in this file due to library upgrade

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
```diff
@@ -122,8 +116,8 @@ object TestSupport extends TestSupport {
     SamDependencies(mockResourceService, policyEvaluatorService, new UserService(directoryDAO, googleExt, registrationDAO), new StatusService(directoryDAO, googleExt, dbRef), mockManagedGroupService, directoryDAO, policyDAO, googleExt)
   }
 
-  def genSamRoutes(samDependencies: SamDependencies)(implicit system: ActorSystem, materializer: Materializer): SamRoutes = new SamRoutes(samDependencies.resourceService, samDependencies.userService, samDependencies.statusService, samDependencies.managedGroupService, null, samDependencies.directoryDAO, samDependencies.policyEvaluatorService, LiquibaseConfig("", false))
-    with StandardUserInfoDirectives
+  def genSamRoutes(samDependencies: SamDependencies, uInfo: UserInfo)(implicit system: ActorSystem, materializer: Materializer): SamRoutes = new SamRoutes(samDependencies.resourceService, samDependencies.userService, samDependencies.statusService, samDependencies.managedGroupService, null, samDependencies.directoryDAO, samDependencies.policyEvaluatorService, LiquibaseConfig("", false))
+    with MockUserInfoDirectives
```
dvoet, line 120: This change is important. This means all tests that use this (pretty much all route testing) no longer use `StandardUserInfoDirectives` and headers to pass in users making requests. This is ok because there is now more comprehensive testing of `StandardUserInfoDirectives` and those other tests did not really care about the user anyway.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesSpec.scala
```diff
@@ -26,42 +24,23 @@ import scala.concurrent.Future
...
-    val (user, headers, _, routes) = createTestUser()
+    val (user, _, routes) = createTestUser()
 
-    Get("/register/user").withHeaders(headers) ~> routes.route ~> check {
-      status shouldEqual StatusCodes.OK
-      responseAs[UserStatus] shouldEqual UserStatus(UserStatusDetails(user.id, user.email), Map("ldap" -> true, "allUsersGroup" -> true, "google" -> true))
-    }
-  }
-
-  "POST /register/user" should "retrieve PET owner's user info if a PET account is provided" in {
```
dvoet, line 52: this is tested in `StandardUserInfoDirectivesSpec`

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectivesSpec.scala
```diff
@@ -207,5 +275,96 @@ class StandardUserInfoDirectivesSpec extends FlatSpec with PropertyBasedTesting
     }
   }
 
-  //skipping positive test for StandardUserInfoDirectives since all other routes mixes in StandardUserInfoDirectives relies on success path.
```
dvoet, line 210: :)

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -117,32 +122,78 @@ object Boot extends IOApp with LazyLogging {
...
+          new GoogleExtensionsInitializer(cloudExtension, googleGroupSynchronizer)
+        }
+      case None => cats.effect.Resource.pure[IO, CloudExtensionsInitializer](NoExtensionsInitializer)
+    }
+
+  private def identityConcentratorResource(appConfig: AppConfig): effect.Resource[IO, Option[IdentityConcentratorService]] =
+    appConfig.identityConcentratorConfig match {
+      case Some(IdentityConcentratorConfig(baseUrl, threadPoolSize)) =>
+        for {
+          identityConcentratorEc <- ExecutionContexts.fixedThreadPool[IO](threadPoolSize)
```
wchamber, line 190: mostly curious: why does IC get its own dedicated threadpool?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -270,6 +322,7 @@ object Boot extends IOApp with LazyLogging {
           val googleExtensions = googleExt
           val cloudExtensions = googleExt
           val googleGroupSynchronizer = synchronizer
+          override val identityConcentratorService = identityConcentrator
```
wchamber, line 325: I don't really understand the scala syntax here, but is identityConcentrator getting stuffed in as a 'GoogleExtensionRoutes'? Doesn't seem very Google-y. Should we rename that?
Is there a reason that the identityConcentratorService is only being set when there are also some GoogleExtensionsInitializer?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -92,47 +128,54 @@ object StandardUserInfoDirectives {
...
+    googleIdentities match {
+      case Seq() =>
+        IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, s"no linked google identities for $identityConcentratorId")))
+      case Seq(singleIdentity) => IO.pure(singleIdentity)
+      case _ =>
+        IO.raiseError(new WorkbenchExceptionWithErrorReport(
+          ErrorReport(StatusCodes.InternalServerError, s"too many linked google identities for $identityConcentratorId: ${googleIdentities.mkString("'")}")))
+    }
+
+  private def parseJwtBearerToken(jwtJson: JsObject): IO[JwtUserInfo] = {
```
wchamber: It's a little confusing that this function has the same name as the private one in JwtAuthorizationHeader. I guess there's levels of how "parsed" our tokens are.

Could we "parse" the incoming authorization token/OAuthBearerToken once into some object that has all the fields for the info inside of it that we want to extract? that would consolidate all of the parsing levels into one place and make the rest of the code care less about how different parts are extracted from jwts and json objects. There are also functions that take bearer tokens and JwtUserInfo, though those two paraemters really need to be related. With a single parameter encapsulating both, we're leveraging our types to prevent silly errors or lazy testing where "I know I only need some of these fields to be right" (which I have personal scars over).
I think the data we want is 1) the OAuthBearerToken, 2) the JwtUserInfo (though I think at the moment it's really just the 'sub' field inside of it.)

src/main/scala/org/broadinstitute/dsde/workbench/sam/identityConcentrator/IdentityConcentratorApi.scala
```diff
@@ -0,0 +1,35 @@
...
+import cats.effect.{ContextShift, IO}
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.workbench.sam.identityConcentrator.IdentityConcentratorModel.{OpenIdConfiguration, UserInfo}
+import org.http4s.{AuthScheme, Credentials, MediaType}
+import org.http4s.Method._
+import org.http4s.client.Client
+import org.http4s.client.dsl.io._
+import org.http4s.headers.{Accept, Authorization}
+
+trait IdentityConcentratorApi {
```
wchamber, line 18: Mind leaving a tiny comment about what role IdentityConcentrator plays in Sam?

src/main/scala/org/broadinstitute/dsde/workbench/sam/identityConcentrator/IdentityConcentratorModel.scala
```diff
@@ -0,0 +1,28 @@
+package org.broadinstitute.dsde.workbench.sam.identityConcentrator
+import cats.effect.IO
+import cats.implicits._
+import io.circe.Decoder
+import io.circe.generic.semiauto.deriveDecoder
+import org.http4s.circe.jsonOf
+import org.http4s.{EntityDecoder, Uri}
+
+object IdentityConcentratorModel {
```
wchamber, line 9: Comment about where these model definitions are coming from?

src/main/scala/org/broadinstitute/dsde/workbench/sam/identityConcentrator/IdentityConcentratorService.scala
```diff
@@ -0,0 +1,75 @@
...
+import cats.implicits._
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.workbench.model.{ErrorReport, GoogleSubjectId, WorkbenchEmail}
+import org.broadinstitute.dsde.workbench.sam._
+import org.broadinstitute.dsde.workbench.sam.identityConcentrator.IdentityConcentratorModel._
+import pdi.jwt.{JwtCirce, JwtOptions}
+
+import scala.util.{Failure, Success}
+
+class IdentityConcentratorService(identityConcentratorApi: IdentityConcentratorApi) extends LazyLogging {
```
wchamber, line 26: Class comment? Worth at least describing why its worth distinguishing from IdentityConcentratorApi.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesSpec.scala
```diff
@@ -329,7 +329,7 @@ class ResourceRoutesSpec extends FlatSpec with Matchers with ScalatestRouteTest
       status shouldEqual StatusCodes.NoContent
     }
 
-    val testUser = CreateWorkbenchUser(WorkbenchUserId("testuser"), defaultGoogleSubjectId, WorkbenchEmail("testuser@foo.com"))
+    val testUser = CreateWorkbenchUser(WorkbenchUserId("testuser"), defaultGoogleSubjectId, WorkbenchEmail("testuser@foo.com"), None)
```
wchamber: Could we define a testUser variable once for this entire file? I don't think the tests care what the id or email strings are.
Ditto other files with lots of mechanical CreateWorkbenchUser changes.
Not going to block the review over, but I think it's worth doing in general and while we're touching a hundred lines anyway.

wchamber (changes requested):
Mind saying a little bit more about needing new http libraries vs akka and IO? What's akka return and is it hard to convert to IO?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: This is essentially constructing an anonymous class of type SamRoutes with StandardUserInfoDirectives and GoogleExtensionRoutes mixed in plus those vals overridden.

Good catch on the second sentence.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: I am not 100% certain if this http client lib performs blocking operations in this thread pool or not. If it does then it is a bad idea to give it the main thread pool. This is what the leonarno project does so I am following their lead.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: the naming is an artifact of the code evolution, will fix. But the parsing steps need to be separate. The code needs to detect a non-opaque bearer token even if it isn't completely parseable to the format we need. This is so we know when to use the OIDC headers or not.

wchamber: Can we still do the parse in one step but allow for the fact that some part of it isn't parseable? We can have Optionals or Eithers or something for further steps in parsing.

dvoet: yeah... I have something better coming up in a minute

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -30,61 +30,62 @@ trait StandardUserInfoDirectives extends UserInfoDirectives with LazyLogging {
...
-      case (_, _, _, _, JwtAuthorizationHeader((jwtJson, bearerToken)), Some(icService)) =>
-        onSuccess {
-          getUserInfoFromJwt(jwtJson, bearerToken, directoryDAO, icService).unsafeToFuture()
-        }
-      case (Some(token), Some(googleSubjectId), Some(expiresIn), Some(email), _, _) =>
-        // request coming through Apache proxy
-        onSuccess {
-          getUserInfoFromOidcHeaders(directoryDAO, googleSubjectId, expiresIn, email, OAuth2BearerToken(token)).unsafeToFuture()
-        }
+      case (JwtAuthorizationHeader(Success(jwtUserInfo), bearerToken), Some(icService)) =>
```
wchamber, line 50: Maybe not a big deal, but the asymmetry here is bugging me. All of the parsing of OIDC -> T is in the fromOIDC function, but this handleAuthHeaders is doing half of the parsing of the header values and leaving the rest to fromJWT after pulling out a jwtUserInfo and a bearerToken.
Can we change the signatures of requireUserInfoFromJwt & requireCreateUserInfoFromJwt to look more like requireUserInfoFromOidc & requireCreateUserInfoFromOidc? I think it's a little different in that we want to allow the authorization header to not be present, so maybe they return some optional or maybe that's handled in this 'handleAuthHeaders' function in choosing where to route it.

Or maybe this is more natural with directives.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -203,17 +194,20 @@ final case class JwtUserInfo(sub: IdentityConcentratorId, exp: Long)
...
+    * - is a jwt but missing required fields: return Some(Failure, OAuth2BearerToken)
+    * - is a jwt that is parsable to JwtUserInfo: return Some(Success(JwtUserInfo), OAuth2BearerToken)
     *
     * @param authorizationHeader unprocessed value of the Authorization http header
     * @return None if the header is not a bearer token or is not a parsable JWT.
-    * Otherwise it will extract the parsed json of the JWT payload and an OAuth2BearerToken.
+    * Otherwise it will Try to parse the JWT payload to a JwtUserInfo and an OAuth2BearerToken.
     */
-  def unapply(authorizationHeader: String): Option[(JsObject, OAuth2BearerToken)] = {
+  def unapply(authorizationHeader: String): Option[(Try[JwtUserInfo], OAuth2BearerToken)] = {
```
wchamber, line 200: Thanks, this is easier for me to follow.
optional consideration: make a case/sum class (whatever you call it) for the three distinct states: the None state, the regrets jwtuserinfo parse case (we never care what the OAuth2BearerToken is if we have regrets about parsing the JwtUserInfo in handleAuthHeaders?), and the tuple (JwtUserInfo, OAuth2BearerToken). I think I draw the line at triple nesting (option, tuple, try), but maybe I'm too faint of heart.

wchamber (approved):

marctalbott (approved):

dvoet:
jenkins retest

adrazhi (approved):

PR #415, opened 2020-03-27
[CA-772] Don't pull in unused members for listAccessPoliciesForUser

marctalbott:
Ticket: [CA-772](https://broadworkbench.atlassian.net/browse/CA-772)
This PR removes the joins in the `listAccessPoliciesForUser` query that pull out membership info. Since the membership info is never used, this was a simple way to make the query faster and simpler. Rather than return an `AccessPolicy` object that has had its membership stripped out, this adds a new case class `AccessPolicyMetadata` that contains all the same info as an `AccessPolicy` but without any membership info.

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -692,14 +692,10 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
       val pr = PolicyRoleTable.syntax("pr")
       val rr = ResourceRoleTable.syntax("rr")
       val pa = PolicyActionTable.syntax("pa")
       val ra = ResourceActionTable.syntax("ra")
-      val sp = PolicyTable.syntax("sp")
-      val sr = ResourceTable.syntax("sr")
-      val srt = ResourceTypeTable.syntax("srt")
 
       val listPoliciesQuery =
         samsql"""with recursive ${ancestorGroupsTable.table}(${agColumn.parentGroupId}, ${agColumn.memberGroupId}) as (
```
marctalbott: The ticket calls for 

> Also the recursive CTE contains the column member_group_id which is not used, remove it

but I'm not sure how to remove the `member_group_id` column without changing how the recursion works, so I've left it alone for now. Looking for guidance on what exactly this requirement means.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -149,9 +149,13 @@ ldap but it felt unnatural.
     roles: Set[ResourceRoleName],
     actions: Set[ResourceAction],
     public: Boolean)
-    extends WorkbenchGroup
+    extends WorkbenchGroup {
+  def metadata: AccessPolicyMetadata = AccessPolicyMetadata(id, email, roles, actions, public)
```
wchamber: Is it time to update the comment on AccessPolicy? We're not constrained by ldap anymore. Or is it useful to still say that's where it comes from?

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -149,9 +149,13 @@ ldap but it felt unnatural.
...
+}
 @Lenses final case class AccessPolicyMembership(memberEmails: Set[WorkbenchEmail], actions: Set[ResourceAction], roles: Set[ResourceRoleName])
 @Lenses final case class AccessPolicyResponseEntry(policyName: AccessPolicyName, policy: AccessPolicyMembership, email: WorkbenchEmail)
+// Access Policy with no membership info
```
wchamber: nit: Maybe say why this is useful to have? does it model the sql table more closely, do clients care less?

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
wchamber: Is it time to update the comment on AccessPolicy above?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
wchamber: the member_group_id column defined by the ancestorsGroupsTable isn't needed. It will make us no longer duplicate recurse. Here, we can just remove the ${agColumn.memberGroupId} from the top this line, and the ${gm/pg.memberGroupId} on lines 202 & 706.
The CTE is selecting the set of unique group id & member group ids i.e. the parent group id and its child. We don't actually care which child group id pulled in the parent group id - we know we're starting from our base user member. As-is, it's selecting extra data about what the child group id is, doing more work and possibly reprocessing the same group on multiple recursions if it's added from multiple children.

The more general change is looking at the SubGroupMemberTable definition and seeing the other places that don't need the memberchild id, but that's out of scope of the ticket. 

Probably easiest to talk it over in a chat/call than through review comments.
I also think it's fine to leave it for a follow on PR of editting the file here, but I'm not picky.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -205,11 +205,12 @@ class PolicyEvaluatorService(
       policies.map(_.groupName)
     }
 
-  def listResourceAccessPoliciesForUser(resource: FullyQualifiedResourceId, userId: WorkbenchUserId, parentSpan: Span = null): IO[Set[AccessPolicy]] =
+  def listResourceAccessPoliciesForUser(resource: FullyQualifiedResourceId, userId: WorkbenchUserId, parentSpan: Span = null): IO[Set[AccessPolicyMetadata]] =
     for {
       policies <- traceIOWithParent("listAccessPoliciesForUser", parentSpan)(_ => accessPolicyDAO.listAccessPoliciesForUser(resource, userId))
       publicPolicies <- traceIOWithParent("listPublicAccessPolicies", parentSpan)(_ => accessPolicyDAO.listPublicAccessPolicies(resource))
```
wchamber, line 211: Is there an equivalent change for listPublicAccessPolicies that we should do? Should it also return a AccessPolicyMetadata and avoid the joins? I think this is it's only use.

wchamber (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -149,9 +149,13 @@ ldap but it felt unnatural.
...
+@Lenses final case class AccessPolicyMetadata(id: FullyQualifiedPolicyId, email: WorkbenchEmail, roles: Set[ResourceRoleName], actions: Set[ResourceAction], public: Boolean)
```
dvoet: I don't really like the term metadata here. How about `AccessPolicyWithoutMembers`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/AccessPolicyDAO.scala
```diff
@@ -23,12 +23,12 @@ trait AccessPolicyDAO {
   def overwritePolicyMembers(id: FullyQualifiedPolicyId, memberList: Set[WorkbenchSubject]): IO[Unit]
   def overwritePolicy(newPolicy: AccessPolicy): IO[AccessPolicy]
   def listPublicAccessPolicies(resourceTypeName: ResourceTypeName): IO[Stream[ResourceIdAndPolicyName]]
-  def listPublicAccessPolicies(resource: FullyQualifiedResourceId): IO[Stream[AccessPolicy]]
+  def listPublicAccessPoliciesWithoutMembers(resource: FullyQualifiedResourceId): IO[Stream[AccessPolicyWithoutMembers]]
```
wchamber: nit: no strong opinion, but this name is no longer symmetric with listAccessPoliciesForUser. It's a little funny that one but not both of them has WithoutMembers.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -205,12 +205,11 @@ class PolicyEvaluatorService(
...
     }
 
-  def listResourceAccessPoliciesForUser(resource: FullyQualifiedResourceId, userId: WorkbenchUserId, parentSpan: Span = null): IO[Set[AccessPolicyMetadata]] =
+  def listResourceAccessPoliciesForUser(resource: FullyQualifiedResourceId, userId: WorkbenchUserId, parentSpan: Span = null): IO[Set[AccessPolicyWithoutMembers]] =
     for {
       policies <- traceIOWithParent("listAccessPoliciesForUser", parentSpan)(_ => accessPolicyDAO.listAccessPoliciesForUser(resource, userId))
-      publicPolicies <- traceIOWithParent("listPublicAccessPolicies", parentSpan)(_ => accessPolicyDAO.listPublicAccessPolicies(resource))
-      publicPoliciesMetadata = publicPolicies.map(_.metadata)
-    } yield policies ++ publicPoliciesMetadata
+      publicPolicies <- traceIOWithParent("listPublicAccessPolicies", parentSpan)(_ => accessPolicyDAO.listPublicAccessPoliciesWithoutMembers(resource))
```
wchamber: nit: worth keeping the traceIOWithParent "methodName" identical to the method name being called?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -538,11 +538,46 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+
+        val (roles, actions) = roleActionResults.unzip
+
+        AccessPolicyWithoutMembers(FullyQualifiedPolicyId(FullyQualifiedResourceId(policyInfo.resourceTypeName, policyInfo.resourceId), policyInfo.name), policyInfo.email, roles.flatten.toSet, actions.flatten.toSet, policyInfo.public)
+      }.toStream
+    }
   }
 
-  private def listPolicies(resource: FullyQualifiedResourceId, publicOnly: Boolean = false, limitOnePolicy: Option[AccessPolicyName] = None): IO[Stream[AccessPolicy]] = {
+  private def listPolicies(resource: FullyQualifiedResourceId, limitOnePolicy: Option[AccessPolicyName] = None): IO[Stream[AccessPolicy]] = {
```
wchamber, line 581: nit: Could we improve this method name? This method name (and lack of comment) doesn't really tell me when I should use this method and when I shouldn't. This class has several methods that 'listAccessPolicies' but only a couple of them use this method.
There's one listAccessPoliciesForUser, two listAccessPolicies, and two listPublicAccessPolicies. However, this method is now used from 'loadPolicy' and one of the 'listAccessPolicies'.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
wchamber: There's another recursive query with the same thing going on in this file. Now line 689 I think. Same, we can remove the memberGroupId. Mind touching that one as well?

wchamber (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/AccessPolicyDAO.scala
dvoet: I think remove without members from function name, return type is explicit

dvoet (approved):

PR #416, opened 2020-03-31
[CA-727] Disable users in ldap when deleting them

marctalbott:
Ticket: [CA-727](https://broadworkbench.atlassian.net/browse/CA-727)
Currently, deleting a user simply deletes their record in ldap. That leaves a membership record in the `enabled-users` group that references a now non-existent user. Since ldap only keeps track of a single group, it is simple enough to clean a user out of the `enabled-users` group when we delete the user. This will also help manage bloat in the `enabled-users` group which has shown to cause performance issues in the perf environment.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapRegistrationDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapRegistrationDAO.scala
@@ -62,6 +62,7 @@ class LdapRegistrationDAO(
       unmarshalUser(results).toOption
     }
 
+  // Deleting a user in ldap will also disable them to clear them out of the enabled-users group
   override def deleteUser(userId: WorkbenchUserId): IO[Unit] =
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/LdapRegistrationDAO.scala
```diff
@@ -63,7 +63,10 @@ class LdapRegistrationDAO(
     }
 
   override def deleteUser(userId: WorkbenchUserId): IO[Unit] =
-    executeLdap(IO(ldapConnectionPool.delete(userDn(userId))))
+    executeLdap(for {
```
yonghaoy, line 67: mind to add one line comment to call out this method will do 2 things, something like (1)remove from enable user group? (2)delete user from ldap?

yonghaoy (approved):

dvoet (approved):

PR #422, opened 2020-04-17
[IA-1583] Add persistent disk resource

akarukappadath:
Ticket: https://broadworkbench.atlassian.net/secure/RapidBoard.jspa?rapidView=35&modal=detail&selectedIssue=IA-1583

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -328,7 +328,7 @@ resourceTypes = {
         roleActions = ["status", "connect", "sync", "delete", "read_policies", "stop_start", "modify"]
       }
     }
-    reuseIds = true
+    reuseIds = false
   }
   persistent-disk = {
     actionPatternObjects = {
@@ -354,7 +354,7 @@ resourceTypes = {
         roleActions = ["read", "attach", "modify", "delete", "read_policies"]
       }
     }
-    reuseIds = true
+    reuseIds = false
   }
   cloud-extension = {
     actionPatternObjects = {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -293,7 +293,7 @@ resourceTypes = {
         roleActions = ["launch_batch_compute"]
       }
       notebook-user = {
-        roleActions = ["launch_notebook_cluster"]
+        roleActions = ["launch_notebook_cluster", "create_persistent_disk"]
       }
     }
     reuseIds = true
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -253,6 +253,15 @@ resourceTypes = {
       stop_start_notebook_cluster = {
         description = "stop and start notebook clusters in this billing-project"
       }
+      list_persistent_disk = {
+        description = "list all persistent disks in this billing-project"
+      }
+      create_persistent_disk = {
+        description = "create persistent disk in this billing-project"
+      }
+      delete_persistent_disk = {
+        description = "delete persistent disk in this billing-project"
+      }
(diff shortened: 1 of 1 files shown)


Qi77Qi (approved):
lgtm

src/main/resources/reference.conf
```diff
@@ -253,6 +253,18 @@ resourceTypes = {
       stop_start_notebook_cluster = {
         description = "stop and start notebook clusters in this billing-project"
       }
+      list_persistent_disk = {
+        description = "list all persistent disks in this billing-project"
+      }
+      create_persistent_disk = {
+        description = "create persistent disk in this billing-project"
+      }
+      modify_persistent_disk = {
```
rtitle: I'm not sure we need "modify" at the project level -- I don't see project owners with the ability to modify (e.g. resize) all disks in the project. Just list/create/delete seems sufficient.

rtitle:
One comment above, overall looks good.

Some context on this for reviewers: we're adding PDs as a new Leo-managed resource type, alongside runtimes. Disks and runtimes have a separate lifecycle: e.g. they can be created and deleted independently. A user can specify a disk to attach at runtime creation time -- this will check the `attach` action on the disk resource. A disk will be permissioned to a single user: there will be no sharing of disks between users, similar to runtimes.

Someone from Sam should review this to make sure we know what we're doing :)

src/main/resources/reference.conf
```diff
@@ -321,6 +330,32 @@ resourceTypes = {
...
+        description = "view all policies and policy details for the persistent disk"
+      }
+    }
+    ownerRoleName = "creator"
+    roles = {
+      creator = {
+        roleActions = ["read", "attach", "modify", "delete", "read_policies"]
+      }
+    }
+    reuseIds = true
```
dvoet: I don't think you want `reuseIds = true`. This would allow someone to delete the resource in sam then allow another to recreate the resource with the same id and take control of the persistent disk

akarukappadath: Gotcha, thanks. The `notebook-cluster` resource has `reuseIds = true`, is that something I should change as well while I'm here then?

rtitle: fwiw Leo now uses UUIDs for sam resource IDs. So I think we could remove `reuseIds = true` for `notebook-cluster` as well?

dvoet: if you are using UUIDs then `reuseIds=false` is recommended as you never intend to reuse anyway

dvoet (approved):

akarukappadath:
jenkins retest

src/main/resources/reference.conf
```diff
@@ -275,7 +284,7 @@ resourceTypes = {
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["view_status", "create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "list_notebook_cluster", "launch_notebook_cluster", "sync_notebook_cluster", "delete_notebook_cluster", "alter_google_role", "add_to_service_perimeter", "delete"]
+        roleActions = ["view_status", "create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "list_notebook_cluster", "launch_notebook_cluster", "sync_notebook_cluster", "delete_notebook_cluster", "list_persistent_disk", "create_persistent_disk", "delete_persistent_disk", "alter_google_role", "add_to_service_perimeter", "delete"]
```
dvoet, line 287: so only project owners can create a persistent disk?

PR #423, opened 2020-04-21
remove psp

gmalkov:
Moving PSP definitions:
https://github.com/broadinstitute/terra-helm/pull/21


wchamber (commented):
We also have to modify role.yaml, as it refer's to the removed sam-psp now.
Does it make sense to also pull a common role up into helm and leave the role-binding in these services?

gmalkov:
@wchamber no, we do not have to modify any other resources, because they are not cluster-wide and need to be created in each namespace. They can still reference the PSP, it will just be created in Helm. `Role` and `RoleBinding` are namespace-specific. If we used `ClusterRole` and `ClusterRoleBinding` then yes we would pull those out also. I think I like this pattern of per-env bindings though.

wchamber (approved):

PR #427, opened 2020-04-29
add endpoint to check user's access on a action

yonghaoy:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -86,7 +86,7 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
                   } ~ pathPrefix("userEmail") {
                     pathPrefix(Segment) { userEmail =>
                       pathEndOrSingleSlash {
-                        getActionPermissionForUserEmail(resource, userInfo, action, userEmail)
+                        getActionPermissionForUserEmail(resource, userInfo, ResourceAction(action), WorkbenchEmail(userEmail))
                       }
                     }
                   }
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -205,6 +211,22 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
       }
     }
 
+  /**
+    * Checks if user has permission by giver user email.
+    *
+    * <p> The caller should have readPolicies, OR testAnyActionAccess or testActionAccess::{action} to make this call.
+    */
+  def getActionPermissionForUserEmail(resource: FullyQualifiedResourceId, userInfo: UserInfo, action: String, userEmail: String): server.Route =
```
marctalbott: It might be nice to wrap the `action` String in a `ResourceAction` and the `userEmail` String in a `WorkbenchEmail` before calling this function just so we aren't passing Strings around

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
```diff
@@ -27,6 +27,10 @@ class PolicyEvaluatorServiceSpec extends FlatSpec with Matchers with TestSupport
   lazy val policyDAO: AccessPolicyDAO = new PostgresAccessPolicyDAO(TestSupport.dbRef, TestSupport.blockingEc)
   val schemaDao = new JndiSchemaDAO(directoryConfig, schemaLockConfig)
 
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+  }
+
```
dvoet: is this required

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
```diff
@@ -217,6 +221,56 @@ class PolicyEvaluatorServiceSpec extends FlatSpec with Matchers with TestSupport
     res.unsafeRunSync()
   }
 
+  "hasPermissionByUserEmail" should "return false if given action is not allowed for a user" in {
```
dvoet: 2 comments
there should be a test for when the user does have access

the structure of this test file is like
```
"function name" should "test case" in {
...
}

it should "another test case" in {
...
}
```
Notice that the second test case starts with `it`. It inherits the name from the prior test case. Where you have inserted your tests interrupts the flow of the other tests. I would move these down to line 480, the first test should be `"hasPermissionByUserEmail"` and the following should be `it`

marctalbott: Is there any value to having a positive test here?

marctalbott (approved):

dvoet (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
yonghaoy: Thanks! That's good to know

yonghaoy: Right! Thanks for the suggestion!  I added one positive test, as all of them are the same as hasPermission() and all scenario are coverd there. My new method just call that one.

yonghaoy:
jenkins retest

dvoet (approved):

PR #435, opened 2020-06-01
[CA-866] prevent googleSubjectIds from being overwritten in Postgres

marctalbott:
Ticket: [CA-866](https://broadworkbench.atlassian.net/browse/CA-866)
OpenDJ already prevents us from overwriting an existing GoogleSubjectId, so this PR changes Postgres to do the same thing as OpenDJ and also prevent existing GoogleSubjectIds from being overwritten. See the ticket for a description of what can happen when this situation arises and Postgres _doesn't_ prevent the overwrite.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
@@ -839,7 +839,7 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
                 where ${u.id} = ${userId} and ${u.googleSubjectId} is null"""
 
       if (updateGoogleSubjectIdQuery.update().apply() != 1) {
-        throw new WorkbenchException(s"Cannot overwrite an existing googleSubjectId for user ${userId}")
+        throw new WorkbenchException(s"Cannot update googleSubjectId for user ${userId} because user does not exist or the googleSubjectId has already been set for this user")
       }
     })
   }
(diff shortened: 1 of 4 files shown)


dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
```diff
@@ -834,7 +834,13 @@ class PostgresDirectoryDAO(protected val dbRef: DbReference,
...
     runInTransaction("setGoogleSubjectId", samRequestContext)({ implicit session =>
       val u = UserTable.column
-      samsql"update ${UserTable.table} set ${u.googleSubjectId} = ${googleSubjectId} where ${u.id} = ${userId}".update().apply()
+      val updateGoogleSubjectIdQuery =
+        samsql"""update ${UserTable.table} set ${u.googleSubjectId} = ${googleSubjectId}
+                where ${u.id} = ${userId} and ${u.googleSubjectId} is null"""
+
+      if (updateGoogleSubjectIdQuery.update().apply() != 1) {
+        throw new WorkbenchException(s"Cannot overwrite an existing googleSubjectId for user ${userId}")
+      }
```
andy7i: Does `updateGoogleSubjectIdQuery.update().apply() ` return a count? If so, is there a realistic chance that the query count is 0 because the id doesn't exist for some reason?

andy7i (approved):
1 Q & 👍

src/main/scala/org/broadinstitute/dsde/workbench/sam/directory/PostgresDirectoryDAO.scala
marctalbott: Yeah, `updateGoogleSubjectIdQuery.update().apply()` returns the count of rows that were updated by the query. So yes, it is entirely possible that the query count is 0 because the id doesn't exist for some reason. The service layer prevents this in practice, but a more accurate exception message here would be "Cannot update googleSubjectId for user ${userId} because user does not exist or the googleSubjectId has already been set". I'll update accordingly

marctalbott: And while we're considering other values for that count, I'm not going to worry too much about `updateGoogleSubjectIdQuery.update().apply()` returning more than 1 since Postgres will enforce <=1 record for a given user Id

marctalbott:
jenkins retest

PR #439, opened 2020-06-26
blacklist new quicklab email accounts

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/CA-920

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -723,4 +723,4 @@ distributedLock {
   maxRetry: 10
 }
 
-domainBlacklist = ["qwiklabs-gsuite.net", "qwiklabs.net"]
+blockedEmailDomains = ["qwiklabs-gsuite.net", "qwiklabs.net"]
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -722,3 +722,5 @@ distributedLock {
   retryInterval: 5s
   maxRetry: 10
 }
+
+domainBlacklist = ["qwiklabs-gsuite.net", "qwiklabs.net"]
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -20,10 +20,11 @@ import scala.util.matching.Regex
 /**
   * Created by dvoet on 7/14/17.
   */
-class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExtensions, val registrationDAO: RegistrationDAO)(implicit val executionContext: ExecutionContext, contextShift: ContextShift[IO]) extends LazyLogging {
+class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExtensions, val registrationDAO: RegistrationDAO, domainBlacklist: Seq[String])(implicit val executionContext: ExecutionContext, contextShift: ContextShift[IO]) extends LazyLogging {
 
   def createUser(user: CreateWorkbenchUser, samRequestContext: SamRequestContext): Future[UserStatus] = {
     for {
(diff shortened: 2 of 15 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -223,8 +224,9 @@ object UserService {
   def genWorkbenchUserId(currentMilli: Long): WorkbenchUserId =
     WorkbenchUserId(genRandom(currentMilli))
 
-  def validateEmailAddress(email: WorkbenchEmail): IO[Unit] = {
+  def validateEmailAddress(email: WorkbenchEmail, domainBlacklist: Seq[String]): IO[Unit] = {
     email.value match {
+      case emailString if domainBlacklist.exists(domain => emailString.endsWith("@" + domain)) => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"email domain not permitted [${email.value}]")))
```
rtitle: I'm not sure the `qwiklabs-gsuite.net` domain blacklist would match this because the emails are of the format:
```
student@goog-test.reseller.gappslabs.co.s-ixlmdsbu.qwiklabs-gsuite.net
```

yonghaoy (approved):

src/main/resources/reference.conf
```diff
@@ -722,3 +722,5 @@ distributedLock {
   retryInterval: 5s
   maxRetry: 10
 }
+
+domainBlacklist = ["qwiklabs-gsuite.net", "qwiklabs.net"]
```
gpolumbo-broad: I'm torn as to whether this information should be in `reference.conf` or `sam.conf`.  I always just thought of `reference.conf` as kind of `resource_types.conf`.  I see now that there are some other settings in there, so I think the question is, what are the guidelines for determining which settings go into `reference.conf` and which go into `sam.conf`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -223,8 +224,9 @@ object UserService {
   def genWorkbenchUserId(currentMilli: Long): WorkbenchUserId =
     WorkbenchUserId(genRandom(currentMilli))
 
-  def validateEmailAddress(email: WorkbenchEmail): IO[Unit] = {
+  def validateEmailAddress(email: WorkbenchEmail, domainBlacklist: Seq[String]): IO[Unit] = {
```
gpolumbo-broad: Is there value in passing the `domainBlacklist` around as an argument to this method?  As far as I understand, the field `UserService.domainBlacklist`will not change on they fly, and so I don't see the value in passing it in here every time it is called.

src/main/resources/reference.conf
gpolumbo-broad:
```suggestion
domainBlocklist = ["qwiklabs-gsuite.net", "qwiklabs.net"]
```
And replace all variables as well, `blacklist` -> `blocklist`

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
dvoet: fixed

src/main/resources/reference.conf
dvoet: new term: blockedEmailDomains

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
dvoet: it is because this is a static function, it does not have access to config

gpolumbo-broad (approved):

rtitle (approved):
LGTM

dvoet:
jenkins retest

PR #441, opened 2020-07-07
[IA-1826] Kubernetes app resource

akarukappadath:
Ticket: https://broadworkbench.atlassian.net/browse/IA-1826
Design doc: https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/595492869/2020-07-07+Kubernetes+App+Sam+Model

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -376,6 +376,9 @@ resourceTypes = {
       start = {
         description = "start kubernetes application"
       }
+      connect = {
+        description = "connect to kubernetes application"
+      }
       update = {
         description = "update kubernetes application"
       }
@@ -383,7 +386,7 @@ resourceTypes = {
     ownerRoleName = "creator"
     roles = {
       creator = {
-        roleActions = ["delete", "stop", "start", "update"]
+        roleActions = ["delete", "stop", "start", "connect", "update"]
       }
       user = {
         roleActions = ["stop", "start"]
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -364,12 +364,6 @@ resourceTypes = {
       delete = {
         description = "delete kubernetes application"
       }
-      stop = {
-        description = "stop kubernetes application"
-      }
-      start = {
-        description = "start kubernetes application"
-      }
       connect = {
         description = "connect to kubernetes application"
       }
@@ -380,7 +374,7 @@ resourceTypes = {
     ownerRoleName = "creator"
     roles = {
       creator = {
-        roleActions = ["delete", "stop", "start", "connect", "update"]
+        roleActions = ["delete", "connect", "update"]
       }
     }
     reuseIds = false
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -259,14 +259,8 @@ resourceTypes = {
       delete_persistent_disk = {
         description = "delete persistent disk in this billing-project"
       }
-      list_kubernetes_app = {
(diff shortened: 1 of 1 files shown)


akarukappadath:
jenkins retest

src/main/resources/reference.conf
```diff
@@ -356,6 +365,32 @@ resourceTypes = {
     }
     reuseIds = false
   }
+  kubernetes-app = {
```
rtitle, line 356: Commented in the design doc as well, but I'm thinking we can:
* drop the `user` role -- there's no distinction between app creator and user; only the creator can use the app (similar to notebook runtimes).
* remove the `stop` and `start` actions -- Leo won't expose this functionality for k8s
* add a `connect` action -- only the app creator should be able to connect to the app, whereas the project owner can delete but not connect.

Let me know if I'm misunderstanding something.

src/main/resources/reference.conf
```diff
@@ -306,9 +312,6 @@ resourceTypes = {
       connect = {
         description = "connect to the Jupyter notebook running on the notebook cluster"
       }
-      sync = {
```
rtitle, line 309: Context: we're not using this action; `connect` covers data syncing as well

rtitle (approved):
This LGTM as we've discussed -- Cloud Integration team should review as well though.

src/main/resources/reference.conf
```diff
@@ -262,6 +259,15 @@ resourceTypes = {
...
         description = "delete persistent disk in this billing-project"
       }
+      list_kubernetes_app = {
+        description = "list all kubernetes applications in this billing-project"
+      }
+      create_kubernetes_app = {
+        description = "create kubernetes application in this billing-project"
+      }
+      delete_kubernetes_app = {
+        description = "delete kubernetes application in this billing-project"
```
dvoet: this is delete _any_ k8s app in the billing-project, correct? But the billing-project owner won't have delete access on the kubernetes-app resource so they won't be able to remove it from sam.

src/main/resources/reference.conf
dvoet: where is the design doc?

akarukappadath: https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/595492869/2020-07-07+Kubernetes+App+Sam+Model

dvoet (commented):
I feel like the only `billing-project` level action should be `create_kubernetes_app`. The `kubernetes_app` resource type should have another role, `manager`, which would be able to delete the app. A list action is not required, just list all the `kubernetes_app`s a user has access to. At the policy level, `kubernetes_app` resources should be created with a creator policy _and_ a manager policy whose member is the email address of the billing-project owner policy. This way we do not infer access to one resource based on access to another.

rtitle:
Interesting. That makes sense as a way to model this. Our other Leo resources (like runtimes, disks) also follow the pattern in this PR. I wonder if we should consider migrating those too (probably separately from the k8s work).

akarukappadath:
jenkins retest

dvoet (approved):

rtitle:
@dvoet, question about this:
> At the policy level, kubernetes_app resources should be created with a creator policy and a manager policy whose member is the email address of the billing-project owner policy.

It just occurred to me, if the app is being created by a user who is NOT the billing-project owner, how can we get the billing-project owner policy email? The calling user does not have permission to look it up in Sam. Can the Leo SA be used to look up arbitrary billing-project policies?

PR #442, opened 2020-07-20
[CA-921][CA-931] Add support for parent resources

marctalbott:
Ticket: [CA-921](https://broadworkbench.atlassian.net/browse/CA-921), [CA-931](https://broadworkbench.atlassian.net/browse/CA-931)
Add 3 endpoints to get, set and remove the parent of a resource. There is now a nullable `resource_parent_id` column on the `sam_resource` table which can be accessed by these endpoints.

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1631,7 +1631,7 @@ paths:
           content:
             application/json:
               schema:
-                type: '#/components/schemas/FullyQualifiedResourceId'
+                $ref: '#/components/schemas/FullyQualifiedResourceId'
         403:
           description: You do not have permission to get the parent for this resource
           content: {}
@@ -2045,6 +2045,7 @@ components:
         resourceId:
           type: string
           description: The id of the resource
+      description: type and id of a resource
     ManagedGroupMembershipEntry:
       required:
         - groupEmail
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1606,6 +1606,121 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/resources/v2/{resourceTypeName}/{resourceId}/parent:
+    get:
+      tags:
+        - Resources
+      summary: Get the parent of a resource if one exists
+      operationId: getResourceParent
+      parameters:
+        - name: resourceTypeName
+          in: path
+          description: Type of resource to query
+          required: true
+          schema:
+            type: string
+        - name: resourceId
+          in: path
+          description: Id of resource to query
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: Id and type of parent resource
+          content:
+            application/json:
+              schema:
+                type: '#/components/schemas/FullyQualifiedResourceId'
+        403:
+          description: You do not have permission to get the parent for this resource
+          content: {}
+        404:
+          description: You do not have access to this resource, it does not exist, or it does not have a parent resource
+          content: {}
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+    put:
+      tags:
+        - Resources
+      summary: Set the parent of a resource
+      operationId: setResourceParent
+      parameters:
+        - name: resourceTypeName
+          in: path
+          description: Type of child resource
+          required: true
+          schema:
+            type: string
+        - name: resourceId
+          in: path
+          description: Id of child resource
+          required: true
+          schema:
+            type: string
+      requestBody:
+        description: The details of the new parent resource
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/FullyQualifiedResourceId'
+      responses:
+        204:
+          description: Resource parent successfully set
+          content: {}
+        400:
+          description: Child resource has an auth domain or new parent would introduce cyclical resource hierarchy
+          content: {}
+        403:
+          description: You do not have permission to set the parent for the child resource, to remove the child of the current parent, or to add a child to the new parent
+          content: {}
+        404:
+          description: You do not have access to the child, new parent, or current resource, or they do not exist
+          content: {}
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+    delete:
+      tags:
+        - Resources
+      summary: Delete the parent of a resource
+      operationId: deleteResourceParent
+      parameters:
+        - name: resourceTypeName
+          in: path
+          description: Type of resource to remove parent from
+          required: true
+          schema:
+            type: string
+        - name: resourceId
+          in: path
+          description: Id of resource to remove parent from
+          required: true
+          schema:
+            type: string
+      responses:
+        204:
+          description: Resource parent successfully removed
+          content: {}
+        403:
+          description: You do not have permission to set the parent for the child resource or to remove the child of the current parent
+          content: {}
+        404:
+          description: You do not have access to this resource, it does not exist, or it does not currently have a parent resource
+          content: {}
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
   /api/users/v1/{email}:
     get:
       tags:
@@ -1918,6 +2033,18 @@ components:
           items:
             $ref: '#/components/schemas/StackTraceElement'
       description: ""
+    FullyQualifiedResourceId:
+      required:
+        - resourceTypeName
+        - resourceId
+      type: object
+      properties:
+        resourceTypeName:
+          type: string
+          description: The type of the resource
+        resourceId:
+          type: string
+          description: The id of the resource
     ManagedGroupMembershipEntry:
       required:
         - groupEmail
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -5,7 +5,7 @@ import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.marshalling.ToResponseMarshallable
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server
-import akka.http.scaladsl.server.{Directive1, Route}
+import akka.http.scaladsl.server.Directive1
 import akka.http.scaladsl.server.Directives._
 import cats.effect.IO
 import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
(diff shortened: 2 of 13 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -78,6 +78,10 @@ object SamResourceActions {
   val setPublic = ResourceAction("set_public")
   val readAuthDomain = ResourceAction("read_auth_domain")
   val testAnyActionAccess = ResourceAction("test_any_action_access")
+  val getParent = ResourceAction("get_parent")
```
marctalbott, line 81: Outstanding question: Do I need to update reference.conf to add these actions to the various resource types listed there?

src/test/scala/org/broadinstitute/dsde/workbench/sam/openam/MockAccessPolicyDAO.scala
```diff
@@ -146,4 +146,30 @@ class MockAccessPolicyDAO(private val policies: mutable.Map[WorkbenchGroupIdenti
       }.toStream
     )
   }
+
+  // Each resource has a list of
+  val resourceParents = TrieMap[FullyQualifiedResourceId, FullyQualifiedResourceId]()
```
marctalbott: I ended up not using this in any of the tests... removing

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
```diff
@@ -0,0 +1,406 @@
+package org.broadinstitute.dsde.workbench.sam.api
```
marctalbott, line 1: This test suite was a bit of a pain to set up. I wanted to mock out a lot of stuff and the option I went with was Mockito. Mockito took a good amount of fiddling to get it working properly and without _too_ much boiler plate, but I'm not entirely sure how clear the `setupRoutesTest` function is to anyone except me, so let me know if you think this is an okay way of mocking these tests or if you'd prefer something else.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -783,6 +783,95 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+      import SamTypeBinders._
+      query.map(rs =>
+          FullyQualifiedResourceId(
+            rs.get[ResourceTypeName](prt.resultName.name),
+            rs.get[ResourceId](pr.resultName.name)))
+        .single.apply()
+    })
+  }
+
+  override def setResourceParent(childResource: FullyQualifiedResourceId, parentResource: FullyQualifiedResourceId, samRequestContext: SamRequestContext): IO[Unit] = {
```
marctalbott, line 813: reworking this query a bit as I realized it allows for a resource to be the parent of itself oops

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -296,4 +316,49 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
...
+
+  private def setResourceParentInternal(resource: FullyQualifiedResourceId, newResourceParent: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route = {
+    requireAction(newResourceParent, SamResourceActions.addChild, userInfo.userId, samRequestContext) {
+      requireAction(resource, SamResourceActions.setParent, userInfo.userId, samRequestContext) {
+        complete(resourceService.setResourceParent(resource, newResourceParent, samRequestContext).map(_ => StatusCodes.NoContent))
+      }
+    }
+  }
+
+  def deleteResourceParent(resource: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route =
```
marctalbott, line 351: This just removes relationship between child and parent, it doesn't remove either resource from the database

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -783,6 +783,95 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+    val ar = ancestorResourceTable.syntax("ar")
+    val arColumn = ancestorResourceTable.column
+
+    val r = ResourceTable.syntax("r")
+    val rt = ResourceTypeTable.syntax("rt")
+    val pr = ResourceTable.syntax("pr")
+    val prt = ResourceTypeTable.syntax("prt")
+    val resourceTableColumn = ResourceTable.column
+
+    val query =
```
marctalbott, line 824: todo: leave a comment explaining what's happening

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -478,4 +478,33 @@ class ResourceService(
     } yield {
       workbenchUsers.map(user => UserIdInfo(user.id, user.email, user.googleSubjectId))
     }
+
+  def getResourceParent(resourceId: FullyQualifiedResourceId, samRequestContext: SamRequestContext): IO[Option[FullyQualifiedResourceId]] = {
+    accessPolicyDAO.getResourceParent(resourceId, samRequestContext)
+  }
+
+  def setResourceParent(childResource: FullyQualifiedResourceId, parentResource: FullyQualifiedResourceId,  samRequestContext: SamRequestContext): IO[Unit] = {
+    for {
```
marctalbott, line 490: todo: link or something explaining _why_ we don't allow auth domains on the child at the moment

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
dvoet: not yet... we don't have any resource types that allow parents are children yet

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
```diff
@@ -0,0 +1,406 @@
...
+    val samRoutes = createSamRoutes()
+
+    setupRoutesTest(samRoutes, fullyQualifiedChildResource, None, actionsOnChild = Set(SamResourceActions.getParent))
+
+    Get(s"/api/resources/v2/${defaultResourceType.name}/${fullyQualifiedChildResource.resourceId.value}/parent") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.NotFound
+    }
+  }
+
+  // this one feels a bit like it's testing the security directives not the route itself but whatever
```
dvoet: think of Route specs as contract tests... we need to make sure that we return responses with the right codes and format

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -296,4 +316,49 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
...
+
+  def setResourceParent(resource: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route =
+    put {
+      entity(as[FullyQualifiedResourceId]) { newResourceParent =>
+        withCurrentParentOption(resource, samRequestContext) {
+          case Some(currentResourceParent) =>
+            requireAction(currentResourceParent, SamResourceActions.removeChild, userInfo.userId, samRequestContext) {
+              setResourceParentInternal(resource, newResourceParent, userInfo, samRequestContext)
+            }
+          case None => setResourceParentInternal(resource, newResourceParent, userInfo, samRequestContext)
```
andy7i, line 338: the last command in each `case` is the same -- would it make sense to write it once, after the `case` statements?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -296,4 +316,49 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
...
+        }
+      }
+    }
+
+  private def setResourceParentInternal(resource: FullyQualifiedResourceId, newResourceParent: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route = {
+    requireAction(newResourceParent, SamResourceActions.addChild, userInfo.userId, samRequestContext) {
+      requireAction(resource, SamResourceActions.setParent, userInfo.userId, samRequestContext) {
```
andy7i, line 345: does anything currently have permission to do these ResourceActions? And if the answer is no is that because right now there is nothing that has parent/child relationships?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
andy7i: the recursive check for cyclical reference is really cool! +1 for greg's suggestion about adding a comment here

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
marctalbott: @dvoet so do you think I should change the tests to not worry about missing specific permissions on a resource and instead to just test that when some permission is missing it returns a 403?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1606,6 +1606,121 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/resources/v2/{resourceTypeName}/{resourceId}/parent:
```
andy7i, line 1609: Do we also need to add parent/child info to getWorkspace?

andy7i (approved):
A couple questions but otherwise 👍

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
marctalbott: I'm not sure if that would work given that the first case statement has `setResourceParentInternal` nested inside of `requireAction` while the second case statement doesn't have that `requireAction` nesting. I can see if something like that would work

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
marctalbott: Nope, nothing currently has permission to do these ResourceActions. I think we will hold off on granting these actions to resource types until we are ready to release the full support for hierarchical resources

src/main/resources/swagger/api-docs.yaml
marctalbott: Hmm, good question. I think that parent/child info from Sam could eventually be bubbled up to the Rawls `getWorkspace` endpoint once Rawls has been updated to use hierarchical resources in Sam. I'm not sure about that though, it will depend on how Rawls ends up using them. At the moment, I could imagine hierarchical resources in Sam being used to track a workspace's namespace as in the parent of all `workspace` resources in Sam are some `billing-project` resource. However, I think Rawls would keep track of that parent/child relationship itself as it already tracks workspace namespaces in its own database. Sam's parent/child knowledge in that case would just be used to verify permissions, I don't think Rawls would necessarily ask Sam to find out which resource is the parent of a workspace. All that said, the `getWorkspace` endpoint in Rawls will eventually be a passthrough to Workspace Manager (I think) and I have no idea how it'll be implemented there.

andy7i: ok -- thanks for the detailed explanation!

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
marctalbott: Was not able to get something to work. Leaving it as is

PR #443, opened 2020-07-22
[CA-925] add listResourceChildren endpoint

marctalbott:
Ticket: [CA-925](https://broadworkbench.atlassian.net/browse/CA-925)
- Add endpoint to Sam to list direct children of a resource

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
@@ -38,7 +38,7 @@ object SamJsonSupport {
 
   implicit val ResourceIdFormat = ValueObjectFormat(ResourceId.apply)
 
-  implicit val ResourceIdentityFormat = jsonFormat2(FullyQualifiedResourceId.apply)
+  implicit val FullyQualifiedResourceIdFormat = jsonFormat2(FullyQualifiedResourceId.apply)
 
   implicit val AccessPolicyMembershipFormat = jsonFormat3(AccessPolicyMembership.apply)
 
@@ -82,6 +82,7 @@ object SamResourceActions {
   val setParent = ResourceAction("set_parent")
   val addChild = ResourceAction("add_child")
   val removeChild = ResourceAction("remove_child")
+  val listChildren = ResourceAction("list_children")
 
   def sharePolicy(policy: AccessPolicyName) = ResourceAction(s"share_policy::${policy.value}")
   def readPolicy(policy: AccessPolicyName) = ResourceAction(s"read_policy::${policy.value}")
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
@@ -7,12 +7,12 @@ import cats.data.NonEmptyList
 import cats.effect.{ContextShift, IO}
 import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
-import org.broadinstitute.dsde.workbench.sam.errorReportSource
 import org.broadinstitute.dsde.workbench.model._
-import org.broadinstitute.dsde.workbench.sam.db.{DbReference, PSQLStateExtensions, SamTypeBinders}
 import org.broadinstitute.dsde.workbench.sam.db.SamParameterBinderFactory._
 import org.broadinstitute.dsde.workbench.sam.db.dao.{PostgresGroupDAO, SubGroupMemberTable}
 import org.broadinstitute.dsde.workbench.sam.db.tables._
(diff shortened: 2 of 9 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -38,7 +38,7 @@ object SamJsonSupport {
 
   implicit val ResourceIdFormat = ValueObjectFormat(ResourceId.apply)
 
-  implicit val ResourceIdentityFormat = jsonFormat2(FullyQualifiedResourceId.apply)
+  implicit val FullyQualifiedResourceIdFormat = jsonFormat2(FullyQualifiedResourceId.apply)
```
marctalbott, line 41: Unrelated to rest of PR, just noticed that this json formatter was named inconsistently

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -875,6 +875,31 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+    val cr = ResourceTable.syntax("cr")
+
+    val query =
+      samsql"""
+         select ${cr.result.name}, ${rt.result.name}
+         from ${ResourceTable as r}
+         join ${ResourceTable as cr} on ${cr.resourceParentId} = ${r.id}
+         join ${ResourceTypeTable as rt} on ${rt.id} = ${cr.resourceTypeId}
+         where ${r.name} = ${resource.resourceId}
+         and ${rt.name} = ${resource.resourceTypeName}"""
```
gpolumbo-broad, line 892: We only want to list child resources with the same Resource Type as the parent?  This surprised me when I read it.  It may be what we want, but it's not what I was expecting.  

If this is what we want, we may want to add a comment here and possibly update the swagger description of the api to explain that it will return only children of the same Resource Type.  

Also, if this is what we want, we should right a test where a resource has children with varied Resource Types.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -901,3 +926,9 @@ final case class AncestorResourceTable(override val tableName: String) extends S
   // need to specify column names explicitly because this table does not actually exist in the database
   override val columnNames: Seq[String] = Seq("resource_parent_id")
 }
+
+final case class ChildResourceRecord(childId: ResourcePK)
+final case class ChildResourceTable(override val tableName: String) extends SQLSyntaxSupport[ChildResourceRecord] {
```
gpolumbo-broad: Are these used anywhere?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
marctalbott: I don't think that's what we want, that would be an unintentional feature (bug) in my query. I'll add a test to make sure we aren't doing that and adjust the query

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
marctalbott: Nope! Actually just pushed up a commit to remove them. They were leftover from when I thought we wanted all descendants of a resource, not just direct children

src/test/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAOSpec.scala
```diff
@@ -931,5 +931,86 @@ class PostgresAccessPolicyDAOSpec extends FreeSpec with Matchers with BeforeAndA
...
+      }
+
+      "can list children of different resource types" in {
+        val parentResourceType = resourceType.copy(name = ResourceTypeName("parentRT"))
+        val childResourceType1 = resourceType.copy(name = ResourceTypeName("childRT1"))
+        val childResourceType2 = resourceType.copy(name = ResourceTypeName("childRT2"))
+
+        val parentResource = Resource(parentResourceType.name, ResourceId("parent"), Set.empty)
+        val childResource1 = Resource(childResourceType1.name, ResourceId("child1"), Set.empty)
+        val childResource2 = Resource(childResourceType1.name, ResourceId("child2"), Set.empty)
```
gpolumbo-broad: I think you meant:
```suggestion
        val childResource2 = Resource(childResourceType2.name, ResourceId("child2"), Set.empty)
```

gpolumbo-broad (approved):

dvoet (approved):

PR #444, opened 2020-07-27
IA-2051 Add actions to kubernetes-application resource type

rtitle:
Ticket: https://broadworkbench.atlassian.net/browse/IA-2051

Added a couple more actions to the `kubernetes-app` resource type:
- `status`: lets you view details about an app, but not actually connect to it. Leo checks this for the `getApp` endpoint. Granted to app creators and managers.
- `read_policies`: I don't think this is actually needed for Leo, but it was useful for me while testing this stuff.

By the way, the approach of having `creator` and `manager` policies at the resource level is working well: it results in cleaner Leo code and fewer Sam calls (don't need to always check things at the resource _and_ project levels). I think I'd like to migrate runtimes and disks to use this approach in the future, but I expect it will require some kind of Sam migration.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -364,14 +364,20 @@ resourceTypes = {
       update = {
         description = "update kubernetes application"
       }
+      status = {
+        description = "view details and configuration of the kubernetes application"
+      }
+      read_policies = {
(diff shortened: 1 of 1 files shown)


rtitle:
jenkins retest

gpolumbo-broad (approved):

PR #446, opened 2020-08-11
[CA-926] Delete policy

marctalbott:
Ticket: [CA-926](https://broadworkbench.atlassian.net/browse/CA-926)
* Add endpoint to Sam to delete policies as long as the resource does not have any children because inherited policies cannot be deleted

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -379,10 +379,8 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
 
   def deletePolicy(policyId: FullyQualifiedPolicyId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route =
     delete {
-      requireAction(policyId.resource, SamResourceActions.alterPolicies, userInfo.userId, samRequestContext) {
-        requireAction(policyId.resource, SamResourceActions.deletePolicy(policyId.accessPolicyName), userInfo.userId, samRequestContext) {
-          complete(resourceService.deletePolicy(policyId, samRequestContext).map(_ => StatusCodes.NoContent))
-        }
+      requireOneOfAction(policyId.resource, Set(SamResourceActions.alterPolicies, SamResourceActions.deletePolicy(policyId.accessPolicyName)), userInfo.userId, samRequestContext) {
+        complete(resourceService.deletePolicy(policyId, samRequestContext).map(_ => StatusCodes.NoContent))
       }
     }
 }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -144,6 +144,12 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
                 } ~
                 path("children") {
                   getResourceChildren(resource, userInfo, samRequestContext)
+                } ~
+                pathPrefix ("policies" / Segment) { policyName =>
+                  val policyId = FullyQualifiedPolicyId(resource, AccessPolicyName(policyName))
+                  pathEndOrSingleSlash {
+                    deletePolicy(policyId, userInfo, samRequestContext)
+                  }
                 }
               }
             }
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -370,4 +376,13 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
         complete(resourceService.listResourceChildren(resource, samRequestContext).map(children => StatusCodes.OK -> children))
       }
     }
+
+  def deletePolicy(policyId: FullyQualifiedPolicyId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route =
+    delete {
+      requireAction(policyId.resource, SamResourceActions.alterPolicies, userInfo.userId, samRequestContext) {
+        requireAction(policyId.resource, SamResourceActions.deletePolicy(policyId.accessPolicyName), userInfo.userId, samRequestContext) {
```
dvoet: I think this should be `requireOneOfAction`. Both actions are not required, only 1.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -214,6 +214,19 @@ class ResourceService(
       _ <- maybeDeleteResource(resource, samRequestContext)
     } yield ()
 
+  def deletePolicy(policyId: FullyQualifiedPolicyId, samRequestContext: SamRequestContext): IO[Unit] = {
+    for {
+      children <- accessPolicyDAO.listResourceChildren(policyId.resource, samRequestContext)
+      _ <- if (children.nonEmpty) {
+        IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "cannot delete inherited policies")))
```
dvoet: should raise this exception if there are children _and_ the policy applies to any of those children. As written if children exist no policies may be deleted.

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
marctalbott: Oh, I was under the impression that all policies on a resource apply to all of that resource's children. How can we tell if a policy applies to a child?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
marctalbott: Just to clarify, the wording in the ticket/design doc is 

> 403 if user has access to resource but is missing one of
> alter_policies
> delete_policy::{policyName}

which I interpreted as both being required. However, now that you mention it, it definitely makes sense to me that if you have either of these actions you'd be able to delete a policy. So is it simply an error in the ticket?

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -370,4 +376,11 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
         complete(resourceService.listResourceChildren(resource, samRequestContext).map(children => StatusCodes.OK -> children))
       }
     }
+
+  def deletePolicy(policyId: FullyQualifiedPolicyId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route =
+    delete {
+      requireOneOfAction(policyId.resource, Set(SamResourceActions.alterPolicies, SamResourceActions.deletePolicy(policyId.accessPolicyName)), userInfo.userId, samRequestContext) {
+        complete(resourceService.deletePolicy(policyId, samRequestContext).map(_ => StatusCodes.NoContent))
```
gpolumbo-broad, line 383: Just to be clear, for now, we are not checking if this policy applies to any children, even though we know that we should check that prior to deleting.  I'm OK with this if that's what was decided for now.  Do we have a ticket to come back and implement this later?  If so, I recommend putting a comment with a TODO here pointing to that ticket.  Either here or in `ResourceService.deletePolicy`.

gpolumbo-broad (changes requested):

marctalbott:
Got a verbal thumb from Greg during standup now that I added the TODO with ticket for additional inheritance work, so merging this

PR #448, opened 2020-08-19
[QA-723] add sleep between creating pet and adding it to user proxy group

marctalbott:
Google support provided this potential explanation of why we are seeing such slow permission propagation time
>What we suspect might have happened is that the Directory API created a separate, "unclaimed id" while trying to create membership for the SA inside the proxy group instead of using the ID that was created during SA creation. For these specific scenario, the account will be in limbo state while the membership get migrated over to the correct id and as such any IAM policy checks will fail  when trying to access any GCP resource (i.e. GCS, BQ, etc...) which may explain the long delay you were experiencing. 

To avoid this situation, this change adds a _long_ sleep between creating the pet and adding it to the user's proxy group. If this seems to be effective, we can test shortening the sleep to see just how short of a duration we can cut it to without sacrificing efficacy.

---

**PR checklist**


marctalbott:

PR #449, opened 2020-08-19
[QA-723] add SHORTER(!) sleep between creating pet and adding it to user proxy group

marctalbott:
same as https://github.com/broadinstitute/sam/pull/448 but with a shorter sleep. If this is the only effective workaround, we want this sleep to be as short as possible

---

**PR checklist**


marctalbott:

PR #451, opened 2020-08-20
[QA-723] 2 second sleep

marctalbott:
2 seconds. 

See https://github.com/broadinstitute/sam/pull/448 for details -- the same motivation applies to this

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -319,7 +319,7 @@ class GoogleExtensions(
           for {
             _ <- assertProjectInTerraOrg(project)
             sa <- IO.fromFuture(IO(googleIamDAO.createServiceAccount(project, petSaName, petSaDisplayName)))
-            _ <- IO.sleep(5 seconds)(IO.timer(executionContext))
+            _ <- IO.sleep(2 seconds)(IO.timer(executionContext))
             r <- IO.fromFuture(IO(withProxyEmail(user.id) { proxyEmail =>
               googleDirectoryDAO.addMemberToGroup(proxyEmail, sa.email)
             }))
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -319,6 +319,10 @@ class GoogleExtensions(
           for {
             _ <- assertProjectInTerraOrg(project)
             sa <- IO.fromFuture(IO(googleIamDAO.createServiceAccount(project, petSaName, petSaDisplayName)))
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -319,6 +319,7 @@ class GoogleExtensions(
           for {
             _ <- assertProjectInTerraOrg(project)
             sa <- IO.fromFuture(IO(googleIamDAO.createServiceAccount(project, petSaName, petSaDisplayName)))
+            _ <- IO.sleep(2 seconds)(IO.timer(executionContext))
```
gpolumbo-broad, line 326: We should at least comment the heck out of this so that people understand why we introduced this sleep and how we arrived at the number that we did.  

Might also want to create a Jira ticket (and link to it) as a place holder to fix this whenever Google finally figures out the root cause.  I see this change as just a bandaid to cover over some underlying Google issue that they will _hopefully_ fix some day, but who knows.

andy7i, line 326: maybe adding a comment here why we're doing this (and when we can remove it) would be helpful.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesV1Spec.scala
```diff
@@ -11,10 +12,14 @@ import org.broadinstitute.dsde.workbench.sam.model._
 import org.broadinstitute.dsde.workbench.sam.service._
 import org.scalatest.concurrent.ScalaFutures
 
+import scala.concurrent.duration._
+
 /**
   * Unit tests of GoogleExtensionRoutes. Can use real Google services. Must mock everything else.
   */
 class GoogleExtensionRoutesV1Spec extends GoogleExtensionRoutesSpecHelper with ScalaFutures{
+  implicit val timeout = RouteTestTimeout(5 seconds) //after using com.google.cloud.storage.contrib.nio.testing.LocalStorageHelper, tests seems to run a bit longer
```
andy7i, line 21: If this is required to account for the 2 second sleep (above) then maybe worth adding a similar comment here about why we're adding this in and when we can remove it. It seems like this would change the timeout for all tests in this suite, which might not be something we want to keep long term?

andy7i (approved):
👍  with comment

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesV1Spec.scala
marctalbott: This timeout already existed in `GoogleExtensionRoutesSpec` with the same explanatory comment seen here. Since that test suite is basically the same as this one, I don't think we'd remove this timeout even after we removed the 2s sleep. I'm guessing that this was just missed or wasn't showing the flakiness for some reason

gpolumbo-broad (approved):

PR #452, opened 2020-08-21
[QA-723] 4 second sleep in case 2 is too short

marctalbott:
We added this sleep in [this PR](https://github.com/broadinstitute/sam/pull/451), but are still seeing ~10% failure rate. We need a pass rate >90% to release, so after a night of test runs, it seems that 2 seconds puts us right on the edge of being able to release. If we see the current 10% failure rate increase at all, we want to be able to go more conservative with the sleep duration so that we can release. That's where this PR comes in. 

By bumping the sleep duration up to 4 seconds, we decrease the odds of hitting the Google race condition and improve our chances of releasing next week. If we see more `bucket did not become readable` failures today (8/21/20), then I will merge this PR in the hopes that our automation pass rate climbs above 90% over the weekend and we can release next week. If we don't see any `bucket did not become readable` failures today, then we will let the current code run over the weekend in the hopes that we remain at a pass rate >90% and can still release without increasing the duration of the sleep.

---

**PR checklist**


andy7i (approved):
👍

gpolumbo-broad (approved):

marctalbott:

PR #453, opened 2020-08-21
No ticket - Hopeful fix for "key not found" error in tests

gpolumbo-broad:
tests were failing with message "Key not found" when trying to cleanup a user in Thurloe after running tests.  Looking at the test, it looks like there was identical duplicated code to remove the user from thurloe and assert its removal.  Seems like it was an accident so I removed the duplicate code.

The other thing we could look at is modifying Workbench-libs to just report a success if it tries to delete a key from thurloe and it gets back a 404.


andy7i (approved):
LGTM 👍 
it's interesting that the tests only errored  sometimes

marctalbott (changes requested):
This changes what the test is doing -- we want to be able to remove multiple times in a row without causing any problems from what I can tell

gpolumbo-broad:
derp, you're right @marctalbott .  I think we need to make the change in workbench-libs probably then.  I think this test is legitimately failing from what I can see.  Not consistently, but there are times when it's not right when the requests stack up just right, so I think we need to make a change to workbench libs to make it resilient to those scenarios

PR #455, opened 2020-09-01
[CA-928, CA-968] Add descendant permissions to policies

marctalbott:
Ticket: [CA-928](https://broadworkbench.atlassian.net/browse/CA-928), [CA-968](https://broadworkbench.atlassian.net/browse/CA-928)
Add descendant permissions to policies so that policies can grant permissions to child resources of a given resource type
Bump Resource routes to v2

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/PolicyTable.scala
```diff
@@ -20,6 +21,7 @@ object PolicyTable extends SQLSyntaxSupportWithDefaultSamDB[PolicyRecord] {
     rs.get(e.resourceId),
     rs.get(e.groupId),
     rs.get(e.name),
-    rs.get(e.public)
+    rs.get(e.public),
+    rs.longOpt(e.descendantResourceTypeId).map(ResourceTypePK)
```
marctalbott: I forgot to update Postgres to actually write this value, but now I'm not sure I even understand this. What if a policy has descendant permissions that apply to multiple different kinds of resource types? Which `ResourceTypePK` would go here?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -322,6 +324,8 @@ class ResourceService(
       validateRoles(resourceType, policy) ++
       validateUrlSafe(policy.policyName.value)
 
+    // todo: validate descendant permissions??
```
marctalbott: As far as I can tell, I do need to add something here to make sure that the policy isn't inherited by any other resources and to throw a 400 if it is

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/PolicyTable.scala
marctalbott: My current thinking is that each role and action in the `PolicyRole` and `PolicyAction` tables has to be associated with a resource type, so there's no need to define which resource type any given policy might apply to using this FK. If there are roles or actions defined in the `PolicyRole` or `PolicyAction` table that descend, then those roles and actions apply to child resources of their given resource type.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20200827_descendant_permissions.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="mtalbott" id="descendant_permissions">
```
dvoet: Put each add column in a different change set. This allows each change set to be atomic and avoid problems where the change set fails part way through and is applied incompletely. This is important for DDL statements since they don't participate in transaction and thus have no rollback.

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/PolicyTable.scala
dvoet: yes!

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -124,11 +124,11 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
   }
 
-  private def insertActions(actions: Set[ResourceAction], resourceTypePK: ResourceTypePK)(implicit session: DBSession): Int = {
+  private def insertActions(actions: Set[(ResourceAction, ResourceTypePK)])(implicit session: DBSession): Int = {
     if (actions.isEmpty) {
-      return 0
+      0
     } else {
-      val uniqueActionValues = actions.map { action =>
+      val uniqueActionValues = actions.collect { case (action, resourceTypePK) =>
```
dvoet: this should still be map, not collect

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -353,7 +353,9 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
       insertGroupMembers(groupId, policy.members)
 
       insertPolicyRoles(policy.roles, policyId)
+      insertDescendantRoles(policy.descendantPermissions.flatMap(permissions => permissions.roles.map((_, permissions.resourceType))), policyId)
```
dvoet: I think we can have a single function instead of 2: `insertPolicyRoles` and `insertDescendantRoles`. The only real difference is that one inserts true for `descends` and the other false. Same goes for actions.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -362,6 +364,60 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
     }
   }
 
+  private def insertDescendantRoles(rolesWithResourceType: Set[(ResourceRoleName, ResourceTypeName)], policyId: PolicyPK)(implicit session: DBSession): Int = {
```
dvoet: I feel like `(ResourceRoleName, ResourceTypeName)` should be its own type: 
```
case class FullyQualifiedResourceRoleName(name: ResourceRoleName, resourceTypeName: ResourceTypeName)
```

maybe a companion object:
```
object FullyQualifiedResourceRoleName {
  def fullyQualify(roles: Set[ResourceRoleName], resourceTypeName: ResourceTypeName) = roles.map(FullyQualifiedResourceRoleName(_, resourceTypeName)
}
```

marctalbott: I tried to work out a way to combine this with `insertPolicyRoles` (and to combine `insertDescendantActions` with `insertPolicyActions`), but I couldn't work out how to do it. If anyone has any ideas, let me know!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -372,6 +387,29 @@ class ResourceService(
     } else None
   }
 
+  private def validatePolicyNotInherited(policy: FullyQualifiedPolicyId, samRequestContext: SamRequestContext): IO[Option[ErrorReport]] = {
+    for {
+      children <- accessPolicyDAO.listResourceChildren(policy.resource, samRequestContext)
+      policyOpt <- accessPolicyDAO.loadPolicy(policy, samRequestContext)
+    } yield {
+      policyOpt.flatMap { policy =>
+        if (policyInheritedByChildren(policy.descendantPermissions, children)) {
```
dvoet: why are we preventing this case?

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20200901_descendant_policy_roles.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="mtalbott" id="descendant_policy_roles">
```
dvoet: you can have more than one `changeSet` in a `databaseChangeLog`

dvoet: in other words, you don't need 2 files

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
marctalbott: According to the ticket, overwriting the policy should "Return 400 if policy is inherited."

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20200901_descendant_policy_roles.xml
marctalbott: yeah, I figured this out when I was making some other liquibase changes, it's back to one file in future commits

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
marctalbott: @dvoet is this still something we want to prevent? Or was the ticket description a relic of when you could reference inherited policies with a '~'?

docker/run-opendj.sh
```diff
@@ -10,6 +10,10 @@ start()  {
             CONTAINER_NAME=opendj-${BUILD_TAG}
         fi
     fi
+
+    echo "attempting to remove old $CONTAINER_NAME container..."
```
marctalbott, line 14: This is an unrelated QoL change that the run-postgres.sh script has

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
marctalbott: Scrapped this as the ticket description was out of date

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -236,41 +329,41 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
...
-        entity(as[AccessPolicyMembership]) { membershipUpdate =>
+      entity(as[AccessPolicyMembership]) { membershipUpdate =>
+        requireAction(policyId.resource, SamResourceActions.alterPolicies, userInfo.userId, samRequestContext) {
           complete(resourceService.overwritePolicy(resourceType, policyId.accessPolicyName, policyId.resource, membershipUpdate, samRequestContext).map(_ => StatusCodes.Created))
         }
       }
     }
+  }
+
+  private def requireActionsForSharePolicy(policyId: FullyQualifiedPolicyId, userInfo: UserInfo, samRequestContext: SamRequestContext)(sharePolicy: server.Route): server.Route =
```
marctalbott, line 342: Previously we were doing the same permissions check in two different places in the v1 routes. I've consolidated that check into a single place so that it will be the same check for all three endpoints in both versions

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2009,6 +2774,17 @@ components:
         policy:
           $ref: '#/components/schemas/AccessPolicyMembership'
       description: ""
+    AccessPolicyResponseEntryV2:
```
marctalbott, line 2777: I opted to make this a separate component instead of just adding descendant permissions as an optional property so that they wouldn't show up at all in the v1 payloads in swagger in an effort to be conscious of the swagger generated client by not changing the v1 endpoints at all

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -156,10 +158,18 @@ consistent "has a" relationship is tracked by this ticket: https://broadworkbenc
...
     public: Boolean)
     extends WorkbenchGroup
 
-@Lenses final case class AccessPolicyMembership(memberEmails: Set[WorkbenchEmail], actions: Set[ResourceAction], roles: Set[ResourceRoleName])
+@Lenses final case class AccessPolicyDescendantPermissions(resourceType: ResourceTypeName, actions: Set[ResourceAction], roles: Set[ResourceRoleName])
+@Lenses final case class AccessPolicyMembership(memberEmails: Set[WorkbenchEmail],
+                                                actions: Set[ResourceAction],
+                                                roles: Set[ResourceRoleName],
+                                                descendantPermissions: Option[Set[AccessPolicyDescendantPermissions]] = None) {
+  def toV1 = this.copy(descendantPermissions = None) // strip out descendantPermissions for backwards compatibility
```
dvoet: I don't think we need to do this. Adding new fields to our response payloads is perfectly backwards compatible. It is also backwards compatible for requests since it is an `Option` and thus not required.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -236,41 +329,41 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
...
-  def getResourcePolicies(resource: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route =
+  private def requireActionsForListPolicies(resource: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext)(listResourcePolicies: server.Route): server.Route =
     get {
       requireAction(resource, SamResourceActions.readPolicies, userInfo.userId, samRequestContext) {
-        complete(resourceService.listResourcePolicies(resource, samRequestContext).map { response =>
-          StatusCodes.OK -> response.toSet
-        })
+        listResourcePolicies
       }
     }
```
dvoet: this would be more akka http friendly if written as a directive:
```
  private def requireActionsForListPolicies(resource: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext): Directive0 =
    get & requireAction(resource, SamResourceActions.readPolicies, userInfo.userId, samRequestContext)
```
BUT I would not include the `get` in there. With the name `requireActionsForListPolicies`, including the `get` is unexpected.

dvoet (changes requested):
I think there is too much change going on the the v1 resource routes. I think the v1 routes should be unchanged. The process should be copy all the v1 routes to v2, add in the new v2 routes. It is ok to change the underlying functions (like `getResourcePolicies`) because they are backwards compatible. The reason to the version bump so far is more of a semantic change instead of a break in compatibility (although that break is coming soon).

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
marctalbott: This is no longer used in the v1 routes because we don't want to change the v1 routes at all in this PR. I'd prefer to keep it for the v2 routes at least, but can also remove this optimization and leave it entirely for a separate PR

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -156,10 +158,17 @@ consistent "has a" relationship is tracked by this ticket: https://broadworkbenc
...
+    descendantPermissions: Set[AccessPolicyDescendantPermissions],
     public: Boolean)
     extends WorkbenchGroup
 
-@Lenses final case class AccessPolicyMembership(memberEmails: Set[WorkbenchEmail], actions: Set[ResourceAction], roles: Set[ResourceRoleName])
+@Lenses final case class AccessPolicyDescendantPermissions(resourceType: ResourceTypeName, actions: Set[ResourceAction], roles: Set[ResourceRoleName])
+@Lenses final case class AccessPolicyMembership(memberEmails: Set[WorkbenchEmail],
+                                                actions: Set[ResourceAction],
+                                                roles: Set[ResourceRoleName],
+                                                descendantPermissions: Option[Set[AccessPolicyDescendantPermissions]] = Option(Set.empty)) {
```
marctalbott, line 169: I changed the default value from `None` to `Option(Set.empty)` so that I wouldn't have to change the v1 routes specs at all. With `None` as the default value, the expected response would be `AccessPolicyMembership(_, _, _, None)` but the response returned by the ResourceService is `AccessPolicyMembership(_, _, _, Some(Set.empty))`

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20200901_descendant_permissions.xml
```diff
@@ -0,0 +1,39 @@
...
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="mtalbott" id="drop_policy_action_PK">
+        <dropPrimaryKey tableName="sam_policy_action" constraintName="sam_policy_action_pkey"/>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="mtalbott" id="descendant_policy_actions">
+        <addColumn tableName="sam_policy_action">
+            <column name="descends" type="BOOLEAN" defaultValue="false">
```
andy7i: `descends` makes me think that this permission would start at the parent and then apply to descendants. I don't have a great alternative but maybe couple options: `descendantsOnly` or `appliesToAllDescendantsOnly`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/PolicyActionTable.scala
```diff
@@ -4,14 +4,16 @@ import scalikejdbc._
 import org.broadinstitute.dsde.workbench.sam.db.SamTypeBinders
 
 final case class PolicyActionRecord(resourcePolicyId: PolicyPK,
-                                    resourceActionId: ResourceActionPK)
+                                    resourceActionId: ResourceActionPK,
+                                    descends: Boolean)
```
andy7i: same comment as above:

> `descends` makes me think that this permission would start at the parent and then apply to descendants. I don't have a great alternative but maybe couple options: `descendantsOnly` or `appliesToAllDescendantsOnly`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/PolicyRoleTable.scala
```diff
@@ -4,14 +4,16 @@ import scalikejdbc._
 import org.broadinstitute.dsde.workbench.sam.db.SamTypeBinders
 
 final case class PolicyRoleRecord(resourcePolicyId: PolicyPK,
-                                  resourceRoleId: ResourceRolePK)
+                                  resourceRoleId: ResourceRolePK,
+                                  descends: Boolean)
```
andy7i: same comment as above:

> `descends` makes me think that this permission would start at the parent and then apply to descendants. I don't have a great alternative but maybe couple options: `descendantsOnly` or `appliesToAllDescendantsOnly`?

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -362,21 +364,19 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
     }
   }
 
-  private def insertPolicyActions(actions: Set[ResourceAction], policyId: PolicyPK)(implicit session: DBSession): Int = {
+  private def insertPolicyActions(actions: Set[FullyQualifiedResourceAction], policyId: PolicyPK, descends: Boolean)(implicit session: DBSession): Int = {
```
andy7i: same comment as above, also applies a couple times in this file:

> `descends` makes me think that this permission would start at the parent and then apply to descendants. I don't have a great alternative but maybe couple options: `descendantsOnly` or `appliesToAllDescendantsOnly`?

andy7i (commented):
adding comments so far, which are all about the naming of `descends` (one comment per file)

(I still need to finish reviewing DAO changes, and then review swagger and tests.)

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -362,21 +364,19 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+        case FullyQualifiedResourceAction(action, resourceType) => samsqls"(${action}, ${resourceType})"
+      }
+      val insertQuery = samsqls"""insert into ${PolicyActionTable.table} (${paCol.resourcePolicyId}, ${paCol.resourceActionId}, ${paCol.descends})
+            select ${policyId}, ${ra.result.id}, ${descends}
             from ${ResourceActionTable as ra}
             join ${ResourceTypeTable as rt} on ${ra.resourceTypeId} = ${rt.id}
-            join ${ResourceTable as r} on ${r.resourceTypeId} = ${rt.id}
-            join ${PolicyTable as p} on ${p.resourceId} = ${r.id}
-            where ${ra.action} in (${actions})
-            and ${p.id} = ${policyId}"""
```
andy7i, line 379: Just to double check that I understand this: the old code here verified that the `policyId` exists. This was removed since this method takes in a `policyId: PolicyPK` which we can assume is valid and exists?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
andy7i: not related to this PR: this function returns an Int, but it doesn't seem to be used anywhere. Do we return the int because it's easy to return that data and might be useful for future use?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -385,9 +385,16 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
-          .map(rs => rs.get[ResourceTypePK](r.resultName.resourceTypeId)).single().apply().getOrElse(throw new WorkbenchException(s"could not find resource type id for policy id $policyId"))
-        insertActions(actions, resourceTypePK)
+        val resourceTypeNames = actions.map(_.resourceTypeName)
+
+        val resourceTypeNameAndPK = samsql"select ${rt.result.name}, ${rt.result.id} from ${ResourceTypeTable as rt} where (${rt.name}) in (${resourceTypeNames})"
+          .map(rs => (rs.get[ResourceTypeName](rt.resultName.name), rs.get[ResourceTypePK](rt.resultName.id))).list().apply().toMap
+        val actionsWithResourceTypePK = actions.map {
+          case FullyQualifiedResourceAction(action, resourceTypeName) =>
+            (action, resourceTypeNameAndPK.getOrElse(resourceTypeName,
+              throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "Resource type not found"))))
```
andy7i: would it be helpful to the user to include the resource type name in the error message?

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -611,34 +620,61 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+
+    val descendantPermissions = descendantRoleResults.groupBy(_.resourceTypeName).collect { case (Some(descendantResourceTypeName), descendantRolesWithResourceType) =>
+      val descendantRoles = descendantRolesWithResourceType.collect { case RoleResult(_, Some(role), _) => role }
+      val descendantActions = descendantActionResults.groupBy(_.resourceTypeName).getOrElse(Option(descendantResourceTypeName), List.empty).collect { case ActionResult(_, Some(action), _) => action }
+      AccessPolicyDescendantPermissions(descendantResourceTypeName, descendantActions.toSet, descendantRoles.toSet)
+    }.toSet
+
+    (roles, actions, descendantPermissions)
+  }
+
```
andy7i: nice 👍

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -920,6 +956,19 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
 }
 
 private final case class PolicyInfo(name: AccessPolicyName, resourceId: ResourceId, resourceTypeName: ResourceTypeName, email: WorkbenchEmail, public: Boolean)
+private final case class MemberResult(userId: Option[WorkbenchUserId], groupName: Option[WorkbenchGroupName], policyName: Option[AccessPolicyName], resourceId: Option[ResourceId], resourceTypeName: Option[ResourceTypeName])
+private final case class RoleResult(resourceTypeName: Option[ResourceTypeName], role: Option[ResourceRoleName], descends: Option[Boolean])
+private final case class ActionResult(resourceTypeName: Option[ResourceTypeName], action: Option[ResourceAction], descends: Option[Boolean])
+
+final case class FullyQualifiedResourceRoleName(roleName: ResourceRoleName, resourceTypeName: ResourceTypeName)
```
andy7i: would `FullyQualifiedResourceRole` be a better name?

src/test/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAOSpec.scala
```diff
@@ -499,10 +499,31 @@ class PostgresAccessPolicyDAOSpec extends FreeSpec with Matchers with BeforeAndA
         dirDao.createGroup(defaultGroup, samRequestContext = samRequestContext).unsafeRunSync()
         dirDao.createUser(defaultUser, samRequestContext).unsafeRunSync()
 
-        val policy = AccessPolicy(FullyQualifiedPolicyId(resource.fullyQualifiedId, AccessPolicyName("policyName")), Set(defaultGroup.id, defaultUser.id), WorkbenchEmail("policy@email.com"), resourceType.roles.map(_.roleName), Set(readAction, writeAction), false)
+        val policy = AccessPolicy(FullyQualifiedPolicyId(resource.fullyQualifiedId, AccessPolicyName("policyName")), Set(defaultGroup.id, defaultUser.id), WorkbenchEmail("policy@email.com"), resourceType.roles.map(_.roleName), Set(readAction, writeAction), Set.empty, false)
         dao.createPolicy(policy, samRequestContext).unsafeRunSync()
         dao.loadPolicy(policy.id, samRequestContext).unsafeRunSync() shouldEqual Option(policy)
       }
+
+      "creates descendant actions and roles" in {
```
andy7i, line 507: would it make sense to add another test where we verify that a "descends" policy cannot be used by the parent resource?

andy7i (approved):
minor questions, nothing blocking. 👍

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
marctalbott: Not exactly. Resource actions are resource type specific. For example, you could have two different `write` resource actions, one for the `workspace` resource type and another for the `billing-project` resource type. When we insert into the `PolicyActionTable`, we need to make sure we are inserting the correct action. 

Before this PR, the correct action was always on the same resource type as the policy. Given that, we would select `${ra.result.id}` based on the resource type of the policy -- that's also why we needed the additional joins to the `ResourceTable` and the `PolicyTable`. 

Now we want to be able to insert actions for arbitrary resource types, not just the resource type of the policy. We are able to do this by passing in the resource type name along with the action and then comparing them to the `(${ra.action}, ${rt.name})` tuples in the conditional statement on l. 379.

So you're right in that we can assume that the `policyId` exists and is valid because we've previously loaded that from Postgres, but that's not why we don't need the conditional anymore

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
marctalbott: Yup, by default the `.apply()` returns the number of rows updated, so we just return that number instead of dropping it. I'm not sure if we ever really use it to be honest, but it's good practice to return it anyways I think

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
marctalbott: I think I do like that better. I went back and forth on that a bit and settled on what I have because it matched with `ResourceRoleName`, but I don't think it has to match and your suggestion rolls off the tongue better

src/test/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAOSpec.scala
marctalbott: Discussed on slack, Doug's PR will include a test to this effect

PR #456, opened 2020-09-02
[CA-930] Don't delete resources with children

andy7i:
Ticket: https://broadworkbench.atlassian.net/browse/CA-930

This PR updates the existing API for delete resource to check for resources with parents or children.

a few notes:
- 403 if the resource has a parent and the user does not have remove_child access on the parent resource
- instead of a new v2 API, this will update the v1 API. Part of the reason for this is that this uses v2 behavior (adding parent info) and also the v1 API would need to be updated with the same or similar behavior to prevent deleting a resource that has children.
- tests for this are being put in v2 spec, partly because it uses v2 behavior, but mostly because v0, v1 behavior should not have changed anyway and we should be putting new tests into the latest version spec file.

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1720,10 +1720,10 @@ paths:
           description: Child resource has an auth domain or new parent would introduce cyclical resource hierarchy
           content: {}
         403:
-          description: You do not have permission to set the parent for the child resource, to remove the child of the current parent, or to add a child to the new parent
+          description: You do not have permission to set the parent for the child resource, to remove the child of the current parent, to add a child to the new parent, or the new parent does not exist
           content: {}
         404:
-          description: You do not have access to the child, new parent, or current resource, or they do not exist
+          description: You do not have access to this resource, it does not exist
           content: {}
         500:
           description: Internal Server Error
@@ -1754,7 +1754,7 @@ paths:
           description: Resource parent successfully removed
           content: {}
         403:
-          description: You do not have permission to set the parent for the child resource or to remove the child of the current parent
+          description: You do not have permission to set the parent for the this resource or to remove this resource of its current parent
           content: {}
         404:
           description: You do not have access to this resource, it does not exist, or it does not currently have a parent resource
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1723,7 +1723,7 @@ paths:
           description: You do not have permission to set the parent for the child resource, to remove the child of the current parent, to add a child to the new parent, or the new parent does not exist
           content: {}
         404:
-          description: You do not have access to this resource, it does not exist
+          description: You do not have access to this resource, or it does not exist
           content: {}
         500:
           description: Internal Server Error
@@ -1754,7 +1754,7 @@ paths:
           description: Resource parent successfully removed
           content: {}
         403:
-          description: You do not have permission to set the parent for the this resource or to remove this resource of its current parent
+          description: You do not have permission to set the parent for this resource or to remove this (child) resource from its current parent
           content: {}
         404:
           description: You do not have access to this resource, it does not exist, or it does not currently have a parent resource
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -970,8 +970,11 @@ paths:
         204:
           description: Successfully deleted resource
           content: {}
+        400:
+          description: Cannot delete a resource with children. Delete the children first then try again.
+          content: {}
         403:
-          description: You do not have permission to perform this action on the resource
+          description: You do not have permission to perform this action on the resource or permissions on the resource's parent
           content: {}
         404:
           description: Resource type does not exist or you are not a member of any
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -36,9 +36,6 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
       case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"resource type ${name.value} not found"))
     }
 
-  def withCurrentParentOption(resource: FullyQualifiedResourceId, samRequestContext: SamRequestContext): Directive1[Option[FullyQualifiedResourceId]] =
-    onSuccess(resourceService.getResourceParent(resource, samRequestContext))
-
   def resourceRoutes: server.Route =
     (pathPrefix("config" / "v1" / "resourceTypes") | pathPrefix("resourceTypes")) {
         requireUserInfo(SamRequestContext(None)) { userInfo => // `SamRequestContext(None)` is used so that we don't trace 1-off boot/init methods ; these in particular are unpublished APIs
@@ -186,18 +183,11 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
 
   def deleteResource(resource: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route =
     delete {
-      withCurrentParentOption(resource, samRequestContext) {
-        case Some(currentResourceParent) =>
-          requireAction(currentResourceParent, SamResourceActions.removeChild, userInfo.userId, samRequestContext) {
-            deleteResourceInternal(resource, userInfo, samRequestContext)
-          }
-        case None => deleteResourceInternal(resource, userInfo, samRequestContext)
+      requireAction(resource, SamResourceActions.delete, userInfo.userId, samRequestContext) {
+        requireParentAction(resource, None, SamResourceActions.removeChild, userInfo.userId, samRequestContext) {
+          complete(resourceService.deleteResource(resource, samRequestContext).map(_ => StatusCodes.NoContent))
+        }
       }
-    }
-
-  def deleteResourceInternal(resource: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext) =
-    requireAction(resource, SamResourceActions.delete, userInfo.userId, samRequestContext) {
-      complete(resourceService.deleteResource(resource, samRequestContext).map(_ => StatusCodes.NoContent))
     }
 
   def postDefaultResource(resourceType: ResourceType, resource: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route =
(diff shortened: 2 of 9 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -196,15 +196,29 @@ class ResourceService(
...
   //      but not the Resource itself, thereby orphaning the Resource so that it cannot be used or accessed anymore and
   //      preventing a new Resource with the same ID from being created
+  // Resources with children cannot be deleted and will throw a 400.
   def deleteResource(resource: FullyQualifiedResourceId, samRequestContext: SamRequestContext): Future[Unit] =
     for {
+      // First check if children exist. If so, then throw a 400.
+      _ <- hasChildren(resource, samRequestContext).unsafeToFuture() map {
+        case true => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Cannot delete a resource with children. Delete the children first then try again."))
+        case false =>
+      }
```
dvoet: I would have a function called `checkNoChildren` instead of `hasChildren` that lists the children and does this check all in one

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -715,6 +715,44 @@ class ResourceServiceSpec extends FlatSpec with Matchers with ScalaFutures with
...
+    runAndWait(service.setResourceParent(childResource, parentResource, samRequestContext))
+
+    assert(policyDAO.listAccessPolicies(childResource, samRequestContext).unsafeRunSync().nonEmpty)
+
+    runAndWait(service.deleteResource(childResource, samRequestContext))
+
+    assert(policyDAO.listAccessPolicies(childResource, samRequestContext).unsafeRunSync().isEmpty)
+  }
+
+  it should "return 400 for a parent resource that has any children" in {
```
dvoet: also have a test that checks for 400 in the route spec. In the test here I mostly care that it fails and does not delete the resource. In the route spec we care that the response code is 400.

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: also an `if` statement is easier to follow than a `case` for boolean (I think)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -196,15 +196,25 @@ class ResourceService(
...
       _ <- maybeDeleteResource(resource, samRequestContext)
     } yield ()
 
+  /** Check if a resource has any children. If so, then throw a 400. */
+  def checkNoChildren(resource: FullyQualifiedResourceId, samRequestContext: SamRequestContext): IO[Unit] = {
+    listResourceChildren(resource, samRequestContext) map { list =>
+      if (list.nonEmpty) throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Cannot delete a resource with children. Delete the children first then try again."))
+    }
+  }
+
```
dvoet, line 218: much cleaner 👍

gpolumbo-broad (approved):

marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
andy7i: thanks for the feedback!

marctalbott (approved):

dvoet (approved):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1720,10 +1720,10 @@ paths:
           description: Child resource has an auth domain or new parent would introduce cyclical resource hierarchy
           content: {}
         403:
-          description: You do not have permission to set the parent for the child resource, to remove the child of the current parent, or to add a child to the new parent
+          description: You do not have permission to set the parent for the child resource, to remove the child of the current parent, to add a child to the new parent, or the new parent does not exist
```
andy7i, line 1723: After looking at IETF's descriptions of http codes (https://tools.ietf.org/html/rfc7231#section-6.5.1) as well as this SO post (https://stackoverflow.com/questions/48449192/rest-http-response-code-for-missing-parent-resource#:~:text=YES%3A%20A%20404%20Not%20Found,%22sorry%2C%20resource%20unavailable%22.), it seems like there isn't a great option, but a 404 might be the most straightforward option. I could also see a 400 as a general catch-all. 403 feels like a bit of a stretch for "parent resource does not exist", but I'm not super tied to any status code as long as the error message is descriptive and useful for the user.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1754,7 +1754,7 @@ paths:
           description: Resource parent successfully removed
           content: {}
         403:
-          description: You do not have permission to set the parent for the child resource or to remove the child of the current parent
+          description: You do not have permission to set the parent for the this resource or to remove this resource of its current parent
```
andy7i:
```suggestion
          description: You do not have permission to set the parent for this resource or to remove this (child) resource from its current parent
```

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1720,10 +1720,10 @@ paths:
           description: Child resource has an auth domain or new parent would introduce cyclical resource hierarchy
           content: {}
         403:
-          description: You do not have permission to set the parent for the child resource, to remove the child of the current parent, or to add a child to the new parent
+          description: You do not have permission to set the parent for the child resource, to remove the child of the current parent, to add a child to the new parent, or the new parent does not exist
           content: {}
         404:
-          description: You do not have access to the child, new parent, or current resource, or they do not exist
+          description: You do not have access to this resource, it does not exist
```
andy7i:
```suggestion
          description: You do not have access to this resource, or it does not exist
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
```diff
@@ -3,48 +3,99 @@ package org.broadinstitute.dsde.workbench.sam.api
...
+    val parentIO = newParent match {
+      case Some(_) => IO.pure(newParent)
+      case None => resourceService.getResourceParent(resource, samRequestContext)
+    }
+
+    parentIO.flatMap {
+      case Some(resourceParent) => hasPermissionOneOf(resourceParent, actions, userId, samRequestContext)
+      case None =>
+        // there is no parent so permission is granted
+        IO.pure(true)
```
andy7i, line 100: why do we return true if no parent exists here? It feels weird since we're checking if the user has some permission on the parent. If the parent doesn't exist they wouldn't be able to make changes on the parent right?

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
```diff
@@ -406,7 +408,7 @@ class ResourceRoutesV2Spec extends FlatSpec with Matchers with TestSupport with
       accessToCurrentParent = false)
 
     Delete(s"/api/resources/v2/${defaultResourceType.name}/${fullyQualifiedChildResource.resourceId.value}/parent") ~> samRoutes.route ~> check {
-      status shouldEqual StatusCodes.NotFound
+      status shouldEqual StatusCodes.Forbidden
     }
   }
```
andy7i, line 431: I think we're missing a test for erroring when a parent is set on a resource, but the parent resource does not exist

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
```diff
@@ -3,48 +3,99 @@ package org.broadinstitute.dsde.workbench.sam.api
...
+    * @param samRequestContext
+    * @return
+    */
+  def requireOneOfParentAction(childResource: FullyQualifiedResourceId, newParent: Option[FullyQualifiedResourceId], parentActions: Set[ResourceAction], userId: WorkbenchUserId, samRequestContext: SamRequestContext): Directive0 =
+    Directives.mapInnerRoute { innerRoute =>
+      onSuccess(hasParentPermissionOneOf(childResource, newParent, parentActions, userId, samRequestContext)) { hasPermission =>
+        if (hasPermission) {
+          innerRoute
+        } else {
+          val forbiddenErrorMessage = s"You may not perform any of ${parentActions.mkString("[", ", ", "]").toUpperCase} on parent of ${childResource.resourceTypeName.value}/${childResource.resourceId.value}"
```
andy7i:
```suggestion
          val forbiddenErrorMessage = s"You may not perform any of ${parentActions.mkString("[", ", ", "]").toUpperCase} on parent of ${childResource.resourceTypeName.value}/${childResource.resourceId.value}. The parent resource may not exist."
```

If a user tries to add/remove a parent that doesn't exist, they will get this message, right? If that is the case then a more descriptive message like this may be helpful?

src/main/resources/swagger/api-docs.yaml
dvoet: The problem is that we can't let our error conditions reveal the existence of a resource. A user should not be able to tell the difference between no access and does not exist. So, yes, 403 is a stretch for parent resource does not exist but 400 is also a stretch for access denied to the parent and we need to have the same response for both.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
dvoet: If there is no parent this effectively disables the parent permission check which seems appropriate.

We hit this case in requiring `remove_child` only at the moment. If there is not parent `remove_child` should be allowed.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
dvoet: can you add what is missing?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
dvoet: I put in a different fix

src/main/resources/swagger/api-docs.yaml
andy7i: That makes sense -- we don't want to reveal the existence of a resource. In other places, for a target resource we use 404 to indicate either `no access` or `does not exist`. It feels like we could be consistent by doing that here for parent resources as well. The user would get an error message saying "404, the parent resource does not exist or you have no access."

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -339,34 +338,28 @@ trait ResourceRoutes extends UserInfoDirectives with SecurityDirectives with Sam
...
-
   def deleteResourceParent(resource: FullyQualifiedResourceId, userInfo: UserInfo, samRequestContext: SamRequestContext): server.Route =
     delete {
-      withCurrentParentOption(resource, samRequestContext) {
-        case Some(resourceParent) =>
-          requireAction(resourceParent, SamResourceActions.removeChild, userInfo.userId, samRequestContext) {
-            requireAction(resource, SamResourceActions.setParent, userInfo.userId, samRequestContext) {
-              complete(resourceService.deleteResourceParent(resource, samRequestContext).map(_ => StatusCodes.NoContent))
+      requireAction(resource, SamResourceActions.setParent, userInfo.userId, samRequestContext) {
+        requireParentAction(resource, None, SamResourceActions.removeChild, userInfo.userId, samRequestContext) {
```
andy7i, line 354: ~I think that `requireParentAction(resource, None` will always pass. Instead of `None` I think we need to to pass in a `currentParent: Option[FullyQualifiedResourceId]` or look for it within `requireParentAction`~

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
```diff
@@ -311,6 +323,22 @@ class ResourceRoutesV2Spec extends FlatSpec with Matchers with TestSupport with
       accessToCurrentParent = false)
 
     Put(s"/api/resources/v2/${defaultResourceType.name}/${fullyQualifiedChildResource.resourceId.value}/parent", newParentResource) ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.Forbidden
+    }
+  }
+
+  it should "404 if the parent resource does not exist" in {
```
andy7i: This is the new unit test. 

~(It seems like if a user wanted to scrape the names of all existing resources, they could use setResourceParent, which currently gives a 404 when the resource does not exist and possibly a 403 when it does. -- ~Needs more investigation~.)~

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
dvoet: see SecurityDirectives line 93: when None is passed it will look up the current parent

src/main/resources/swagger/api-docs.yaml
dvoet: I think that is inconsistent with the HTTP spec. 404 means that some part of the URL does not exist. That is not what is happening here.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
dvoet: I don't think this is right. We want a 403 if the parent resource does not exist. It should behave exactly the same if it exists but there is not access. The user should not be able to distinguish between no access to parent or parent does not exist.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
andy7i: For `setResourceParent`, if we pass in `Option(newResourceParent)` instead of `None`, the resulting error would be: ~`StatusCodes.BadRequest, "Cannot set parent as this would introduce a cyclical resource hierarchy"`.~

it should get caught as not having the right permissions on the parent and then 403

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
andy7i: ~looks like ^ is the behavior at the moment, mostly because we pass `None` into `requireParentAction(resource, None` which ends up calling getResourceParent. If we change that, we get the `cyclical resource hierarchy ` error message mentioned in the above comment.~

Nope, nevermind.. seems like it should 403 in either case

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
andy7i: (I missed the parameter name `newParent` in `requireParentAction`)

andy7i:
jenkins retest -- failure in orch seems unrelated

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
```diff
@@ -107,20 +107,22 @@ class ResourceRoutesV2Spec extends FlatSpec with Matchers with TestSupport with
...
+        missingActionsOnNewParent.map(action => when(samRoutes.policyEvaluatorService.hasPermission(mockitoEq(newParent), mockitoEq(action), mockitoEq(defaultUserInfo.userId), any[SamRequestContext]))
+          .thenReturn(IO(false)))
+        if (accessToNewParent) {
+          when(samRoutes.policyEvaluatorService.listResourceAccessPoliciesForUser(mockitoEq(newParent), mockitoEq(defaultUserInfo.userId), any[SamRequestContext]))
+            .thenReturn(IO(Set(otherPolicy)))
+        } else {
+          when(samRoutes.policyEvaluatorService.listResourceAccessPoliciesForUser(mockitoEq(newParent), mockitoEq(defaultUserInfo.userId), any[SamRequestContext]))
+            .thenReturn(IO(Set[AccessPolicyWithoutMembers]()))
+        }
+      case None =>
```
dvoet: since there is nothing actually in the None case, does this actually change anything? If not, I would undo.

andy7i: yeah that sounds good. When I was reviewing my code, I thought it was a little easier to read this way so I left it as is, but I'm not super attached to it!

PR #464, opened 2020-10-15
perf fixes

dvoet:
Ticket: <Link to Jira ticket>
2 fixes that came out of looking at traces of the AoU 100 concurrent user test:
- `hasPermissionOneOf` used to has `hasPermission` for each action which does the same work multiply times. Fixed to do the work once
- Auth domain check list all the groups a user is in. It is cheaper to get the roles only on the groups in question.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
@@ -40,12 +40,18 @@ class PolicyEvaluatorService(
       }
   }
 
+  def hasPermissionOneOf(resource: FullyQualifiedResourceId, actions: Iterable[ResourceAction], userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Boolean] = traceIOWithContext("hasPermissionOneOf", samRequestContext)(samRequestContext => {
+    listUserResourceActions(resource, userId, samRequestContext).map { userActions =>
+      actions.toSet.intersect(userActions).nonEmpty
+    }
+  })
+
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
```diff
@@ -92,6 +95,18 @@ class PolicyEvaluatorService(
     } yield res
   }
 
+  private def isMemberOfAllAuthDomainGroups(authDomains: NonEmptyList[WorkbenchGroupName], userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Boolean] = {
```
marctalbott, line 98: Could this be accomplished by using `directoryDAO.listIntersectionGroupUsers` and checking to see if the user is in it? I'm not sure how that would compare from an efficiency perspective, but it seems like it'd be a more straightforward way to find out if the user is in all the auth domain groups

marctalbott (approved):
It all looks good to me, your call if you want to try out my suggested alternative

gpolumbo-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorService.scala
dvoet: I think `listIntersectionGroupUsers` will fail if there is only 1 group (it is meant to intersect the policy group with all the AD groups, not the AD groups alone). However it _might_ scale better when there are 2 or more AD groups. Also my initial implementation which just replaced the body of this function with
```
directoryDAO.listIntersectionGroupUsers(authDomains.toList.toSet, samRequestContext).map(_.contains(userId))
```
failed some unit tests so I am going to save this suggestion for future consideration.

araina2018:
For compliance: seems this fix was a product change and should've been linked to JIRA.

araina2018:
Relevant  JIRA https://broadworkbench.atlassian.net/browse/QA-1285

PR #465, opened 2020-10-16
DR-1400: Add spend profile resource

snf2ye:
Ticket: [DR-1400 - Add Sam definition of the billing profile resource](https://broadworkbench.atlassian.net/browse/DR-1400)
- Add new "spend-profile" resource as a part of work for [DR-1370](https://broadworkbench.atlassian.net/browse/DR-1370).  The aim of this work is to add authorization around the use of billing profiles
- The actions listed were defined as a part of our work to clarify and [update our permissions model](https://broadworkbench.atlassian.net/browse/DR-663). 

---

**PR checklist**

N/A - I don't think I have made any changes to the api nor any security related changes. There also should not be any manual steps required.

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -511,6 +511,38 @@ resourceTypes = {
     }
     reuseIds = true
   }
+  spendProfile = {
+    actionPatternObjects = {
+      "share_policy::admin" = {
+        description = "Can grant and revoke a users' admin permission"
+      }
+      "share_policy::steward" = {
+        description = "Can grant and revoke a users' steward permission"
+      }
+      "read_policies" = {
+        description = "Can read policies"
+      }
+      "read_policy::steward" = {
+        description = "Can read the steward policy"
+      }
+      "alter_policies" = {
+        description = "Can alter policies"
+      }
+      "delete" = {
+        description = "Can delete policies/resources"
+      }
+    }
+    ownerRoleName = "admin"
+    roles = {
+      admin = {
+        roleActions = ["share_policy::steward", "share_policy::admin", "read_policies", "alter_policies"]
+      }
+      steward = {
+        roleActions = ["share_policy::steward", "read_policy::steward"]
+      }
+    }
+    reuseIds = true
+  }
   study = {
     actionPatternObjects = {
       delete = {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -513,35 +513,35 @@ resourceTypes = {
   }
   spend-profile = {
     actionPatternObjects = {
-      "create" = {
-        description = "Can create profiles"
-      }
-      "update" = {
-        description = "Can update profiles"
+      "update_billing_account" = {
+        description = "Ability to update the spend profile to use a different billing account"
       }
       "delete" = {
-        description = "Can delete profiles"
-      }
-      "list" = {
-        description = "Can read profiles"
+        description = "Ability to delete profiles"
       }
       "link" = {
-        description = "Can link profiles"
+        description = "Ability to link a spend profile to an object, thus allowing the object to spend funds from the designated billing account"
       }
       "share_policy::owner" = {
-        description = "Can grant and revoke a users' owner permission"
+        description = "Ability to grant and revoke a users' owner permission"
+      }
+      "share_policy::user" = {
+        description = "Ability to grant and revoke a users' user permission"
+      }
+      "share_policy::admin" = {
+        description = "Ability to grant and revoke a users' admin permission"
       }
     }
     ownerRoleName = "owner"
     roles = {
       admin = {
-        roleActions = ["share_policy::owner"]
+        roleActions = ["share_policy::owner", "share_policy::user", "share_policy::admin"]
       }
       owner = {
-        roleActions = ["create", "update", "delete", "list", "link", "share_policy::owner"]
+        roleActions = ["update_billing_account", "delete", "link", "share_policy::owner", "share_policy::user", "share_policy::admin"]
       }
       user = {
-        roleActions = ["list", "link", "share_policy::owner"]
+        roleActions = ["link"]
       }
     }
     reuseIds = true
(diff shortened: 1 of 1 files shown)


nmalfroy (approved):
lgtm!

src/main/resources/reference.conf
```diff
@@ -511,6 +511,41 @@ resourceTypes = {
     }
     reuseIds = true
   }
+  spend-profile = {
+    actionPatternObjects = {
+      "create" = {
```
dvoet: all users can create resources so `create` action is not required

src/main/resources/reference.conf
```diff
@@ -511,6 +511,41 @@ resourceTypes = {
...
+        description = "Can create profiles"
+      }
+      "update" = {
+        description = "Can update profiles"
+      }
+      "delete" = {
+        description = "Can delete profiles"
+      }
+      "list" = {
```
dvoet: all users can list resources they have access to so `list` is not requried

src/main/resources/reference.conf
```diff
@@ -511,6 +511,41 @@ resourceTypes = {
     }
     reuseIds = true
   }
+  spend-profile = {
+    actionPatternObjects = {
+      "create" = {
+        description = "Can create profiles"
+      }
+      "update" = {
```
dvoet: what is being updated? It is a best practice to be granular in defining actions.

src/main/resources/reference.conf
```diff
@@ -511,6 +511,41 @@ resourceTypes = {
...
+        description = "Can update profiles"
+      }
+      "delete" = {
+        description = "Can delete profiles"
+      }
+      "list" = {
+        description = "Can read profiles"
+      }
+      "link" = {
```
dvoet, line 525: what does it mean to link a profile?

src/main/resources/reference.conf
```diff
@@ -511,6 +511,41 @@ resourceTypes = {
...
+        description = "Can link profiles"
+      }
+      "share_policy::owner" = {
+        description = "Can grant and revoke a users' owner permission"
+      }
+    }
+    ownerRoleName = "owner"
+    roles = {
+      admin = {
```
dvoet: do we need admin in addition to owner? If a user has `share_policy::owner` they can add themselves to the owner policy

src/main/resources/reference.conf
```diff
@@ -511,6 +511,41 @@ resourceTypes = {
...
+        roleActions = ["share_policy::owner"]
+      }
+      owner = {
+        roleActions = ["create", "update", "delete", "list", "link", "share_policy::owner"]
+      }
+      user = {
+        roleActions = ["list", "link", "share_policy::owner"]
```
dvoet: `share_policy::owner` does not belong in the user role

src/main/resources/reference.conf
```diff
@@ -511,6 +511,41 @@ resourceTypes = {
...
+      "share_policy::owner" = {
+        description = "Can grant and revoke a users' owner permission"
+      }
+    }
+    ownerRoleName = "owner"
+    roles = {
+      admin = {
+        roleActions = ["share_policy::owner"]
+      }
+      owner = {
```
dvoet, line 537: probably want `share_policy::user` in here unless there are only owners

dvoet (changes requested):

src/main/resources/reference.conf
snf2ye: Thank you for catching this!

src/main/resources/reference.conf
snf2ye: Does this mean we also need to assign the owner & admin a "share_policy::admin" action?

dvoet: yes, but why do we need the admin role?

src/main/resources/reference.conf
ddietterich: We were debating this ourselves. How have you handled the case where there is no one in the owner role? For example, someone leaves the company and they are the sole owner.

We put the admin in with the ability to make a new owner. That is all it is for.

src/main/resources/reference.conf
ddietterich: So you would recommend making an update permission for every updatable field in the data structure?
That would be name, billingAccountId, and description.

I guess i just don't know where the edge of granularity is. At this point we only have one update operation.

src/main/resources/reference.conf
dvoet: We don't handle orphaned resources well at the moment but I don't see how an admin role fixes it. Some other user would still have to be granted the admin role... why not just grant the other user the owner role (owners can make other owners too)?

ddietterich: Our idea was this: when TDR creates a resource, it grants admin role to a group - some managed group like JadeAdmins. The owner would not be able to change that grant; we would not give the owner share_policy:admin. The only action the admin would have is share_policy:owner to add a new owner. (Of course, in practice that is all permissions, but it would require auditable actions.)

What do you think?

src/main/resources/reference.conf
ddietterich: We have a similar action for other TDR objects: `read_dataset` action for seeing the dataset definition.
We use the action check on the resource id when accessing a single dataset with this:
`/api/resources/v1/{resourceTypeName}/{resourceId}/action/{action}`

When enumerating, we get all accessible resources using this:
`/api/resources/v1/{resourceTypeName}`

So if I understand, you would recommend using `/api/resources/v1/{resourceTypeName}/{resourceId}/actions` for checking if a user has some access to the object rather than making an action. Yes?

src/main/resources/reference.conf
snf2ye: Updated to two update actions: update_billing_account and update_metadata.

src/main/resources/reference.conf
snf2ye: After talking with Dan, I've removed the "admin" role.

dvoet:
you will need to rebase develop to get the jenkins automation tests to pass

dvoet (approved):

PR #467, opened 2020-10-22
[CA-913] Add google project to reference.conf

s-rubenstein:
Ticket: https://broadworkbench.atlassian.net/browse/CA-913
I added the google-project resource type to reference.conf. Two things I wasn't super sure about was what access levels there should be (so I just started with only owner), and also what all permissions were compute related. (There was one called `launch_batch_compute` but I assume that shouldn't be there, cause I read compute as specifically Leo compute?)

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -226,6 +226,12 @@ resourceTypes = {
   }
   google-project = {
     actionPatternObjects = {
+      alter_policies = {
+        description = "create and delete policies for this billing-project"
+      }
+      read_policies = {
+        description = "list all policies and policy details for this billing-project"
+      }
       list_notebook_cluster = {
         description = "list all notebook clusters in this billing-project"
       }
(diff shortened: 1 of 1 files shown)


s-rubenstein:
I don't think there's any fisma considerations, but would like confirmation on that.

src/main/resources/reference.conf
```diff
@@ -224,6 +224,41 @@ resourceTypes = {
...
+      }
+      delete_persistent_disk = {
+        description = "delete persistent disk in this billing-project"
+      }
+      create_kubernetes_app = {
+        description = "create kubernetes application in this billing-project"
+      }
+    }
+    ownerRoleName = "owner"
+    roles = {
```
marctalbott, line 261: I believe we also want the `notebook-user` role to be copied onto this resource type from `billing-project`. I think we don't need to bother with `batch-compute-user` since we're removing the one check that I know about. I leave it up to Doug to confirm whether or not that's true.

src/main/resources/reference.conf
```diff
@@ -224,6 +224,41 @@ resourceTypes = {
...
+      owner = {
+        roleActions = ["list_notebook_cluster", "launch_notebook_cluster", "delete_notebook_cluster", "list_persistent_disk", "create_persistent_disk", "delete_persistent_disk", "create_kubernetes_app"]
```
marctalbott: There are a few non-Leo compute actions that should probably be added to this list: `alter_policies`, `read_policies`, `add_to_service_perimeter`, and `delete`.

One could make the case for `alter_google_role` to be added here as well, but I think it's not really used anymore after Doug's recent Rawls PR

marctalbott (commented):
Couple comments on stuff that wasn't really covered in the ticket but that I think should be included

marctalbott:
> I don't think there's any fisma considerations, but would like confirmation on that.

Agreed on no FISMA consideration

src/main/resources/reference.conf
dvoet: yeah, drop alter_google_role
add_to_service_perimeter should not be added, that is remaining at the billing-project level

marctalbott (approved):

s-rubenstein:
Jenkins Retest

dvoet (approved):

PR #469, opened 2020-10-26
CA-1047 create resource with parent

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/CA-1047
update api to allow specification of parent upon creation and allow ownerless resources when create with a parent

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
@@ -7,7 +7,7 @@ import cats.effect.IO
 import org.broadinstitute.dsde.workbench.model.{ErrorReport, WorkbenchExceptionWithErrorReport, WorkbenchUserId}
 import org.broadinstitute.dsde.workbench.sam.ImplicitConversions.ioOnSuccessMagnet
 import org.broadinstitute.dsde.workbench.sam._
-import org.broadinstitute.dsde.workbench.sam.model.{FullyQualifiedResourceId, ResourceAction}
+import org.broadinstitute.dsde.workbench.sam.model.{FullyQualifiedResourceId, ResourceAction, ResourceType, SamResourceActions}
 import org.broadinstitute.dsde.workbench.sam.service.{PolicyEvaluatorService, ResourceService}
 import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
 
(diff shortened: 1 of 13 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
```diff
@@ -51,6 +51,33 @@ trait SecurityDirectives {
       }
     }
 
+  def requireCreateWithParent(maybeParent: Option[FullyQualifiedResourceId], resourceType: ResourceType, userId: WorkbenchUserId, samRequestContext: SamRequestContext): Directive0 = {
```
s-rubenstein: From the name of the method, I would expect this to require a parent to be passed in, but it seems to work without one specified?

s-rubenstein: Just to understand this a bit, is this example correct?
I am a user, and I want to create a workspace with billing project Y as its parent. I might end up calling this method?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
```diff
@@ -51,6 +51,33 @@ trait SecurityDirectives {
...
+    maybeParent match {
+      case None => Directives.pass // no parent specified, proceed
+      case Some(parent) =>
+        // parents are allowed for a resource type if the owner role contains the SamResourceActions.setParent action
+        val parentAllowed = resourceType.roles.find(_.roleName == resourceType.ownerRoleName).exists(_.actions.contains(SamResourceActions.setParent))
```
s-rubenstein, line 59: Is this just checking if the resource has parent options available? It seems weird to have a permission on a child to be whether you can set the parent or not.

src/main/scala/org/broadinstitute/dsde/workbench/sam/openam/PostgresAccessPolicyDAO.scala
```diff
@@ -948,37 +978,33 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+                  join ${ResourceTypeTable as rt} on ${r.resourceTypeId} = ${rt.id}
+                  where ${r.name} = ${parentResource.resourceId}
+                  and ${rt.name} = ${parentResource.resourceTypeName}
+                  union
+                  select ${pr.resourceParentId}
+                  from ${ResourceTable as pr}
+                  join ${ancestorResourceTable as ar} on ${ar.resourceId} = ${pr.id}
+                  where ${pr.resourceParentId} is not null
+        ) update ${ResourceTable as r}
+          set ${resourceTableColumn.resourceParentId} = $parentResourcePK
```
s-rubenstein, line 999: Nice. I really like this abstraction.

s-rubenstein (commented):
Nice, I like a lot of this. Asked a few questions, but mostly looks good.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
dvoet: the calling code looks better if this handles parent both cases (with and without parent) in 1 function... this could be called `maybeRequireCreateWithParent` but that did not make me feel good either. `requireCreateWithParentIfSpecified`?

dvoet: well... users don't call methods, they call apis. But the api the user calls would go through this method.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
dvoet: why is that weird?

s-rubenstein: I would expect that to be a permission on the parent, not on the child. (And then it checks if there are parents available, and if there are none, its irrelevant.)

Unless this is about like… switching billing projects for a workspace or something? (Or moving a notebook to another workspace or something? I could see that being a thing, but it seems like it might be better represented as permissions on the parent)

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
s-rubenstein: How about `requireCreateWithOptionalParent`

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
s-rubenstein: Resolved in voice.

dvoet:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -120,42 +122,50 @@ class ResourceService(
...
+        // make sure there is an owner policy if there is no parent
+        policies.exists { policy =>
+          policy.roles.contains(resourceType.ownerRoleName) && policy.emailsToSubjects.nonEmpty
+        } match {
+          case true => None
+          case false =>
+            Option(ErrorReport(s"Cannot create resource without at least 1 policy with ${resourceType.ownerRoleName.value} role and non-empty membership"))
+        }
+
+      case Some(_) => None // if a parent exists, the parent's owners are effectively owners of this resource
```
gpolumbo-broad, line 168: And we're sure that there's no way that someone could empty out the members of the Owner policy on the parent?

dvoet: just as sure as no one could empty out the Owner policy on the child

gpolumbo-broad (approved):

s-rubenstein (approved):
LGTM

PR #474, opened 2020-11-02
[CA-1062, CA-1075] Add nested roles to config

marctalbott:
Ticket: [CA-1062](https://broadworkbench.atlassian.net/browse/CA-1062), [CA-1075](https://broadworkbench.atlassian.net/browse/CA-1075)

- read descendant and included roles from config
- add descendant and included roles for `google-project` resource types
- update `ResourceRole` to include descendant and included roles
- upsert nested resource roles during resource type init 

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -261,6 +261,7 @@ resourceTypes = {
     roles = {
       owner = {
         roleActions = ["read_policies", "list_notebook_cluster", "launch_notebook_cluster", "delete_notebook_cluster", "list_persistent_disk", "create_persistent_disk", "delete_persistent_disk", "create_kubernetes_app", "delete"]
+        includedRoles = ["notebook-user"]
       }
       notebook-user = {
         roleActions = ["launch_notebook_cluster", "create_persistent_disk", "create_kubernetes_app"]
@@ -329,6 +330,9 @@ resourceTypes = {
     roles = {
       owner = {
         roleActions = ["view_status", "create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "list_notebook_cluster", "launch_notebook_cluster", "delete_notebook_cluster", "list_persistent_disk", "create_persistent_disk", "delete_persistent_disk", "create_kubernetes_app", "alter_google_role", "add_to_service_perimeter", "delete"]
+        descendantRoles = {
+          google-project = ["owner"]
+        }
       }
       workspace-creator = {
         roleActions = ["view_status", "create_workspace", "share_policy::can-compute-user", "read_policy::can-compute-user"]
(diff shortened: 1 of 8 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
```diff
@@ -44,7 +44,12 @@ object AppConfig {
       val uqPath = unquoteAndEscape(path)
       ResourceRole(
         ResourceRoleName(uqPath),
-        config.as[Set[String]](s"$uqPath.roleActions").map(ResourceAction.apply)
+        config.as[Set[String]](s"$uqPath.roleActions").map(ResourceAction.apply),
+        config.as[Option[Set[String]]](s"$uqPath.includedRoles").getOrElse(Set.empty).map(ResourceRoleName.apply),
+        config.as[Option[Map[String, Set[String]]]](s"$uqPath.descendantRoles").getOrElse(Map.empty)
+          .map { case (resourceTypeName, roleNames) =>
+            (ResourceTypeName(resourceTypeName), roleNames.map(ResourceRoleName.apply))
+          }
```
dvoet, line 51: is there a reason why the ValueReaders extend `ValueReader` instead of using `ValueReader.relative` like `directoryConfigReader` below? I think we could get rid of the `uqPath` stuff.

src/main/resources/reference.conf
```diff
@@ -329,6 +330,9 @@ resourceTypes = {
     roles = {
       owner = {
         roleActions = ["view_status", "create_workspace", "alter_policies", "read_policies", "launch_batch_compute", "list_notebook_cluster", "launch_notebook_cluster", "delete_notebook_cluster", "list_persistent_disk", "create_persistent_disk", "delete_persistent_disk", "create_kubernetes_app", "alter_google_role", "add_to_service_perimeter", "delete"]
+        descendantRoles = {
```
dvoet, line 339: add `descendantRoles` to `workspace` too?

andy7i (approved):
👍  with feedback addressed

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
marctalbott: The `ValueReader.relative` stuff knows the config path as opposed to the readers we have here where the config path defines the role and resource type names. I'd guess that there's some way to use the `ValueReader.relative` instead of extending `ValueReader` and still figure out the path, but it's not immediately apparent to me. From what I've seen, it doesn't seem like we could necessarily get rid of `uqPath` either since we'd still need to unquote and escape whatever path we were able to glean from the `Config` object before constructing our case classes.

dvoet (approved):

PR #482, opened 2020-11-13
DR-1484: Add read & alter policy actions to spend profile

snf2ye:
Ticket: DR-1484
Follow on to [DR-1400 - Add Sam definition of the billing profile resource](https://broadworkbench.atlassian.net/browse/DR-1400). These changes will add the ability to read and alter spend profile policies. 

---

**PR checklist**
(None of these items should be applicable) 

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -594,6 +594,15 @@ resourceTypes = {
   }
   spend-profile = {
     actionPatterns = {
+      "read_policies" = {
+        description = "Can read policies"
+      }
+      "read_policy::owner" = {
+        description = "Can read the owner policy"
+      }
+      "alter_policies" = {
+        description = "Can alter policies"
+      }
       "update_billing_account" = {
         description = "Ability to update the spend profile to use a different billing account"
       }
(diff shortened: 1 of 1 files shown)


ddietterich (approved):
lgtm

src/main/resources/reference.conf
```diff
@@ -616,7 +622,7 @@ resourceTypes = {
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["update_billing_account", "update_metadata", "delete", "link", "share_policy::owner", "share_policy::user"]
+        roleActions = ["update_billing_account", "update_metadata", "delete", "link", "share_policy::owner", "share_policy::user", "read_policies", "alter_policies"]
```
dvoet: Given that you have `"share_policy::owner", "share_policy::user"` probably don't want `alter_policies`. The difference is that `alter_policies` allows owners to change any aspect of policies (add/delete, change roles). The share actions only allow membership changes but the policy structure cannot be changed.

snf2ye: Ok, thank you! Do you think we cover all of the actions that alter_policies would cover? It doesn't look like we have an action to change roles. (But, I'm not sure if that's needed)

dvoet (approved):

src/main/resources/reference.conf
dvoet: You guys are the ones creating the policies when you create the resource. Once created you are not giving anyone permissions to change what policies exist or the roles they have. This is the current best practice because it leaves your app in control of the structure but still allows users to change memberships.

dvoet:
jenkins retest

PR #483, opened 2020-11-16
[CA-951] Add update permission to billing project

s-rubenstein:
Ticket: <Link to Jira ticket>
I wanted to add update to Sam to use in Rawls for updating billing account for a billing project. I figure we can use this to update other metadata as well, when/if we want to add other updating to this.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -346,6 +346,9 @@ resourceTypes = {
       delete = {
         description = "delete this billing-project"
       }
+      update = {
+        description = "update this billing-project
+      }
       add_child = {
         description = "add google-project"
       }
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -346,6 +346,9 @@ resourceTypes = {
       delete = {
         description = "delete this billing-project"
       }
+      update = {
```
dvoet: actions should be as granular as possible so it is easier to create roles that do just what we want. So this should be `update_billing_account`

dvoet (changes requested):

src/main/resources/reference.conf
s-rubenstein: I was thinking that we may want an update billing project endpoint (that would cover full updates), and this could serve as the permission for it all, but that may be too much future proofing.

dvoet (approved):

andy7i (approved):
👍  with tests passing

PR #487, opened 2020-11-25
Flat queries

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/CA-315
* There are some docs in the README, start there
* I created a new package `org/broadinstitute/dsde/workbench/sam/dataAccess` and moved most of the DAOs there

---

**PR checklist**


dvoet:
jenkins retest

dvoet:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/DbReference.scala
```diff
@@ -42,38 +49,72 @@ object DbReference extends LazyLogging {
...
-    DbReference(dbName)
+    DbReference(dbName, dbExecutionContext)
   }
 
-  def resource(liquibaseConfig: LiquibaseConfig, dbName: DatabaseName): Resource[IO, DbReference] = Resource.make(
-    IO(init(liquibaseConfig, dbName))
-  )(_ => IO(DBs.close(dbName.name)))
+  def resource(liquibaseConfig: LiquibaseConfig, dbName: DatabaseName): Resource[IO, DbReference] = {
+    for {
+      dbExecutionContext <- ExecutionContexts.fixedThreadPool[IO](DBs.config.getInt(s"db.${dbName.name.name}.poolMaxSize"))
```
dvoet, line 64: note this change moves the thread pool associated to a db connection internal to the db ref.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -125,11 +124,21 @@ class ResourceService(
...
     */
-  private def persistResource(resourceType: ResourceType, resourceId: ResourceId, policies: Set[ValidatableAccessPolicy], authDomain: Set[WorkbenchGroupName], parentOpt: Option[FullyQualifiedResourceId], samRequestContext: SamRequestContext) =
-    for {
-      resource <- accessPolicyDAO.createResource(Resource(resourceType.name, resourceId, authDomain, parent = parentOpt), samRequestContext)
-      policies <- policies.toList.traverse(p => createOrUpdatePolicy(FullyQualifiedPolicyId(resource.fullyQualifiedId, p.policyName), p, samRequestContext))
-    } yield resource.copy(accessPolicies = policies.toSet)
+  private def persistResource(resourceType: ResourceType, resourceId: ResourceId, policies: Set[ValidatableAccessPolicy], authDomain: Set[WorkbenchGroupName], parentOpt: Option[FullyQualifiedResourceId], samRequestContext: SamRequestContext) = {
+    val accessPolicies = policies.map(constructAccessPolicy(resourceType, resourceId, _, public = false)) // can't set public at create time
+    accessPolicyDAO.createResource(Resource(resourceType.name, resourceId, authDomain, accessPolicies = accessPolicies, parent = parentOpt), samRequestContext)
+  }
```
dvoet, line 130: change to create a resource plus all policies in a single transaction

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -211,10 +220,10 @@ class ResourceService(
       _ <- checkNoChildren(resource, samRequestContext).unsafeToFuture()
 
       // remove from cloud extensions first so a failure there does not leave ldap in a bad state
-      policiesToDelete <- cloudDeletePolicies(resource, samRequestContext)
+      _ <- cloudDeletePolicies(resource, samRequestContext)
 
-      _ <- policiesToDelete.toList.parTraverse(p => accessPolicyDAO.deletePolicy(p.id, samRequestContext)).unsafeToFuture()
-      _ <- maybeDeleteResource(resource, samRequestContext)
+      _ <- accessPolicyDAO.deleteAllResourcePolicies(resource, samRequestContext).unsafeToFuture()
```
dvoet, line 225: delete all policies in a single transaction

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAOSpec.scala
```diff
@@ -751,8 +761,8 @@ class PostgresDirectoryDAOSpec extends AnyFreeSpec with Matchers with BeforeAndA
       }
 
       "intersect lots of groups with lots of dups and overlaps" in {
-        val groupCount = 80
-        val userCount = 100
+        val groupCount = 40
+        val userCount = 50
```
dvoet, line 764: scaled down for the sake of elapsed time of test

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
```diff
@@ -144,7 +144,7 @@ class PolicyEvaluatorServiceSpec extends AnyFlatSpec with Matchers with TestSupp
   protected def clearDatabase(): IO[Unit] = IO(TestSupport.truncateAll).void
 
   private[service] def savePolicyMembers(policy: AccessPolicy) = {
-    policy.members.toList.parTraverse {
+    policy.members.toList.traverse {
```
dvoet, line 147: given the small database there are too many serializable failures with `parTraverse` so it turns out to be slower

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
```diff
@@ -0,0 +1,119 @@
...
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="effective_policy_table">
+        <createTable tableName="SAM_EFFECTIVE_RESOURCE_POLICY">
+            <column name="id" type="BIGINT" autoIncrement="true">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="resource_id" type="BIGINT">
+                <constraints nullable="false" foreignKeyName="FK_SERP_RESOURCE" referencedTableName="SAM_RESOURCE" referencedColumnNames="id"/>
+            </column>
+            <column name="source_policy_id" type="BIGINT">
```
gpolumbo-broad, line 16: Should be`resource_policy_id`?

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
```diff
@@ -0,0 +1,119 @@
...
+        <sql stripComments="true">
+            insert into SAM_EFFECTIVE_POLICY_ROLE(resource_policy_id, resource_role_id)
+            select ep.id, role.id
+            from SAM_EFFECTIVE_RESOURCE_POLICY ep
+            join SAM_POLICY_ROLE pr on ep.source_policy_id = pr.resource_policy_id
+            join SAM_RESOURCE resource on ep.resource_id = resource.id
+            join sam_flattened_role flat_role on pr.resource_role_id = flat_role.base_role_id
+            join SAM_RESOURCE_ROLE role on flat_role.nested_role_id = role.id and resource.resource_type_id = role.resource_type_id
+            join sam_resource_policy sourcePolicy on ep.source_policy_id = sourcePolicy.id
+            where resource.resource_type_id = role.resource_type_id
```
marctalbott: nit: This seems to be duplicated in the join on l. 98

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/RegistrationDAO.scala
```diff
@@ -1,4 +1,5 @@
-package org.broadinstitute.dsde.workbench.sam.directory
+package org.broadinstitute.dsde.workbench.sam.dataAccess
```
marctalbott, line 1: Should the LdapRegistrationDAO also be moved here?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+          from ${ResourceTable as resource}
+          where ${resource.id} = ${childResourcePK} and ${resource.resourceParentId} is not null
+          union
+          select ${parentResource.resourceParentId}
+          from ${ResourceTable as parentResource}
+          join ${ancestorResourceTable as ancestorResource} on ${ancestorResource.resourceId} = ${parentResource.id}
+          where ${parentResource.resourceParentId} is not null)"""
+  }
+
+  private def descendantsRecursiveQuery(childResourcePK: ResourcePK, descendantResourceTable: AncestorResourceTable) = {
```
marctalbott: Can we change `childResourcePK` to `parentResourcePK`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+              """.update().apply()
+  }
+
+  /**
+    * Delete all inherited effective policies on the specified resource and descendants.
+    * @param childResourcePK
+    * @param session
+    * @return
+    */
+  protected def deleteDescendantInheritedEffectivePolicies(childResourcePK: ResourcePK)(implicit session: DBSession): Int = {
```
marctalbott, line 48: Delete any effective policies that may have been granted to descendants by this resource or any of its ancestors

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+              join ${PolicyActionTable as pa} on ${pa.resourcePolicyId} = ${ep.sourcePolicyId}
+              join ${ResourceActionTable as ra} on ${pa.resourceActionId} = ${ra.id}
+              join ${ResourceTable as r} on ${ep.resourceId} = ${r.id}
+              join ${PolicyTable as sourcePolicy} on ${sourcePolicy.id} = ${ep.sourcePolicyId}
+              where ${pa.descendantsOnly} = (${sourcePolicy.resourceId} != ${ep.resourceId})
+              and ${r.resourceTypeId} = ${ra.resourceTypeId}
+              on conflict do nothing""".update().apply()
+  }
+
+  private def insertEffectivePoliciesForChildAndDescendants(childResourcePK: ResourcePK)(implicit session: DBSession): Int = {
```
marctalbott, line 166: For this resource and all of its descendants, insert a record in the EP table for every policy on this resource or any of its ancestors

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+              join ${FlattenedRoleMaterializedView as fr} on ${pr.resourceRoleId} = ${fr.baseRoleId}
+              join ${ResourceRoleTable as rr} on ${fr.nestedRoleId} = ${rr.id}
+              join ${ResourceTable as r} on ${ep.resourceId} = ${r.id}
+              join ${PolicyTable as sourcePolicy} on ${sourcePolicy.id} = ${ep.sourcePolicyId}
+              where ${r.resourceTypeId} = ${rr.resourceTypeId}
+              and ${roleAppliesToResource(pr, fr, ep, sourcePolicy)}
+              on conflict do nothing""".update().apply()
+  }
+
+  private def insertEffectivePolicyActionsForChildAndDescendants(childResourcePK: ResourcePK)(implicit session: DBSession): Int = {
```
marctalbott, line 131: For this resource and all of its descendants, find all of the policies that could possibly apply to them and grab all the actions on those policies and then insert them into the EPA table if they are applicable (correct resource type and descendant or not)

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+    * @param session
+    * @return
+    */
+  protected def populateInheritedEffectivePolicies(childResourcePK: ResourcePK)(implicit session: DBSession): Int = {
+    insertEffectivePoliciesForChildAndDescendants(childResourcePK) +
+      insertEffectivePolicyActionsForChildAndDescendants(childResourcePK) +
+      insertEffectivePolicyRolesForChildAndDescendants(childResourcePK)
+  }
+
+  private def insertEffectivePolicyRolesForChildAndDescendants(childResourcePK: ResourcePK)(implicit session: DBSession): Int = {
```
marctalbott, line 93: For this resource and all of its descendants, find all of the effective policies and unroll any roles that those policies grant. Check that they are applicable to the resource in question and insert them if they are

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+    val ar = ancestorResourceTable.syntax("ar")
+
+    samsql"""with recursive ${descendantResourcesCTE(policyPK, ancestorResourceTable)}
+            insert into ${EffectivePolicyTable.table} (${epCol.resourceId}, ${epCol.sourcePolicyId}, ${epCol.groupId}, ${epCol.public})
+            select ${ar.resourceId}, ${policyPK}, ${groupId}, ${policy.public}
+            from ${ancestorResourceTable as ar}
+            """.update().apply()
+  }
+
+  private def descendantResourcesCTE(policyPK: PolicyPK, ancestorResourceTable: AncestorResourceTable) = {
```
marctalbott: It seems like this does the exact same thing that `descendantsRecursiveQuery` does, would it make sense to overload that function name?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  /**
+    * Insert effective policy roles from the specified policy for all descendants.
+    * @param policyId
+    * @param session
+    * @return
+    */
+  protected def insertEffectivePolicyRoles(policyId: PolicyPK)(implicit session: DBSession): Int = {
+    val ancestorResourceTable = AncestorResourceTable("ancestor_resource")
```
marctalbott, line 239: nit: I keep getting slightly confused at the use of `ancestorResourceTable` for the `descendantResourcesCTE` here and in the other policy-specific inserts

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+
+              insert into ${EffectivePolicyRoleTable.table} (${erCol.resourcePolicyId}, ${erCol.resourceRoleId})
+              select ${ep.id}, ${rr.id}
+              from ${ancestorResourceTable as ar}
+              join ${EffectivePolicyTable as ep} on ${ar.resourceId} = ${ep.resourceId} and ${ep.sourcePolicyId} = ${policyId}
+              join ${PolicyRoleTable as pr} on ${pr.resourcePolicyId} = ${policyId}
+              join ${FlattenedRoleMaterializedView as fr} on ${pr.resourceRoleId} = ${fr.baseRoleId}
+              join ${ResourceRoleTable as rr} on ${fr.nestedRoleId} = ${rr.id}
+              join ${ResourceTable as r} on ${ep.resourceId} = ${r.id},
+              ${PolicyTable as policy}
```
marctalbott: consider joining on `policy.id = ep.sourcePolicyId` and removing the first clause of the `where`

marctalbott: Same thing with the `insertEffectivePolicyActions`

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,324 @@
...
+import org.postgresql.util.PSQLException
+import scalikejdbc.DBSession
+import scalikejdbc.interpolation.SQLSyntax
+
+import scala.util.{Failure, Try}
+
+/**
+  * The sam group model is implemented using 2 tables GroupTable and GroupMemberTable. GroupTable stores
+  * the name, email address and other top level information of the group. GroupMemberTable stores
+  * member groups and emails. Note that this is a recursive structure; groups can contain groups and
```
marctalbott: `member groups and users` instead of `emails`

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_flat_group_membership.xml
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+    <changeSet logicalFilePath="dummy" author="thibault" id="flat_group_member_table">
+        <createTable tableName="SAM_GROUP_MEMBER_FLAT">
```
marctalbott, line 8: Should we add a constraint that exactly 1 of `member_user_id` or `member_group_id` can be null like we have for the `sam_group_member` table? I'm not sure that it's necessary since I believe that constraint was added for the sake of the opendj -> postgres migration and I suppose that constraint also effectively protects this table since we're populating this table from that table, but wanted to pose the question

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,324 @@
...
+  * --------+----------+-----------------+----------------+-----------------------
+  *  345985 |     7795 |                 | userid         | {7795}
+  *  345986 |     7798 |                 | userid         | {7798}
+  *  345987 |     7801 |            7798 |                | {7801}
+  *  345988 |     7801 |            7799 |                | {7801}
+  *  345989 |     7801 |                 | userid         | {7801,7798}
+  *
+  * It is crucial that all group updates are in serializable transactions to avoid race conditions when
+  * concurrent modifications are made affecting the same group structure.
+  */
```
marctalbott, line 60: Not sure if this will make a substantial difference, but could you update this comment to include info about the `last_group_membership_element` and what it's used for

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,324 @@
...
+
+    if (memberGroupPKs.size != memberGroupNames.size + memberPolicyIdTuples.size) {
+      throw new WorkbenchException(s"Some member groups not found.")
+    }
+    memberGroupPKs
+  }
+
+  def workbenchGroupIdentityToGroupPK(groupId: WorkbenchGroupIdentity): SQLSyntax = {
+    groupId match {
+      case group: WorkbenchGroupName => GroupTable.groupPKQueryForGroup(group)
```
marctalbott: Can we move `groupPkQueryForGroup` into the GroupDAO for consistency with `groupPKQueryForPolicy`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,324 @@
...
+             from (values ${directUserValues ++ directGroupValues}) AS insertValues (member_user_id, member_group_id),
+             ${FlatGroupMemberTable as fgm}
+             where ${fgm.memberGroupId} = $groupId""".update().apply()
+
+    if (memberGroupPKs.nonEmpty) {
+      // insert subgroup memberships
+      val tail = FlatGroupMemberTable.syntax("tail")
+      val head = FlatGroupMemberTable.syntax("head")
+      samsql"""insert into ${FlatGroupMemberTable.table} (${fgmColumn.groupId}, ${fgmColumn.memberUserId}, ${fgmColumn.memberGroupId}, ${fgmColumn.groupMembershipPath}, ${fgmColumn.lastGroupMembershipElement})
+             select ${head.groupId}, ${tail.memberUserId}, ${tail.memberGroupId}, array_cat(${head.groupMembershipPath}, ${tail.groupMembershipPath}), ${tail.groupMembershipPath}[array_upper(${tail.groupMembershipPath}, 1)]
```
marctalbott, line 186: Confused on the syntax here, what is the first `${tail.groupMembershipPath}` for in `${tail.groupMembershipPath}[array_upper(${tail.groupMembershipPath}, 1)]`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,324 @@
...
+    samsql"""delete from ${GroupMemberTable.table}
+                where ${groupMemberColumn.groupId} = (${workbenchGroupIdentityToGroupPK(groupId)})
+                and ${groupMemberColumn.memberUserId} = ${memberUser}""".update().apply() > 0
+  }
+
+  private def removeMemberUserFromFlatGroup(groupId: WorkbenchGroupIdentity, memberUser: WorkbenchUserId)(implicit session: DBSession) = {
+    val f = FlatGroupMemberTable.syntax("f")
+    samsql"""delete from ${FlatGroupMemberTable as f}
+                where ${f.memberUserId} = $memberUser
+                and ${f.groupMembershipPath}[array_upper(${f.groupMembershipPath}, 1)] = (${workbenchGroupIdentityToGroupPK(groupId)})""".update().apply()
```
marctalbott: could this use the `lastGroupMembershipElement` column instead of the `array_upper`?

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAOSpec.scala
```diff
@@ -287,7 +285,8 @@ class PostgresDirectoryDAOSpec extends AnyFreeSpec with Matchers with BeforeAndA
         val subGroup = emptyWorkbenchGroup("subGroup")
         dao.createGroup(defaultGroup, samRequestContext = samRequestContext).unsafeRunSync()
 
-        assertThrows[PSQLException] {
+        // TODO: changed this from PSQLException.  This seems OK but let's double check
```
marctalbott: TODO

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+  * effective for all resources so that at query time the lookup is straight forward, less stress on the db and fast.
+  *
+  * An effective policy applies to one and only one resource, never to descendants. It contains roles and actions
+  * that are applicable only to the resource; it is not required to check if roles are descendant only or apply
+  * to other resource types.
+  *
+  * It is crucial that all resource and policy updates are in serializable transactions to avoid race conditions when
+  * concurrent modifications are made affecting the same resource structure.
+  */
+trait EffectivePolicyMutationStatements {
```
marctalbott, line 21: I found it helpful while reading through this to try and describe in my own words what a bunch of the different functions do. Leaving them in the review in case you want to include any as comments, but feel free to disregard if you don't find them helpful!

marctalbott (commented):
Nothing blocking, mostly just questions and nits

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
```diff
@@ -0,0 +1,119 @@
...
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="index_erp_resource_id">
+        <createIndex tableName="SAM_EFFECTIVE_RESOURCE_POLICY" indexName="IDX_ERP_RESOURCE_ID">
+            <column name="resource_id"/>
+        </createIndex>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="index_erp_public">
+        <createIndex tableName="SAM_EFFECTIVE_RESOURCE_POLICY" indexName="IDX_ERP_PUBLIC">
+            <column name="public"/>
```
gpolumbo-broad, line 30: While we're at it, can we add the index to the source `RESOURCE_POLICY.public` column too?  https://broadworkbench.atlassian.net/browse/CA-796

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
```diff
@@ -0,0 +1,119 @@
...
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="resource_id" type="BIGINT">
+                <constraints nullable="false" foreignKeyName="FK_SERP_RESOURCE" referencedTableName="SAM_RESOURCE" referencedColumnNames="id"/>
+            </column>
+            <column name="source_policy_id" type="BIGINT">
+                <constraints nullable="false" foreignKeyName="FK_SERP_SOURCE_POLICY" referencedTableName="SAM_RESOURCE_POLICY" referencedColumnNames="id" deleteCascade="true"/>
+            </column>
+            <column name="group_id" type="BIGINT">
+                <constraints nullable="false" foreignKeyName="FK_SERP_GROUP" referencedTableName="SAM_GROUP" referencedColumnNames="id"/>
```
gpolumbo-broad: Does this column really need to be duplicated into this table?   (also `public`, but that column I'm less sure about)

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
```diff
@@ -0,0 +1,119 @@
...
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="index_sr_parent_resource_id">
+        <createIndex tableName="SAM_RESOURCE" indexName="IDX_SR_PARENT_RESOURCE_ID">
+            <column name="resource_parent_id"/>
+        </createIndex>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="effective_policy_role_table">
+        <createTable tableName="SAM_EFFECTIVE_POLICY_ROLE">
+            <column name="resource_policy_id" type="BIGINT">
```
gpolumbo-broad: rename to `effective_resource_policy_id` to avoid confusion.  As named, I thought this was a foreign key for `SAM_RESOURCE_POLICY.id`

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
```diff
@@ -0,0 +1,119 @@
...
+            </column>
+            <column name="resource_role_id" type="BIGINT">
+                <constraints nullable="false" foreignKeyName="FK_SEPR_ROLE" referencedTableName="SAM_RESOURCE_ROLE" referencedColumnNames="id" primaryKey="true"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="effective_policy_action_table">
+        <createTable tableName="SAM_EFFECTIVE_POLICY_ACTION">
+            <column name="resource_policy_id" type="BIGINT">
```
gpolumbo-broad: See comment for same FK in `SAM_EFFECTIVE_POLICY_ROLE`

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
```diff
@@ -0,0 +1,119 @@
...
+            select ar.base_resource_id, policy.id, policy.group_id, policy.public
+            from SAM_RESOURCE_POLICY policy join ancestor_resource ar on policy.resource_id = ar.resource_id
+        </sql>
+        <sql stripComments="true">
+            insert into SAM_EFFECTIVE_POLICY_ROLE(resource_policy_id, resource_role_id)
+            select ep.id, role.id
+            from SAM_EFFECTIVE_RESOURCE_POLICY ep
+            join SAM_POLICY_ROLE pr on ep.source_policy_id = pr.resource_policy_id
+            join SAM_RESOURCE resource on ep.resource_id = resource.id
+            join sam_flattened_role flat_role on pr.resource_role_id = flat_role.base_role_id
```
gpolumbo-broad: I think we should add materialized views to the ERD.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
```diff
@@ -0,0 +1,119 @@
...
+            join SAM_RESOURCE resource on ep.resource_id = resource.id
+            join SAM_RESOURCE_ACTION action on pa.resource_action_id = action.id and resource.resource_type_id = action.resource_type_id
+            join sam_resource_policy sourcePolicy on ep.source_policy_id = sourcePolicy.id
+            where pa.descendants_only = (sourcePolicy.resource_id != ep.resource_id)
+            and resource.resource_type_id = action.resource_type_id
+            on conflict do nothing
+        </sql>
+    </changeSet>
+
+</databaseChangeLog>
```
gpolumbo-broad, line 106: A thought occurs to me as I read these queries for populating the `EFFECTIVE_*` tables.  Since these tables are sort of "read optimized views" of data already stored in the database should these queries that we are  using to populate them live in some location where we can easily trigger and rerun them if needed?  For example, if the tables somehow fall out of sync with the source tables from which they're populated?   

I'm pretty sure the answer is no (I'm thinking out loud here),  but there's something niggling at my brain here.  Something feels different about what we're doing here.  There's a fairly complex relation/dependency between the source `RESOURCE_*` tables and the `EFFECTIVE_*` tables.  Maybe it is that just from looking at the schema, there is no immediate way to understand the relationship between `RESOURCE_*` and `EFFECTIVE_*`.  

You can probably just ignore this comment, it's probably fine.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_flat_group_membership.xml
```diff
@@ -0,0 +1,67 @@
...
+            </column>
+            <column name="member_group_id" type="BIGINT">
+                <constraints foreignKeyName="FK_SGM_MEMBER_GROUP" referencedTableName="SAM_GROUP" referencedColumnNames="id"/>
+            </column>
+            <column name="member_user_id" type="VARCHAR">
+                <constraints foreignKeyName="FK_SGM_MEMBER_USER" referencedTableName="SAM_USER" referencedColumnNames="id" deleteCascade="true"/>
+            </column>
+            <column name="group_membership_path" type="BIGINT ARRAY"/>
+            <column name="last_group_membership_element" type="BIGINT">
+                <constraints foreignKeyName="FK_SGM_LAST_GROUP_MEMBERSHIP_ELEMENT" referencedTableName="SAM_GROUP" referencedColumnNames="id" deleteCascade="true"/>
```
gpolumbo-broad, line 23: column missing in ERD

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -309,7 +301,7 @@ object Boot extends IOApp with LazyLogging {
     val policyEvaluatorService = PolicyEvaluatorService(config.emailDomain, resourceTypeMap, accessPolicyDAO, directoryDAO)
     val resourceService = new ResourceService(resourceTypeMap, policyEvaluatorService, accessPolicyDAO, directoryDAO, cloudExtensionsInitializer.cloudExtensions, config.emailDomain)
     val userService = new UserService(directoryDAO, cloudExtensionsInitializer.cloudExtensions, registrationDAO, config.blockedEmailDomains)
-    val statusService = new StatusService(directoryDAO, cloudExtensionsInitializer.cloudExtensions, DbReference(DatabaseNames.Foreground), 10 seconds)
+    val statusService = new StatusService(directoryDAO, cloudExtensionsInitializer.cloudExtensions, DbReference(DatabaseNames.Read, implicitly), 10 seconds)
```
gpolumbo-broad, line 303: I have no idea what the answer is here, but what do we want to be checking when we check the status of the database?  Is it good enough to say that there is at least one valid connection in the `READ` pool?  Or do we need to check the `WRITE` and `BACKGROUND` pools too?

Looking at https://github.com/broadinstitute/firecloud-develop/pull/2402/files, the pools are configured similarly enough that the check as-is is probably fine.  However, if the settings for the pools start to differ any meaningful way, such as a specific user/pw combo for writing, then we might want to modify the status check to verify that each pool is able to connect properly.  I'm fine leaving this as-is though so long as nobody can think of another reason why we should check each pool.

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/EffectivePolicyTable.scala
```diff
@@ -0,0 +1,24 @@
...
+import scalikejdbc._
+
+final case class EffectivePolicyPK(value: Long) extends DatabaseKey
+final case class EffectivePolicyRecord(id: EffectivePolicyPK,
+                                       resourceId: ResourcePK,
+                                       sourcePolicyId: PolicyPK,
+                                       groupId: GroupPK,
+                                       public: Boolean)
+
+object EffectivePolicyTable extends SQLSyntaxSupportWithDefaultSamDB[EffectivePolicyRecord] {
```
gpolumbo-broad: perhaps a nit, but going back and forth between schemas and debugging queries, I would rather see these names match, in other words, rename the object `EffectiveResourcePolicyTable`.  I think it was named shorter because it's less verbose and Policies of course only apply to resources, so it is effectively communicative.  I don't disagree with that, but my opinion here is that fudging the name is more likely to cause confusion than to help when debugging queries.

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/FlatGroupMemberTable.scala
```diff
@@ -0,0 +1,42 @@
...
+
+// lastGroupMembershipElement is added so it can be indexed as a performance optimization, it is always the last element of groupMembershipPath
+final case class FlatGroupMemberRecord(id: FlatGroupMemberPK,
+                                       groupId: GroupPK,
+                                       memberUserId: Option[WorkbenchUserId],
+                                       memberGroupId: Option[GroupPK],
+                                       groupMembershipPath: FlatGroupMembershipPath,
+                                       lastGroupMembershipElement: Option[GroupPK])
+object FlatGroupMemberTable extends SQLSyntaxSupportWithDefaultSamDB[FlatGroupMemberRecord] {
+  override def tableName: String = "SAM_GROUP_MEMBER_FLAT"
```
gpolumbo-broad: nit: again, preference is that the name of the object and the table match if possible.  This is close enough to avoid confusion I think, but this flip/flopped naming seems to indicate that one is preferred to the other, so let's just make them match and avoid that one extra translation we need to do in our heads, however minor it may be.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -125,11 +124,21 @@ class ResourceService(
...
+  }
+
+  private def constructAccessPolicy(resourceType: ResourceType, resourceId: ResourceId, validatableAccessPolicy: ValidatableAccessPolicy, public: Boolean) = {
+    AccessPolicy(FullyQualifiedPolicyId(FullyQualifiedResourceId(resourceType.name, resourceId), validatableAccessPolicy.policyName),
+      validatableAccessPolicy.emailsToSubjects.values.flatten.toSet,
+      generateGroupEmail(),
+      validatableAccessPolicy.roles,
+      validatableAccessPolicy.actions,
+      validatableAccessPolicy.descendantPermissions,
+      public
```
gpolumbo-broad, line 139: L128: "can't set public at create time"

Why?  Is this something we intend to change/fix?

When do we expect to use `constructAccessPolicy` in lieu of directly creating an instance of `AccessPolicy`?  Given the statement about policies not being permitted to be public at create time, should we just not allow the parameter to be passed to `constructAccessPolicy` and just always set `AccessPolicy.public = false`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -211,10 +220,10 @@ class ResourceService(
       _ <- checkNoChildren(resource, samRequestContext).unsafeToFuture()
 
       // remove from cloud extensions first so a failure there does not leave ldap in a bad state
```
gpolumbo-broad:
```suggestion
      // remove from cloud first so a failure there does not leave sam in a bad state
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DatabaseSupport.scala
```diff
@@ -19,12 +20,54 @@ trait DatabaseSupport {
...
+  /**
+    * Run a database transaction with isolation level set to serializable.
+    * See https://www.postgresql.org/docs/9.6/transaction-iso.html.
+    * Use this when you need to be sure there are no concurrent changes that affect the outcome of databaseFunction.
+    * When a transaction fails due to a serialization failure it may be retried.
+    *
+    * Serializable transactions are used primarily by updates to the hierarchical group and resource/policy models
+    * to avoid race conditions where there are concurrent changes to those structures.
+    *
+    * Example: group A contains group B contains group C, add user U to C at the same time as removing A from B.
```
gpolumbo-broad:
```suggestion
    * Example: group A contains group B contains group C, add user U to C at the same time as removing B from A.
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DatabaseSupport.scala
```diff
@@ -19,12 +20,54 @@ trait DatabaseSupport {
...
+  }
+
+  /**
+    * Attempts to run a serializable transaction, retrying any serialization failures with a total number
+    * of maxTries, doubling the sleep duration between each try.
+    */
+  private def attemptSerializableTransaction[A](samRequestContext: SamRequestContext, dbQueryName: String, transactionIO: IO[A], maxTries: Int, trialNumber: Int = 1, sleepDuration: FiniteDuration = 10 millis)(implicit timer: Timer[IO], cs: ContextShift[IO]): IO[A] = {
+    writeDbRef.runDatabaseIO(dbQueryName, samRequestContext, transactionIO, cs, Map("trial" -> AttributeValue.longAttributeValue(trialNumber.longValue()))).handleErrorWith {
+      case psqlE: PSQLException if psqlE.getSQLState == PSQLStateExtensions.SERIALIZATION_FAILURE && trialNumber < maxTries =>
+        timer.sleep(addJitter(sleepDuration, sleepDuration/2)) *> attemptSerializableTransaction(samRequestContext, dbQueryName, transactionIO, maxTries, trialNumber+1, sleepDuration)
```
gpolumbo-broad, line 68: Do we care at all about logging retry attempts?  I could see where it might be very noisy, but might also be helpful in diagnosing issues.  Maybe at least a DEBUG level?

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
```diff
@@ -137,7 +136,7 @@ object TestSupport extends TestSupport {
 
   def genSamRoutesWithDefault(implicit system: ActorSystem, materializer: Materializer): SamRoutes = genSamRoutes(genSamDependencies(), UserInfo(OAuth2BearerToken(""), genWorkbenchUserId(System.currentTimeMillis()), defaultUserEmail, 3600))
 
-  lazy val dbRef = DbReference.init(config.as[LiquibaseConfig]("liquibase"), DatabaseNames.Foreground)
+  lazy val dbRef = DbReference.init(config.as[LiquibaseConfig]("liquibase"), DatabaseNames.Read, TestSupport.blockingEc)
```
gpolumbo-broad, line 146: should this be `readDbRef` maybe?  I'm looking at line 142 below and we're using the read pool to truncate the database.  In test world, this is fine, but things might start getting confusing in the tests if we're using read and write pools willy nilly.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,324 @@
...
+  }
+
+  protected def insertGroupMemberPKs(groupId: GroupPK, memberGroupPKs: List[GroupPK], memberUserIds: List[WorkbenchUserId])(implicit session: DBSession): Int = {
+    if (memberGroupPKs.isEmpty && memberUserIds.isEmpty) {
+      0
+    } else {
+      val insertCount = Try {
+        insertGroupMembersIntoHierarchical(groupId, memberGroupPKs, memberUserIds)
+      }.recover {
+        case duplicateException: PSQLException if duplicateException.getSQLState == PSQLStateExtensions.UNIQUE_VIOLATION => 0
```
gpolumbo-broad: I always get antsy when swallowing exceptions.  If inserting multiple rows and one of them causes problems, it might be nice to have logged that error so we know which row is the culprit.

Furthermore, what triggers transaction failures?  Does the fact that a transaction was thrown, handled, and swallowed force any applicable transactions to fail?  Or do we need to keep bubbling up the exception and or manually trigger the failure/rollback?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -674,64 +740,81 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
-    val groupId = samsql"${workbenchGroupIdentityToGroupPK(id)}".map(rs => rs.long(1)).single().apply().getOrElse {
+    val groupId = samsql"${workbenchGroupIdentityToGroupPK(id)}".map(rs => rs.get[GroupPK](1)).single().apply().getOrElse {
       throw new WorkbenchException(s"Group for policy [$id] not found")
     }
-
-    samsql"delete from ${GroupMemberTable as gm} where ${gm.groupId} = ${groupId}".update().apply()
-
-    insertGroupMembers(GroupPK(groupId.toLong), memberList)
+    removeAllGroupMembers(groupId)
+    insertGroupMembers(groupId, memberList)
```
gpolumbo-broad, line 783: This insert has the potential to fail due to `PSQLStateExtensions.UNIQUE_VIOLATION` and just `0` will be returned.  In that case we would have just deleted all the group members and then "successfully" inserted 0 members?  I guess it all boils down to what triggers the transaction failure.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,324 @@
...
+    * Inserting a user/group into a group requires 2 inserts into the flat group model:
+    *   1) insert direct membership - group_id is given group, member_group_id/member_user_id is give member, path contains only given group
+    *   2) insert indirect memberships - insert a record for every record where member_group_id is the given group, group_id is the same, member_group_id/member_user_id is the member, path is the same with given group appended
+    *
+    * Inserting a subgroup into a group requires an additional insert to connect the subgroup's lower hierarchy:
+    * insert a record for every record where
+    *   - member_group_id is the given group id (all containers of given group), call these the head
+    *   - group_id is the given group id (all members of given group), call these the tail
+    * new record group_id is the head group_id, member id is the tail member id, path is the head path + tail path
+    *
```
gpolumbo-broad, line 155: I _think_ I understand what the structure is for the flattened table....but I have read this comment several times and I still haven't figured out what you mean here.  I think explanation here is valuable, I think we just need to wordsmith and/or include some examples.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
dvoet: no, `source` as in where did this effective policy come from

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
dvoet: `listPublicAccessPolicies` is deprecated so I don't think we should do this at all

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: I can't think of a reason that makes it worth the effort to do separate status checks for each pool at this time

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: setting a policy public requires an access check (not everyone has permission to do so). Will we ever want to implement that? I don't know.

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/DatabaseSupport.scala
dvoet: it does get captured in the trace... do you think that is good enough?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
dvoet: would adding `on conflict do nothing` to the query instead make you feel better?

dvoet: actually `on conflict do nothing` is functionally correct, as implemented it is probably a bug

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
```diff
@@ -0,0 +1,119 @@
...
+            <column name="resource_action_id" type="BIGINT">
+                <constraints nullable="false" foreignKeyName="FK_SEPA_ACTION" referencedTableName="SAM_RESOURCE_ACTION" referencedColumnNames="id" primaryKey="true"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="populate_effective_policies" runInTransaction="true">
+        <sql stripComments="true">
+            with recursive
+                ancestor_resource(resource_id, base_resource_id) as (
```
s-rubenstein, line 65: Just to make sure I am understanding this set of the query correctly, this is pulling the resource_id from `resource.id` in the resource table in the bit before the union, and then the base_resource_id is getting pulled from `parentResource.resource_parent_id` in the bit after the union?

What does `with recursive` do? Does it mean do this recursively for each resource in the table?

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
s-rubenstein: Nit: Out of curiosity, should `sam_flattened_role` be in all caps here?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
dvoet: `tail.groupMembershipPath` is the column name. This is the equivalent of `ary[len(ary)]` to get the last element of an array `ary` with a 1 based index

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_flat_group_membership.xml
```diff
@@ -0,0 +1,67 @@
...
+        </createIndex>
+    </changeSet>
+    <changeSet logicalFilePath="dummy" author="dvoet" id="index_gmf_member_group_id">
+        <createIndex tableName="SAM_GROUP_MEMBER_FLAT" indexName="IDX_GMF_MEMBER_GROUP_ID">
+            <column name="member_group_id"/>
+        </createIndex>
+    </changeSet>
+    <changeSet logicalFilePath="dummy" author="dvoet" id="index_gmf_group_membership_path">
+        <sql>
+            CREATE INDEX IDX_GMF_MEMBERSHIP_PATH ON SAM_GROUP_MEMBER_FLAT USING GIN (group_membership_path)
```
s-rubenstein, line 39: Why can't this be done using the liquibase syntax for createIndex?

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
dvoet: I take that back since removing the `public` column from `SAM_EFFECTIVE_RESOURCE_POLICY`

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
dvoet: In unit tests there really is not a difference between read and write pools. Ideally I would not even have it. But I also want to have `DatabaseNames` enum and `DbReference.init` to use it. So the situation is a little messy and I favor having more mess on the test side than the production side (i.e. I don't want to add a new database name just for tests).

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
dvoet: added `on conflict do nothing` so there will be no failure

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+    samsql"""delete from ${EffectivePolicyTable as ep}
+               using ${ResourceTable as r}
+               where ${r.name} = ${resource.resourceId}
+               and ${r.resourceTypeId} = ${resourceTypePKsByName(resource.resourceTypeName)}
+               and ${r.id} = ${ep.resourceId}
+              """.update().apply()
+  }
+
+  /**
+    * Delete all inherited effective policies on the specified resource and descendants.
```
s-rubenstein: This is different than Marc's understanding below, and what I understand the method name to be. Say we have a resource A, and a child B. Does this:
A) Delete all policies on just children (AKA just the policies on B)
B) Delete all policies on the object and its children (AKA policies on A and B)

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+
+    samsql"""with recursive
+        ${ancestorsRecursiveQuery(childResourcePK, ancestorResourceTable)},
+        ${descendantsRecursiveQuery(childResourcePK, descendantResourceTable)}
+
+        insert into ${EffectivePolicyTable.table} (${epCol.resourceId}, ${epCol.sourcePolicyId}, ${epCol.groupId}, ${epCol.public})
+        select ${descendantResource.resourceId}, ${p.id}, ${p.groupId}, ${p.public}
+        from ${descendantResourceTable as descendantResource},
+        ${ancestorResourceTable as ancestorResource}
+        join ${PolicyTable as p} on ${ancestorResource.resourceId} = ${p.resourceId}""".update().apply()
```
s-rubenstein, line 184: Why doesn't this need the 'on conflict do nothing'?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
s-rubenstein: Isn't it pulling childResourcePK though? So shouldn't this be the child resource? (I may not be clearly following this, though)

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
```diff
@@ -0,0 +1,363 @@
...
+    *         inherited             |  policyRole.descendantsOnly  |  flattenedRole.descendantsOnly  |  result
+    *            T                  |             T                |                T                |    T
+    *            T                  |             T                |                F                |    T
+    *            T                  |             F                |                T                |    T
+    *            T                  |             F                |                F                |    F
+    *            F                  |             T                |                T                |    F
+    *            F                  |             T                |                F                |    F
+    *            F                  |             F                |                T                |    F
+    *            F                  |             F                |                F                |    T
+    *
```
s-rubenstein, line 392: Just to make sure I am understanding the logic behind this table, it is:
`I xor not(PR or FR)`

If so, that may be a useful thing to put in the comment.

s-rubenstein: That may also be a simpler way to write the query? If that is the logic?

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
gpolumbo-broad: Can we leave a comment explaining this then?  I think it is likely for someone to raise a question about this in the future.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
gpolumbo-broad: Right, I guess that's my point.  You just removed all group members, then `insertGroupMembers` doesn't fail but also didn't insert anything, so at the end of the function, all that we did was delete all the group members, which sounds like a bug to me.

marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -63,10 +72,22 @@ class PostgresAccessPolicyDAO(protected val dbRef: DbReference,
...
+    result <* reloadResourceTypePKs(samRequestContext)
+  }
+
+  private def reloadResourceTypePKs(samRequestContext: SamRequestContext): IO[Unit] = {
+    readOnlyTransaction("loadResourceTypePKsByName", samRequestContext) { implicit session =>
+      val rt = ResourceTypeTable.syntax("rt")
+      samsql"""select ${rt.result.name}, ${rt.result.id} from ${ResourceTypeTable as rt}""".map(rs =>
+        rs.get[ResourceTypeName](rt.resultName.name) -> rs.get[ResourceTypePK](rt.resultName.id)).list().apply().toMap
+    }.map { results =>
+      results.foreach(kv => resourceTypePKsByName.put(kv._1, kv._2))
```
s-rubenstein: Could this be done with a `resourceTypePKsByName.addAll(results)`? (I just know there is a hesitance sometimes to avoid `.foreach` sometimes, and that could be a clean way to look at doing that in a single way (because then you could do a `.filterInPlace` chained from that.)

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,366 @@
...
+    *   ((H, T, {H}), (A, T, {A,H}), (B, T, {B,H}))
+    * Group | Member Group | Path
+    * ------|--------------|------
+    *   H   |      X       | {H,T}
+    *   H   |      Y       | {H,T}
+    *   A   |      X       | {A,H,T}
+    *   A   |      Y       | {A,H,T}
+    *   B   |      X       | {B,H,T}
+    *   B   |      Y       | {B,H,T}
+    *
```
s-rubenstein, line 155: This is a really helpful comment, thank you!

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
```diff
@@ -0,0 +1,366 @@
...
+    val rt = ResourceTypeTable.syntax("rt")
+    val r = ResourceTable.syntax("r")
+    val p = PolicyTable.syntax("p")
+    val policyGroupPKStatement = samsqls"""select ${p.groupId} as group_id
+               from ${ResourceTypeTable as rt}
+               join ${ResourceTable as r} on ${rt.id} = ${r.resourceTypeId}
+               join ${PolicyTable as p} on ${r.id} = ${p.resourceId}
+               where (${rt.name}, ${r.name}, ${p.name}) in ($memberPolicyIdTuples)"""
+
+    // there are 4 scenarios: there are both member groups and policies, only one or the other or neither
```
s-rubenstein, line 320: Optional nit: This feels explained in the match case below, so I don't know if we need the comment. (And could get out of date if we added a different type)

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_flat_group_membership.xml
dvoet: I think because of `USING GIN`

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
dvoet: https://www.postgresql.org/docs/9.6/queries-with.html

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
dvoet: doesn't matter

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
dvoet: @marctalbott figured this out, if there were XOR in SQL it might be better but there isn't :(

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
dvoet: there isn't a unique constraint that would trigger a conflict

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresGroupDAO.scala
dvoet: I hear you but I am inclined to leave it because I feel like the `case (true, true)` stuff is messy and needs explaination

dvoet:
jenkins retest - strange ivy thing on rawls

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala
s-rubenstein: Yeah, the query looks good, but I was wondering about putting it above the truth table in the comment, just to make it more clear at a first glance. Although since we can't represent the sql that way, I'm okay with leaving it for now.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
s-rubenstein: Neat! Thank you!

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml
s-rubenstein: If it doesn't matter, would you mind moving it to caps for consistency?

s-rubenstein (approved):
LGTM at this point, don't think any of my nits should hold this up.

zbedo:
👍

dvoet:
jenkins retest

PR #494, opened 2021-01-04
WIP IA-2370 Disable user

Qi77Qi:
Ticket: https://broadworkbench.atlassian.net/browse/IA-2370

Before the break, we decided on the approach where leo will publish the bad user and sam will subscribe to the topic

Depends on https://github.com/broadinstitute/firecloud-develop/pull/2372

Test:
Manually sent a message to `projects/broad-dsde-dev/topics/terra-cryptomining` topic, and sam receives the message properly.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -792,8 +792,8 @@ resourceTypes = {
 
 
 distributedLock {
-  retryInterval: 5s
-  maxRetry: 10
+  retryInterval = 5s
+  maxRetry = 10
 }
 
 blockedEmailDomains = ["qwiklabs-gsuite.net", "qwiklabs.net"]
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -2,7 +2,6 @@ package org.broadinstitute.dsde.workbench.sam
 
 import java.io.File
 import java.net.URI
-
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import cats.data.NonEmptyList
@@ -12,17 +11,20 @@ import cats.implicits._
 import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
 import com.unboundid.ldap.sdk._
+import fs2.concurrent.InspectableQueue
 import io.chrisdavenport.log4cats.StructuredLogger
 import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
+
 import javax.net.SocketFactory
 import javax.net.ssl.SSLContext
 import org.broadinstitute.dsde.workbench.dataaccess.PubSubNotificationDAO
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.{Json, Pem}
 import org.broadinstitute.dsde.workbench.google2.util.DistributedLock
 import org.broadinstitute.dsde.workbench.google.{GoogleDirectoryDAO, GoogleKmsInterpreter, GoogleKmsService, HttpGoogleDirectoryDAO, HttpGoogleIamDAO, HttpGoogleProjectDAO, HttpGooglePubSubDAO, HttpGoogleStorageDAO}
-import org.broadinstitute.dsde.workbench.google2.{GoogleFirestoreInterpreter, GoogleStorageInterpreter, GoogleStorageService}
+import org.broadinstitute.dsde.workbench.google2.{Event, GoogleFirestoreInterpreter, GoogleStorageInterpreter, GoogleStorageService, GoogleSubscriber, SubscriberConfig}
 import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchException}
 import org.broadinstitute.dsde.workbench.sam.api.{SamRoutes, StandardUserInfoDirectives}
+import org.broadinstitute.dsde.workbench.sam.backgroundProcess.{CryptominingSubscriber, CryptominingUserMessage}
 import org.broadinstitute.dsde.workbench.sam.config.{AppConfig, GoogleConfig, IdentityConcentratorConfig}
 import org.broadinstitute.dsde.workbench.sam.db.DatabaseNames.DatabaseName
 import org.broadinstitute.dsde.workbench.sam.db.{DatabaseNames, DbReference}
@@ -37,6 +39,7 @@ import org.broadinstitute.dsde.workbench.util.DelegatePool
 import org.broadinstitute.dsde.workbench.util2.ExecutionContexts
 import org.http4s.client.blaze.BlazeClientBuilder
 import scalikejdbc.config.DBs
+import CryptominingSubscriber.cryptominingUserMessageDecoder
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -61,9 +64,9 @@ object Boot extends IOApp with LazyLogging {
 
     val schemaDAO = new JndiSchemaDAO(appConfig.directoryConfig, appConfig.schemaLockConfig)
 
-    val appDependencies = createAppDependencies(appConfig)
+    implicit val loggerIO: StructuredLogger[IO] = Slf4jLogger.getLogger[IO]
 
-    appDependencies.use { dependencies => // this is where the resource is used
+    createAppDependencies(appConfig).use { dependencies => // this is where the resource is used
       for {
         _ <- IO.fromFuture(IO(schemaDAO.init())).onError {
           case e: WorkbenchException =>
@@ -81,11 +84,11 @@ object Boot extends IOApp with LazyLogging {
 
         _ <- dependencies.cloudExtensionsInitializer.onBoot(dependencies.samApplication)
 
-        binding <- IO.fromFuture(IO(Http().newServerAt("0.0.0.0", 8080).bind(dependencies.samRoutes.route))).onError {
+        httpServer = IO.fromFuture(IO(Http().newServerAt("0.0.0.0", 8080).bind(dependencies.samRoutes.route))).onError {
           case t: Throwable => IO(logger.error("FATAL - failure starting http server", t)) *> IO.raiseError(t)
-        }
-        _ <- IO.fromFuture(IO(binding.whenTerminated))
-        _ <- IO(system.terminate())
+        }.void
+        allStreams = fs2.Stream.eval(httpServer) :: dependencies.cryptominingSubscriberProcess
+        _ <- fs2.Stream.emits(allStreams).covary[IO].parJoin(allStreams.length).compile.drain
       } yield ()
     }
   }
@@ -112,7 +115,7 @@ object Boot extends IOApp with LazyLogging {
   }
 
   private[sam] def createAppDependencies(
-      appConfig: AppConfig)(implicit actorSystem: ActorSystem): cats.effect.Resource[IO, AppDependencies] =
+      appConfig: AppConfig)(implicit actorSystem: ActorSystem, logger: StructuredLogger[IO]): cats.effect.Resource[IO, AppDependencies] =
     for {
       (foregroundDirectoryDAO, foregroundAccessPolicyDAO, _, registrationDAO) <- createDAOs(appConfig, DatabaseNames.Foreground, appConfig.directoryConfig.connectionPoolSize, "foreground")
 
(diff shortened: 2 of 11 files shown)


src/main/resources/reference.conf
```diff
@@ -792,8 +792,8 @@ resourceTypes = {
 
 
 distributedLock {
-  retryInterval: 5s
-  maxRetry: 10
+  retryInterval = 5s
```
Qi77Qi, line 795: was this working before? 🤔

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -81,11 +84,11 @@ object Boot extends IOApp with LazyLogging {
 
         _ <- dependencies.cloudExtensionsInitializer.onBoot(dependencies.samApplication)
 
-        binding <- IO.fromFuture(IO(Http().newServerAt("0.0.0.0", 8080).bind(dependencies.samRoutes.route))).onError {
+        httpServer = IO.fromFuture(IO(Http().newServerAt("0.0.0.0", 8080).bind(dependencies.samRoutes.route))).onError {
           case t: Throwable => IO(logger.error("FATAL - failure starting http server", t)) *> IO.raiseError(t)
-        }
-        _ <- IO.fromFuture(IO(binding.whenTerminated))
-        _ <- IO(system.terminate())
+        }.void
```
andy7i, line 89:
```
  /*
   * Ignores the result of this IO.
   */
  def void
```

(I'm a bit confused) isn't httpServer just IO[Unit]? Why would that be useful to pass into the next line?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -81,11 +84,11 @@ object Boot extends IOApp with LazyLogging {
...
+        allStreams = fs2.Stream.eval(httpServer) :: dependencies.cryptominingSubscriberProcess
+        _ <- fs2.Stream.emits(allStreams).covary[IO].parJoin(allStreams.length).compile.drain
```
andy7i, line 91: not knowing a lot about streams, could you summarize what is happening here? (and maybe I should read up on Streams)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -507,11 +507,9 @@ class ResourceService(
           // resources with auth domains logically can't have public policies but also technically allowing them poses a problem
           // because the logic for public resources is different. However, sharing with the auth domain should have the
           // exact same effect as making a policy public: anyone in the auth domain can access.
-          if (public)
```
andy7i: why was this removed? based on the function description seems useful to have

src/main/scala/org/broadinstitute/dsde/workbench/sam/backgroundProcess/CryptominingSubscriber.scala
```diff
@@ -0,0 +1,47 @@
...
+import fs2.Stream
+import io.chrisdavenport.log4cats.StructuredLogger
+import io.circe.Decoder
+import org.broadinstitute.dsde.workbench.google2.{Event, GoogleSubscriber}
+import org.broadinstitute.dsde.workbench.model.WorkbenchUserId
+import org.broadinstitute.dsde.workbench.sam.service.UserService
+import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
+
+// Each instance of Sam will run a subscriber, since they all subscribe to the same topic via same subscription,
+// messages will be distributed among subscribers instead of distributed to all of them, which is what we want.
```
andy7i, line 14: thanks for the explanation 👍

src/main/scala/org/broadinstitute/dsde/workbench/sam/backgroundProcess/CryptominingSubscriber.scala
```diff
@@ -0,0 +1,47 @@
...
+      // To conditionally acking message, we'll need to config a dead letter topic where messages reach max retry can go to,
+      // which we're not doing for this use case.
+    _ <- IO{event.consumer.ack()}
+    } yield ()
+  }
+
+  private def messageResponder(
+                                         message: CryptominingUserMessage
+                                       ): IO[Unit] = IO.fromFuture(IO(userService.disableUser(message.userSubjectId, SamRequestContext(None)))).void
+}
```
andy7i, line 40: is it possible to make a unit test for this? secondly, would it be worth it?

andy7i (commented):
comments / questions

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
Qi77Qi: This is a way to run various processes concurrently. In this case, it's the cryptomining subscriber, and the main sam http server.

For your question above, because the Streams here are expected to be all `Stream[IO, Unit]`, hence we're using `void` up there. Without `void`, it'll return `IO[Http.ServerBinding]`, which can't be combined with other Streams for running concurrently.

src/main/scala/org/broadinstitute/dsde/workbench/sam/backgroundProcess/CryptominingSubscriber.scala
```diff
@@ -0,0 +1,47 @@
...
+import io.chrisdavenport.log4cats.StructuredLogger
+import io.circe.Decoder
+import org.broadinstitute.dsde.workbench.google2.{Event, GoogleSubscriber}
+import org.broadinstitute.dsde.workbench.model.WorkbenchUserId
+import org.broadinstitute.dsde.workbench.sam.service.UserService
+import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
+
+// Each instance of Sam will run a subscriber, since they all subscribe to the same topic via same subscription,
+// messages will be distributed among subscribers instead of distributed to all of them, which is what we want.
+class CryptominingSubscriber(subscriber: GoogleSubscriber[IO, CryptominingUserMessage],
```
dvoet, line 15: I don't think we should mention cryptomining. This is a general purpose mechanism for any automated external process to tell sam to disable a user. Cryptomining is just the first use of this.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -133,7 +136,23 @@ object Boot extends IOApp with LazyLogging {
...
+      }
+    } yield createAppDepenciesWithSamRoutes(appConfig, cloudExtensionsInitializer, userService, cryptominingSubscriberProcess, foregroundAccessPolicyDAO, foregroundDirectoryDAO, identityConcentrator)
+
+  private def initCryptominingSubscriber(cryptominingSubscriberConfig: SubscriberConfig, userService: UserService)
+                                        (implicit logger: StructuredLogger[IO]): effect.Resource[IO, List[fs2.Stream[IO, Unit]]] = for {
+    subscriberQueue <- cats.effect.Resource.liftF(InspectableQueue.bounded[IO, Event[CryptominingUserMessage]](200))
+    cryptominingSubscriber <- GoogleSubscriber.resource[IO, CryptominingUserMessage](cryptominingSubscriberConfig, subscriberQueue)
+  } yield {
+    val subscriber = new CryptominingSubscriber(cryptominingSubscriber, userService)
+    List(subscriber.process, fs2.Stream.eval(cryptominingSubscriber.start))
```
dvoet, line 154: can you explain why this is a list of streams?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -133,7 +136,23 @@ object Boot extends IOApp with LazyLogging {
...
+
+      cryptominingSubscriberProcess <- appConfig.googleConfig match {
+        case Some(value) => initCryptominingSubscriber(value.googleServicesConfig.cryptominingSubscriber, userService)
+        case None => effect.Resource.liftF[IO, List[fs2.Stream[IO, Unit]]](IO(List.empty))
+      }
+    } yield createAppDepenciesWithSamRoutes(appConfig, cloudExtensionsInitializer, userService, cryptominingSubscriberProcess, foregroundAccessPolicyDAO, foregroundDirectoryDAO, identityConcentrator)
+
+  private def initCryptominingSubscriber(cryptominingSubscriberConfig: SubscriberConfig, userService: UserService)
+                                        (implicit logger: StructuredLogger[IO]): effect.Resource[IO, List[fs2.Stream[IO, Unit]]] = for {
+    subscriberQueue <- cats.effect.Resource.liftF(InspectableQueue.bounded[IO, Event[CryptominingUserMessage]](200))
```
dvoet, line 150: why 200?

src/main/scala/org/broadinstitute/dsde/workbench/sam/backgroundProcess/CryptominingSubscriber.scala
```diff
@@ -0,0 +1,47 @@
...
+                             userService: UserService
+                            )(implicit logger: StructuredLogger[IO], cs: ContextShift[IO]) {
+  val process: Stream[IO, Unit] = subscriber.messages
+    .parEvalMapUnordered(2)(messageHandler) // concurrency is low since we don't need high performance in disabling cryptomining users.
```
dvoet: why not `evalMap` and forget the question of concurrency? There are other sam instances too so this is never single threaded anyway.

src/main/scala/org/broadinstitute/dsde/workbench/sam/backgroundProcess/CryptominingSubscriber.scala
```diff
@@ -0,0 +1,47 @@
...
+    for {
+      res <- messageResponder(event.msg).attempt
+      _ <- res match {
+        case Left(e)  => logger.error(e)("Fail to process pubsub message")
+        case Right(_) => IO.unit
+      }
+      // acking the message whether or not we successfully processed the message since this is not essential part of
+      // terra functionality. That is, if we didn't disable a bad user, it's no big deal.
+      // To conditionally acking message, we'll need to config a dead letter topic where messages reach max retry can go to,
+      // which we're not doing for this use case.
```
dvoet, line 32: I disagree with this. Maybe for cryptomining it is not a big deal but I think the general contract for sam should be stronger. I think we should only ack on success.

dvoet (changes requested):
There are a few detailed comments. In general though, this is a departure from how most of the sam code works. Since this team needs to maintain it I am going to discuss it with them.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
Qi77Qi: good catch! this is probably from bad merge

src/main/scala/org/broadinstitute/dsde/workbench/sam/backgroundProcess/CryptominingSubscriber.scala
Qi77Qi: hmm..think my perspective is a bit different...this is a subscriber specifically for `terra-cryptomining` topic (which I was assuming we can potentially have other messages related to cryptomining activity other than disabling users), we can't really use this for another topic if there's another use case for disabling users.

src/main/scala/org/broadinstitute/dsde/workbench/sam/backgroundProcess/CryptominingSubscriber.scala
Qi77Qi: I made it 2 just so if in the future there's need to increase concurrency, people don't have to figure out how to use `parEvalMapUnordered`...happy to change this to no concurrency

src/main/scala/org/broadinstitute/dsde/workbench/sam/backgroundProcess/CryptominingSubscriber.scala
Qi77Qi: this ties back to https://github.com/broadinstitute/sam/pull/494/files/75635f99381bdba2035a10e72098ac34e11fd052#r552697473...

The reason why I'm choosing to ack blindly here is because if we choose to nack, then we should set up max retry, which means we should set up a deadlettertopic for receiving messages that have exceeded max retry, otherwise, nacked messages will be resent forever (I'm not certain if it's forever or there's a default max at some point, but from what we see in leo, it looked like forever). I didn't want to introduce more complexity if not necessary, but if you think it's worth it, I'm happy to add those

src/main/scala/org/broadinstitute/dsde/workbench/sam/backgroundProcess/CryptominingSubscriber.scala
Qi77Qi: what is the test in your mind trying to validate? if it's just disabling user, I think it should be covered in userService's unit test

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
Qi77Qi: `cryptominingSubscriber.start` is for starting up the underlying google subscriber, `subscriber.process` is for actually running the stream that processes messages that the subscriber receives

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
Qi77Qi: it's just best guess to start. It's the number of messages that the subscriber will queue up before back pressure happens.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
andy7i: OH ok. I think I also missed the `eval` in `fs2.Stream.eval(httpServer)` -- which would allow the stream to evaluate the httpServer IO.

src/main/scala/org/broadinstitute/dsde/workbench/sam/backgroundProcess/CryptominingSubscriber.scala
andy7i: that makes sense. Would it be useful to test this subscriber? (I don't know how we would test it either, but just wanted to check and ask to understand why we would or wouldn't want to test this)

Qi77Qi: anything specific you want to test about the subscriber? (asking in case I'm missing something)

Subscriber has unit test in google2 to make sure it can receive message properly.
The only thing worth testing here is the actual handler, but the handler here is quite simple cuz it's just calling userSerivce.disableUser, which I assume is covered by userSerivce's unit test

src/main/resources/reference.conf
gpolumbo-broad: https://github.com/lightbend/config/blob/master/HOCON.md#key-value-separator

Yes

gpolumbo-broad (changes requested):
Hi Qi, we discussed this PR pretty extensively as a team and we are apprehensive about introducing further usage of cats or fs2 unless it is clearly justified.  This was not an easy decision for us to make.  We want to learn and use new technology, but paramount to that we want the code to be easily readable and maintainable for engineers of all seniority levels.  Our experience has shown us a steep learning curve with these libraries and the pure FP principles behind them.  Given these points and the direction of the organization towards more Java work, we would prefer to not introduce the usage of fs2 streams at this time and instead implement this feature using Akka Actors so that it is in alignment with existing code in Sam.

We are very appreciative that you took the time to implement this feature for us and we know that this feature is important to the Leo team and the health of the platform in general.  Our team can take on the reimplementation of this feature using Akka Actors, unless you or your team is particularly keen on doing it, in which case you are welcome to take it on.   Please let us know how you would like to proceed so that we can prioritize this work accordingly.

On a personal level I want to acknowledge your expertise in FP and the training and guidance you are providing to all of us.  You are clearly a leader in this space and I hope that you understand that even though our team is requesting this change, we all appreciate your FP knowledge and expertise.

Qi77Qi:
@gpolumbo-broad thanks for the reply. in that case, I'll leave the work to you guys...

PR #500, opened 2021-02-02
[CA-1146] skip tests - readme - UML key

andy7i:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Group - Create, delete, read, list, add/remove users and groups. Nested groups a
 ### UML Model
 ![Sam Model](model.png)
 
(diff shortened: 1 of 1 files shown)


gpolumbo-broad (approved):
I'm slightly worried that this is a 3rd party site that might change at any given time.  I might point at something like the wikipedia page https://en.wikipedia.org/wiki/Class_diagram which is less likely to go away or become unreachable.  That said, it's perhaps less digestible too.

andy7i:
@gpolumbo-broad that's a very legit concern. I would propose that we cross that bridge once we get to it? Maybe the wikipedia page will have improved by then 🙃

dvoet (approved):
ok to use admin to merge

andy7i:
using admin permissions to merge without tests (this is a doc change so tests aren't needed)

PR #504, opened 2021-02-08
[CA-1149] Add initial pass at user disable monitor

s-rubenstein:
Ticket: <Link to Jira ticket>

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
```diff
@@ -0,0 +1,189 @@
...
+      // start the process by pulling a message and sending it back to self
+      pubSubDao.pullMessages(pubSubSubscriptionName, 1).map(_.headOption) pipeTo self
+
+    case Some(message: PubSubMessage) =>
+      logger.debug(s"received disable user message: $message")
+      import Tracing._
+      trace("DisableUsersMonitor-PubSubMessage") { span =>
+        userService
+          .disableUser(message.contents.parseJson.convertTo[WorkbenchUserId], samRequestContext = SamRequestContext(Option(span)))
+      }
```
dvoet: I think you are missing something here to message back to the actor to ack the message. Something ending in `pipeTo self`

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
```diff
@@ -0,0 +1,189 @@
...
+        Stop
+    }
+
+}
+
+object DisableUsersMonitor {
+  case object StartMonitorPass
+
+  sealed abstract class DisableUserResult(ackId: String)
+  final case class ReportMessage(value: Map[WorkbenchEmail, Seq[SyncReportItem]], ackId: String) extends DisableUserResult(ackId = ackId)
```
dvoet: I think `SyncReportItem` is specific to google group sync. The `value` here probably needs to be different and constructed at the end of `case Some(message: PubSubMessage)`.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
```diff
@@ -0,0 +1,189 @@
...
+  final case class FailToSynchronize(t: Throwable, ackId: String) extends DisableUserResult(ackId = ackId)
```
dvoet: `FailToDisable`?

dvoet (changes requested):
some tests too please

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
```diff
@@ -0,0 +1,192 @@
...
+
+        case regrets: Throwable => throw regrets
+      }
+
+    case Status.Failure(t) => throw t
+
+    case ReceiveTimeout =>
+      throw new WorkbenchException("DisableUsersMonitorActor has received no messages for too long")
+
+    case x => logger.info(s"unhandled $x")
```
gpolumbo-broad: I would make this message a little more descriptive, maybe: 

```suggestion
    case x => logger.info(s"Received unhandleable message in DisableUsersMonitor: $x")
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  private def handleDisableUserResponse(disableUserResponse: DisableUserResponse, ackId: String) = {
+    import Tracing._
```
gpolumbo-broad: inline import of `Tracing._` in a couple of places.  Why not import at the top of the file?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
```diff
@@ -0,0 +1,192 @@
...
+    trace("DisableUsersMonitor-ReportMessage") { _ =>
+      disableUserResponse.value match {
+        case Some(_) =>
+          // If we have gotten to this point, the disableUser method has run, so we can assume the user is disabled
+          logger.info(s"disabled user: ${disableUserResponse.userId} ")
+        case None =>
+          logger.info(s"user to disable not found: ${disableUserResponse.userId}")
```
gpolumbo-broad, line 150: If I am reading this right, if we end up here, things have probably gone wrong in a weird way, right?  The `attemptToDisableUser` claims that it succeeded, but there's something wrong with the Response object.  I don't know what we should do in this situation then.  Maybe just logging a message is fine and go ahead and ack the message.  Or maybe that's the wrong thing to do, maybe `attemptToDisableUser` thinks it succeeded, but it didn't and the user is still enabled?  I just want to make sure if we end up at this line of code, we understand the hows and whys and what it means and at the very least we're logging as meaningful a message as possible.

gpolumbo-broad (commented):
certainly do not need to wait for my thumb, I just wanted to take a look and had a few questions.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
s-rubenstein: This is what happens if `directoryDAO.loadUser(userId, samRequestContext)` returns None. So in that case, we are cool just logging that the user doesn't exist, and being happy. Because that means the user has been deleted or somesuch before we got to it.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
```diff
@@ -0,0 +1,188 @@
...
+        startOne()
+        Stop
+    }
+
+}
+
+object DisableUsersMonitor {
+  case object StartMonitorPass
+
+  sealed abstract class DisableUserResult(ackId: String)
```
dvoet: is there benefit to this having `ackId`? The common pattern is for this to be a `sealed trait` and used only as a marker so that when the compiler sees a match on `DisableUserResult` it can determine that the cases are comprehensive or not. But actually embedding the common field `ackId` does not buy us anything as we never treat these objects generically.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
dvoet: actually doing `$x` is a big no no because it writes only the exception message and loses critical information: the exception class and stack trace.

`logger.info(s"Received unhandleable message in DisableUsersMonitor", x)` is correct

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
```diff
@@ -0,0 +1,188 @@
...
+    case ReportMessage(disableUserResponse, ackId) =>
+      handleDisableUserResponse(disableUserResponse, ackId) pipeTo self
+
+    case FailToDisable(t, ackId) =>
+      t match {
+        case userNotFound: WorkbenchExceptionWithErrorReport if userNotFound.errorReport.statusCode.contains(StatusCodes.NotFound) =>
+          // this can happen if a user is deleted before the disable message is handled
+          // acknowledge it so we don't have to handle it again
+          acknowledgeMessage(ackId).map(_ => StartMonitorPass) pipeTo self
+          logger.info(s"user to disable not found: ${userNotFound.errorReport}")
```
dvoet: Is this actually true? It looks like disabling a user that does not exist just returns `None` from `userService.disableUser`

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitorSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+      ))
+
+    mockGooglePubSubDAO.publishMessages(topicName, Seq(userId.value))
+
+    // `eventually` now requires an implicit `Retrying` instance. When the statement inside returns future, it'll
+
+    // try to use `Retrying[Future[T]]`, which gets weird when we're using mockito together with it.
+    // Hence adding ascribing [Unit] explicitly here so that `eventually` will use `Retrying[Unit]`
+    eventually[Unit] {
+      assertResult(1) { mockGooglePubSubDAO.acks.size() }
```
dvoet: if you move this to after the call the `verify` you can remove the `()` below and `[Unit]` above and get rid of the comment

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitorSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+  it should "handle missing user" in {
+    system.actorOf(DisableUsersMonitorSupervisor.props(10 milliseconds, 0 seconds, mockGooglePubSubDAO, topicName, subscriptionName, 1, userService))
+    when(userService.disableUser(mockitoEq(userId), any[SamRequestContext])).thenReturn(Future.failed(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "not found"))))
+
+    mockGooglePubSubDAO.publishMessages(topicName, Seq(userId.value))
+
+    // `eventually` now requires an implicit `Retrying` instance. When the statement inside returns future, it'll
+
+    // try to use `Retrying[Future[T]]`, which gets weird when we're using mockito together with it.
+    // Hence adding ascribing [Unit] explicitly here so that `eventually` will use `Retrying[Unit]`
```
dvoet: copy pasta error

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
s-rubenstein: Yeah, I noticed that. I think maybe there's no special casing here.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
```diff
@@ -0,0 +1,178 @@
...
+
+}
+
+object DisableUsersMonitor {
+  case object StartMonitorPass
+
+  sealed abstract class DisableUserResult()
+  final case class DisableUserResponse(userId: WorkbenchUserId, value: Option[UserStatus])
+  final case class ReportMessage(value: DisableUserResponse, ackId: String) extends DisableUserResult()
+  final case class FailToDisable(t: Throwable, ackId: String) extends DisableUserResult()
```
dvoet:
```suggestion
  final case class DisableUserResponse(userId: WorkbenchUserId, value: Option[UserStatus])
  final case class ReportMessage(value: DisableUserResponse, ackId: String)
  final case class FailToDisable(t: Throwable, ackId: String)
```

dvoet: I don't think `DisableUserResult` is really doing anything for us

dvoet (approved):

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitorSpec.scala
```diff
@@ -0,0 +1,92 @@
+package org.broadinstitute.dsde.workbench.sam.google
```
marctalbott, line 1: Would it be possible/valuable to add a test that goes through the `FailToDisable` path? It doesn't seem like that path will really happen since `disableUser` always seems to return a successful Future, but I'm not positive

marctalbott (approved):
Looks good! I suggested one more test, but not sure if the test is viable/useful so consider it nonblocking

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitorSpec.scala
s-rubenstein: I don't think it will ever go through that path, since either `disableUser` will either return a successful future or a none. I could add a test which throws an exception just in case we add some exceptions to the endpoint? But that feels future-proof-y, and not necessarily correct for current behavior.

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/GoogleServicesConfig.scala
```diff
@@ -31,6 +31,11 @@ final case class GoogleServicesConfig(
     groupSyncTopic: String,
     groupSyncSubscription: String,
     groupSyncWorkerCount: Int,
```
marctalbott: Should there be a `disableUsersPubSubProject` value here too?

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/GoogleServicesConfig.scala
```diff
@@ -31,6 +31,11 @@ final case class GoogleServicesConfig(
...
+    disableUsersPollInterval: FiniteDuration,
+    disableUsersPollJitter: FiniteDuration,
+    disableUsersTopic: String,
+    disableUsersSubscription: String,
+    disableUsersWorkerCount: Int,
```
marctalbott: I also wouldn't mind seeing this  and the config for `groupSync` broken into separate `PubSubConfig` classes like we have with the `GoogleKeyCacheConfig`, especially if we'd need these same values for any future pubsub topics we might add to Sam

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/GoogleServicesConfig.scala
s-rubenstein: So, it wasn't used. I did a bit of searching though, and the only usage of the `groupSyncPubSubProject` was in setting up the `googlePubSubDao`, so maybe we should do a refactor of that in general.

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/GoogleServicesConfig.scala
s-rubenstein: Yeah, will do some abstraction into a separate config.

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

marctalbott (approved):

PR #506, opened 2021-02-12
PubSub Config proposal, wrap all pub sub config into single object

s-rubenstein:
Ticket: <Link to Jira ticket>

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -247,7 +247,7 @@ object Boot extends IOApp with LazyLogging {
       config.googleServicesConfig.appName,
       Pem(WorkbenchEmail(config.googleServicesConfig.serviceAccountClientId), new File(config.googleServicesConfig.pemFile)),
       "google",
-      config.googleServicesConfig.groupSyncPubSubProject
+      config.googleServicesConfig.googlePubSubConfig.pubSubProject
```
s-rubenstein, line 250: This entire PR is because of this one line here. We define the project when we set up the googlePubSubDAO, but provide a project value for each of the pub sub config objects. That seemed weird to me. The goal here is to make it a bit more honest about what is happening, which is that we need separate googlePubSubDao objects for separate google projects.

dvoet: why not go the other way and specify a project for each pubsub topic? I don't really like assuming they are all in the same project.

s-rubenstein: I would be okay with that, but that would involve creating three `pubSubDao`s, which would be an actual behavior change, as opposed to this which is just making the configs reflect reality. I can bring it to standup if we want to talk over approach. (I think that would be my preference as well, but I would want team buyin/heads up for that sort of a change)

I'll make a PR for that change as well, just so we can look at them side by side.

PR #507, opened 2021-02-16
[CA-1190] Change pubSubDAO to be one DAO for each topic

s-rubenstein:
Ticket: <Link to Jira ticket>
This creates a separate DAO for each pub sub, so that we are using the individual projects.
---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -265,18 +277,20 @@ object Boot extends IOApp with LazyLogging {
         googleIamDAO,
         googleStorageDAO,
         googleStorageNew,
-        googlePubSubDAO,
+        googleKeyCachePubSubDAO,
         config.googleServicesConfig,
         config.petServiceAccountConfig)
-    val notificationDAO = new PubSubNotificationDAO(googlePubSubDAO, config.googleServicesConfig.notificationTopic)
+    // Which pubSubProject/Dao should this be using?
+    val notificationDAO = new PubSubNotificationDAO(googleGroupSyncPubSubDAO, config.googleServicesConfig.notificationTopic)
```
s-rubenstein: I wasn't sure which pubsubDAO this should use? Maybe we need a new project config value for that?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
```diff
@@ -28,15 +28,15 @@ object GoogleGroupSyncMonitorSupervisor {
...
-      pollInterval: FiniteDuration,
-      pollIntervalJitter: FiniteDuration,
-      pubSubDao: GooglePubSubDAO,
-      pubSubTopicName: String,
-      pubSubSubscriptionName: String,
-      workerCount: Int,
-      groupSynchronizer: GoogleGroupSynchronizer): Props =
+             pollInterval: FiniteDuration,
+             pollIntervalJitter: FiniteDuration,
+             groupSyncPubSubDao: GooglePubSubDAO,
```
s-rubenstein, line 33: I renamed the Daos in the monitors to make it clear which pubSubDAO they are using.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -558,8 +559,8 @@ class GoogleExtensions(
     }
 
     def checkPubsub: Future[SubsystemStatus] = {
-      logger.debug("Checking Google PubSub...")
-      googlePubSubDAO.getTopic(googleServicesConfig.groupSyncTopic).map {
+      logger.debug("Checking Google Group Sync PubSub...")
+      googleGroupSyncPubSubDAO.getTopic(googleServicesConfig.groupSyncTopic).map {
```
s-rubenstein: Wasn't sure what to do for this one? Should we checkPubsub for the other projects as well?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -228,7 +229,7 @@ class GoogleExtensions(
   }
 
   private def publishMessages(messages: Seq[String]): Future[Unit] = {
```
s-rubenstein, line 232: This name should probably also be changed.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
marctalbott: A new project config value and accompanying PubSubDAO would make sense to me -- it's currently the same project as everything else, but I suppose it could differ some day

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/GoogleServicesConfig.scala
```diff
@@ -31,6 +31,12 @@ final case class GoogleServicesConfig(
     groupSyncTopic: String,
     groupSyncSubscription: String,
     groupSyncWorkerCount: Int,
```
marctalbott: Could we contain all of these in a `PubSubConfig` case class like in the other PR? Otherwise, I think I like this approach more

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
marctalbott: Seems like we should yeah since I think this gets reported in status

marctalbott (commented):
I think I prefer this version over the other approach. Github seems to be showing extra stuff in the diff, so it's a bit challenging to tell what's part of this PR and what's a part of the main PR

s-rubenstein:
Oh noooo! I hoped it would head that off, but I think since I made changes to the main branch it is showing extra changes. I can try rebasing off of that branch, if we think that's necessary, but my inclination would be to do this as its own follow up PR

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/GoogleServicesConfig.scala
s-rubenstein: Absolutely. Although I think that could be its own independent PR.

dvoet (approved):

andy7i (approved):
LGTM!

PR #509, opened 2021-02-18
[PF-413] Support for WSM controlled resources

zloery:
Ticket: [PF-413](https://broadworkbench.atlassian.net/browse/PF-413)
This change adds controlled resource types and expands the workspace resource model to also grant certain inherited roles on child resources of a workspace. I didn't use `authDomainConstrainable` for new workspace roles or the new resources - I'm not sure if we need this or not.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -228,7 +228,6 @@ resourceTypes = {
         roleActions = ["read"]
       }
     }
-    authDomainConstrainable = true
     reuseIds = false
   }
   controlled-user-private-workspace-resource = {
@@ -288,7 +287,6 @@ resourceTypes = {
         roleActions = ["read"]
       }
     }
-    authDomainConstrainable = true
     reuseIds = false
   }
   controlled-application-shared-workspace-resource = {
@@ -342,7 +340,6 @@ resourceTypes = {
         roleActions = ["read"]
       }
     }
-    authDomainConstrainable = true
     reuseIds = false
   }
   controlled-application-private-workspace-resource = {
@@ -402,7 +399,6 @@ resourceTypes = {
         roleActions = ["read"]
       }
     }
-    authDomainConstrainable = true
     reuseIds = false
   }
   managed-group = {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -33,7 +33,6 @@ resourceTypes = {
     reuseIds = false
   }
   workspace = {
-;   todo forcing travis retest
     actionPatterns = {
       delete = {
         description = "delete this workspace"
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -33,6 +33,7 @@ resourceTypes = {
     reuseIds = false
   }
   workspace = {
+;   todo forcing travis retest
     actionPatterns = {
       delete = {
         description = "delete this workspace"
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -128,7 +128,8 @@ resourceTypes = {
           google-project = ["owner"]
           controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
           controlled-user-private-workspace-resource = ["assigner", "editor"]
-          controlled-application-shared-workspace-resource = ["writer", "reader"]
+          controlled-application-shared-workspace-resource = ["editor", "writer", "reader"]
+          controlled-application-private-workspace-resource = ["editor"]
         }
       }
       editor = {
@@ -136,7 +137,8 @@ resourceTypes = {
         descendantRoles = {
           controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
           controlled-user-private-workspace-resource = ["editor"]
-          controlled-application-shared-workspace-resource = ["writer", "reader"]
+          controlled-application-shared-workspace-resource = ["editor", "writer", "reader"]
+          controlled-application-private-workspace-resource = ["editor"]
         }
       }
       application = {
@@ -214,7 +216,7 @@ resourceTypes = {
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "set_parent"]
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "set_parent"]
       }
       editor = {
         roleActions = ["delete", "edit"]
@@ -270,7 +272,7 @@ resourceTypes = {
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "manage_private_user", "set_parent"]
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "manage_private_user", "set_parent"]
       }
       assigner = {
         roleActions = ["manage_private_user"]
@@ -326,7 +328,7 @@ resourceTypes = {
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "set_parent"]
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "set_parent"]
       }
       editor = {
         roleActions = ["delete", "edit"]
@@ -382,7 +384,7 @@ resourceTypes = {
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "manage_private_user", "set_parent"]
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "manage_private_user", "set_parent"]
       }
       assigner = {
         roleActions = ["manage_private_user"]
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -228,6 +228,7 @@ resourceTypes = {
         roleActions = ["read"]
       }
     }
+    authDomainConstrainable = true
     reuseIds = false
   }
   controlled-user-private-workspace-resource = {
@@ -287,6 +288,7 @@ resourceTypes = {
         roleActions = ["read"]
       }
     }
+    authDomainConstrainable = true
     reuseIds = false
   }
   controlled-application-shared-workspace-resource = {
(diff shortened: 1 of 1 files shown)


zloery:
jenkins retest

zloery:
A few specific questions I want to check on:

Dan: does this match the current spec for controlled resources?
David: this change affects workspaces, which Rawls also uses. Is everything here compatible?
Doug: is this the right way to use hierarchical resources? I haven't followed development over the past few months, so my understanding might be off.

src/main/resources/reference.conf
```diff
@@ -93,16 +123,38 @@ resourceTypes = {
...
       }
       owner = {
-        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::writer", "share_policy::reader", "own", "write", "read", "compute", "share_policy::share-reader", "share_policy::share-writer", "share_policy::can-compute", "share_policy::can-catalog", "read_auth_domain", "list_children", "remove_child", "add_child"]
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::writer", "share_policy::reader", "own", "edit", "write", "read", "compute", "share_policy::share-reader", "share_policy::share-writer", "share_policy::can-compute", "share_policy::can-catalog", "read_auth_domain", "create_controlled_user_shared", "create_controlled_user_private", "create_referenced_resource", "update_referenced_resource", "delete_referenced_resource", "list_children", "remove_child", "add_child"]
         descendantRoles = {
           google-project = ["owner"]
+          controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
+          controlled-user-private-workspace-resource = ["assigner", "editor"]
+          controlled-application-shared-workspace-resource = ["writer", "reader"]
```
ddietterich: I think owner needs editor role on controlled-application-*-workspace-resource.

When WSM does permission checks for update and delete operations on application-* resources, we will always require two tokens on the call and do two checks:
1. Make sure the application making the call has editor role on the resource
2. Make sure the user making the call has editor role on the resource

We want owners to be able to delete anything, so owner needs editor role on application-*.

src/main/resources/reference.conf
```diff
@@ -93,16 +123,38 @@ resourceTypes = {
...
         }
       }
+      editor = {
+        roleActions = ["read_policy::owner", "edit", "write", "read", "create_controlled_user_shared", "create_controlled_user_private", "create_referenced_resource", "update_referenced_resource", "delete_referenced_resource", "add_child", "remove_child", "read_auth_domain"]
+        descendantRoles = {
+          controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
+          controlled-user-private-workspace-resource = ["editor"]
+          controlled-application-shared-workspace-resource = ["writer", "reader"]
```
ddietterich: Same goes for editor as for owner. They will need editor.

src/main/resources/reference.conf
```diff
@@ -123,6 +175,230 @@ resourceTypes = {
...
+        description = "perform editor actions on the resource"
+      }
+      "own" = {
+        description = "perform owner actions on the resource"
+      }
+    }
+    ownerRoleName = "owner"
+    roles = {
+      owner = {
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "set_parent"]
```
ddietterich: Does owner need edit? I think it is missed in all cases below.

src/main/resources/reference.conf
```diff
@@ -123,6 +175,230 @@ resourceTypes = {
     authDomainConstrainable = true
     reuseIds = false
   }
+  controlled-user-shared-workspace-resource = {
```
ddietterich, line 180: Is there any reason to apply auth domains to these resources?

ddietterich (commented):
Hey, I made some comments!

src/main/resources/reference.conf
zloery: I don't know, I didn't include them here because I wasn't sure. @davidangb or @dvoet, do we want controlled resources to be auth domain constrainable?

src/main/resources/reference.conf
zloery: To confirm, editors in a workspace should be allowed to edit/delete (but not read/write) all controlled resources in that workspace, including private/application owned ones?

src/main/resources/reference.conf
zloery: Ah, I had been thinking of applications as users themselves, so we would only use one token. If workspace owners should be able to delete any controlled resource in that workspace, then yes they need to get editor on child resources.

zloery: Related question: should workspace owners be allowed to reassign anything? They definitely should be assigners on user-private resources, but I don't know about app-private.

src/main/resources/reference.conf
zloery: iiuc the owner role on a resource is only granted to the WSM service account - it definitely needs delete permission, but I'm less sure about edit. Added for now, but that might change

src/main/resources/reference.conf
davidangb: I think that when/if Rawls (or current Terra as a whole) wants to use controlled resources, they will need to support auth domains. Else, we can't properly constrain access to the resource.

davidangb (commented):
I don't see anything in here that looks incompatible, though I admit my low-level Sam expertise is somewhat lacking. 

I would love to see some code comments explaining each resource/role; I know the rest of the conf file doesn't have these, so this request goes against tradition.

src/main/resources/reference.conf
zloery: Sounds good, added this to the four types of controlled resources

zloery:
> I would love to see some code comments explaining each resource/role; I know the rest of the conf file doesn't have these, so this request goes against tradition.

We do have more explanation in the controlled resources design doc and I'm hesitant to double-document things that are likely to change frequently - does that seem like a good place for that explanation to live?

src/main/resources/reference.conf
davidangb: I'm now second-guessing my comment, and suggest getting an authoritative answer from Cloud Integration. My concern here is making sure that the cloud ACLs on the resource itself in Google or Azure or wherever are correct, and that if a user creates a resource inside an auth-domained workspace, cloud-native access to the resource is prevented for anyone not in the auth domain.  I don't know if this requires the Sam resource to support  auth domains or not;  I *assume* it does but don't know for sure.

davidangb:
> We do have more explanation in the controlled resources design doc and I'm hesitant to double-document things that are likely to change frequently - does that seem like a good place for that explanation to live?

I like seeing the documentation close to t he implementation -  but I won't push on this; explanation in the design doc  is fine.

zloery:
jenkins retest

src/main/resources/reference.conf
zloery: From CI:
> Synchronization with google groups gets extra complicated when there are auth domains involved so we are not allowing this for now. This remains a problem to figure out

at the moment, that means we can't add controlled resources to workspaces with auth domains. I don't know if the current plan is to replace auth domains with policy enforcement or fix this restriction in Sam, but either way this is a blocker on migrating existing Rawls workspaces to WSM. 
Given that, my inclination for now is to disallow auth domains on resources until we know we need them. I'm still pretty shaky on this whole area, though

zloery:
jenkins retest

src/main/resources/reference.conf
```diff
@@ -63,6 +69,9 @@ resourceTypes = {
         description = "perform writer actions on the workspace"
         authDomainConstrainable = true
       }
+      "edit" = {
```
ddietterich, line 72: Do we need an "application" action here to go along with the application role?

src/main/resources/reference.conf
```diff
@@ -93,16 +123,40 @@ resourceTypes = {
...
+      editor = {
+        roleActions = ["read_policy::owner", "edit", "write", "read", "create_controlled_user_shared", "create_controlled_user_private", "create_referenced_resource", "update_referenced_resource", "delete_referenced_resource", "add_child", "remove_child", "read_auth_domain"]
+        descendantRoles = {
+          controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
+          controlled-user-private-workspace-resource = ["editor"]
+          controlled-application-shared-workspace-resource = ["editor", "writer", "reader"]
+          controlled-application-private-workspace-resource = ["editor"]
+        }
+      }
+      application = {
```
s-rubenstein, line 144: What role is `application`? Is that like `this user is an application, e.g., rawls`?

zloery: Yep, it's `this user is a registered application in this workspace`

src/main/resources/reference.conf
```diff
@@ -123,6 +177,230 @@ resourceTypes = {
...
+    ownerRoleName = "owner"
+    roles = {
+      owner = {
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "set_parent"]
+      }
+      editor = {
+        roleActions = ["delete", "edit"]
+      }
+      writer = {
+        roleActions = ["write"]
```
ddietterich, line 225: I wonder if a better way to organize this is to have the writer have "read" and "write" actions. Then have the writer role in the workspace just map to the writer role in the resource.

ddietterich (commented):
Two questions - otherwise, looks good.

src/main/resources/reference.conf
zloery: I don't think so, I don't know when we would check an application action.
Rather than asking Sam "is this user an application", I think we'll probably ask e.g. "can this user create an application-owned shared resource"

src/main/resources/reference.conf
zloery: If we want all writers to also be readers, then yeah that's better. 
Controlled resources don't have the tiered access like workspaces do (e.g. editors explicitly don't have reader/writer access) so I was hesitant to add that here, but it's fine if that's how we want to structure it

ddietterich: Good point. Let's leave it just as it is.

ddietterich (approved):
LGTM

zloery:
jenkins retest

zloery:
jenkins retest

src/main/resources/reference.conf
s-rubenstein: When in doubt, our philosophy in Sam has been to have roles contain the minimal set of information, because we can always grant multiple roles. I think leaving as is seems like the right call.

dvoet (approved):
FYI - sam supports `includedRoles` in addition to `descendantRoles`, [example](https://github.com/broadinstitute/sam/blob/93c23ffef58d3ba43beba7e7600c775ab7e79470/src/main/resources/reference.conf#L566)

If you want to reformulate the workspace roles such that editor includes writer includes reader that is possible (but up to you).

zloery:
jenkins retest

zloery:
jenkins retest

zloery:
jenkins retest

zloery:
jenkins retest

zloery:
jenkins retest

zloery:
jenkins retest

zloery:
I don't know why Jenkins has been failing - the same code passed a few commits ago before I added and then removed auth domains from controlled resources. I also can't read the Jenkins report, so my inclination is to keep re-running for now, but I'm wondering if I broke something by accident.
Edit: larger known issue with FIABs. I'll re-run when this is fixed

zloery:
jenkins retest

zloery:
jenkins retest

PR #514, opened 2021-03-11
[CA-1056] Upgrade Sam to Java 11

s-rubenstein:
Ticket: <Link to Jira ticket>

---

**PR checklist**


project/Merging.scala
```diff
@@ -7,7 +7,7 @@ object Merging {
     case PathList("google", "protobuf", _ @ _*) => MergeStrategy.first
     case PathList("META-INF", "versions", "9", "module-info.class") => MergeStrategy.first
     case "module-info.class" =>
-      MergeStrategy.discard // JDK 8 does not use the file module-info.class so it is safe to discard the file.
+      MergeStrategy.discard // JDK 11 does not use the file module-info.class so it is safe to discard the file.
```
davidangb: I believe module-info was introduced in Java 9, I don't believe this comment is true any more

s-rubenstein: Ah! Thank you! I couldn't find info on it!

s-rubenstein:
https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/46486/ swatomation run with the firecloud develop branch.

project/Settings.scala
```diff
@@ -18,14 +18,14 @@ object Settings {
   //coreDefaultSettings + defaultConfigs = the now deprecated defaultSettings
   lazy val commonBuildSettings = Defaults.coreDefaultSettings ++ Defaults.defaultConfigs ++ Seq(
     javaOptions += "-Xmx2G",
-    javacOptions ++= Seq("-source", "1.8", "-target", "1.8"),
+    javacOptions ++= Seq("--release", "11"),
```
andy7i, line 21: what's the difference between source and target vs release?

andy7i: answering my own question: https://stackoverflow.com/questions/54843571/java-11-what-is-the-difference-between-javac-release-and-source-and-target

> The --release flag not only sets source and target versions, but it also causes the compiler to use the symbol table for the JDK libraries corresponding to the specified release, preventing you from accidentally using APIs that are present in the compilation JDK but not present in the specified version. The --release flag was added later, and in most cases, should replace uses of --source and --target.

andy7i (approved):
tested locally, updating intellij jdk and running a test and things seem to work 👍  with passing jenkins tests

https://www.jetbrains.com/help/idea/sdk.html#change-project-sdk

docker/build.sh
```diff
@@ -98,10 +98,10 @@ function make_jar()
...
+    docker run --rm --link postgres:postgres --link $OPENDJ:opendj -e DIRECTORY_URL=$DIRECTORY_URL -e GIT_MODEL_HASH=$GIT_MODEL_HASH -e DIRECTORY_PASSWORD=$DIRECTORY_PASSWORD -v $PWD:/working -v jar-cache:/root/.ivy -v jar-cache:/root/.ivy2 broadinstitute/scala-baseimage:jdk11-2.13.5-1.4.7 /working/docker/init_schema.sh /working
     docker restart $OPENDJ
     sleep 40
-    docker run --rm --link postgres:postgres --link $OPENDJ:opendj -e DIRECTORY_URL=$DIRECTORY_URL -e GIT_MODEL_HASH=$GIT_MODEL_HASH -e DIRECTORY_PASSWORD=$DIRECTORY_PASSWORD -v $PWD:/working -v jar-cache:/root/.ivy -v jar-cache:/root/.ivy2 broadinstitute/scala-baseimage /working/docker/install.sh /working
+    docker run --rm --link postgres:postgres --link $OPENDJ:opendj -e DIRECTORY_URL=$DIRECTORY_URL -e GIT_MODEL_HASH=$GIT_MODEL_HASH -e DIRECTORY_PASSWORD=$DIRECTORY_PASSWORD -v $PWD:/working -v jar-cache:/root/.ivy -v jar-cache:/root/.ivy2 broadinstitute/scala-baseimage:jdk11-2.13.5-1.4.7 /working/docker/install.sh /working
```
andy7i, line 104: part of the ticket says:
>We need to make sure that the client libraries build Java 8 and 11 version. 

do we still create the java8 client library?

docker/build.sh
```diff
@@ -98,10 +98,10 @@ function make_jar()
 
     # make jar.  cache sbt dependencies.
     set +e # Turn off error detection so that opendj has a chance to get stopped before exiting
-    docker run --rm --link postgres:postgres --link $OPENDJ:opendj -e DIRECTORY_URL=$DIRECTORY_URL -e GIT_MODEL_HASH=$GIT_MODEL_HASH -e DIRECTORY_PASSWORD=$DIRECTORY_PASSWORD -v $PWD:/working -v jar-cache:/root/.ivy -v jar-cache:/root/.ivy2 broadinstitute/scala-baseimage /working/docker/init_schema.sh /working
+    docker run --rm --link postgres:postgres --link $OPENDJ:opendj -e DIRECTORY_URL=$DIRECTORY_URL -e GIT_MODEL_HASH=$GIT_MODEL_HASH -e DIRECTORY_PASSWORD=$DIRECTORY_PASSWORD -v $PWD:/working -v jar-cache:/root/.ivy -v jar-cache:/root/.ivy2 broadinstitute/scala-baseimage:jdk11-2.13.5-1.4.7 /working/docker/init_schema.sh /working
```
gpolumbo-broad, line 101: Firecloud develop PR seemed to be removing the `.ivy` cache and just using `.ivy2`.  My very brief googling didn't seem to give me a clear answer on what we should be doing.  Hopefully, we can figure it out and be consistent in all docker commands where we map the local cache into a container.

docker/build.sh
s-rubenstein: I don't think so, but I am also questioning whether we need to? I think users of the client library should just pin their version? (I think I remember chatting about that?) But maybe we can discuss that.

gpolumbo-broad: Great question.  I do not recall which service(s) rely on the java8 library nor do I remember if that dependency was just on Rawls or also Sam.  Perhaps we can ask Doug if he knows the answer or maybe Isaac?  We should document whatever we learn and write a ticket that enumerates those services so that when they are upgraded, we can assert that the java8 artifact is no longer needed and we can remove any code for building java8 artifacts.

s-rubenstein: I will ask Doug!

s-rubenstein: WSM, Data Repo, and Agora all use it. WSM has the version pinned, and I am checking Data Repo and Agora, but I _think_ we should be okay.

s-rubenstein: Data Repo and Agora both have the version pinned. Workbench CLI also uses it, and has the version pinned. I think we can safely upgrade, and let all the services know that if they are upgrading, we don't promise Java8 backwards compatibility.

Most of these services are in Java 11 anyway, so that seems like the right approach.

s-rubenstein:
@andy7i https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/46613/ fresh test run

project/Merging.scala
```diff
@@ -7,7 +7,7 @@ object Merging {
     case PathList("google", "protobuf", _ @ _*) => MergeStrategy.first
     case PathList("META-INF", "versions", "9", "module-info.class") => MergeStrategy.first
     case "module-info.class" =>
-      MergeStrategy.discard // JDK 8 does not use the file module-info.class so it is safe to discard the file.
+      MergeStrategy.discard
```
jacmrob: is this safe for JDK 11 too?  should this note still stand?

project/Settings.scala
jacmrob: nice

docker/build.sh
jacmrob: this is all set then?  will this theory be tested in dev?

also how are you notifying other services?

jacmrob (approved):
LGTM, assuming all the version numbers are the correct ones.  had a few minor questions.

jacmrob:
Is the failed test related or flaky?

docker/build.sh
s-rubenstein: This is all set, and I am planning on emailing those teams after this goes live to let them know.

project/Merging.scala
s-rubenstein: I believe this is safe. I was surprised that it was, but tests/compilation fail if this isn't here, and succeeds if it is, so I think the original comment may have been wrong? Or I am misunderstanding.

s-rubenstein:
@jacmrob the failed tests are related, this has to go out simultaneously with a `firecloud-develop` change, so the CI-CD is failing because it points at dev firecloud-develop

PR #513, opened 2021-03-04
[PF-504] Followup changes to WSM controlled resources

zloery:
Ticket: [PF-504](https://broadworkbench.atlassian.net/browse/PF-504)
This PR makes a number of tweaks to the controlled resource object model introduced previously in #509:

- First, this removes the `inheritedRole` mappings on workspace roles. In order to sync Sam policies (which only exist on workspace objects and not individual resources) to cloud resources, WSM still needs to maintain its own mapping of the workspace role -> resource role inheritance. Rather than duplicate that mapping in both WSM and Sam, it's more appropriate to have one canonical mapping in WSM and use policy-based inheritance in Sam (where the specification can live in WSM), rather than role-based inheritance.
- Second, this grants workspace owners the ability to share the new editor and application workspace roles. Workspace owners should have the ability to add other users to a workspace, and I forgot to add this permission when I previously introduced the new workspace roles.
- Third, this grants workspace editors and applications permission to list children of the workspace (i.e. the resources in that workspace). They already have the ability to add or remove children, and listing existing children is convenient.

---

**PR checklist**

- [X] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [X] I've filled out the [Security Risk Assessment](https://broadinstitute.github.io/dsp-appsec-security-risk-assessment/) and attached the result to the JIRA ticket

PR #516, opened 2021-03-17
[PF-578] Add role-based inheritance back to workspaces

zloery:
Ticket: [PF-578](https://broadworkbench.atlassian.net/browse/PF-578)
After discussion of WSM's approach to hierarchical resources, we decided to move back from policy-based inheritance to role-based inheritance. This change adds back the inheritance introduced in #509 and removed in #513. Since then we've removed the workspace `editor` role and given its permissions to the `writer` role, so the exact mapping is slightly different than it was in the previous change.

Because we never merged the policy-based inheritance change to WSM, there are no workspaces with old policy configurations to clean up.

This also modifies controlled resource roles in the following ways:

- For convenience, the `writer` role also includes the `read` action. Previously, we would always grant the `reader` role alongside `writer` instead.
- The `assigner` role includes direct permission for modifying membership of the `reader`, `writer`, and `editor` policies so that we do not rely on the WSM service account to actually execute changes to these policies

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -120,6 +120,10 @@ resourceTypes = {
         roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::application", "share_policy::writer", "share_policy::reader", "own", "write", "read", "compute", "share_policy::share-reader", "share_policy::share-writer", "share_policy::can-compute", "share_policy::can-catalog", "read_auth_domain", "create_controlled_user_shared", "create_controlled_user_private", "create_referenced_resource", "update_referenced_resource", "delete_referenced_resource", "list_children", "remove_child", "add_child"]
         descendantRoles = {
           google-project = ["owner"]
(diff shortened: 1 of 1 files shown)


ddietterich (commented):
This looks right to me.

Doug had suggested that we add a separate role for the WSM SA instead of using the owner role as the way to handle the private resource checking. Should we add that to the controlled resources as part of this change?

zloery:
I previously added the `owner` role to all four controlled resource types as an "admin" type role that's never granted to users, and I intended to create a policy with only the WSM SA for that role on each resource. That's slightly different than Doug's doc describes though.

Unrelated but this discussion also makes me think we should grant `share_policy::{reader, writer, editor}` to the assigner role. Otherwise we're relying on a stopgap admin control for the normal assigner flow.

ddietterich:
You are right about the `share_policy:*` comment above.

dvoet (approved):

ddietterich (approved):
lgtm

PR #519, opened 2021-04-08
[CA-1258] Add a status check to LdapRegistrationDAO and use to check openDJ

s-rubenstein:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ jdk:
 dist: trusty
 language: scala
 scala:
-- 2.12.8
+- 2.13.5
 sudo: required
 services:
 - docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ script:
 - export SBT_OPTS="-J-Xmx3g -Denv.type=test -Ddirectory.url=ldap://localhost:3389
   -Ddirectory.password=testtesttest -Dpostgres.host=localhost -Dpostgres.port=15432"
 - sbt clean "testOnly -- -n org.broadinstitute.tags.SchemaInit"
+- echo FINDTHIS
+- sbt --version
 - docker restart opendj
 - sleep 40
 - sbt coverage "testOnly -- -l org.broadinstitute.tags.SchemaInit" coverageReport
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,6 @@ script:
 - export SBT_OPTS="-J-Xmx3g -Denv.type=test -Ddirectory.url=ldap://localhost:3389
   -Ddirectory.password=testtesttest -Dpostgres.host=localhost -Dpostgres.port=15432"
 - sbt clean "testOnly -- -n org.broadinstitute.tags.SchemaInit"
-- echo FINDTHIS
-- sbt --version
-- scala -version
 - docker restart opendj
 - sleep 40
 - sbt coverage "testOnly -- -l org.broadinstitute.tags.SchemaInit" coverageReport
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ jdk:
 dist: trusty
 language: scala
 scala:
-- 2.13.5
+- 2.12.8
 sudo: required
 services:
 - docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ script:
 - sbt clean "testOnly -- -n org.broadinstitute.tags.SchemaInit"
 - echo FINDTHIS
 - sbt --version
+- scala -version
 - docker restart opendj
 - sleep 40
 - sbt coverage "testOnly -- -l org.broadinstitute.tags.SchemaInit" coverageReport
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
@@ -9,6 +9,7 @@ import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.model.google.{ServiceAccount, ServiceAccountDisplayName, ServiceAccountSubjectId}
 import org.broadinstitute.dsde.workbench.sam._
 import org.broadinstitute.dsde.workbench.sam.config.DirectoryConfig
+import org.broadinstitute.dsde.workbench.sam.dataAccess.ConnectionType.ConnectionType
 import org.broadinstitute.dsde.workbench.sam.schema.JndiSchemaDAO.{Attr, ObjectClass}
 import org.broadinstitute.dsde.workbench.sam.util.{LdapSupport, SamRequestContext}
 
@@ -40,6 +41,8 @@ class LdapRegistrationDAO(
       }
     }
   }
+
+  override def getConnectionTarget(): ConnectionType = ConnectionType.LDAP
 
   override def createUser(user: WorkbenchUser, samRequestContext: SamRequestContext): IO[WorkbenchUser] = {
     val attrs = List(
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
@@ -42,7 +42,7 @@ class LdapRegistrationDAO(
     }
   }
 
-  override def getConnectionTarget(): ConnectionType = ConnectionType.LDAP
+  override def getConnectionType(): ConnectionType = ConnectionType.LDAP
 
   override def createUser(user: WorkbenchUser, samRequestContext: SamRequestContext): IO[WorkbenchUser] = {
     val attrs = List(
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
@@ -10,6 +10,7 @@ import org.broadinstitute.dsde.workbench.model.google.{ServiceAccount, ServiceAc
 import org.broadinstitute.dsde.workbench.sam._
 import org.broadinstitute.dsde.workbench.sam.config.DirectoryConfig
 import org.broadinstitute.dsde.workbench.sam.schema.JndiSchemaDAO.{Attr, ObjectClass}
+import org.broadinstitute.dsde.workbench.sam.service.NoExtensionsInitializer.cloudExtensions
 import org.broadinstitute.dsde.workbench.sam.util.{LdapSupport, SamRequestContext}
 
 import scala.concurrent.ExecutionContext
@@ -159,7 +160,7 @@ class LdapRegistrationDAO(
 
   override def checkStatus(samRequestContext: SamRequestContext): IO[Boolean] =
     for {
-      entry <- executeLdap(IO(ldapConnectionPool.getEntry(directoryConfig.enabledUsersGroupDn, Attr.member)), "isEnabled", samRequestContext)
+      entry <- executeLdap(IO(ldapConnectionPool.getEntry(directoryConfig.enabledUsersGroupDn, Attr.member)), cloudExtensions.allUsersGroupName.toString(), samRequestContext)
     } yield {
       val result = for {
         e <- Option(entry)
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
@@ -156,4 +156,15 @@ class LdapRegistrationDAO(
 
   override def setGoogleSubjectId(userId: WorkbenchUserId, googleSubjectId: GoogleSubjectId, samRequestContext: SamRequestContext): IO[Unit] =
     executeLdap(IO(ldapConnectionPool.modify(userDn(userId), new Modification(ModificationType.ADD, Attr.googleSubjectId, googleSubjectId.value))), "setGoogleSubjectId", samRequestContext)
+
+  override def checkStatus(samRequestContext: SamRequestContext): IO[Boolean] =
+    for {
+      entry <- executeLdap(IO(ldapConnectionPool.getEntry(directoryConfig.enabledUsersGroupDn, Attr.member)), "isEnabled", samRequestContext)
+    } yield {
+      val result = for {
+        e <- Option(entry)
+        members = entry.getAttributeValues(Attr.member)
+      } yield members.nonEmpty
+      result.getOrElse(false)
+    }
 }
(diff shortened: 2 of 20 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
```diff
@@ -156,4 +156,15 @@ class LdapRegistrationDAO(
...
+
+  override def checkStatus(samRequestContext: SamRequestContext): IO[Boolean] =
+    for {
+      entry <- executeLdap(IO(ldapConnectionPool.getEntry(directoryConfig.enabledUsersGroupDn, Attr.member)), "isEnabled", samRequestContext)
+    } yield {
+      val result = for {
+        e <- Option(entry)
+        members = entry.getAttributeValues(Attr.member)
+      } yield members.nonEmpty
+      result.getOrElse(false)
```
gpolumbo-broad: This can be simplified to get rid of the `for` comp inside the `yield` and get rid of `e <- Option(entry)` which is not being used.

gpolumbo-broad (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/TestSamRoutes.scala
```diff
@@ -68,11 +74,38 @@ object TestSamRoutes {
   )
 
   def apply(resourceTypes: Map[ResourceTypeName, ResourceType], userInfo: UserInfo = defaultUserInfo, policyAccessDAO: Option[AccessPolicyDAO] = None, policies: Option[mutable.Map[WorkbenchGroupIdentity, WorkbenchGroup]] = None)(implicit system: ActorSystem, materializer: Materializer, executionContext: ExecutionContext, contextShift: ContextShift[IO]) = {
+    val dbRef = TestSupport.dbRef
+    val directoryConfig: DirectoryConfig = TestSupport.directoryConfig
+    val dirURI = new URI(directoryConfig.directoryUrl)
+    val ldapConnectionPool = new LDAPConnectionPool(new LDAPConnection(dirURI.getHost, dirURI.getPort, directoryConfig.user, directoryConfig.password), directoryConfig.connectionPoolSize)
```
s-rubenstein: I'm a bit concerned about this, because there's no close on the pool? But also I don't know another way to ensure we are testing the real logic. Is there an afterAll or something hook in this file where cleanup can happen?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/StatusService.scala
```diff
@@ -31,24 +31,23 @@ class StatusService(
...
   private def checkDatabase(): IO[SubsystemStatus] = IO {
     logger.info("checking database connection")
-    dbReference.inLocalTransaction { session =>
-      if (session.connection.isValid((2 seconds).toSeconds.intValue())) {
-        HealthMonitor.OkStatus
-      } else {
-        HealthMonitor.failedStatus("database connection invalid or timed out checking")
-      }
-    }
+    if (directoryDAO.checkStatus(SamRequestContext(None)))
```
marctalbott: I mentioned this while we were pairing, but I'll repeat that I like that you moved the status check into the DAOs. Gives some nice symmetry and the DAOs should know how to check the connection status

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -761,4 +762,14 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
     })
   }
+
+  override def checkStatus(samRequestContext: SamRequestContext): Boolean = {
+    writeDbRef.inLocalTransaction { session =>
+      if (session.connection.isValid((2 seconds).toSeconds.intValue())) {
+        true
+      } else {
+        false
+      }
```
marctalbott: This can be simplified by just returning `session.connection.isValid((2 seconds).toSeconds.intValue())`

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
```diff
@@ -156,4 +157,17 @@ class LdapRegistrationDAO(
 
   override def setGoogleSubjectId(userId: WorkbenchUserId, googleSubjectId: GoogleSubjectId, samRequestContext: SamRequestContext): IO[Unit] =
     executeLdap(IO(ldapConnectionPool.modify(userDn(userId), new Modification(ModificationType.ADD, Attr.googleSubjectId, googleSubjectId.value))), "setGoogleSubjectId", samRequestContext)
+
+  override def checkStatus(samRequestContext: SamRequestContext): Boolean = {
+    val ldapIsHealthy = Try {
+      ldapConnectionPool.getHealthCheck
+      val connection = ldapConnectionPool.getConnection
+      ldapConnectionPool.getHealthCheck.ensureNewConnectionValid(connection)
```
gpolumbo-broad, line 168: Was there a decision made between using `ensureNewConnectionValid` and `ensureConnectionValidAfterAuthentication`?  Without digging into this too deeply, I would think that we'd want to ensure that the credentials are also right.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/StatusService.scala
```diff
@@ -18,6 +17,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class StatusService(
     val directoryDAO: DirectoryDAO,
+    val registrationDAO: RegistrationDAO,
```
gpolumbo-broad: Wondering if we shouldn't just be explicit here and say:

```suggestion
    val ldapRegistrationDAO: LdapRegistrationDAO,
```

For this class, we don't want a generic `RegistrationDAO`, we explicitly need it to be an `LdapRegistrationDAO` in order to satisfy the business requirement.  This will require a little refactoring in Boot.scala too I think, but I think it would be worth it.  As it is here, I had to trace all the calls back through Boot.scala just to make sure we were passing in the right kind of `RegistrationDAO` here, which was annoying in itself, and also doesn't prevent anyone from accidentally breaking this again in the future. 

I realize this will mess with tests where we use `MockRegistrationDAO`.  I don't know the answer off the top of my head and maybe it's something we can sort out with a discussion/mob/pair to see if we can make this better in some way.

gpolumbo-broad (changes requested):
Didn't get a chance to finish reviewing before I had to run to meetings, but I had a comment here that I wanted to get out for discussion at least.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/StatusService.scala
s-rubenstein: We don't actually have a `MockRegistrationDAO`, interestingly enough! That said, I think changing it assumes that LDAP is how we communicate to our registration service. I might rather have it be `checkRegistrationDBStatus` rather than `checkOpenDJStatus`? Given that things changed with Postgres, I don't want to assume LDAP will be the way forwards forever.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
s-rubenstein: I don't _think_ that `ensureConnectionValidAfterAuthentication` is the right method. It seems to use a BindResult https://docs.ldap.com/ldap-sdk/docs/javadoc/com/unboundid/ldap/sdk/BindResult.html which doesn't actually look like its about auth? I am not 100% certain on that, though. I couldn't find any usages of BindResult anywhere in the code base, it looks like we just pass in a user and a password when setting up the pool, the once.

I would think that the thing to test that might be an integration test, not as part of our status check, because those values shouldn't be set anywhere, just once when the service stands up.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/TestSamRoutes.scala
```diff
@@ -68,11 +74,38 @@ object TestSamRoutes {
...
     val resourceTypesWithAdmin = resourceTypes + (resourceTypeAdmin.name -> resourceTypeAdmin)
     // need to make sure MockDirectoryDAO and MockAccessPolicyDAO share the same groups
     val groups: mutable.Map[WorkbenchGroupIdentity, WorkbenchGroup] = policies.getOrElse(new TrieMap())
-    val directoryDAO = new MockDirectoryDAO(groups)
-    val registrationDAO = new MockDirectoryDAO()
+    val directoryDAO = new MockDirectoryDAO(groups) {
```
marctalbott: Do we need to override `checkStatus`? Getting rid of these overrides would let us remove the `LdapConnectionPool` that you're concerned about as well.

I tried to test it myself, but I'm having all kinds of environment issues

marctalbott: Also, in general, I would love to avoid overriding `checkStatus` in multiple places if possible. I'd prefer to see us either use a real `LDAPRegistrationDAO` or to use the mocked out `checkStatus`. Duplicating the implementation will make it difficult to change the `checkStatus` implementation in the future

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/StatusServiceSpec.scala
```diff
@@ -25,52 +28,74 @@ class StatusServiceSpec extends AnyFreeSpec with Matchers with BeforeAndAfterAll
...
+
   override def afterAll(): Unit = {
+    connectionPool.close()
     system.terminate()
   }
 
-  private def newStatusService(directoryDAO: DirectoryDAO) = {
-    new StatusService(directoryDAO, new NoExtensions {
+  // We need to mock this because Postgres does some automatic setup and gets unhappy with multiple instances of its DAO existing at once.
+  private def mockDirectoryDAOWithPostgresStatusCheck(dbReferenceOverride: DbReference = dbReference) = {
```
marctalbott: Same comment here re: duplicating the `checkStatus` implementation

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/StatusService.scala
marctalbott: Agreed that we typically don't want to assume that LDAP will be the way forwards forever (I hope the RegistrationDAO goes away entirely one day). However, in this case, I'd argue that we want the `StatusService` checks to be implementation-specific. Keeping it generic was how we missed this in the first place and didn't check OpenDJ's status for >1 year. If we change the implementation of the RegistrationDAO, we'd want to be forced to change the `StatusService` to check the new implementation of it so that we couldn't accidentally leave it checking OpenDJ

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
marctalbott: Seconding Sky, I'm not sure how we'd go about getting the `BindResult`. It'd be great to have our status check make sure we can auth with LDAP, but `ensureConnectionValidAfterAuthentication` may not be the answer

marctalbott (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/TestSamRoutes.scala
s-rubenstein: So, I could just have the mock return true on status, which would satisfy the tests, I believe. I can't have the mock actually have an LdapConnectionPool because it doesn't have a real db. But I think that is fine.

marctalbott: I'm fine with the mock DAOs defaulting to true with individual overrides to return false as needed/when appropriate

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/StatusService.scala
s-rubenstein: Ack. That seems reasonable in rationale to me. I will make the change.

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

.travis.yml
```diff
@@ -3,7 +3,7 @@ jdk:
 dist: trusty
 language: scala
 scala:
-- 2.13.5
+- 2.12.8
```
marctalbott: Is this related to the Github actions stuff?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/StatusServiceSpec.scala
s-rubenstein: So, this one we can't use the real postgres db, because postgres makes some assumptions that it will only get stood up once within the whole test suite, and we don't want to use a persistent postgres across all the suites here, because we want to have different DBRef objects to manage bringing it down or up, so we can't actually override directoryDao itself. (or at least, I couldn't figure out how to)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/StatusService.scala
s-rubenstein: I went down a rabbit hole on this, specifying it as `LDAPRegistrationDao` makes mocking it exceptionally hard, because SBT/Mockito doesn't know how to override a class, only the traits that we use. So the best we could do is spy on it and then override the spy. This is going to be super clunky across our tests. I think the best thing we can do is to rename the variable and make a comment, which will be both simpler, and also provide a check for us against future changes.

.travis.yml
gpolumbo-broad: Was this an accident?  I think Sam is using Scala 2.13.  Was this done just because of issues on Travis?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/StatusServiceSpec.scala
marctalbott: Could we fully mock out the DAOs here then? It doesn't feel right to be overriding our mocks with the real implementation. I'd prefer that we either use the real implementation with a real `PostgresDirectoryDAO` or we use the mock rather than relying on this combination. This goes for the `MockDirectoryDAO` in `TestSamRoutes` as well

.travis.yml
s-rubenstein: Yes, will revert.

marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/RegistrationDAO.scala
```diff
@@ -11,6 +12,7 @@ import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
   * away from a solution that requires that the Apache proxies query this group, we can remove the RegistrationDAO.
   */
 trait RegistrationDAO {
+  def getConnectionTarget(): ConnectionType
```
gpolumbo-broad: Why not `getConnectionType`?

gpolumbo-broad (approved):
One method naming nit, but it's fine.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/RegistrationDAO.scala
s-rubenstein: Because I am bad at naming :D

.travis.yml
s-rubenstein: It was done because of Travis, and I was going to revert and then forgot. Will update

s-rubenstein:
Jenkin's retest

s-rubenstein:
Jenkins retest

s-rubenstein:
Jenkins retest

PR #521, opened 2021-04-21
[CA-1327] Github Actions

s-rubenstein:
Ticket: <Link to Jira ticket>

---

**PR checklist**


s-rubenstein:
I will make the Travis CI piece not required, and make the GHA required before merging, just that is a repo-wide setting so wanted to wait for approvals before futzing with it.

andy7i:
could you add a bit more details about the PR: any differences between travis vs GHA that we should know about? (also, I think we're required to have a ticket # for each PR now)

s-rubenstein:
Yes, thank you. Forgot to update the title after requesting review.

The big differences between GHA and Travis are:

1. There's no separate script for publishing the client library, instead we have some branching logic in the actions file.
2. Test logic and client logic have been separated into two jobs for clarity
3. Artifactory login information is stored in GitHub secrets rather than in Travis env variables.

.github/workflows/build.yml
```diff
@@ -0,0 +1,113 @@
...
+on:
+  pull_request:
+    paths-ignore: ['**.md']
+  push:
+    paths-ignore: ['**.md']
+
+jobs:
+  test:
+
+    runs-on: ubuntu-20.04
```
marctalbott, line 12: In Travis we had to specify the Scala version, does GHA know to just default to what Sam uses?

.github/workflows/build.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Secrets check
+        run: |
+          ./minnie-kenny.sh --force
+          git secrets --scan-history
+
+      - name: Integration tests
+        id: tests
+        run: sbt clean "testOnly -- -n org.broadinstitute.tags.SchemaInit"
+
+      - name: Generate coverage report
```
marctalbott: nit: this runs the unit tests and generates the coverage report

.github/workflows/build.yml
```diff
@@ -0,0 +1,113 @@
...
+          cd ~/git-secrets
+          git checkout ad82d68ee924906a0401dfd48de5057731a9bc84
+          sudo make install
+
+      - name: Secrets check
+        run: |
+          ./minnie-kenny.sh --force
+          git secrets --scan-history
+
+      - name: Integration tests
```
marctalbott: nit: these aren't integration tests, this command just runs the tests tagged with `SchemaInit`. Maybe rename to 'Run JndiSchemaInit tests' or something like that

.github/workflows/build.yml
```diff
@@ -0,0 +1,113 @@
...
+        id: extractBranch
+
+      - name: Generate java client
+        id: generateJavaClient
+        run: bash scripts/gen_java_client.sh
+
+      - name: Publish java client for merge
+        working-directory: codegen_java
+        id: publishJavaClientSnapshot
+        if: ${{steps.extractBranch.outputs.branch == 'develop'}}
```
marctalbott, line 106: do we need to worry about the pull request part of the conditional in the publishJavaClient script?

.github/workflows/build.yml
```diff
@@ -0,0 +1,113 @@
...
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extractBranch
+
+      - name: Generate java client
+        id: generateJavaClient
+        run: bash scripts/gen_java_client.sh
+
+      - name: Publish java client for merge
+        working-directory: codegen_java
+        id: publishJavaClientSnapshot
```
marctalbott: nit: the ids are swapped, this one isn't a snapshot

.github/workflows/build.yml
s-rubenstein: I believe so? If not, it is dictated by the coursier/cache-action, which caches at least sbt.

.github/workflows/build.yml
s-rubenstein: Sure, can rename.

.github/workflows/build.yml
s-rubenstein: That script goes away with this. Since it had travis env variables anyway.

marctalbott: Yeah, I was just concerned that the changed conditional would make it so we aren't publishing snapshots in the correct situation. But the `develop` branch seems like the most important part to me, so I guess the pull request part was irrelevant?

marctalbott (approved):

.github/workflows/build.yml
s-rubenstein: Ah! No, sorry! That is covered by this line https://github.com/broadinstitute/sam/pull/521/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R4

s-rubenstein: It will only run on push or pull request, due to the nature of the workflow. It will only ever be push on dev.

s-rubenstein: Although… I guess if we pull request from dev to a branch, it would publish the dev snapshot? But I think that is okay, since it will only ever release a main version from up to date dev code.

.github/workflows/build.yml
```diff
@@ -0,0 +1,113 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      # coursier cache action caches both coursier and sbt caches
+      - name: coursier-cache-action
+        uses: coursier/cache-action@v5
+
+      - name: Set up opendj
+        run: |
+          sh docker/run-opendj.sh start
```
andy7i, line 47: do we know why travis had this:

```
- docker restart opendj
- sleep 40
```

and whether we need to do something similar (whether restarting or waiting 40 seconds for opendj to start up)? I see that it passed in this test but just checking in case it could cause a race condition and test flakiness.

.github/workflows/build.yml
```diff
@@ -0,0 +1,113 @@
...
+  test:
+
+    runs-on: ubuntu-20.04
+
+    env:
+      SBT_OPTS: >-
+        -Xmx3g
+        -Denv.type=test
+        -Ddirectory.url=ldap://localhost:3389 -Ddirectory.password=testtesttest
+        -Dpostgres.host=localhost -Dpostgres.port=5432
```
andy7i, line 19: any reason for swapping to port 5432 from 15342?

andy7i: oh maybe this was why travis used 15432 instead of postgres's default 5432 port:
> Port 15432 is used here for the local machine, rather than using 5432, in case an instance of PostgreSQL was also running on the local machine. If an instance of PostgreSQL was running on the local machine and the same port was used, setting up the connection would fail as the port would already be in use.
https://www.openshift.com/blog/openshift-connecting-database-using-port-forwarding

this probably isn't an issue for us here though

.travis.yml
```diff
@@ -1,39 +0,0 @@
...
-- export SBT_OPTS="-J-Xmx3g -Denv.type=test -Ddirectory.url=ldap://localhost:3389
-  -Ddirectory.password=testtesttest -Dpostgres.host=localhost -Dpostgres.port=15432"
-- sbt clean "testOnly -- -n org.broadinstitute.tags.SchemaInit"
-- docker restart opendj
-- sleep 40
-- sbt coverage "testOnly -- -l org.broadinstitute.tags.SchemaInit" coverageReport
-- bash scripts/gen_java_client.sh
-after_success:
-- export SBT_OPTS="-J-Xmx3g"
-- sbt coveralls
```
andy7i, line 32: i _think_ we can remove this dependency https://github.com/broadinstitute/sam/blob/c34ff18ddb109aa15e4f4cade8377f582414c86c/project/plugins.sbt#L5

.github/workflows/build.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Extract branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extractBranch
+
+      - name: Generate java client
+        id: generateJavaClient
+        run: bash scripts/gen_java_client.sh
+
+      - name: Publish java client for merge
```
andy7i:
```suggestion
      - name: Publish java client for merge to develop branch
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,113 @@
+name: Scala tests with coverage
+
+on:
+  pull_request:
+    paths-ignore: ['**.md']
+  push:
+    paths-ignore: ['**.md']
```
andy7i, line 7: not sure why, but the tests for this PR seem to be running duplicate tests for each(?) commit, one for push and one for pull_request? maybe we really just need `push` if that runs on all pushed commits. not a big deal if we like it this way

```
Scala tests with coverage / test (pull_request) Successful in 6m
Scala tests with coverage / test (push) Successful in 7m

Scala tests with coverage / generateAndPublishClient (pull_request) Successful in 1m
Scala tests with coverage / generateAndPublishClient (push) Successful in 1m
```

.github/workflows/build.yml
s-rubenstein: I am not sure. Since it is unit tests, I am not too worried? Since otherwise it would be a sign of data leakage being an issue in our testing, which I would want to solve anyway.

.github/workflows/build.yml
s-rubenstein: This ^. Seemed like not an issue for GHA

.github/workflows/build.yml
s-rubenstein: IDK, but it mirrors what travis did (with two identical test runs), and I am trying to keep it as similar as possible for the moment. I think we can make later improvements down the line if it makes sense to.

andy7i (approved):
👍 thanks for making these changes / answering questions!

PR #522, opened 2021-04-30
[CA-1328] Update opendj run script

s-rubenstein:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/docker/opendjsetup.sh
+++ b/docker/opendjsetup.sh
@@ -12,6 +12,7 @@ if [ -n "${NUMBER_SAMPLE_USERS+set}" ]; then
 fi
 
 /opt/opendj/setup directory-server --ldapPort 389 \
+  --instancePath ${INSTANCE_ROOT} \
   --adminConnectorPort 4444 \
   --baseDN $BASE_DN -h opendj.dsde-dev.broadinstitute.org --rootUserPassword "${ROOTPASS}" \
   --acceptLicense \
--- a/docker/opendjsetup.sh
+++ b/docker/opendjsetup.sh
@@ -3,7 +3,6 @@
 #
 # Copyright (c) 2016-2017 ForgeRock AS. Use of this source code is subject to the
 # Common Development and Distribution License (CDDL) that can be found in the LICENSE file
-
 echo "Setting up default OpenDJ instance."
 
 INIT_OPTION="--addBaseEntry"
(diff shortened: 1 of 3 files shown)


marctalbott (approved):

andy7i (approved):

s-rubenstein:
Jenkins retest

docker/stand-alone/docker-compose.yaml
```diff
@@ -23,7 +23,7 @@ services:
       - sam-app:app
       - sam-opendj:opendj
   sam-opendj:
-    image: broadinstitute/openam:opendj5
```
s-rubenstein, line 26: No clue what this was… a different tag? We should probably just use all the same tag

marctalbott: I'm not familiar with this file/directory at all, maybe @gpolumbo-broad can weigh in as a past contributor to the file? Is it used for anything? I wouldn't mind removing it entirely if it's not used anymore (not in this PR but as a future tech debt item)

marctalbott (approved):

docker/opendjsetup.sh
```diff
@@ -13,13 +12,13 @@ if [ -n "${NUMBER_SAMPLE_USERS+set}" ]; then
 fi
 
 /opt/opendj/setup directory-server --ldapPort 389 \
+  --instancePath ${INSTANCE_ROOT} \
```
andy7i, line 15: how does this get the value for ${INSTANCE_ROOT}? I see that env variable defined in the firecloud-opendj repo but not in this repo.

andy7i: another question: why do we need to add `instancePath` here?

docker/stand-alone/docker-compose.yaml
gpolumbo-broad: Oh, the old stand-alone stuff was from way back when I first started and there was a request to set up stand-alone sam.  I don't remember why or if it was ever used.  I suspect that we can probably remove it now, but I'm not sure.

gpolumbo-broad (approved):

docker/opendjsetup.sh
s-rubenstein: `instancePath` seems to have not been default in the GCR dockers? It may be that there was a command getting run in our previous image that we are missing here, but figuring out is basically impossible at this point.

${INSTANCE_ROOT} is defined at the dockerfile level here https://github.com/broadinstitute/firecloud-opendj/blob/dev/Dockerfile#L54, so it will be on the docker container's environment variables, which means that this will get picked up with when run on the docker container.

andy7i: oh I see, this file is copied into the docker container (https://github.com/broadinstitute/sam/blob/c34ff18ddb109aa15e4f4cade8377f582414c86c/docker/stand-alone/docker-compose.yaml#L33) and is run within that container, so that makes sense.

andy7i (approved):
LGTM

PR #524, opened 2021-06-16
[CA-1143] Sr update bouncy castle

s-rubenstein:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -68,6 +68,8 @@ object Dependencies {
 
   val unboundid: ModuleID = "com.unboundid" % "unboundid-ldapsdk" % "4.0.12"
 
+  val bouncyCastleProvider: ModuleID = "org.bouncycastle" % "bcprov-jdk15on" % "1.68"
+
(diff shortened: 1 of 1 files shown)


andy7i (approved):

MatthewBemis (approved):

gpolumbo-broad (changes requested):
I'm wondering if just _adding_ the dependency to Sam is the right solution here?  If the problem is in Workbench Libs, then it seems to me that the solution should exist in Workbench Libs.  I wonder if appsec found this error in all of their scans of all services that use Workbench Libs?  Anyway, I think that is the better place to implement the fix.

s-rubenstein:
@gpolumbo-broad ah! I didn't realize it was a workbench libs problem. The ticket seemed to make it seem like it was just a Sam and Rawls issue. I can look into that, but I have no clue how to verify that it solves the problem. Is the right process for that to ping AppSec after making the change, and have them tell us?

PR #531, opened 2021-07-12
CA-1243: Admin resource, policies, roles

rushtong:
Ticket: https://broadworkbench.atlassian.net/browse/CA-1243

Added an admin resource type with policies and roles

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -2,6 +2,26 @@ akka.http.server.idle-timeout = 180 s
 akka.http.server.request-timeout = 60 s
 
 resourceTypes = {
+  admin {
+    actionPatterns = {
+      admin_read_policies = {
+        description = "view all policies and policy details"
+      }
+      admin_add_member = {
+        description = "add membership to group"
+      }
+      admin_remove_member = {
+        description = "remove membership from group"
+      }
+    }
+    ownerRoleName = "resource_type_admin"
+    roles = {
+      resource_type_admin = {
+        roleActions = ["admin_read_policies", "admin_add_member", "admin_remove_member"]
+      }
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -2,6 +2,26 @@ akka.http.server.idle-timeout = 180 s
 akka.http.server.request-timeout = 60 s
 
 resourceTypes = {
+  admin {
```
s-rubenstein: We decided to go with `admin` here, because from what we could tell, `resource_type_admin` below is about administering to specific resource types. I am not super happy with the role as the same name as the resource. @marctalbott do you have any sense of what the original resource type of `resource_type_admin` is for?

s-rubenstein (approved):
This seems reasonable to me, but I want to look into `resource_type.resource_type_admin` first.

src/main/resources/reference.conf
rushtong: Agreed @s-rubenstein .... the duplicate naming is super confusing and I think we need more explicit terminology.

marctalbott: Instead of adding a new `admin` resource type, we should just add these actions and roles to the existing `resource_type_admin` resource type. The `resource_type_admin` resource type has a resource for every other resource type in Sam, and we can use those resources to grant per-resource-type admin access

rushtong:
@marctalbott Updated per feedback. Will move this to ready for review.

marctalbott (commented):
Changes look good. I want to check with the rest of the CI team before we starting merging any jamboree changes to `develop` though in case they'd prefer we keep it contained within a feature branch

PR #533, opened 2021-07-13
[CA-1245][CA-1246] Sr admin add to policy

s-rubenstein:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminResourceRoutes.scala
@@ -97,9 +97,9 @@ trait AdminResourceRoutes extends UserInfoDirectives with SecurityDirectives wit
     }
 
   private def requireActionsForAdminPutUserInPolicy(policyId: FullyQualifiedPolicyId, userInfo: UserInfo, samRequestContext: SamRequestContext)(addUserToPolicy: server.Route): server.Route =
-    requireOneOfAction(
+    requireAction(
       FullyQualifiedResourceId(SamResourceTypes.resourceTypeAdminName, ResourceId(policyId.resource.resourceTypeName.value)),
-      Set(SamResourceActions.adminAddMember),
+      SamResourceActions.adminAddMember,
       userInfo.userId,
       samRequestContext
     ) {
@@ -107,9 +107,9 @@ trait AdminResourceRoutes extends UserInfoDirectives with SecurityDirectives wit
     }
 
   private def requireActionsForAdminRemoveUserFromPolicy(policyId: FullyQualifiedPolicyId, userInfo: UserInfo, samRequestContext: SamRequestContext)(removeUserFromPolicy: server.Route): server.Route =
-    requireOneOfAction(
+    requireAction(
       FullyQualifiedResourceId(SamResourceTypes.resourceTypeAdminName, ResourceId(policyId.resource.resourceTypeName.value)),
-      Set(SamResourceActions.adminRemoveMember),
+      SamResourceActions.adminRemoveMember,
       userInfo.userId,
       samRequestContext
     ) {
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -89,6 +89,9 @@ object SamResourceActions {
   val addChild = ResourceAction("add_child")
   val removeChild = ResourceAction("remove_child")
   val listChildren = ResourceAction("list_children")
+  val adminReadPolicies = ResourceAction("admin_read_policies")
```
s-rubenstein, line 92: I realized we never updated the Sam model. I think this is what we need? @marctalbott does this look right to you?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminResourceRoutes.scala
```diff
@@ -78,4 +90,33 @@ trait AdminResourceRoutes extends UserInfoDirectives with SecurityDirectives wit
...
+    requireOneOfAction(
+      FullyQualifiedResourceId(SamResourceTypes.resourceTypeAdminName, ResourceId(policyId.resource.resourceTypeName.value)),
+      Set(SamResourceActions.adminAddMember),
+      userInfo.userId,
+      samRequestContext
+    ) {
+      addUserToPolicy
+    }
+
+  private def requireActionsForAdminRemoveUserFromPolicy(policyId: FullyQualifiedPolicyId, userInfo: UserInfo, samRequestContext: SamRequestContext)(removeUserFromPolicy: server.Route): server.Route =
```
s-rubenstein, line 109: This logic is borrowed from here: https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala#L389 which I _think_ is similar to what we want here?

marctalbott: `requireOneOfAction` probably isn't needed here. Since we're only checking one action, you can use `requireAction` instead similar to https://github.com/broadinstitute/sam/blob/314eee8aae77032123df5264b6d484967c7e1843/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala#L413

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminResourceRoutes.scala
```diff
@@ -78,4 +90,33 @@ trait AdminResourceRoutes extends UserInfoDirectives with SecurityDirectives wit
...
+      FullyQualifiedResourceId(SamResourceTypes.resourceTypeAdminName, ResourceId(policyId.resource.resourceTypeName.value)),
+      Set(SamResourceActions.adminRemoveMember),
+      userInfo.userId,
+      samRequestContext
+    ) {
+      removeUserFromPolicy
+    }
+
+  def adminPutUserInPolicy(policyId: FullyQualifiedPolicyId, subject: WorkbenchSubject, samRequestContext: SamRequestContext): server.Route =
+    put {
```
marctalbott: Duplicates the `put` above, probably best to remove this one

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminResourceRoutes.scala
```diff
@@ -78,4 +90,33 @@ trait AdminResourceRoutes extends UserInfoDirectives with SecurityDirectives wit
...
+      complete(resourceService.addSubjectToPolicy(policyId, subject, samRequestContext).map(_ => StatusCodes.NoContent))
+    }
+
+  def adminRemoveUserFromPolicy(policyId: FullyQualifiedPolicyId, subject: WorkbenchSubject, samRequestContext: SamRequestContext): server.Route =
+    delete {
```
marctalbott: Same here, this duplicates the `delete` above and can be removed

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
marctalbott: Yeah, this is correct, we need this to perform the auth checks that you added in the routes

marctalbott (approved):
Couple minor things to change, but looks good!

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminResourceRoutes.scala
s-rubenstein: Done!

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
rushtong: I saw this too and had this in my branch, good catch

rushtong (approved):
Pending Marcus' feedback, 👍🏽

s-rubenstein:
@marctalbott I would love a quick extra round of eyes on tests.

marctalbott (commented):
Looks good, I think there are a few more test cases that might be worth writing:

- user we're adding doesn't exist returns 404
- user we're adding is already a member returns successfully(? actually not sure maybe it should error)
- policy doesn't exist returns 404
- user we're removing doesn't exist returns 404
- user we're removing isn't a member of the policy returns successfully (? same question here as if the user is already a member when adding)
- negative test case where user doesn't have permissions and gets a 403 as a result (both endpoints)

s-rubenstein:
@marctalbott I think I have added all of your test ideas! Do you want to take another look?

marctalbott (approved):
Looks great! Thanks for adding those!

PR #538, opened 2021-09-14
AS-886: Add optional ToS payload to register routes

MatthewBemis:
Ticket: https://broadworkbench.atlassian.net/browse/AS-886

Tech design: https://docs.google.com/document/d/1hhSbqJyzHWYgNh95B5Vqq_L-cv1Yi3ytP2PoVSZcTRo/edit#heading=h.a2h76s64ql73

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/RejectionHandlers.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/RejectionHandlers.scala
@@ -10,11 +10,9 @@ import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 
 object RejectionHandlers {
 
-  def termsOfServiceErrorReport(tosUrl: String) = ErrorReport(StatusCodes.Forbidden, s"You must accept the Terms of Service in order to register. See ${tosUrl}")
-
   def termsOfServiceRejectionHandler(tosUrl: String) = RejectionHandler.newBuilder().handle {
     case RequestEntityExpectedRejection | MalformedRequestContentRejection(_, _) =>
-      complete(StatusCodes.Forbidden, termsOfServiceErrorReport(tosUrl))
+      complete(StatusCodes.Forbidden, ErrorReport(StatusCodes.Forbidden, s"You must accept the Terms of Service in order to register. See ${tosUrl}"))
   }.result()
 
 }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/RejectionHandlers.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/RejectionHandlers.scala
@@ -1,0 +1,20 @@
+package org.broadinstitute.dsde.workbench.sam
+package api
+
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.Directives.complete
+import akka.http.scaladsl.server.{MalformedRequestContentRejection, RejectionHandler, RequestEntityExpectedRejection}
+import org.broadinstitute.dsde.workbench.model.ErrorReport
+import org.broadinstitute.dsde.workbench.model.ErrorReportJsonSupport.ErrorReportFormat
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+
+object RejectionHandlers {
+
+  def termsOfServiceErrorReport(tosUrl: String) = ErrorReport(StatusCodes.Forbidden, s"You must accept the Terms of Service in order to register. See ${tosUrl}")
+
+  def termsOfServiceRejectionHandler(tosUrl: String) = RejectionHandler.newBuilder().handle {
+    case RequestEntityExpectedRejection | MalformedRequestContentRejection(_, _) =>
+      complete(StatusCodes.Forbidden, termsOfServiceErrorReport(tosUrl))
+  }.result()
+
+}
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserInfoDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserInfoDirectives.scala
@@ -1,12 +1,17 @@
 package org.broadinstitute.dsde.workbench.sam.api
 
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.StatusCodes
-import akka.http.scaladsl.server.Directives.onSuccess
-import akka.http.scaladsl.server.{Directive0, Directive1, Directives}
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.{Directive0, Directive1, Directives, MalformedRequestContentRejection}
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.sam.service.CloudExtensions
 import org.broadinstitute.dsde.workbench.sam._
+import org.broadinstitute.dsde.workbench.sam.api.RejectionHandlers.termsOfServiceRejectionHandler
+import org.broadinstitute.dsde.workbench.sam.config.TermsOfServiceConfig
 import org.broadinstitute.dsde.workbench.sam.dataAccess.DirectoryDAO
+import org.broadinstitute.dsde.workbench.sam.model.SamJsonSupport._
+import org.broadinstitute.dsde.workbench.sam.model.TermsOfServiceAcceptance
 import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
 
 /**
@@ -15,6 +20,7 @@ import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
 trait UserInfoDirectives {
   val directoryDAO: DirectoryDAO
   val cloudExtensions: CloudExtensions
+  val termsOfServiceConfig: TermsOfServiceConfig
 
   def requireUserInfo(samRequestContext: SamRequestContext): Directive1[UserInfo]
 
@@ -27,5 +33,20 @@ trait UserInfoDirectives {
         else r
       }
     }
+
+  def withTermsOfServiceAcceptance: Directive0 = {
+    Directives.mapInnerRoute { r =>
+      if(termsOfServiceConfig.enabled) {
+        handleRejections(termsOfServiceRejectionHandler(termsOfServiceConfig.url)) {
+          requestEntityPresent {
+            entity(as[TermsOfServiceAcceptance]) { tos =>
+              if (tos.value.equalsIgnoreCase(termsOfServiceConfig.url)) r
+              else reject(MalformedRequestContentRejection(s"Invalid ToS acceptance", new WorkbenchException(s"ToS URL did not match ${termsOfServiceConfig.url}")))
+            }
+          }
+        }
+      } else r
+    }
+  }
 
 }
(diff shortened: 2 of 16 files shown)


MatthewBemis:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
```diff
@@ -38,6 +38,32 @@ class UserRoutesV1Spec extends UserRoutesSpecHelper{
     }
   }
 
+  it should "create a user if ToS is enabled and the user specifies the ToS parameter correctly" in withTosEnabledRoutes { samRoutes =>
+    Post("/register/user/v1?tos=app.terra.bio/#terms-of-service") ~> samRoutes.route ~> check {
```
MatthewBemis: this test (and the v2 equivalent) are failing because `app.terra.bio/#terms-of-service` is not actually an allowed URL parameter, since it contains a forward slash. we'll need to either URL encode this parameter or change the URL that we accept. the reason i didn't jump right on URL encoding this was because that would also force the users to URL encode this parameter, and I'm not sure that we want to do that.

as-is, `/register/user/v1?tos=app.terra.bio/#terms-of-service` gets read as `/register/user/v1?tos=app.terra.bio`

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserInfoDirectives.scala
```diff
@@ -28,4 +30,20 @@ trait UserInfoDirectives {
...
+    val failDirective = Directives.failWith(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden, s"You must accept the Terms of Service in order to register. See ${termsOfServiceConfig.url}")))
+
+    Directives.mapInnerRoute { r =>
+      parameter("tos".?) { termsOfServiceUrlOpt =>
+        if(!termsOfServiceConfig.enabled) r //ToS-enforcement is not enabled, so let the request through
+        else {
+          termsOfServiceUrlOpt match {
+            case Some(url) => if(url.equals(termsOfServiceConfig.url)) r else failDirective
+            case None => failDirective //User didn't supply any acknowledgement of ToS, fail directive
+          }
```
davidangb, line 46: your call, you could Scala-simplify this to:
```
if (!termsOfServiceConfig.enabled || termsOfServiceUrlOpt.contains(termsOfServiceConfig.url))
    r
else
    failDirective
```

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
```diff
@@ -38,6 +38,32 @@ class UserRoutesV1Spec extends UserRoutesSpecHelper{
...
+    }
+
+    Post("/register/user/v1?tos=app.terra.bio/#terms-of-service") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.Conflict
+    }
+  }
+
+  it should "forbid the registration if ToS is enabled and the user doesn't specify a ToS parameter" in withTosEnabledRoutes { samRoutes =>
+    Post("/register/user/v1") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.Forbidden
```
davidangb, line 59: I suggest adding `responseAs[ErrorReport].message should startWith "You must accept the Terms of Service in order to register."` or similar here, to distinguish between a Forbidden thrown by the TOS check and a forbidden thrown by something else. This applies to all your new tests that look for Forbidden.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
davidangb: hm, good point! Going back to the original intent, the goal here is to prove _informed_ consent of the TOS … what can we ask the user to pass in that proves they know where to read the TOS? Some options I can think of:
* include the url `app.terra.bio/#terms-of-service` as-is, unencoded, in a POST body
* include the url, encoded, as a query param
* include two separate query params, something like "?toshost=app.terra.bio&toslocation=terms-of-service"

lastly - this may not be possible - use `/register/user/v1?tos=app.terra.bio#terms-of-service`. Note carefully that this omits the slash between "bio" and "terms-of-service". This relies on two things that need testing:
1. if a user enters `app.terra.bio#terms-of-service` in their browser, does it properly redirect to `https://app.terra.bio/#terms-of-service`? It does in my Chrome, but this may not be universal.
2. Can akka even see/retrieve the "#terms-of-service" portion of that URL? Technically it should be interpreted as a URL fragment and should not even be sent from the browser to the server, so akka may not have access to it at all.

davidangb: one more thought, maybe overkill: what if we stood up a new subdomain/webpage at something like "tos.terra.bio" and asked users to submit that value as the query param?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserInfoDirectives.scala
```diff
@@ -28,4 +33,27 @@ trait UserInfoDirectives {
...
+
+  def optionalEntity: Directive1[Option[TermsOfServiceAcceptance]] = {
+    entity(as[String]).flatMap { stringEntity =>
+      if(stringEntity == null || stringEntity.isEmpty) {
+        provide(Option.empty[TermsOfServiceAcceptance])
+      } else {
+        entity(as[TermsOfServiceAcceptance]).flatMap(e => provide(Some(e)))
+      }
+    }
+  }
```
davidangb, line 50: extracting the entity as String, testing the String, and then (sometimes) extracting the entity as TermsOfServiceAcceptance feels awkward here. Could this run into any `cannot be materialized more than once` errors?

How about setting a custom `RejectionHandler` on the 
`entity(as[TermsOfServiceAcceptance])`, so if the entity could not be unmarshalled to a TOS acceptance, we still send the proper error message? Maybe mix the `requestEntityPresent` directive in too?

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
```diff
@@ -38,6 +39,36 @@ class UserRoutesV1Spec extends UserRoutesSpecHelper{
...
+    }
+  }
+
+  it should "forbid the registration if ToS is enabled and the user doesn't specify the correct ToS url" in withTosEnabledRoutes { samRoutes =>
+    val tos = TermsOfServiceAcceptance("onemillionpats.com")
+    Post("/register/user/v1", tos) ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.Forbidden
+      responseAs[ErrorReport].message should startWith("You must accept the Terms of Service in order to register.")
+    }
+  }
```
davidangb, line 70: I suggest another test that posts a payload that won't unmarshal to a `TermsOfServiceAcceptance` class - for instance if the user posted some invalid json, or if they posted some random non-TermsOfServiceAcceptance case class

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
```diff
@@ -38,6 +39,36 @@ class UserRoutesV2Spec extends UserRoutesSpecHelper {
...
+  }
+
+  it should "forbid the registration if ToS is enabled and the user doesn't specify the correct ToS url" in withTosEnabledRoutes { samRoutes =>
+    val tos = TermsOfServiceAcceptance("onemillionpats.com")
+    Post("/register/user/v2/self", tos) ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.Forbidden
+      responseAs[ErrorReport].message should startWith("You must accept the Terms of Service in order to register.")
+    }
+  }
+
```
davidangb, line 71: same here - add a test for a non-TermsOfServiceAcceptance payload

davidangb (commented):
A few hopefully not-too-big comments inline, mostly this looks good, thank you!

sagehen03 (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserInfoDirectives.scala
MatthewBemis: That's much better, just pushed this change

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/RejectionHandlers.scala
```diff
@@ -0,0 +1,22 @@
...
+
+object RejectionHandlers {
+
+  def termsOfServiceErrorReport(tosUrl: String) = ErrorReport(StatusCodes.Forbidden, s"You must accept the Terms of Service in order to register. See ${tosUrl}")
+
+  def termsOfServiceRejectionHandler(tosUrl: String) = RejectionHandler.newBuilder().handle {
+    case RequestEntityExpectedRejection =>
+      complete(StatusCodes.Forbidden, termsOfServiceErrorReport(tosUrl))
+    case MalformedRequestContentRejection(_, _) =>
+      complete(StatusCodes.Forbidden, termsOfServiceErrorReport(tosUrl))
```
davidangb: nit: combine these into one case, since they do the same thing:
```suggestion
    case RequestEntityExpectedRejection | MalformedRequestContentRejection(_, _) =>
      complete(StatusCodes.Forbidden, termsOfServiceErrorReport(tosUrl))
```

davidangb (approved):
one syntax nit inline, otherwise :+1:. I don't need to re-review if you don't make any other changes!

MatthewBemis:
unrelated leo test failed

jenkins retest for the heck of it

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/RejectionHandlers.scala
```diff
@@ -0,0 +1,20 @@
...
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.Directives.complete
+import akka.http.scaladsl.server.{MalformedRequestContentRejection, RejectionHandler, RequestEntityExpectedRejection}
+import org.broadinstitute.dsde.workbench.model.ErrorReport
+import org.broadinstitute.dsde.workbench.model.ErrorReportJsonSupport.ErrorReportFormat
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+
+object RejectionHandlers {
+
+  def termsOfServiceErrorReport(tosUrl: String) = ErrorReport(StatusCodes.Forbidden, s"You must accept the Terms of Service in order to register. See ${tosUrl}")
```
s-rubenstein: Is this only used in one place? Is there any reason not to inline it?

s-rubenstein (approved):
LGTM other than one stylistic piece.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/RejectionHandlers.scala
MatthewBemis: ah, it was previously used in 2 places but is no longer! I'll fix!

MatthewBemis:
i'll give this one jenkins retest

PR #539, opened 2021-09-15
AS-884

sagehen03:
Ticket: https://broadworkbench.atlassian.net/browse/AS-884

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
@@ -17,7 +17,7 @@ class TosService (val directoryDao: DirectoryDAO, val appsDomain: String) extend
         case None =>
           logger.info("creating new ToS group")
           directoryDao.createGroup(BasicWorkbenchGroup(WorkbenchGroupName(getGroupName(currentVersion)),
-            Set.empty, WorkbenchEmail(s"${getGroupName(currentVersion)}_GROUP@${appsDomain}")), samRequestContext = SamRequestContext(None)).map(Option(_))
+            Set.empty, WorkbenchEmail(s"GROUP_${getGroupName(currentVersion)}@${appsDomain}")), samRequestContext = SamRequestContext(None)).map(Option(_))
       }
     } else
       IO.none
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
@@ -9,16 +9,20 @@ import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
 
 class TosService (val directoryDao: DirectoryDAO, val appsDomain: String) extends LazyLogging {
 
-  def createNewGroupIfNeeded(currentVersion:Int, isEnabled:Boolean): IO[Option[BasicWorkbenchGroup]] = {
-    if(!isEnabled){
+  def createNewGroupIfNeeded(currentVersion: Int, isEnabled: Boolean): IO[Option[BasicWorkbenchGroup]] = {
+    if(isEnabled) {
+      getTosGroup(currentVersion).flatMap {
+        case Some(_) =>
+          logger.info("Returning already created group")
+          IO.none
+        case None =>
+          logger.info("creating new ToS group")
+          directoryDao.createGroup(BasicWorkbenchGroup(WorkbenchGroupName(getGroupName(currentVersion)),
+            Set.empty, WorkbenchEmail(s"${getGroupName(currentVersion)}_GROUP@${appsDomain}")), samRequestContext = SamRequestContext(None)).map(Option(_))
+      }
+    } else {
+      logger.info("Else being hit")
       IO.none
-    }
-    getTosGroup(currentVersion).flatMap {
-      case Some(_) => IO.none
-      case None =>
-        logger.info("creating new ToS group")
-        directoryDao.createGroup(BasicWorkbenchGroup(WorkbenchGroupName(getGroupName(currentVersion)),
-          Set.empty, WorkbenchEmail(s"${getGroupName(currentVersion)}_GROUP@${appsDomain}")), samRequestContext = SamRequestContext(None)).map(Option(_))
     }
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
@@ -13,17 +13,14 @@ class TosService (val directoryDao: DirectoryDAO, val appsDomain: String) extend
     if(isEnabled) {
       getTosGroup(currentVersion).flatMap {
         case Some(_) =>
-          logger.info("Returning already created group")
           IO.none
         case None =>
           logger.info("creating new ToS group")
           directoryDao.createGroup(BasicWorkbenchGroup(WorkbenchGroupName(getGroupName(currentVersion)),
             Set.empty, WorkbenchEmail(s"${getGroupName(currentVersion)}_GROUP@${appsDomain}")), samRequestContext = SamRequestContext(None)).map(Option(_))
       }
-    } else {
-      logger.info("Else being hit")
+    } else
       IO.none
-    }
   }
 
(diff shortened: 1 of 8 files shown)


sagehen03:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -0,0 +1,31 @@
...
+
+class TosService (val directoryDao: DirectoryDAO) extends LazyLogging {
+
+  def createNewGroupIfNeeded(currentVersion:Int, isEnabled:Boolean): IO[Any] = {
+    if(!isEnabled || groupExists(currentVersion)){
+      IO.unit
+    } else {
+      logger.info("creating new ToS group")
+      directoryDao.createGroup(BasicWorkbenchGroup(WorkbenchGroupName(getGroupName(currentVersion)),
+        Set.empty, WorkbenchEmail("foo@bar.com")), samRequestContext = SamRequestContext(None))
```
sagehen03: Looking for suggestions about what to do with the email for the new group.  Should we move that to config?  Will the email be different per environment?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -0,0 +1,31 @@
...
+    }
+  }
+
+  def getGroupName(currentVersion:Int): String = {
+    s"tos_accepted_${currentVersion}"
```
sagehen03, line 27: Will the group name depend upon environment?  Should this get moved to config?

sagehen03:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
MatthewBemis: I don't believe this needs to be environment specific, so I don't think it's necessary to move this to config. If you follow the `All_Users` pattern (which I think is a reasonable pattern), it defines `WorkbenchGroupName` [with a hardcoded name in code](https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/CloudExtensions.scala#L25).

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -0,0 +1,31 @@
...
+import cats.effect.IO
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchGroupName}
+import org.broadinstitute.dsde.workbench.sam.dataAccess.DirectoryDAO
+import org.broadinstitute.dsde.workbench.sam.model.BasicWorkbenchGroup
+import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
+
+class TosService (val directoryDao: DirectoryDAO) extends LazyLogging {
+
+  def createNewGroupIfNeeded(currentVersion:Int, isEnabled:Boolean): IO[Any] = {
```
MatthewBemis: I think this might be a bit more idiomatic:

```
  def createNewGroupIfNeeded(currentVersion: Int, isEnabled: Boolean): IO[Option[BasicWorkbenchGroup]] = {
    if(isEnabled) {
      getTosGroup(currentVersion).flatMap {
        case Some(_) => IO.none
        case None =>
          logger.info("creating new ToS group")
          directoryDao.createGroup(BasicWorkbenchGroup(WorkbenchGroupName(getGroupName(currentVersion)),
            Set.empty, WorkbenchEmail("foo@bar.com")), samRequestContext = SamRequestContext(None)).map(Option(_))
      }
    } else IO.none
  }
```

And then a second change, turning `groupExists` into `getTosGroup`:

```
  def getTosGroup(currentVersion: Int): IO[Option[BasicWorkbenchGroup]] = {
    directoryDao.loadGroup(WorkbenchGroupName(getGroupName(currentVersion)), SamRequestContext(None))
  }
```

IMO the main benefits are:
1) Keeps everything within an IO context, so there's no need to do any `unsafeRunSync` weirdness which breaks up the execution chain
2) An `Any` return type is a bit sketchy in Scala. This change gives it a concrete return type. While the value isn't actually used upstream, it's nice to have a definitive return type, at least for the sake of readability (IMO)
3) Utilizes pattern matching instead of `isDefined`, which is pretty powerful in Scala and more idiomatic

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
MatthewBemis: You could do something similar to [how the All_Users group does it.](https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala#L80) The email address _domain_ will be unique per env (because we've got a different subdomain for each env), but the prefix won't be. And the domain is already in config.

All_Users is probably the closest analog we've got to these ToS groups.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
sagehen03: Thanks, Matt, I've incorporated this very useful feedback.

sagehen03:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  def createNewGroupIfNeeded(currentVersion: Int, isEnabled: Boolean): IO[Option[BasicWorkbenchGroup]] = {
+    if(isEnabled) {
+      getTosGroup(currentVersion).flatMap {
+        case Some(_) =>
+          IO.none
+        case None =>
+          logger.info("creating new ToS group")
+          directoryDao.createGroup(BasicWorkbenchGroup(WorkbenchGroupName(getGroupName(currentVersion)),
+            Set.empty, WorkbenchEmail(s"${getGroupName(currentVersion)}_GROUP@${appsDomain}")), samRequestContext = SamRequestContext(None)).map(Option(_))
```
MatthewBemis: nit: the naming convention is typically the prefix `GROUP_` followed by the name

MatthewBemis (approved):
LGTM, just have one nit but I don't need to re-review that one

s-rubenstein (commented):
Just asking for some context, why is Sam chosen to manage the life-cycle of this group? Is it treated differently than other groups in some way? Or could it be managed by the ToS service (which I think exists? But not 100% sure, so please correct me if I am wrong). Just want to make sure that this is all being put in the right place. It doesn't look much like it is treated that differently from other groups, so I am wondering why it needs special logic.

sagehen03:
> Just asking for some context, why is Sam chosen to manage the life-cycle of this group? Is it treated differently than other groups in some way? Or could it be managed by the ToS service (which I think exists? But not 100% sure, so please correct me if I am wrong). Just want to make sure that this is all being put in the right place. It doesn't look much like it is treated that differently from other groups, so I am wondering why it needs special logic.

@ansingh7115 I know we had some discussions about this, but I forgot why we settled on the approach that we did.  Could give Sky some context?

ansingh7115:
@s-rubenstein  Good question! We decided that the best place for ToS enforcement was Sam because we want all publicly available Terra services to be gated to only users who have agreed to the ToS. As such, the best place for this enforcement is the apache proxy. The apache proxy already talks to Sam to get whether a user is "enabled", so we decided that it made sense for the definition of "enabled" to also include the user signing the ToS. We will actually be removing the current ToS service and moving all functionality to Sam. Here's more about the design and implementation: https://docs.google.com/document/d/1hhSbqJyzHWYgNh95B5Vqq_L-cv1Yi3ytP2PoVSZcTRo/edit

s-rubenstein (approved):
Code all looks good to me. Are there designed endpoints in Sam that will allow users to sign the ToS? Otherwise, I would be concerned that API users of Sam won't have a way to use it without going to another service to sign ToS.

ansingh7115:
Adding the ability to sign the ToS is in this ticket: https://broadworkbench.atlassian.net/browse/AS-886

PR #540, opened 2021-09-20
[QA-1571] adding additional logging for createResource, foreign key violation

andy7i:
Ticket: https://broadworkbench.atlassian.net/browse/QA-1571

Adds a bit more logging to help with debugging (in general, and for QA-1571)

foreign key violation error message example (from local):
```
{"causes":[],"message":"Foreign Key Violation(s) while deleting group(s): List(Map(groupId -> 7238732, groupName -> billing-project_aj-ca-1206_owner, still used in group(s): -> {workspace_1b86b9cd-35f2-4867-ad53-54be57a65856_project-owner,workspace_77fa1cce-b004-45c9-8429-da403cf905eb_project-owner,workspace_9845724b-34d9-40ca-9452-d391677ff277_project-owner}))","source":"sam","stackTrace":[],"statusCode":500}%
```

example of new createResource:
```
{"@timestamp":"2021-09-21T21:06:01.389Z","@version":"1","message":"Creating new `billing-project` with resourceId: `gpalloc-qa-master-xnh3dtz`","logger_name":"org.broadinstitute.dsde.workbench.sam.service.ResourceService","thread_name":"scala-execution-context-global-81","severity":"INFO","level_value":20000}

```

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -53,7 +53,7 @@ resourceTypes = {
         description = "change the membership of the reader policy for this workspace"
       }
       "share_policy::can-compute" = {
-        description = "change the member of the can-compute policy for this workspace"
+        description = "change the membership of the can-compute policy for this workspace"
       }
       "read_policy::owner" = {
         description = "view the details of the owner policy for this workspace"
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
@@ -744,6 +744,7 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
   override def deleteAllResourcePolicies(resourceId: FullyQualifiedResourceId, samRequestContext: SamRequestContext): IO[Unit] = {
     val p = PolicyTable.syntax("p")
     val g = GroupTable.syntax("g")
+    val pg = GroupTable.syntax("pg") // problematic groups
     val gm = GroupMemberTable.syntax("gm")
 
     val fromAndWhereFragment = samsqls"""from ${PolicyTable as p} where ${p.resourceId} = (${loadResourcePKSubQuery(resourceId)})"""
@@ -766,26 +767,25 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
       _ <- writeResult match {
         case Left(fkViolation: PSQLException) if fkViolation.getSQLState == PSQLStateExtensions.FOREIGN_KEY_VIOLATION =>
           for {
-            (problematicGroupIdsToNames, problematicGroupMemberships) <- readOnlyTransaction("deletePolicyErrorFindProblematicGroups", samRequestContext) { implicit session =>
-            val problematicGroupIdsToNames: Map[GroupPK, String] =
-              samsql"""select ${g.result.id}, ${g.result.name}
+            problematicGroupIdsToNamesAndMembership <- readOnlyTransaction("deletePolicyErrorFindProblematicGroups", samRequestContext) { implicit session =>
+            val problematicGroupIdsToNames: List[Map[String, String]] =
+              samsql"""select ${g.result.id}, ${g.result.name}, ${pg.result.name}
                      from ${GroupTable as g}
+                     join ${GroupMemberTable as gm} on ${g.id} = ${gm.groupId}
+                     join ${GroupTable as pg} on ${gm.memberGroupId} = ${pg.groupId}
                      where ${g.id} in
                          (select distinct ${gm.result.memberGroupId}
                           from ${GroupMemberTable as gm}
-                          where ${gm.memberGroupId} in ($groupPKsToDelete))
-                  """
-                .map(rs => rs.get[GroupPK](g.resultName.id) -> rs.get[String](g.resultName.name)).list().apply().toMap
-            val problematicGroupMemberships: List[String] =
-              samsql"""select ${g.result.name}
-                     from ${GroupTable as g}
-                     join ${GroupMemberTable as gm} on ${g.id} = ${gm.groupId}
-                     where ${gm.memberGroupId} in (${problematicGroupIdsToNames.keys})"""
-                .map(rs => rs.get[String](g.resultName.name)).list().apply()
-            (problematicGroupIdsToNames, problematicGroupMemberships)
+                          where ${gm.memberGroupId} in ($groupPKsToDelete))"""
+                .map(rs =>
+                  List("groupId" -> rs.get[GroupPK](g.resultName.id).value.toString,
+                    "groupName" -> rs.get[String](g.resultName.name),
+                    "still used in:" -> rs.get[String](pg.resultName.name)))
+                    .list().apply().map(_.toMap)
+              problematicGroupIdsToNames
           }
           _ <- IO.raiseError[Unit](new WorkbenchExceptionWithErrorReport( // throws a 500 since that's the current behavior
-            ErrorReport(StatusCodes.InternalServerError, s"Foreign Key Violation -- Cannot delete group(s) ${problematicGroupIdsToNames.values.toList} because they are a member of group(s): ${problematicGroupMemberships}")))
+            ErrorReport(StatusCodes.InternalServerError, s"Foreign Key Violation while deleting groups: ${problematicGroupIdsToNamesAndMembership}")))
         } yield ()
         case Left(e) => IO.raiseError(e)
         case Right(_) => IO.unit
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
@@ -771,7 +771,7 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
               }.list().apply().toMap
             val problematicGroupMemberships: List[String] = samsql"""select ${g.result.name} from ${GroupTable as g}
                     join ${GroupMemberTable as gm} on ${g.id} = ${gm.groupId}
-                    where ${gm.memberGroupId} in ${problematicGroupIdsToNames.keys.toList}"""
+                    where ${gm.memberGroupId} in (${problematicGroupIdsToNames.keys})"""
               .map(_.toString).list().apply()
             Failure(new WorkbenchExceptionWithErrorReport(
               ErrorReport(StatusCodes.Conflict, s"Foreign Key Violation -- Cannot delete group(s) ${problematicGroupIdsToNames.values.toList} because they are a member of group(s): ${problematicGroupMemberships}")))
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
@@ -764,32 +764,43 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
                  where ${g.id} in ($groupPKsToDelete)""".update().apply()
         }
       }).attempt
-      _ <- writeResult match {
-        case Left(fkViolation: PSQLException) if fkViolation.getSQLState == PSQLStateExtensions.FOREIGN_KEY_VIOLATION =>
-          for {
-            problematicGroupIdsToNamesAndMembership <- readOnlyTransaction("deletePolicyErrorFindProblematicGroups", samRequestContext) { implicit session =>
-            val problematicGroupIdsToNames: List[Map[String, String]] =
-              samsql"""select ${g.result.id}, ${g.result.name}, ${pg.result.name}
+      _ <- handleResult(samRequestContext, writeResult, groupPKsToDelete)
+    } yield ()
+  }
+
+  private def handleResult(samRequestContext: SamRequestContext, writeResult: Either[Throwable, AnyVal], groupPKsToDelete: List[GroupPK]) = {
+    writeResult match {
+      case Left(fkViolation: PSQLException) if fkViolation.getSQLState == PSQLStateExtensions.FOREIGN_KEY_VIOLATION =>
+        handleForeignKeyViolation(samRequestContext, groupPKsToDelete)
+      case Left(e) => IO.raiseError(e)
+      case Right(_) => IO.unit
+    }
+  }
+
+  private def handleForeignKeyViolation(samRequestContext: SamRequestContext, groupPKsToDelete: List[GroupPK]) = {
+    val g = GroupTable.syntax("g")
+    val pg = GroupTable.syntax("pg") // problematic group
+    val gm = GroupMemberTable.syntax("gm")
+    for {
+      problematicGroupIdsToNamesAndMembership <- readOnlyTransaction("deletePolicyErrorFindProblematicGroups", samRequestContext) { implicit session =>
+        val problematicGroupIdsToNames: List[Map[String, String]] =
+          samsql"""select ${g.result.id}, ${g.result.name}, ${pg.result.name}
                      from ${GroupTable as g}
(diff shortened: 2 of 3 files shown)


src/main/resources/reference.conf
```diff
@@ -53,7 +53,7 @@ resourceTypes = {
         description = "change the membership of the reader policy for this workspace"
       }
       "share_policy::can-compute" = {
-        description = "change the member of the can-compute policy for this workspace"
+        description = "change the membership of the can-compute policy for this workspace"
```
andy7i, line 56: minor text fix

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -744,23 +744,53 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
   override def deleteAllResourcePolicies(resourceId: FullyQualifiedResourceId, samRequestContext: SamRequestContext): IO[Unit] = {
     val p = PolicyTable.syntax("p")
     val g = GroupTable.syntax("g")
+    val gm = GroupMemberTable.syntax("gm")
 
-    serializableWriteTransaction("deletePolicy", samRequestContext)({ implicit session =>
```
andy7i, line 748: moved this write transaction into a for comp.
this write transaction was split into 1 read transaction and 1 write transaction. this is so that we can still use groupPKsToDelete if the write transaction fails

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -102,14 +102,16 @@ class ResourceService(
     * @param userId
     * @return Future[Resource]
     */
-  def createResource(resourceType: ResourceType, resourceId: ResourceId, policiesMap: Map[AccessPolicyName, AccessPolicyMembership], authDomain: Set[WorkbenchGroupName], parentOpt: Option[FullyQualifiedResourceId], userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Resource] =
+  def createResource(resourceType: ResourceType, resourceId: ResourceId, policiesMap: Map[AccessPolicyName, AccessPolicyMembership], authDomain: Set[WorkbenchGroupName], parentOpt: Option[FullyQualifiedResourceId], userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Resource] = {
+    logger.info(s"Creating new `${resourceType}` with resourceId: `${resourceId}`")
```
andy7i: tech debt item -- it's difficult to find the createResource call made to sam.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -744,23 +744,53 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
-
-      samsql"""delete $fromAndWhereFragment""".update().apply()
+    val fromAndWhereFragment = samsqls"""from ${PolicyTable as p} where ${p.resourceId} = (${loadResourcePKSubQuery(resourceId)})"""
 
-      if (groupPKsToDelete.nonEmpty) {
-        samsql"""delete from ${GroupTable as g}
+    for {
+      groupPKsToDelete <- readOnlyTransaction("deletePolicyFindGroups", samRequestContext) { implicit session =>
+          samsql"""select ${p.result.groupId} $fromAndWhereFragment""".map(_.get[GroupPK](p.resultName.groupId)).list().apply()
+        }
```
gpolumbo-broad: I know the intention here was to split this into multiple transactions, but I'm not sure that's a good idea either.  Technically, if someone did something like adding a Policy to this resource between these transactions, then we would fail to properly clean up the Group behind that new Policy.  

If we had a cascading delete on the `sam_resource_policy.group_id` then I think we would be OK and we could just delete the policies and not have to worry about also deleting the groups, but that does not seem to be the case.  I do not know why the cascading delete is missing, so we would need to do some research before considering adding it.  

It seems like we should be able to read the `groupPKsToDelete` from within a single write transaction, but I don't know much about Slick and don't know what you were wrestling with, so I appreciate that it might not seem as straight forward as I think.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -744,23 +744,53 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+          samsql"""delete from ${GroupTable as g}
                  where ${g.id} in ($groupPKsToDelete)""".update().apply()
+        }
+      }).attempt
+      _ <- writeResult match {
+        case Left(fkViolation: PSQLException) if fkViolation.getSQLState == PSQLStateExtensions.FOREIGN_KEY_VIOLATION =>
+          for {
+            (problematicGroupIdsToNames, problematicGroupMemberships) <- readOnlyTransaction("deletePolicyErrorFindProblematicGroups", samRequestContext) { implicit session =>
+            val problematicGroupIdsToNames: Map[GroupPK, String] =
+              samsql"""select ${g.result.id}, ${g.result.name}
```
marctalbott: These queries feel like they could be condensed into a single query -- possibly something like this? While we're grabbing all the problematic group ids and names, also grab all of the groups that they're a member of.

```
select ${g.result.id}, ${g.result.name}, ${pg.result.name}
                     from ${GroupTable as g}
                     join ${GroupMemberTable as gm} on ${g.id} = ${gm.memberGroupId}
                     join ${GroupTable as pg} on ${pg.id} = ${gm.groupId}
                     where ${g.id} in
                         (select distinct ${gm.result.memberGroupId}
                          from ${GroupMemberTable as gm}
                          where ${gm.memberGroupId} in ($groupPKsToDelete))
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -744,23 +744,53 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+        // then delete all the resource policies
+        // then delete all the groups from the first query
+        samsql"""delete $fromAndWhereFragment""".update().apply()
+
+        if (groupPKsToDelete.nonEmpty) {
+          samsql"""delete from ${GroupTable as g}
                  where ${g.id} in ($groupPKsToDelete)""".update().apply()
+        }
+      }).attempt
+      _ <- writeResult match {
```
marctalbott: It might be worth breaking this match out into a separate function to avoid the nested match and for comp

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -744,23 +744,53 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+        case Left(fkViolation: PSQLException) if fkViolation.getSQLState == PSQLStateExtensions.FOREIGN_KEY_VIOLATION =>
```
gpolumbo-broad: In typical Greg fashion, I suggest extracting this case into its own method

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
gpolumbo-broad: +1

gpolumbo-broad (commented):
I'm a little worried about the transaction change, but I'm not sure the problems you were seeing either, so maybe we need to talk about it some more.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
gpolumbo-broad: +1

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
andy7i: > Technically, if someone did something like adding a Policy to this resource between these transactions, then we would fail to properly clean up the Group behind that new Policy.

good point, that thought didn't occur to me.

> If we had a cascading delete on the sam_resource_policy.group_id then I think we would be OK and we could just delete the policies and not have to worry about also deleting the groups, but that does not seem to be the case. I do not know why the cascading delete is missing, so we would need to do some research before considering adding it.

you might be right about this. Based on the dev db, we have a 1 to 1 mapping of policies to distinct groups. 
```
sam=> select COUNT(*) from sam_resource_policy ; select COUNT(*) from (select distinct group_id from sam_resource_policy) as nq;
 602731

 602731
```
Maybe worth adding this to the tech debt backlog?


>It seems like we should be able to read the groupPKsToDelete from within a single write transaction, but I don't know much about Slick and don't know what you were wrestling with, so I appreciate that it might not seem as straight forward as I think.

If `groupPKsToDelete` is set inside the write transaction, and the write transaction succeeds, I think getting `groupPKsToDelete` is simple. However, we need `groupPKsToDelete` when the write transaction fails. Is there a way to get `groupPKsToDelete` if the write transaction fails?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -745,22 +745,64 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+                     join ${GroupMemberTable as gm} on ${g.id} = ${gm.memberGroupId}
+                     join ${GroupTable as pg} on ${gm.groupId} = ${pg.id}
+                     where ${g.id} in
+                         (select distinct ${gm.result.memberGroupId}
+                          from ${GroupMemberTable as gm}
+                          where ${gm.memberGroupId} in ($groupPKsToDelete))"""
+        .map(rs =>
+          Map("groupId" -> rs.get[GroupPK](g.resultName.id).value.toString,
+            "groupName" -> rs.get[String](g.resultName.name),
+            "still used in group:" -> rs.get[String](pg.resultName.name)))
```
marctalbott: might be more readable if these were collected into a single list so the output looked something like 
`groupId -> 12, groupName -> 'example', still used in group: -> [34, 45, 56]` , but not blocking

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
marctalbott: you could just get `groupPKstoDelete` again in the error handling transaction. Not the most efficient, but at this point things have already gone wrong so as long as it doesn't take down Sam it's not like users will mind if an error gets back to them slightly slower.

Agreed that it'd be nice to have the cascading delete, but like Greg I can't remember why we omitted it in the first place and I'd want to do a bit of research to be sure we aren't overlooking some important reason

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -745,22 +745,64 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+        if (groupPKsToDelete.nonEmpty) {
+          samsql"""delete from ${GroupTable as g}
+                 where ${g.id} in ($groupPKsToDelete)""".update().apply()
+        }
+      }).attempt
+      _ <- handleResult(samRequestContext, writeResult, groupPKsToDelete)
+    } yield ()
+  }
+
+  private def handleResult(samRequestContext: SamRequestContext, writeResult: Either[Throwable, AnyVal], groupPKsToDelete: List[GroupPK]) = {
```
marctalbott: could you add type annotations to these private functions?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -745,22 +745,64 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+      case Left(e) => IO.raiseError(e)
+      case Right(_) => IO.unit
+    }
+  }
 
-      val groupPKsToDelete = samsql"""select ${p.result.groupId} $fromAndWhereFragment""".map(_.get[GroupPK](p.resultName.groupId)).list().apply()
+  private def handleForeignKeyViolation(samRequestContext: SamRequestContext, groupPKsToDelete: List[GroupPK]) = {
+    for {
+      problematicGroups <- getGroupsCausingForeignKeyViolation(samRequestContext, groupPKsToDelete)
+      _ <- IO.raiseError[Unit](new WorkbenchExceptionWithErrorReport( // throws a 500 since that's the current behavior
```
marctalbott, line 782: Out of curiosity, why is `[Unit]` needed here but not when we're using `IO.raiseError` above?

andy7i: I'm not sure why, but without the `[Unit]` the compiler says `dead code following this construct`.

```
[error] /app/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala:782:9: dead code following this construct
[error]       _ <- IO.raiseError(new WorkbenchExceptionWithErrorReport( // throws a 500 since that's the current behavior
[error]         ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

I also get the same error if I put add a [Nothing]. (`IO.raiseError(e)` return type is IO[Nothing]). 

Screen Shot 2021-09-23 at 9 29 38 AM

Seems like there's technically a type mismatch but in the rest of sam code it seems like we don't usually need to typecast it and it doesn't cause any issues.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
gpolumbo-broad: I know it's a pain, but I think it's worth keeping these queries in the same transaction, and following Marc's suggestion of rerunning the query when generating the error message.  Not ideal, but should work.

As for trying to get the intermediate result of `groupPKstoDelete`, I would try starting the transaction and then try using `zip` per https://scala-slick.org/doc/3.2.0/dbio.html#sequential-execution.  I'm not 100% that that will work for you, but that's where I'd start at least.

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
andy7i: totally makes sense. I combined the transactions and pulled out the extracted the query into a val and now it is run twice.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
andy7i: updated:
```
{"causes":[],"message":"Foreign Key Violation(s) while deleting group(s): List(Map(groupId -> 7238732, groupName -> billing-project_aj-ca-1206_owner, still used in group(s): -> {workspace_1b86b9cd-35f2-4867-ad53-54be57a65856_project-owner,workspace_77fa1cce-b004-45c9-8429-da403cf905eb_project-owner,workspace_9845724b-34d9-40ca-9452-d391677ff277_project-owner}))","source":"sam","stackTrace":[],"statusCode":500}%
```

gpolumbo-broad (approved):

marctalbott (approved):

PR #542, opened 2021-10-12
Work on adding a pet conditional

s-rubenstein:
Ticket: https://broadworkbench.atlassian.net/browse/CA-1274
This adds a check for v2 google projects _only_ that someone trying to create a pet service account has pet creator access. This is granted via reference.conf via any permission on the workspace. The goal here is to block use of this endpoint on v2 projects for people who shouldn't have any access to the google project, while allowing anyone to create pets in v1 projects, because there is no direct path to the google project for all users who need to create pets.
---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
@@ -95,7 +95,7 @@ trait SecurityDirectives {
       checkPermissionAndReturnRoute(innerRoute, resource, requestedActions, userId, samRequestContext)
     }
 
-  def requireOneOfActionIfWorkspace(resource: FullyQualifiedResourceId, requestedActions: Set[ResourceAction], userId: WorkbenchUserId, samRequestContext: SamRequestContext): Directive0 = Directives.mapInnerRoute { innerRoute =>
+  def requireOneOfActionIfParentIsWorkspace(resource: FullyQualifiedResourceId, requestedActions: Set[ResourceAction], userId: WorkbenchUserId, samRequestContext: SamRequestContext): Directive0 = Directives.mapInnerRoute { innerRoute =>
     onSuccess(resourceService.getResourceParent(resource, samRequestContext)) {
       case Some(parent) => if (parent.resourceTypeName == SamResourceTypes.workspaceName) {
         checkPermissionAndReturnRoute(innerRoute, resource, requestedActions, userId, samRequestContext)
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
@@ -8,7 +8,7 @@ import cats.effect.IO
 import org.broadinstitute.dsde.workbench.model.{ErrorReport, WorkbenchExceptionWithErrorReport, WorkbenchUserId}
 import org.broadinstitute.dsde.workbench.sam.ImplicitConversions.ioOnSuccessMagnet
 import org.broadinstitute.dsde.workbench.sam._
-import org.broadinstitute.dsde.workbench.sam.model.{FullyQualifiedResourceId, ResourceAction, ResourceType, ResourceTypeName, SamResourceActions}
+import org.broadinstitute.dsde.workbench.sam.model.{FullyQualifiedResourceId, ResourceAction, ResourceType, ResourceTypeName, SamResourceActions, SamResourceTypes}
 import org.broadinstitute.dsde.workbench.sam.service.{PolicyEvaluatorService, ResourceService}
 import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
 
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
```diff
@@ -95,14 +95,20 @@ trait GoogleExtensionRoutes extends ExtensionRoutes with UserInfoDirectives with
...
-                      post {
-                        entity(as[Set[String]]) { scopes =>
-                          complete {
-                            googleExtensions
-                              .getPetServiceAccountToken(WorkbenchUser(userInfo.userId, None, userInfo.userEmail, None), GoogleProject(project), scopes, samRequestContext)
-                              .map { token =>
-                                StatusCodes.OK -> JsString(token)
-                              }
+                      requireOneOfAction(
+                        FullyQualifiedResourceId(ResourceTypeName("google-project"), ResourceId("GOOGLE PROJECT")),
```
s-rubenstein: I put these on here as hard coded strings specifically to flag them to ask for input here. There's gotta be a better way to get our resource type names than this. However, when I tried looking, there were only scattered assigning to constants across the codebase. I am curious if we have a better method by now.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
```diff
@@ -77,18 +78,33 @@ trait SecurityDirectives {
...
-        } else {
-          val forbiddenErrorMessage = s"You may not perform any of ${requestedActions.mkString("[", ", ", "]").toUpperCase} on ${resource.resourceTypeName.value}/${resource.resourceId.value}"
-          determineErrorMessage(resource, userId, forbiddenErrorMessage, samRequestContext)
-        }
+      checkPermissionAndReturnRoute(innerRoute, resource, requestedActions, userId, samRequestContext)
+    }
+
+  def requireOneOfActionIfWorkspace(resource: FullyQualifiedResourceId, requestedActions: Set[ResourceAction], userId: WorkbenchUserId, samRequestContext: SamRequestContext): Directive0 = Directives.mapInnerRoute { innerRoute =>
+    onSuccess(resourceService.getResourceParent(resource, samRequestContext)) {
+      case Some(parent) => if (parent.resourceTypeName == ResourceTypeName("Workspace")) {
```
s-rubenstein: Same as below, we've gotta have a better method to get these names than this. Anyone know what it is? (Also, I know this won't work rn, I just wanted to get feedback now!)

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
marctalbott: I would add them as `ResourceTypeName`s in the SamModel like `resource_type_admin`
https://github.com/broadinstitute/sam/blob/e642c8c6118e83e365cd8cb5337d74760bf36b61/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala#L101

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
```diff
@@ -73,15 +73,21 @@ trait GoogleExtensionRoutes extends ExtensionRoutes with UserInfoDirectives with
...
-                        googleExtensions
-                          .getPetServiceAccountKey(WorkbenchUser(userInfo.userId, None, userInfo.userEmail, None), GoogleProject(project), samRequestContext)
-                          .map { key =>
-                            StatusCodes.OK -> key.parseJson
-                          }
+                    requireOneOfActionIfParentIsWorkspace(
+                      FullyQualifiedResourceId(SamResourceTypes.googleProjectName, ResourceId(project)),
+                      Set(SamResourceActions.createPet),
+                      userInfo.userId,
+                      samRequestContext) {
```
ehigham, line 80: I'm wondering if this is the right place to add this. I perfer to keep the routes definition fairly minimal and leave the logic to the handler - there's already a lot going on in this file. Is there a reason for keeping it here?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
```diff
@@ -77,18 +78,33 @@ trait SecurityDirectives {
     }
   }
 
+  def checkPermissionAndReturnRoute(route: server.Route, resource: FullyQualifiedResourceId, requestedActions: Set[ResourceAction], userId: WorkbenchUserId, samRequestContext: SamRequestContext): server.Route = {
```
ehigham: maybe `failUnlessUserHasPermission`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
s-rubenstein: So, my preference is generally for permission checks to live at the route/controller level, because that way we:

1. Do as little work as possible before checking permissions
2. Keep the business logic of the actual endpoint separate from the permissions checking (which is often fairly unrelated from the permissions check)
3. Can wrap an entire section of routing in a permissions check if we want.

I can see an argument for moving this into the service level as a one-liner at the top of most functions, but I think there's enough reason to keep it here that I prefer that. I think this also aligns more with how permissions are already done in Sam (https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala#L94)

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
s-rubenstein: I like that! Will change.

ehigham (approved):

andy7i (approved):
we may want to update the access policies doc with these changes: https://docs.google.com/document/d/1uOGtmw_l_Ve2gqJohLGntJmQz69md2h3T2EAvyKTm20/edit#

s-rubenstein:
@andy7i done

PR #543, opened 2021-10-12
POL-97 Azure b2c

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/POL-97

This removes all Identity Concentrator and JWT stuff which never panned out. Then it adds Azure B2C integration.

The Azure B2C integration looks much the same as Google Auth, it relies on the same http headers. The different is that the Azure B2C Id is a UUID and the Google Subject Id is a 21 digit integer. In addition, an Azure login may be accompanied with an idp access token which is an opaque google access token. This happens when a user selects Google as the idp in Azure B2C. This token is used in the event that the Azure B2C Id is not found in Sam in order to lookup an account in Sam that might be associated to the opaque google access token. This can happen when a user has previously logged in directly with Google Authentication and is now logging in with Azure B2C via the Google idp. If a user is found then we update their record with the Azure B2C id.

I would start reviewing with src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala. A bunch of the changes are removing identity concentrator stuff. Many other changes are making the google subject id fully optional.

also see https://github.com/broadinstitute/firecloud-orchestration/pull/894

and also https://github.com/broadinstitute/firecloud-develop/pull/2690 (eventually)

---

**PR checklist**


dvoet:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -67,17 +71,26 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
               case Some(uid: WorkbenchUserId) =>
                 for {
                   groups <- directoryDAO.listUserDirectMemberships(uid, samRequestContext)
-                  _ <- directoryDAO.setGoogleSubjectId(uid, user.googleSubjectId, samRequestContext)
-                  _ <- registrationDAO.setGoogleSubjectId(uid, user.googleSubjectId, samRequestContext)
+                  _ <- user.googleSubjectId match {
```
Qi77Qi: fyi, you can use `.traverse`..sth like `user.googleSubjectId.traverse{sid => ???}`..this will only execute ??? if `googleSubjectId` exists

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -7,159 +7,141 @@ import akka.http.scaladsl.server.Directives._
...
 }
 
 object StandardUserInfoDirectives {
-  val SAdomain: Regex = "\\S+@\\S+\\.iam\\.gserviceaccount\\.com".r
+  val SAdomain: Regex = "(\\S+@\\S+\\.iam\\.gserviceaccount\\.com$)".r
   val accessTokenHeader = "OIDC_access_token"
   val expiresInHeader = "OIDC_CLAIM_expires_in"
   val emailHeader = "OIDC_CLAIM_email"
-  val googleSubjectIdHeader = "OIDC_CLAIM_user_id"
-  val authorizationHeader = "Authorization"
```
marctalbott, line 100: Is this header duplicated by `OIDC_access_token`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -7,159 +7,141 @@ import akka.http.scaladsl.server.Directives._
...
-
-        case _ =>
-          // request coming through Apache proxy with OIDC headers
-          withChildTraceSpan("OIDC-request", samRequestContext).flatMap { samRequestContext =>
-            (headerValueByName(accessTokenHeader).as(OAuth2BearerToken) &
-              headerValueByName(googleSubjectIdHeader).as(GoogleSubjectId) &
-              headerValueByName(expiresInHeader) &
-              headerValueByName(emailHeader).as(WorkbenchEmail)).as(OIDCHeaders).flatMap(fromOIDC(_, samRequestContext))
-          }
+  private def verifyAzureUserIsNotAlreadyGoogleUser(oidcHeaders: OIDCHeaders, samRequestContext: SamRequestContext): Directive1[OIDCHeaders] = {
```
marctalbott: Do we need any kind of verification that a Google user is not already an Azure user? Do we not care?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -57,8 +57,12 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
     */
```
marctalbott, line 54: would you mind updating this comment to include the azure b2c id?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleOpaqueTokenResolver.scala
```diff
@@ -55,15 +61,17 @@ class StandardGoogleOpaqueTokenResolver(directoryDAO: DirectoryDAO, googleTokenI
...
-  private def lookupUser(maybeTokenInfoResponse: Option[TokenInfoResponse], samRequestContext: SamRequestContext): IO[Option[WorkbenchUser]] = {
+  private def lookupUser(maybeTokenInfoResponse: Option[TokenInfoResponse], samRequestContext: SamRequestContext): IO[Option[GoogleTokenInfo]] = {
     maybeTokenInfoResponse.map {
-      case TokenInfoResponse(googleSubjectId, email) =>
+      case TokenInfoResponse(googleSubjectId) =>
         directoryDAO.loadSubjectFromGoogleSubjectId(googleSubjectId, samRequestContext).map(_.flatMap {
-          case userId: WorkbenchUserId => Option(WorkbenchUser(userId, Option(googleSubjectId), email, None))
-          case _ => None
+          case userId: WorkbenchUserId => Option(GoogleTokenInfo(Option(userId), googleSubjectId))
+          case PetServiceAccountId(userId, _) => Option(GoogleTokenInfo(Option(userId), googleSubjectId))
```
marctalbott: Should we block pets here?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: I assume you mean `OAUTH2_CLAIM_idp_access_token`. No it is not. `OIDC_access_token` is the access token in the authorization header. `OAUTH2_CLAIM_idp_access_token` is the idp specific access token (i.e. google access token) embedded in the azure b2c jwt.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: I updated the functionality so that when a user is created via an azure access token which contains a google access token both the azure b2c id and google subject id will be populated. Then the user is free to use either direct google or azure authn.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleOpaqueTokenResolver.scala
dvoet: I am happy for other opinions here. At the moment I don't think pets can actually make it here because they can't authenticate via azure b2c. If that were to change, would this code be right? I think my mind is changing here.

dvoet:
jenkins retest after orch PR merged

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: So, if a user registers with their azure token, their google id will be populated and if they try to register with a google token the google subject id will already exist and an error will be thrown (just like if they try to register twice with the same google token).

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleOpaqueTokenResolver.scala
dvoet: I removed this line

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
marctalbott: Sorry, wasn't clear, I was just checking that the access token in the `Authorization` header was also in the `OIDC_access_token` header

dvoet:
jenkins retest

dvoet:
jenkins retest

dvoet:
jenkins retest

dvoet:
jenkins retest - what do I need to do to get a clean run around here

marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -7,159 +7,150 @@ import akka.http.scaladsl.server.Directives._
...
-        case None => lookUpByGoogleSubjectId(googleSubjectId, directoryDAO, samRequestContext).map(uid => UserInfo(token, uid, email, expiresIn))
-      }
-    } else {
-      lookUpByGoogleSubjectId(googleSubjectId, directoryDAO, samRequestContext).map(uid => UserInfo(token, uid, email, expiresIn))
+  val userIdHeader = "OIDC_CLAIM_user_id"
+  val idpAccessTokenHeader = "OAUTH2_CLAIM_idp_access_token"
+
+  def getUserInfo(directoryDAO: DirectoryDAO, maybeGoogleOpaqueTokenResolver: Option[GoogleOpaqueTokenResolver], oidcHeaders: OIDCHeaders, samRequestContext: SamRequestContext): IO[UserInfo] = {
+    oidcHeaders match {
+      case OIDCHeaders(token, Left(googleSubjectId), _, saEmail@WorkbenchEmail(SAdomain(_)), _) =>
```
gpolumbo-broad: For these pattern matches, I'm wondering if it wouldn't be better to write it as:
```suggestion
      case OIDCHeaders(_, Left(googleSubjectId), _, saEmail@WorkbenchEmail(SAdomain(_)), _) =>
```
And same thing in the other two `case`s to use `_` for the headers we don't care about in the pattern match?  I believe it would make it easier to read if the only things the patterns are checking are the fields we care about, which seems to be the `externalId` and `email`.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -7,159 +7,150 @@ import akka.http.scaladsl.server.Directives._
...
-  }
-
-  private def updateUserIdentityConcentratorId(jwtUserInfo: JwtUserInfo, bearerToken: OAuth2BearerToken, directoryDAO: DirectoryDAO, identityConcentratorService: IdentityConcentratorService, samRequestContext: SamRequestContext): IO[Option[WorkbenchUser]] = {
-    for {
-      googleIdentities <- identityConcentratorService.getGoogleIdentities(bearerToken)
-      (googleSubjectId, _) <- singleGoogleIdentity(jwtUserInfo.sub, googleIdentities)
-      _ <- directoryDAO.setUserIdentityConcentratorId(googleSubjectId, jwtUserInfo.sub, samRequestContext)
-      maybeUser <- directoryDAO.loadUserByIdentityConcentratorId(jwtUserInfo.sub, samRequestContext)
-    } yield maybeUser
+    } yield maybeUserAgain.getOrElse(throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"user Id $azureB2CId not found in sam")))
```
gpolumbo-broad: Does a `401` make more sense here instead of a `404`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -7,159 +7,150 @@ import akka.http.scaladsl.server.Directives._
...
+  private def populateGoogleIdAndThrowIfUserExists(samRequestContext: SamRequestContext, accessToken: OAuth2BearerToken, googleOpaqueTokenResolver: GoogleOpaqueTokenResolver, azureB2CId: AzureB2CId, createWorkbenchUser: CreateWorkbenchUser): Directive1[CreateWorkbenchUser] = {
     onSuccess {
-      getUserInfoFromJwt(jwtUserInfo, bearerToken, directoryDAO, icService, samRequestContext).unsafeToFuture()
+      googleOpaqueTokenResolver.getWorkbenchUser(accessToken, samRequestContext).flatMap {
+        case None =>
+          IO.pure(createWorkbenchUser) // access token not valid
+        case Some(GoogleTokenInfo(None, googleSubjectId)) => // access token valid but not already a user
+          IO.pure(createWorkbenchUser.copy(googleSubjectId = Option(googleSubjectId)))
+        case Some(GoogleTokenInfo(Some(_), _)) => // already a user
+          IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"user $azureB2CId already exists")))
```
gpolumbo-broad: Perhaps modify the pattern match to capture the `googleSubjectId` and then:

```suggestion
          IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"user with Google Subject ID $googleSubjectId already exists")))
```
Because if I'm reading this right, we're looking up the user record based on their `googleSubjectId`, not their `azureB2CId`.  And I think it's important that in any of these log statements now, we be explicit as to which user ID we are printing.  Yeah, we should be able to tell by if it's a `long` or a `uuid`, but I think it is better to explicit.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -172,57 +163,14 @@ object StandardUserInfoDirectives {
...
-    val bearerPrefix = "bearer "
-    // need to be case insensitive when looking for the 'bearer' prefix so can't use stripPrefix
-    if (authorizationHeader.toLowerCase.startsWith(bearerPrefix)) {
-      Option(OAuth2BearerToken(authorizationHeader.drop(bearerPrefix.length)))
-    } else {
-      logger.debug("authorization header not a bearer token")
-      None
-    }
-  }
+final case class CreateWorkbenchUser(id: WorkbenchUserId, googleSubjectId: Option[GoogleSubjectId], email: WorkbenchEmail, azureB2CId: Option[AzureB2CId])
```
gpolumbo-broad: Perhaps a nit, but I'd prefer this class be called `NewWorkbenchUser` or maybe `PendingWorkbenchUser` or something like that.  When I saw the word "create" here, I was thinking verb, and expecting this to do more than just be a value object.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20211010_azure_b2c_id.xml
```diff
@@ -0,0 +1,18 @@
...
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="drop_identity_concentrator_id">
+        <dropColumn tableName="sam_user" columnName="identity_concentrator_id"/>
+    </changeSet>
+    <changeSet logicalFilePath="dummy" author="dvoet" id="add_azure_b2c_id">
+        <addColumn tableName="sam_user">
+            <column name="azure_b2c_id" type="VARCHAR(40)">
+                <constraints unique="true"/>
```
gpolumbo-broad, line 14: This seems like a column we will want indexed

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/IdentityConcentratorConfig.scala
```diff
@@ -1,3 +1,3 @@
 package org.broadinstitute.dsde.workbench.sam.config
 
-final case class IdentityConcentratorConfig(baseUrl: String, threadPoolSize: Int)
\ No newline at end of file
+final case class GoogleOpaqueTokenResolverConfig(googleTokenInfoUrl: String, threadPoolSize: Int)
```
gpolumbo-broad: Rename File too

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -46,19 +47,19 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+    * this is a new user, create them.
     */
   protected[service] def registerUser(user: CreateWorkbenchUser, samRequestContext: SamRequestContext): IO[WorkbenchUser] =
     for {
-      existingSubFromGoogleSubjectId <- directoryDAO.loadSubjectFromGoogleSubjectId(user.googleSubjectId, samRequestContext)
-      user <- existingSubFromGoogleSubjectId match {
+      existingSubject <- user match {
+        case CreateWorkbenchUser(_, Some(googleSubjectId), _, _) => directoryDAO.loadSubjectFromGoogleSubjectId(googleSubjectId, samRequestContext)
+        case CreateWorkbenchUser(_, _, _, Some(azureB2CId)) => directoryDAO.loadUserByAzureB2CId(azureB2CId, samRequestContext).map(_.map(_.id))
+        case _ => IO.raiseError(new WorkbenchException("cannot create user when neither google subject id nor azure b2c id exists"))
```
gpolumbo-broad: Is there a reason why this exception is not wrapped up in a `WorkbenchExceptionWithErrorReport`?  

If left as is, I assume this will just bubble up as a `500`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: yes, but the `Authorization` header value starts with `Bearer ` whereas `OIDC_access_token` has that stripped out already

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -46,19 +47,19 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+    * First lookup user by either googleSubjectId or azureB2DId, whichever is populated. If the user exists
+    * throw a conflict error. If the user does not exist look them up by email. If the user email exists
+    * then this is an invited user, update their googleSubjectId and/or azureB2CId. If the email does not exist,
+    * this is a new user, create them.
     */
   protected[service] def registerUser(user: CreateWorkbenchUser, samRequestContext: SamRequestContext): IO[WorkbenchUser] =
     for {
-      existingSubFromGoogleSubjectId <- directoryDAO.loadSubjectFromGoogleSubjectId(user.googleSubjectId, samRequestContext)
-      user <- existingSubFromGoogleSubjectId match {
+      existingSubject <- user match {
```
gpolumbo-broad: This method already has a lot of nested `for` comps and `match`es already.  My general preference would be to refactor the whole method, but I'll settle for not continuing making it any worse, which means could we extract this `user match { … }` into its own method, maybe something like `maybeLoadExistingUser` or something like that.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
gpolumbo-broad: Furthermore, just looked at the existing `WorkbenchUser` and they seem identical.  So I don't think we need this case class

gpolumbo-broad (changes requested):
Made it through the main code.  I'll keep going through the tests, but wanted to get this back to you now because I know it has taken a while.  Sorry for the delay.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
dvoet: this is an unexpected situation, it will just be a 500 which is appropriate, there is nothing the user can do about this

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: interesting… this only became the case in this PR :)

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: I agree 404 is not right. I would like to avoid 401 though because only the proxy returns 401 at this point making it really easy to pinpoint that as a source of error. So I am inclined to make this a 403 or a 400.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: the change is easy enough to make here and I will do so to make the cases clear. However in general this is not always desired because this is not only pattern matching but extracting values out of the pattern. So in this case I will have a bunch more `oidcHeaders.` but in general it can be a pain.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: how do you feel about printing the email address instead? I don't think the id is particularly helpful to the user anyway and neither is the subtly of google or azure id

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20211010_azure_b2c_id.xml
dvoet: the unique constraint creates an index

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -52,45 +52,55 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+        case None =>
+          createUserInternal(WorkbenchUser(user.id, user.googleSubjectId, user.email, user.azureB2CId), samRequestContext)
+
+        case Some(_) =>
+          //We don't support inviting a group account or pet service account
+          IO.raiseError[WorkbenchUser](
+            new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"$user is not a regular user. Please use a different endpoint")))
+
+      }
+    } yield updated
```
gpolumbo-broad, line 72: This refactoring here makes me very happy!  To me, this is much much easier to read.  Thanks!

dvoet:
jenkins retest

gpolumbo-broad (approved):

dvoet:
jenkins retest - last one all passed but fiab-stop failed with some git issue

PR #545, opened 2021-10-26
AS-907: Change GET Register endpoints to return tosAccepted status

ansingh7115:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
@@ -198,19 +198,17 @@ trait UserRoutes extends UserInfoDirectives with SamRequestContextDirectives {
     pathPrefix("v1") {
       withSamRequestContext { samRequestContext =>
         requireUserInfo(samRequestContext) { userInfo =>
-          pathPrefix("tos") {
-            pathPrefix("accept") {
-              pathEndOrSingleSlash {
-                post {
-                  withTermsOfServiceAcceptance {
-                    complete {
-                      userService.acceptTermsOfService(userInfo.userId, samRequestContext).map { statusOption =>
-                        statusOption
-                          .map { status =>
-                            StatusCodes.OK -> Option(status)
-                          }
-                          .getOrElse(StatusCodes.NotFound -> None)
-                      }
+          pathPrefix("tos" / "accept") {
+            pathEndOrSingleSlash {
+              post {
+                withTermsOfServiceAcceptance {
+                  complete {
+                    userService.acceptTermsOfService(userInfo.userId, samRequestContext).map { statusOption =>
+                      statusOption
+                        .map { status =>
+                          StatusCodes.OK -> Option(status)
+                        }
+                        .getOrElse(StatusCodes.NotFound -> None)
                     }
                   }
                 }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
@@ -53,9 +53,7 @@ trait UserRoutes extends UserInfoDirectives with SamRequestContextDirectives {
               get {
                 parameter("userDetailsOnly".?) { userDetailsOnly =>
                   complete {
-                    println("USER DETAILS: " + user.toString)
                     userService.getUserStatus(user.userId, userDetailsOnly.exists(_.equalsIgnoreCase("true")), samRequestContext).map { statusOption =>
-                      println("SAM ROUTES STATUS OPTION: " + statusOption)
                       statusOption
                         .map { status =>
                           StatusCodes.OK -> Option(status)
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
@@ -39,22 +39,13 @@ trait UserRoutes extends UserInfoDirectives with SamRequestContextDirectives {
       (pathPrefix("v1") | pathEndOrSingleSlash) {
         pathEndOrSingleSlash {
           post {
-            withTermsOfServiceAcceptance {
-              withSamRequestContext { samRequestContext =>
-                requireCreateUser(samRequestContext) { createUser =>
-                  complete {
-                    userService.createUser(createUser, true, samRequestContext).map(userStatus => StatusCodes.Created -> userStatus)
-                  }
+            withSamRequestContext { samRequestContext =>
+              requireCreateUser(samRequestContext) { createUser =>
+                complete {
+                  userService.createUser(createUser, samRequestContext).map(userStatus => StatusCodes.Created -> userStatus)
                 }
               }
-            } ~
-              withSamRequestContext { samRequestContext =>
-                requireCreateUser(samRequestContext) { createUser =>
-                  complete {
-                    userService.createUser(createUser, false, samRequestContext).map(userStatus => StatusCodes.Created -> userStatus)
-                  }
-                }
-              }
+            }
           } ~ withSamRequestContext { samRequestContext =>
             (changeForbiddenToNotFound & requireUserInfo(samRequestContext)) { user =>
               get {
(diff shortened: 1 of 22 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -1,23 +1,31 @@
...
+
+import org.broadinstitute.dsde.workbench.model.{ErrorReport, WorkbenchEmail, WorkbenchExceptionWithErrorReport, WorkbenchGroupName, WorkbenchSubject}
 import org.broadinstitute.dsde.workbench.sam.dataAccess.DirectoryDAO
 import org.broadinstitute.dsde.workbench.sam.model.BasicWorkbenchGroup
 import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
+import org.broadinstitute.dsde.workbench.sam.config.TermsOfServiceConfig
+
+import scala.concurrent.ExecutionContext
+
+//import scala.concurrent.ExecutionContext
```
MatthewBemis: nit: this can be removed

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -1,23 +1,31 @@
...
-    if(isEnabled) {
-      getTosGroup(currentVersion).flatMap {
-        case Some(_) =>
-          IO.none
+class TosService (val directoryDao: DirectoryDAO, val appsDomain: String, val tosConfig: TermsOfServiceConfig)(implicit val executionContext: ExecutionContext) extends LazyLogging {
+
+  def createNewGroupIfNeeded(): IO[Option[BasicWorkbenchGroup]] = {
+    if(tosConfig.enabled) {
+      getTosGroup().flatMap {
         case None =>
```
jordansafer, line 26: Will we only get NONE if the group doesn't exist yet? What happens if we get another type of exception, like a networking error, when trying to look up the group? Is this returned as an exception?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -116,29 +112,40 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
-        if(userDetailsOnly)
+        if (userDetailsOnly)
           Future.successful(Option(UserStatus(UserStatusDetails(user.id, user.email), Map.empty)))
         else
           for {
             googleStatus <- cloudExtensions.getUserStatus(user)
             allUsersGroup <- cloudExtensions.getOrCreateAllUsersGroup(directoryDAO, samRequestContext)
             allUsersStatus <- directoryDAO.isGroupMember(allUsersGroup.id, user.id, samRequestContext).unsafeToFuture() recover { case _: NameNotFoundException => false }
+            tosAcceptedStatus <- {
+              tosService.getTosStatus(user.id).unsafeToFuture()
```
jordansafer: I saw that unsafeToFuture forces evaluation of the IO at this time, is there a reason to do the evaluation now vs later? In the UserStatusDiagnostics function I see the evaluations were pulled out of the for comprehension for parallelism. Would that make sense to do here with allUsersStatus, tosAcceptedStatus, and ldapStatus?

jordansafer (commented):
Looks really good!!!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -116,29 +112,40 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
       case Some(user) =>
-        if(userDetailsOnly)
+        if (userDetailsOnly)
           Future.successful(Option(UserStatus(UserStatusDetails(user.id, user.email), Map.empty)))
         else
           for {
             googleStatus <- cloudExtensions.getUserStatus(user)
             allUsersGroup <- cloudExtensions.getOrCreateAllUsersGroup(directoryDAO, samRequestContext)
             allUsersStatus <- directoryDAO.isGroupMember(allUsersGroup.id, user.id, samRequestContext).unsafeToFuture() recover { case _: NameNotFoundException => false }
+            tosAcceptedStatus <- {
```
MatthewBemis: nit: this can be one line

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -27,8 +35,26 @@ class TosService (val directoryDao: DirectoryDAO, val appsDomain: String) extend
...
+      createNewGroupIfNeeded().flatMap {
+        case Some(group) => directoryDao.addGroupMember(group.id, user, SamRequestContext(None))
+        case None => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Terms of Service group ${getGroupName(tosConfig.version)} failed to create.")))
+      }
+    } else IO.pure(false)
+  }
+
+  def getTosStatus(user: WorkbenchSubject): IO[Option[Boolean]] = {
+    if (tosConfig.enabled) {
+      createNewGroupIfNeeded().flatMap {
```
sagehen03: Does this mean we'll query the database to see if the Tos group exists each time we check a user's tos agreement status?  Or am I missing some caching?

sagehen03:
Should we update the swagger to mention the new `tosAccepted` field for UserStatus and UserStatusInfo definitions?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
```diff
@@ -39,15 +40,24 @@ trait UserRoutes extends UserInfoDirectives with SamRequestContextDirectives {
       (pathPrefix("v1") | pathEndOrSingleSlash) {
         pathEndOrSingleSlash {
           post {
-            withTermsOfServiceAcceptance {
+            entity(as[TermsOfServiceAcceptance]) { tos =>
```
MatthewBemis: I'm curious- why break this line out of the `withTermsOfServiceAcceptance` directive if it's ultimately going to be passed into the directive?

MatthewBemis (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
ansingh7115: I think it would depend on what's important to us. If made all of these parallel, it might save some time. However, this way I think we decrease the number of calls we're making to the postgres and opendj dbs (if a previous call fails, there's no need to check the rest). I'm going to punt making a decision on parallel vs sequential in this PR though, if that's alright!

jordansafer: That makes sense! I think it's always something we could look into if we needed to improve latency for some use-case. And definitely simplifies logic to be sequential-

jordansafer (approved):

ansingh7115:
@sagehen03 I'm not sure about adding swagger changes, as these changes only apply if the instance of Sam has ToS enabled. If not, then the swagger wouldn't align with the behavior of those endpoints. I think there's a bit of discussion we should do to figure out what we want the swagger to look like given that ToS can be disabled. Perhaps we could generate it based on the tos config https://broadworkbench.atlassian.net/browse/AS-1050

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
ansingh7115: I think this only returns None if the group does not exist yet

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
```diff
@@ -202,6 +198,25 @@ trait UserRoutes extends UserInfoDirectives with SamRequestContextDirectives {
     pathPrefix("v1") {
       withSamRequestContext { samRequestContext =>
         requireUserInfo(samRequestContext) { userInfo =>
+          pathPrefix("tos") {
+            pathPrefix("accept") {
```
MatthewBemis: nit: this can be condensed into `pathPrefix("tos" / "accept")`

jordansafer:
jenkins retest

MatthewBemis (approved):
LGTM. One minor nit, but it's totally optional

jordansafer (approved):

sagehen03 (approved):
Can confirm that this branch provides the necessary functionality to remove the existing tos service calls from terra ui.

sagehen03:
> @sagehen03 I'm not sure about adding swagger changes, as these changes only apply if the instance of Sam has ToS enabled. If not, then the swagger wouldn't align with the behavior of those endpoints. I think there's a bit of discussion we should do to figure out what we want the swagger to look like given that ToS can be disabled. Perhaps we could generate it based on the tos config https://broadworkbench.atlassian.net/browse/AS-1050

Great point, makes sense to wait.

PR #546, opened 2021-10-28
Mob qa 1559

gpolumbo-broad:
https://broadworkbench.atlassian.net/browse/QA-1559

It looks like there are two automation tests that both use the `Temps` UserPool.  

https://github.com/broadinstitute/firecloud-orchestration/blob/3cb3cb15090e37e72d546a4cdb5d592c4d065a2a/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala#L34

and

https://github.com/broadinstitute/workbench-libs/blob/2906dd05df65e7003fbf1ef7d0aa4aad94c24fd5/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/UserPool.scala#L61

Notice that the Orch test always explicitly uses Luna.  The Sam test always grabs the `head` of the Temps list, which is also always Luna.  

This means that if these two tests happen to get run at the same time, they both do checks to see if the User already exists or not and will try to remove the User at the start of the test if they already exist and then each test will try to create (and perhaps delete) that User as part of test.  If these tests are running at the same time, they could each be trying to create/remove Luna at the same time and thus cannot actually guarantee that the states they are trying to create.

This fixes the Sam test at least, so that if it fails, it will retry a few times.  I think this is OK to do in this case because if something changed in the logic of create or delete to break the idempotency of the api, the test should fail and keep failing.  If it passes just once, then we know it is good.  

There should be a corresponding change for QA-1016 as well.

This PR should be revisited after #549 merges.

---

**PR checklist**

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
@@ -104,15 +104,11 @@ class SamApiSpec extends AnyFreeSpec with BillingFixtures with Matchers with Sca
       tempUserInfo.userEmail shouldBe tempUser.email
 
       // Remove user
-
-      userStatus.userInfo.userEmail shouldBe tempUser.email
-
-      removeUser(userStatus.userInfo.userSubjectId)
+      removeUser(tempUserInfo.userSubjectId)
       Sam.user.status()(tempAuthToken) shouldBe None
 
       // OK to re-remove
-
-      removeUser(userStatus.userInfo.userSubjectId)
+      removeUser(tempUserInfo.userSubjectId)
       Sam.user.status()(tempAuthToken) shouldBe None
     }
   }
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
@@ -1,4 +1,4 @@
-package org.broadinstitute.dsde.workbench.test.api
+package org.broadinstitute.dsde.test.api
 
 
 import java.util.UUID
@@ -55,7 +55,7 @@ class SamApiSpec extends AnyFreeSpec with BillingFixtures with Matchers with Sca
     Thurloe.keyValuePairs.deleteAll(subjectId)
   }
 
-  "Sam test utilities" - {
+  "Sam User apis" - {
     "should be idempotent for user registration and removal" in {
 
       // use a temp user because they should not be registered.  Remove them after!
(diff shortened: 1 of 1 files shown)


automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
```diff
@@ -12,24 +10,44 @@ import org.broadinstitute.dsde.workbench.fixture.BillingFixtures
...
   val gcsConfig = SamConfig.GCS
 
+  val retries = 4
+
+  // We are intentionally not using trait `FailedTestRetryable` from WorkbenchLibs here because that trait only retries
+  // failures (not cancels) and if it succeeds on a retry, it will only put the test into a Canceled state instead of a
+  // passing state, which for our purposes here, is what we want.
+  override def withFixture(test: NoArgTest): Outcome = {
+    if (isRetryable(test)) withFixture(test, retries) else super.withFixture(test)
+  }
```
ehigham, line 41: These mixed function calls are a bit confusing - hard to see who is calling whom. Perhaps this might be a simplification?

```suggestion
  override def withFixture(test: NoArgTest): Outcome = {
    val maxAttempts = if (isRetryable(test)) 4 else 1

    @tailrec
    def go(attempt: Int): Outcome =
      super.withFixture(test) match {
        case failure@Exceptional(_) => if (attempt < maxAttempts) go(attempt + 1) else failure
        case other => other
      }

    go(attempt = 0)
  }
```

automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
```diff
@@ -12,24 +10,44 @@ import org.broadinstitute.dsde.workbench.fixture.BillingFixtures
...
   implicit override val patienceConfig: PatienceConfig = PatienceConfig(timeout = scaled(Span(5, Seconds)))
   implicit lazy val system = ActorSystem()
 
   val gcsConfig = SamConfig.GCS
 
+  val retries = 4
+
+  // We are intentionally not using trait `FailedTestRetryable` from WorkbenchLibs here because that trait only retries
+  // failures (not cancels) and if it succeeds on a retry, it will only put the test into a Canceled state instead of a
+  // passing state, which for our purposes here, is what we want.
```
ehigham, line 38: Or is it not what we want?

automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
```diff
@@ -12,24 +10,44 @@ import org.broadinstitute.dsde.workbench.fixture.BillingFixtures
...
+class SamApiSpec extends AnyFreeSpec with BillingFixtures with Matchers with ScalaFutures with CleanUp with Eventually with TestKitBase with Retries {
   implicit override val patienceConfig: PatienceConfig = PatienceConfig(timeout = scaled(Span(5, Seconds)))
   implicit lazy val system = ActorSystem()
 
   val gcsConfig = SamConfig.GCS
 
+  val retries = 4
+
+  // We are intentionally not using trait `FailedTestRetryable` from WorkbenchLibs here because that trait only retries
+  // failures (not cancels) and if it succeeds on a retry, it will only put the test into a Canceled state instead of a
```
ehigham, line 37: Why do we want to retry cancelations?

gpolumbo-broad:
I'm closing this PR.  We don't really use retries in the tests anywhere and there's legitimate reason, which is that by just retrying things, we may be obfuscating real problems in the code.  It still might be useful when used appropriately, but I think that is even up for debate.  

This PR came out during a major QA test stability blockage during late October and early November 2021.  The tests have since been fixed up, so this PR isn't technically necessary anymore.

PR #547, opened 2021-10-28
QA-1500 all write txns are serializable, stop using default isolation level

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/QA-1500

Removes the `writeTransaction` function and for remaining code using `inLocalTransaction` use read committed isolation instead of default because default in scalikejdbc does not actually do anything and just reuses whatever isolation level was used last.

Switching to serializable isolation level is ok because in effect that is what has been happening anyway. The write database connection pool is small and was shared between `writeTransaction` and `serializableWriteTransaction` meaning all connections would quickly flip serializable.

This is important because write transactions were being executed in serializable isolation _without_ retries. `serializableWriteTransaction` retries appropriately.

---

**PR checklist**


jacmrob (approved):
LGTM

gpolumbo-broad (approved):

PR #548, opened 2021-10-28
[PF-1037] Update WSM role inheritance and controlled resource permissions

ddietterich:
Ticket: PF-1037
Update the WSM role inheritance and controlled resource permissions.
This is documented in [WSM Controlled Resource Design](https://docs.google.com/document/d/1F28KHeVpTh0Zuqdux30ONabMgjj6CNmyv7B8iqcXnsE/edit?resourcekey=0-5ndsiOmBJdG7bB1gZXOy1g#heading=h.qw1bsdn5v5n7)

A summary spreadsheet of the roles and actions is here [dd Role Definition changes](https://docs.google.com/spreadsheets/d/1rHSQ8ecJehVbs591Fw-4scWMsThyyBGrbsuQ-89eSiY/edit?resourcekey=0-VmhXLfvYDRi4xPRxfYATow#gid=1430007751). The sheet "dd Inherited Role changes" shows the inheritance changes.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -124,7 +124,6 @@ resourceTypes = {
           controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
           controlled-user-private-workspace-resource = ["deleter"]
           controlled-application-shared-workspace-resource = ["writer", "reader"]
-          controlled-application-private-workspace-resource = []
         }
       }
       application = {
@@ -134,9 +133,7 @@ resourceTypes = {
         roleActions = ["read_policy::owner", "write", "read", "create_controlled_user_shared", "create_controlled_user_private", "create_referenced_resource", "update_referenced_resource", "delete_referenced_resource", "list_children", "add_child", "remove_child", "read_auth_domain"]
         descendantRoles = {
           controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
-          controlled-user-private-workspace-resource = []
           controlled-application-shared-workspace-resource = ["writer", "reader"]
-          controlled-application-private-workspace-resource = []
         }
       }
       reader = {
@@ -270,6 +267,9 @@ resourceTypes = {
       }
       reader = {
         roleActions = ["read"]
+      }
+      deleter = {
+        roleActions = ["delete"]
       }
     }
     reuseIds = false
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -122,9 +122,7 @@ resourceTypes = {
         descendantRoles = {
           google-project = ["owner"]
           controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
-          controlled-user-private-workspace-resource = ["assigner", "editor"]
-          controlled-application-shared-workspace-resource = ["editor", "writer", "reader"]
-          controlled-application-private-workspace-resource = ["editor"]
+          controlled-application-shared-workspace-resource = ["writer", "reader"]
         }
       }
       application = {
@@ -134,9 +132,7 @@ resourceTypes = {
         roleActions = ["read_policy::owner", "write", "read", "create_controlled_user_shared", "create_controlled_user_private", "create_referenced_resource", "update_referenced_resource", "delete_referenced_resource", "list_children", "add_child", "remove_child", "read_auth_domain"]
         descendantRoles = {
           controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
-          controlled-user-private-workspace-resource = ["editor"]
-          controlled-application-shared-workspace-resource = ["editor", "writer", "reader"]
-          controlled-application-private-workspace-resource = ["editor"]
+          controlled-application-shared-workspace-resource = ["writer", "reader"]
         }
       }
       reader = {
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -253,21 +250,18 @@ resourceTypes = {
...
-      assigner = {
-        roleActions = ["manage_private_user", "read_policies", "share_policy::editor", "share_policy::writer", "share_policy::reader"]
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "set_parent", "read", "write"]
       }
+      # editor is always the resource creator
       editor = {
-        roleActions = ["delete", "edit"]
+        roleActions = ["delete", "edit", "read", "write"]
       }
+      # we don't use writer or reader
```
s-rubenstein: I would just clarify "we" here. I assume that is "Workspace Manger doesn't use writer or reader"?

src/main/resources/reference.conf
```diff
@@ -371,14 +369,21 @@ resourceTypes = {
...
-        roleActions = ["delete", "edit"]
+        roleActions = ["delete", "edit", "read", "write", "manage_private_user", "read_policies", "share_policy::writer", "share_policy::reader"]
+      }
+      # workspace owner gets new role 'deleter' so they can remove private resources
+      deleter = {
+        roleActions = ["delete"]
+      }
+      # assign becomes an action, not a role. Delete it?
+      assigner = {
+        roleActions = []
```
zloery: +1, I don't think we need a separate assigner role. the permission is part of the editor role which goes to the app SA

src/main/resources/reference.conf
```diff
@@ -371,14 +369,21 @@ resourceTypes = {
...
-        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "manage_private_user", "set_parent"]
-      }
-      assigner = {
-        roleActions = ["manage_private_user", "read_policies", "share_policy::editor", "share_policy::writer", "share_policy::reader"]
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "manage_private_user", "set_parent", "read", "write"]
       }
+      # editor is application SA. Can assign a reader or a writer
       editor = {
-        roleActions = ["delete", "edit"]
+        roleActions = ["delete", "edit", "read", "write", "manage_private_user", "read_policies", "share_policy::writer", "share_policy::reader"]
```
zloery: I don't think we should give applications direct access to modify writer and reader policies (and you don't in the spreadsheet). my understanding was that applications would have to do resource assignment through WSM (which is why we have a `manage_private_user` action). Otherwise, we're now expecting applications to understand and enforce private resource semantics (exactly one reader/writer), which seems like something they shouldn't be expected to do.

src/main/resources/reference.conf
ddietterich: Removed

src/main/resources/reference.conf
ddietterich: Good catch. I removed the share_policy permissions. I left read_policies. I am thinking the application should be able to call Sam directly to check that a user has access to the resource.

src/main/resources/reference.conf
ddietterich: Yes, I fixed the comment. Not sure how much commenting is common in the file. It was mostly for me as I was making the changes.

zloery (approved):
WSM business logic LGTM, I'll defer to CI team on this as a Sam change

src/main/resources/reference.conf
```diff
@@ -253,27 +250,27 @@ resourceTypes = {
       "own" = {
         description = "perform owner actions on the resource"
       }
-      "manage_private_user" = {
-        description = "assign, remove, and reassign the private user"
-      }
     }
     ownerRoleName = "owner"
     roles = {
+      # owner is always WSM SA
```
gpolumbo-broad: Not sure some of these comments are useful here.  The fact that the owner should always be a WSM SA is not enforced in Sam (nor should it be).  I usually favor more commenting over less, but in this situation, I don't want anyone getting confused if they read these comments and think Sam or something here in reference.conf is enforcing any kinds of roles as to who or what kind of users can be granted these roles.

src/main/resources/reference.conf
```diff
@@ -253,27 +250,27 @@ resourceTypes = {
...
-      assigner = {
-        roleActions = ["manage_private_user", "read_policies", "share_policy::editor", "share_policy::writer", "share_policy::reader"]
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "set_parent", "read", "write"]
       }
+      # editor is always the resource creator
       editor = {
-        roleActions = ["delete", "edit"]
+        roleActions = ["delete", "edit", "read", "write"]
       }
+      # WSM does not use these reader and writer roles
```
gpolumbo-broad: Is WSM the only service using `controlled-user-private-workspace-resource`?  If so and you're not using these roles, can you just remove them?  Sam does not care if they're here or not.

gpolumbo-broad (changes requested):
I don't have a lot of context for WSM and what the overall goal of what you're trying to achieve is, but the edits here seem logical enough.  I would just ask that you clean up some of the comments to avoid confusion.

src/main/resources/reference.conf
ddietterich: They were more for my benefit to keep my place. I'll remove the comments. We have a document that has a better explanation.

src/main/resources/reference.conf
ddietterich: Turns out I was wrong about that, so I am going to leave them in.

ddietterich:
jenkins retest

ddietterich:
jenkins retest

gpolumbo-broad (approved):

PR #550, opened 2021-10-29
AS-887

sagehen03:
Ticket: https://broadworkbench.atlassian.net/browse/AS-887

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -21,15 +21,16 @@ import scala.util.matching.Regex
 /**
   * Created by dvoet on 7/14/17.
   */
-class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExtensions, val registrationDAO: RegistrationDAO, blockedEmailDomains: Seq[String])(implicit val executionContext: ExecutionContext, contextShift: ContextShift[IO]) extends LazyLogging {
+class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExtensions, val registrationDAO: RegistrationDAO, blockedEmailDomains: Seq[String], tosService: TosService)(implicit val executionContext: ExecutionContext, contextShift: ContextShift[IO]) extends LazyLogging {
 
-  def createUser(user: WorkbenchUser, samRequestContext: SamRequestContext): Future[UserStatus] = {
+  def createUser(user: WorkbenchUser, samRequestContext: SamRequestContext, tosEnforcementEnabled: Boolean = false, tosVersion: Int = 0): Future[UserStatus] = {
     for {
       _ <- UserService.validateEmailAddress(user.email, blockedEmailDomains).unsafeToFuture()
       allUsersGroup <- cloudExtensions.getOrCreateAllUsersGroup(directoryDAO, samRequestContext)
       createdUser <- registerUser(user, samRequestContext).unsafeToFuture()
       _ <- enableUserInternal(createdUser, samRequestContext)
       _ <- directoryDAO.addGroupMember(allUsersGroup.id, createdUser.id, samRequestContext).unsafeToFuture()
+      _ <- if (tosEnforcementEnabled) directoryDAO.addGroupMember(tosService.getTosGroup(tosVersion).unsafeRunSync().get.id, createdUser.id, samRequestContext).unsafeToFuture() else Future.successful(IO.pure(true))
       userStatus <- getUserStatus(createdUser.id, samRequestContext = samRequestContext)
       res <- userStatus.toRight(new WorkbenchException("getUserStatus returned None after user was created")).fold(Future.failed, Future.successful)
     } yield res
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -30,7 +30,11 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
       createdUser <- registerUser(user, samRequestContext).unsafeToFuture()
       _ <- enableUserInternal(createdUser, samRequestContext)
       _ <- directoryDAO.addGroupMember(allUsersGroup.id, createdUser.id, samRequestContext).unsafeToFuture()
-      _ <- if (tosEnforcementEnabled) directoryDAO.addGroupMember(tosService.getTosGroup(tosVersion).unsafeRunSync().get.id, createdUser.id, samRequestContext).unsafeToFuture() else Future.successful(IO.pure(true))
+      tosGroupOpt <- if (tosEnforcementEnabled) tosService.getTosGroup(tosVersion).unsafeToFuture() else Future.successful(None)
+      _ <- (tosEnforcementEnabled, tosGroupOpt) match {
+        case (true, Some(tosGroup)) => directoryDAO.addGroupMember(tosGroup.id, createdUser.id, samRequestContext).unsafeToFuture()
+        case _ => Future.successful(true)
+      }
       userStatus <- getUserStatus(createdUser.id, samRequestContext = samRequestContext)
       res <- userStatus.toRight(new WorkbenchException("getUserStatus returned None after user was created")).fold(Future.failed, Future.successful)
     } yield res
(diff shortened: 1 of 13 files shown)


sagehen03:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
```diff
@@ -120,7 +120,7 @@ object TestSupport extends TestSupport {
     val mockResourceService = new ResourceService(resourceTypes, policyEvaluatorService, policyDAO, directoryDAO, googleExt, "example.com")
     val mockManagedGroupService = new ManagedGroupService(mockResourceService, policyEvaluatorService, resourceTypes, policyDAO, directoryDAO, googleExt, "example.com")
 
-    SamDependencies(mockResourceService, policyEvaluatorService, new UserService(directoryDAO, googleExt, registrationDAO, Seq.empty), new StatusService(directoryDAO, registrationDAO, googleExt, dbRef), mockManagedGroupService, directoryDAO, policyDAO, googleExt)
+    SamDependencies(mockResourceService, policyEvaluatorService, new UserService(directoryDAO, googleExt, registrationDAO, Seq.empty, new TosService(directoryDAO, appConfig.googleConfig.get.googleServicesConfig.appsDomain)), new StatusService(directoryDAO, registrationDAO, googleExt, dbRef), mockManagedGroupService, directoryDAO, policyDAO, googleExt)
```
davidangb: can you use `googleServicesConfig.appsDomain` instead of `appConfig.googleConfig.get.googleServicesConfig.appsDomain` here? Note that these are subtly different - the `googleServicesConfig` can be overridden by the caller calling `genSamDependencies(…)`. Though that's exactly why I think you should use it - if the caller is overriding, this should respect the override.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesSpec.scala
```diff
@@ -53,7 +54,7 @@ class ResourceRoutesSpec extends AnyFlatSpec with Matchers with ScalatestRouteTe
     val emailDomain = "example.com"
     val policyEvaluatorService = PolicyEvaluatorService(emailDomain, resourceTypes, accessPolicyDAO, directoryDAO)
     val mockResourceService = new ResourceService(resourceTypes, policyEvaluatorService, accessPolicyDAO, directoryDAO, NoExtensions, emailDomain)
-    val mockUserService = new UserService(directoryDAO, NoExtensions, registrationDAO, Seq.empty)
+    val mockUserService = new UserService(directoryDAO, NoExtensions, registrationDAO, Seq.empty, new TosService(directoryDAO, appConfig.googleConfig.get.googleServicesConfig.appsDomain))
```
davidangb: can you change the import to
```
import org.broadinstitute.dsde.workbench.sam.TestSupport.googleServicesConfig
```
and then use `googleServicesConfig.appsDomain` instead of `appConfig.googleConfig.get.googleServicesConfig`?

Compared to my comment in TestSupport, this should be the exact same value - it's just a bit less code

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
```diff
@@ -53,7 +53,7 @@ class ResourceRoutesV2Spec extends AnyFlatSpec with Matchers with TestSupport wi
     resourceTypes.map { case (resourceTypeName, resourceType) =>
       when(mockResourceService.getResourceType(resourceTypeName)).thenReturn(IO(Option(resourceType)))
     }
-    val mockUserService = new UserService(directoryDAO, NoExtensions, registrationDAO, Seq.empty)
+    val mockUserService = new UserService(directoryDAO, NoExtensions, registrationDAO, Seq.empty, new TosService(directoryDAO, appConfig.googleConfig.get.googleServicesConfig.appsDomain))
```
davidangb: same, use `googleServicesConfig` if possible. I see this in other files too - I'll stop commenting about it but hope you can change everywhere

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
```diff
@@ -109,6 +111,23 @@ class UserServiceSpec extends AnyFlatSpec with Matchers with TestSupport with Mo
     }.errorReport.statusCode shouldBe Some(StatusCodes.BadRequest)
   }
 
+  it should "create user and add user to ToS group" in {
+    val tosVersion = 1
+    tos.createNewGroupIfNeeded(tosVersion, true).unsafeToFuture()
```
davidangb: because this is a Future, it's fire-and-forget and is not guaranteed to have completed by the time the next lines run. Change `.unsafeToFuture()` to `.unsafeRunSync()` to force it to complete

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
```diff
@@ -109,6 +111,23 @@ class UserServiceSpec extends AnyFlatSpec with Matchers with TestSupport with Mo
...
+    val user = service.createUser(defaultUser, samRequestContext, true, tosVersion).futureValue
+    user.enabled("allUsersGroup") shouldBe true
```
davidangb: minor nit: if something fails during user creation, and the `user.enabled` map does not contain a key "allUsersGroup", this line will fail with a less-helpful error message, something like `java.util.NoSuchElementException: key not found: allUsersGroup`. Consider rewriting this as:
```
user.enabled should contain (org.scalatest.Entry("allUsersGroup", true))
```
this isn't a big deal, since the less-helpful error message is still reasonably clear, but less so imho than what would be generated by the other syntaxes. I'm on a pet-peeve effort right now to improve our tests' error-messaging!

davidangb (commented):
nit: add a PR titie!

A few comments inline, nothing big I hope

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -21,15 +21,16 @@ import scala.util.matching.Regex
...
 
-  def createUser(user: WorkbenchUser, samRequestContext: SamRequestContext): Future[UserStatus] = {
+  def createUser(user: WorkbenchUser, samRequestContext: SamRequestContext, tosEnforcementEnabled: Boolean = false, tosVersion: Int = 0): Future[UserStatus] = {
     for {
       _ <- UserService.validateEmailAddress(user.email, blockedEmailDomains).unsafeToFuture()
       allUsersGroup <- cloudExtensions.getOrCreateAllUsersGroup(directoryDAO, samRequestContext)
       createdUser <- registerUser(user, samRequestContext).unsafeToFuture()
       _ <- enableUserInternal(createdUser, samRequestContext)
       _ <- directoryDAO.addGroupMember(allUsersGroup.id, createdUser.id, samRequestContext).unsafeToFuture()
+      _ <- if (tosEnforcementEnabled) directoryDAO.addGroupMember(tosService.getTosGroup(tosVersion).unsafeRunSync().get.id, createdUser.id, samRequestContext).unsafeToFuture() else Future.successful(IO.pure(true))
```
MatthewBemis: a couple of things:

1) the value returned by the `else` clause only needs to be a `Future.successful(true)`. the value is ultimately discarded but if you assign this line to a variable you'll see that it returns `Any` because the two possible return types are `Future[Boolean]` and `Future[IO[Boolean]]`

2) I think i would prefer to see this split out into it's individual steps:

```
      tosGroup <- if (tosEnforcementEnabled) tosService.getTosGroup(tosVersion).unsafeToFuture() else Future.successful(None)
      _ <- if (tosEnforcementEnabled && tosGroup.nonEmpty) directoryDAO.addGroupMember(tosGroup.get.id, createdUser.id, samRequestContext).unsafeToFuture() else Future.successful(true)
```

there is another benefit: using `.get` on `Option`s is something that i personally find to be a bit dangerous since it can throw a `NoSuchElementException` and then there's no handling in place to deal with that. doing the `nonEmpty` check at least grants some reassurance that there will always be a ToS group returned. testing for emptiness is logically safe but pattern matching or `getOrElse` are usually preferred, so there is another way to write this that takes it a step further and protects against these exception situations:

```
      tosGroupOpt <- if (tosEnforcementEnabled) tosService.getTosGroup(tosVersion).unsafeToFuture() else Future.successful(None)
      _ <- (tosEnforcementEnabled, tosGroupOpt) match {
        case (true, Some(tosGroup)) => directoryDAO.addGroupMember(tosGroup.id, createdUser.id, samRequestContext).unsafeToFuture()
        case _ => Future.successful(true)
      }
```

disclaimer: both of these compile but i haven't run unit tests or anything. but i _think_ they should both work

i personally prefer the 2nd option but i understand that it's not quite as readable, so i'll leave it up to you!

davidangb (approved):
lgtm, I see this needs a rebase unfortunately!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
sagehen03: Thanks, Matt! I've updated the code to use the 2nd option.

MatthewBemis (approved):
lgtm!

PR #551, opened 2021-10-29
[QA-1240] Only sync constrained policies that have been synced before

marctalbott:
Ticket: [QA-1240](https://broadworkbench.atlassian.net/browse/QA-1240)
When we update an auth domain group, we add all of the policies constrained by that auth domain to the group synchronizer queue. The group synchronizer doesn't check whether a group has been synced in the past because it is responsible for first time syncs and just blindly syncs any group it gets off the queue. This can lead to a timing issue that interferes with workspace deletion in tests with constrained resources.

1. Constrained workspace is created with all 8 of its policies and Rawls tells Sam to sync 5 of those policies immediately (`project-owner`, `owner`, `writer`, `reader`, `can-compute`). The logs indicate that the Google call to create the Google groups for these policies (aka syncing them) occurs <1 second after the workspace is created.
2. Auth domain group is modified later. This adds the auth domain group to the group sync queue along with all of the policies it constrains. This includes all of the workspace policies, not just the 5 that have been synced before.
3. The synchronizer eventually gets around to syncing the messages generated above. For the 3 previously unsynced policy groups (`can-catalog`, `share-reader`, `share-writer`), the synchronizer creates new Google groups.
4. While this is going on in the background, the test completes and begins cleaning up the workspace. This logic in Sam tries to delete all Google groups for all policies on the resource, whether those policies have been synced or not.
5. The requests to delete the policy groups are issued right around the same time as the requests to create the previously unsynced policy groups. These requests conflict with each other, and Google rightfully returns a 409. This failure prevents Sam from deleting the policies in Postgres and the workspace resource is not. deleted.
6. Since the workspace isn't deleted correctly in Sam, the billing-project can't be deleted a thse billing-project owner policy is a member of the workspace project-owner policy.

This PR makes it so we only sync policies that have been synced previously by checking for a last synchronized date before adding the policy to the group sync queue. If a constrained policy hasn't been synced in the past, then there's no point in syncing it when the auth domain group gets synced since nobody is using the unsynced policy to grant access to cloud resources.

---

**PR checklist**

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAccessPolicyDAO.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAccessPolicyDAO.scala
@@ -94,7 +94,7 @@ class MockAccessPolicyDAO(private val resourceTypes: mutable.Map[ResourceTypeNam
     resourceToRemoveOpt.map( resourceToRemove => resources += resource -> resourceToRemove.copy(authDomain = Set.empty))
   }
 
-  override def listResourcesConstrainedByGroup(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Set[Resource]] = IO.pure(Set.empty)
+  override def listSyncedAccessPolicyIdsOnResourcesConstrainedByGroup(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Set[FullyQualifiedPolicyId]] = ???
 
   override def createPolicy(policy: AccessPolicy, samRequestContext: SamRequestContext): IO[AccessPolicy] = IO {
     policies += policy.id -> policy
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAccessPolicyDAO.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAccessPolicyDAO.scala
@@ -94,7 +94,22 @@ class MockAccessPolicyDAO(private val resourceTypes: mutable.Map[ResourceTypeNam
     resourceToRemoveOpt.map( resourceToRemove => resources += resource -> resourceToRemove.copy(authDomain = Set.empty))
   }
 
-  override def listSyncedAccessPolicyIdsOnResourcesConstrainedByGroup(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Set[FullyQualifiedPolicyId]] = ???
+  override def listSyncedAccessPolicyIdsOnResourcesConstrainedByGroup(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Set[FullyQualifiedPolicyId]] = IO {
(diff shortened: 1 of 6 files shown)


ehigham (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAccessPolicyDAO.scala
```diff
@@ -94,7 +94,7 @@ class MockAccessPolicyDAO(private val resourceTypes: mutable.Map[ResourceTypeNam
     resourceToRemoveOpt.map( resourceToRemove => resources += resource -> resourceToRemove.copy(authDomain = Set.empty))
   }
 
-  override def listResourcesConstrainedByGroup(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Set[Resource]] = IO.pure(Set.empty)
+  override def listSyncedAccessPolicyIdsOnResourcesConstrainedByGroup(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Set[FullyQualifiedPolicyId]] = ???
```
gpolumbo-broad: Looks like this is never used directly and is always mocked, which is fine, but leaving the `???` here forced me to go do that check.  I'm fine leaving the `???`, but a comment might be a good idea to briefly explain why this was done.  Or maybe just have it return an empty `Set`.

gpolumbo-broad (changes requested):

gpolumbo-broad:
FWIW - talked faces with @marctalbott on this to clarify some things.  A few key takeaways:

- The Sam `sync policy` endpoint does not use the synchronizer queue.  It syncs the group immediately.  This is how policies can/should be sync'd for the first time.  (This is how a group's synchronized date will get set for the first time)
- We want the synchronizer (the thing popping messages off the queue) to remain sort of naive, it syncs whatever its told to sync
- The logic for whether something should be sync'd or not should happen when trying to determine what should or should not be put onto the synchronizer queue
- We believe that the `can-catalog`, `share-reader`, and `share-writer` polices should not be getting sync'd and the fact that they _are_ being sync'd today is a bug.  This PR fixes that bug
- _IF_ there is logic in production code or tests that counts on those policies being sync'd, this PR _will_ break that code and whatever that code is, it should be fixed or it can directly call the Sam `sync` endpoint to explicitly request that the policy be sync'd, which is perfectly valid.  (Explicitly requesting a policy to be sync'd is fine.  Sam auto-magically sync'ing policies that were _never_ explicitly requested to be sync'd is the bug)
- In theory, we _could_ write an integration test in Rawls or something that creates an auth domain constrained resource, updates the auth domain, and then checks that Google still only knows about exact policies it asked Sam to sync.  There should be no "bonus" policies for that resource that showed up in Google as sync'd.  We are not going to write that test now though.

gpolumbo-broad (approved):

araina2018:
Thanks @marctalbott @gpolumbo-broad  for detailed takeaways. Noting the last bullet which is beyond the current unblocking monolith scope but a key takeaway - Sam/ Rawls/Orch test suites review to ensure we test the integration of these components adequately without conflating / confusing the purpose of each test in these test suites.

PR #553, opened 2021-11-12
[AS-908] Add Terms of Service text to SAM

jordansafer:
Ticket: https://broadworkbench.atlassian.net/browse/AS-908
Part of the effort to move Terms of service (ToS) text out of the Terra UI and into SAM:
- Added ToS text as a resource file
- Added ToS API to get the text
- Added Integration test to verify API works as expected

Testing: Unit testing, Swagger UI testing, Integration testing

---

**PR checklist**


jordansafer:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRoutes.scala
```diff
@@ -0,0 +1,24 @@
...
+trait TermsOfServiceRoutes {
+  val tosService: TosService
+  implicit val executionContext: ExecutionContext
+
+  def termsOfServiceRoutes: server.Route =
+    pathPrefix("tos") {
+      path("text") {
+        pathEndOrSingleSlash {
+          get {
+            complete(HttpResponse(entity = tosService.getText))
```
davidangb: I believe the `HttpResponse` is extraneous here; you should be able to do just `complete(tosService.getText)`

automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
```diff
@@ -97,6 +99,19 @@ class SamApiSpec extends AnyFreeSpec with BillingFixtures with Matchers with Sca
...
+
+      val response = Sam.getRequest(Sam.url + s"tos/text")(userAuthToken)
+      val textFuture = Unmarshal(response.entity).to[String]
+
+      response.status shouldEqual StatusCodes.OK
+      whenReady(textFuture) { text =>
+        text should include("Terms as of February 12, 2020.")
+      }
+    }
+
```
davidangb, line 109: the ToS text should be available to users who haven't yet registered, since it's part of the registration flow. Could you add a test that checks anonymous access to /tos/text too?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -31,4 +37,23 @@ class TosService (val directoryDao: DirectoryDAO, val appsDomain: String) extend
...
   }
 
+  /**
+    * Get the terms of service text and send it to the caller
+    * @return terms of service text
+    */
+  def getText: String = {
+    val tosFileStream = try {
+      logger.debug("Reading terms of service")
+      Source.fromURL(getClass.getResource(pathToTermsOfService))
```
davidangb: `Source.fromResource` encapsulates this for you; either way is fine

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -31,4 +37,23 @@ class TosService (val directoryDao: DirectoryDAO, val appsDomain: String) extend
...
+    } catch {
+      case e: FileNotFoundException => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, e))
+      case e: IOException => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, e))
```
davidangb: might wanna add logging statements for each case so that we as devs can distinguish between the two errors

davidangb (commented):
here's some preliminary comments, since I was looking at this PR anyway!

automation/src/test/scala/org/broadinstitute/dsde/test/api/SamApiSpec.scala
jordansafer: Yeah, I'll add an anonymous test-

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRoutes.scala
jordansafer: Fixing!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
jordansafer: Will do!

sagehen03 (commented):
Looks pretty good to me, but I think this implementation means that every request to tos/text will read the markdown file from disk?  I'm wondering if we can add caching in the route definition or potentially have the tos service hold the string in memory without causing other issues.  It's a pretty small issue and I'm actually a little hesitant to bring it up because it might be a premature optimization, but let's discuss.

jordansafer:
> Looks pretty good to me, but I think this implementation means that every request to tos/text will read the markdown file from disk? I'm wondering if we can add caching in the route definition or potentially have the tos service hold the string in memory without causing other issues. It's a pretty small issue and I'm actually a little hesitant to bring it up because it might be a premature optimization, but let's discuss.

I was thinking about this, my initial impression was that this would be good since the file size is a tiny thing to hold in memory, but after considering the relatively low call rate to the API (assuming less than 1 call per minute, potentially much less than that), and finding that the file read only adds some microseconds to of computation time, I decided to go with the simpler implementation. What do you think?

EDIT: Actually, from a latency standpoint, it could be interesting if the UI cached this, although given the low call rate, I'm still hesitant to add it. If our cached entry expires every 24h anyway, I'm not sure how much re-use that would get-

sagehen03:
> > Looks pretty good to me, but I think this implementation means that every request to tos/text will read the markdown file from disk? I'm wondering if we can add caching in the route definition or potentially have the tos service hold the string in memory without causing other issues. It's a pretty small issue and I'm actually a little hesitant to bring it up because it might be a premature optimization, but let's discuss.
> 
> I was thinking about this, my initial impression was that this would be good since the file size is a tiny thing to hold in memory, but after considering the relatively low call rate to the API (assuming less than 1 call per minute, potentially much less than that), and finding that the file read only adds some microseconds to of computation time, I decided to go with the simpler implementation. What do you think?
> 
> EDIT: Actually, from a latency standpoint, it could be interesting if the UI cached this, although given the low call rate, I'm still hesitant to add it. If our cached entry expires every 24h anyway, I'm not sure how much re-use that would get-

Works for me!

sagehen03 (approved):

jordansafer:
jenkins retest

jordansafer:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -31,4 +37,28 @@ class TosService (val directoryDao: DirectoryDAO, val appsDomain: String) extend
...
   }
 
+  /**
+    * Get the terms of service text and send it to the caller
+    * @return terms of service text
+    */
+  def getText: String = {
+    val tosFileStream = try {
+      logger.debug("Reading terms of service")
+      logger.info(System.getProperty("user.dir"))
```
davidangb: nit: can this log statement be removed?

davidangb (approved):
one nit inline, otherwise :+1:. I don't need to re-review if all you change is removing logging statement(s)!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
jordansafer: Oops! Yes, it shouldn't be there-

PR #555, opened 2021-11-18
[CA-1562] update policy role table on boot

s-rubenstein:
Ticket: https://broadworkbench.atlassian.net/browse/CA-1562
This PR adds a command to initResourceTypes that involves updating the `sam_effective_policy_role` table. This PR does not add anything to update the `sam_effective_policy_action` table, because it turns out that we don't actually seem to support descendant actions in reference.conf, so that table isn't actually effected by that bug _on boot_. It is possible that the bug also manifests when people call the `updatePolicy` endpoint in Sam, which is how that could actually be updated, but I would push to deprioritize that in favor of other work, because to my understanding we have basically deprecated that in favor of the configuration file.

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1350,7 +1350,59 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+               ${PolicyTable as policy},
+               ${ResourceTable as policyResource},
+               ${ResourceTypeTable as resourceTypeOne},
+               ${ResourceTypeTable as resourceTypeTwo}
+               where ${effectivePolicyRole.effectiveResourcePolicyId} = ${effectiveResourcePolicy.id}
+               and ${effectivePolicyRole.resourceRoleId} = ${resourceRole.id}
+               and ${effectiveResourcePolicy.sourcePolicyId} = ${policy.id}
+               and ${policy.resourceId} = ${policyResource.id}
+               and ${policyResource.resourceTypeId} = ${resourceTypeOne.id}
+               and ${resourceRole.resourceTypeId} = ${resourceTypeTwo.id}
```
s-rubenstein: We want to make these changes where _either_ the role definition (aka the child resource) has been changed, or where the source policy (aka the parent resource) has been changed.

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAccessPolicyDAO.scala
```diff
@@ -250,4 +250,6 @@ class MockAccessPolicyDAO(private val resourceTypes: mutable.Map[ResourceTypeNam
 
     aggregateByResource(forEachPolicy)
   }
+
+  def recreateEffectivePolicyRolesTableEntry(resourceTypeName: ResourceTypeName, samRequestContext: SamRequestContext): IO[Unit] = IO.pure(None)
```
s-rubenstein: I didn't bother implementing this here, because the tests for this endpoint are exclusively through service tests, which want to use the real DAO.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -48,12 +48,23 @@ class ResourceServiceSpec extends AnyFlatSpec with Matchers with ScalaFutures wi
   lazy val policyDAO: AccessPolicyDAO = new PostgresAccessPolicyDAO(TestSupport.dbRef, TestSupport.dbRef)
   val schemaDao = new JndiSchemaDAO(directoryConfig, schemaLockConfig)
 
+  private val ownerRoleName = ResourceRoleName("owner")
```
s-rubenstein, line 51: I pulled out any static definitions into this helper, because it seemed wrong to re-define the string all over the place. I could have done this to other variables as well, but that seemed like a bigger refactor. I didn't effect places where it is overridden locally, because I figure it is there locally for a reason.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1350,7 +1350,59 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+               and (
+                 ${resourceTypeOne.name} = ${resourceTypeName}
+                 or ${resourceTypeTwo.name} = ${resourceTypeName}
+               )
+            """.update().apply()
+
+      samsql"""insert into ${EffectivePolicyRoleTable.table}(${EffectivePolicyRoleTable.column.effectiveResourcePolicyId}, ${EffectivePolicyRoleTable.column.resourceRoleId})
```
s-rubenstein: This is mostly from here: https://github.com/broadinstitute/sam/blob/develop/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20201123_effective_policy.xml#L79

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAccessPolicyDAO.scala
andy7i:
```suggestion
  override def recreateEffectivePolicyRolesTableEntry(resourceTypeName: ResourceTypeName, samRequestContext: SamRequestContext): IO[Unit] = IO.pure(None)
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1350,7 +1350,59 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+    val resourceTypeOne = ResourceTypeTable.syntax("resourceTypeOne")
+    val resourceTypeTwo = ResourceTypeTable.syntax("resourceTypeTwo")
+    val policy = PolicyTable.syntax("policy")
+    val policyRole = PolicyRoleTable.syntax("policyRole")
+    val effectivePolicyRole = EffectivePolicyRoleTable.syntax("effectivePolicyRole")
+    val effectiveResourcePolicy = EffectiveResourcePolicyTable.syntax("effectiveResourcePolicy")
+
+    serializableWriteTransaction("recreateEffectivePolicyRolesTableEntry", samRequestContext)({ implicit session =>
+      samsql"""delete from ${EffectivePolicyRoleTable as effectivePolicyRole}
+               using ${EffectiveResourcePolicyTable as effectiveResourcePolicy},
```
andy7i: I think due to the use of `using` and many tables being joined, I'm having some difficulty seeing how the tables are joined and where the important filters are. 

I think using `join ... on ...` would make this more clear. (for example):
delete from ${EffectivePolicyRoleTable as effectivePolicyRole}
               join ${EffectiveResourcePolicyTable as effectiveResourcePolicy} on effectivePolicyRole = effectiveResourcePolicy.id
               join ...
where ${resourceType.name} = ${resourceTypeName}



thoughts?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -68,9 +67,20 @@ class ResourceService(
                 Resource(resourceTypeAdmin.name, ResourceId(rtName.value), Set.empty)
             }
           }
+          _ <- updateEffectivePolicyTables(samRequestContext, newOrUpdatedResourceTypeNames)
         } yield resourceTypes.values
     }
 
+  def updateEffectivePolicyTables(samRequestContext: SamRequestContext, newOrUpdatedResourceTypeNames: Set[ResourceTypeName]): IO[List[Unit]] = {
```
andy7i: >This PR does not add anything to update the sam_effective_policy_action table, because it turns out that we don't actually seem to support descendant actions in reference.conf, so that table isn't actually effected by that bug on boot. It is possible that the bug also manifests when people call the updatePolicy endpoint in Sam, which is how that could actually be updated, but I would push to deprioritize that in favor of other work, because to my understanding we have basically deprecated that in favor of the configuration file.

should we create a tech debt ticket?  or capture this in a code comment around here

andy7i (commented):
Commenting here for now, but happy to be a reviewer.

the sql debugging tips are 🔥 

PR comments are helpful too!

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
marctalbott: Postgres doesn't support joins on delete statements unfortunately, so the `using` syntax is the only way to do joins afaik

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1350,7 +1350,59 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
   }
 
-  private def setPolicyIsPublicInternal(policyPK: PolicyPK, isPublic: Boolean)(implicit session: DBSession): Int = {
+  override def recreateEffectivePolicyRolesTableEntry(resourceTypeName: ResourceTypeName, samRequestContext: SamRequestContext): IO[Unit] = {
+    val resource = ResourceTable.syntax("resource")
+    val policyResource = ResourceTable.syntax("policyResource")
+    val resourceRole = ResourceRoleTable.syntax("resourceRole")
+    val flattenedRole = FlattenedRoleMaterializedView.syntax("flattenedRole")
+    val resourceTypeOne = ResourceTypeTable.syntax("resourceTypeOne")
+    val resourceTypeTwo = ResourceTypeTable.syntax("resourceTypeTwo")
```
marctalbott: Can we rename these to `policyResourceType` and `roleResourceType` respectively?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1350,7 +1350,59 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
     })
   }
 
-  private def setPolicyIsPublicInternal(policyPK: PolicyPK, isPublic: Boolean)(implicit session: DBSession): Int = {
+  override def recreateEffectivePolicyRolesTableEntry(resourceTypeName: ResourceTypeName, samRequestContext: SamRequestContext): IO[Unit] = {
+    val resource = ResourceTable.syntax("resource")
```
marctalbott, line 1355: Is there a more informative name we could use here to distinguish it from `policyResource`? Maybe `effectiveResource`? Not sure if that's more helpful or less...

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
andy7i: looks like there's a workaround for that, if we want to go this route: https://stackoverflow.com/questions/21662726/delete-using-left-outer-join-in-postgres

marctalbott: TOL: I wonder if adjusting the order of the `and` clauses would make them easier to follow? Something like:
```
where ${effectivePolicyRole.effectiveResourcePolicyId} = ${effectiveResourcePolicy.id}
               and ${effectiveResourcePolicy.sourcePolicyId} = ${policy.id}
               and ${policy.resourceId} = ${policyResource.id}
               and ${policyResource.resourceTypeId} = ${resourceTypeOne.id}
               and ${effectivePolicyRole.resourceRoleId} = ${resourceRole.id} # keep role related joins together?
               and ${resourceRole.resourceTypeId} = ${resourceTypeTwo.id}
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
marctalbott: We update the effective policy tables when we create or overwrite policies

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -68,9 +67,20 @@ class ResourceService(
                 Resource(resourceTypeAdmin.name, ResourceId(rtName.value), Set.empty)
             }
           }
+          _ <- updateEffectivePolicyTables(samRequestContext, newOrUpdatedResourceTypeNames)
```
marctalbott: I think we want this to be in the same transaction as `upsertResourceTypes`. We obtain a lock there that prevents concurrent writes when multiple Sam's are starting up and that will be useful here as well

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -68,9 +67,20 @@ class ResourceService(
...
         } yield resourceTypes.values
     }
 
+  def updateEffectivePolicyTables(samRequestContext: SamRequestContext, newOrUpdatedResourceTypeNames: Set[ResourceTypeName]): IO[List[Unit]] = {
+    for {
+      response <- newOrUpdatedResourceTypeNames.filterNot(_ == SamResourceTypes.resourceTypeAdminName).toList.traverse { rtName => {
```
marctalbott: We should include `resource-type-admin` here in case we ever add hierarchical roles to that resource type

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
s-rubenstein: I will try the workaround, and if that doesn't work will do the reordering.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
s-rubenstein: I think we need to do it after `persistResource`. I tried adding it to the `upsertResourceTypes` method, and that failed my tests, so I think it needs to happen there. I am going to add a lock to it, but I think it should be okay to do it outside of the other lock, because things will catch up if changes are made between upsert and updating the effective policy tables.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
s-rubenstein: I can try the workaround, otherwise I will readjust the order.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
s-rubenstein: I don't actually think that's more helpful. I don't think it is any less helpful, but verbosity has its own cost. My inclination is to leave it. Unfortunately I don't think there's a more descriptive name, because it kind of does just describe the resource.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
s-rubenstein: I tried a few things, and got:
`There is an entry for table "effectivepolicyrole", but it cannot be referenced from this part of the query.`

Given that we have used `using` throughout the codebase, and its more idiomatic for psql, I'm going to suggest that I just readjust the order and we stay with using.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
s-rubenstein: Thanks for adding that context! I was going to go looking but that sounds good.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
marctalbott: What failed in the tests? Persisting the `resource-type-admin` resources shouldn't impact our ability to update effective policy tables from what I can tell

s-rubenstein: The permissions check at the end of the test failed, which means the insert query and/or the delete query returned a different response. I didn't dig into why, but I can.

andy7i (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
s-rubenstein: (This was the test failure. Looks like maybe the delete failed?)

true was not equal to false
ScalaTestFailureLocation: org.broadinstitute.dsde.workbench.sam.service.ResourceServiceSpec at (ResourceServiceSpec.scala:1132)
Expected :false
Actual   :true
<Click to see difference>

marctalbott (approved):

s-rubenstein:
Jenkins retest

PR #559, opened 2021-11-24
AS-903 When a user is enabled check TOS Acceptance before adding them OpenDJ 

jordansafer:
Ticket: https://broadworkbench.atlassian.net/browse/AS-903
When a user is enabled, check TOS:
- If TOS is disabled, enable the user in OpenDJ (old behavior)
- If TOS is enabled and the user has accepted the latest TOS, enable in OpenDJ
- If TOS is enabled and the user hasn't accepted the latest TOS, don't enable in OpenDJ

Testing:
- Local tests passing and additional added tests verifying the above 3 cases

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -198,7 +198,7 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
       case _ => Future.successful(None)
     }
 
-  def enableUser(userId: WorkbenchUserId, userInfo: UserInfo, samRequestContext: SamRequestContext): Future[Option[UserStatus]] =
+  def enableUser(userId: WorkbenchUserId, samRequestContext: SamRequestContext): Future[Option[UserStatus]] =
     directoryDAO.loadUser(userId, samRequestContext).unsafeToFuture().flatMap {
       case Some(user) =>
         for {
@@ -211,9 +211,17 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
   private def enableUserInternal(user: WorkbenchUser, samRequestContext: SamRequestContext): Future[Unit] = {
     for {
       _ <- directoryDAO.enableIdentity(user.id, samRequestContext).unsafeToFuture()
-      _ <- registrationDAO.enableIdentity(user.id, samRequestContext).unsafeToFuture()
+      _ <- enableIdentityIfTosAccepted(user.id, samRequestContext).unsafeToFuture()
       _ <- cloudExtensions.onUserEnable(user, samRequestContext)
     } yield ()
+  }
+
+  private def enableIdentityIfTosAccepted(userId: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Unit] = {
+    tosService.getTosStatus(userId)
+      .flatMap {
+        case Some(true) => registrationDAO.enableIdentity(userId, samRequestContext)
+        case None => IO.unit
+      }
   }
 
   def disableUser(userId: WorkbenchUserId, samRequestContext: SamRequestContext): Future[Option[UserStatus]] =
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -124,10 +124,11 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
             allUsersStatus <- directoryDAO.isGroupMember(allUsersGroup.id, user.id, samRequestContext).unsafeToFuture() recover { case _: NameNotFoundException => false }
             tosAcceptedStatus <- tosService.getTosStatus(user.id).unsafeToFuture()
             ldapStatus <- registrationDAO.isEnabled(user.id, samRequestContext).unsafeToFuture()
+            adminEnabled <- directoryDAO.isEnabled(user.id, samRequestContext).unsafeToFuture()
           } yield {
             val enabledMap = Map("ldap" -> ldapStatus, "allUsersGroup" -> allUsersStatus, "google" -> googleStatus)
             val enabledStatuses = tosAcceptedStatus match {
-              case Some(status) => enabledMap + ("tosAccepted" -> status)
+              case Some(status) => enabledMap + ("tosAccepted" -> status, "adminEnabled" -> adminEnabled)
               case None => enabledMap
             }
             val res = Option(UserStatus(UserStatusDetails(user.id, user.email), enabledStatuses))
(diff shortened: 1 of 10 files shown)


jordansafer:
jenkins retest

jordansafer:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
```diff
@@ -61,7 +61,7 @@ class UserRoutesV1Spec extends UserRoutesSpecHelper{
       val res = responseAs[UserStatus]
       res.userInfo.userSubjectId.value.length shouldBe 21
       res.userInfo.userEmail shouldBe defaultUserEmail
-      res.enabled shouldBe Map("ldap" -> true, "allUsersGroup" -> true, "google" -> true, "tosAccepted" -> false)
+      res.enabled shouldBe Map("ldap" -> false, "allUsersGroup" -> true, "google" -> true, "tosAccepted" -> false)
```
jordansafer: Right now, when we return a response map, we read from OpenDJ for the value we return in ldap, so this value is equivalent to (postgres.isEnabled && postgres.tosAccepted). I think it would be more correct to read/return the value from postgres, so we would return the postgres.isEnabled() value. Thoughts?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -192,7 +192,7 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
       case _ => Future.successful(None)
     }
 
-  def enableUser(userId: WorkbenchUserId, userInfo: UserInfo, samRequestContext: SamRequestContext): Future[Option[UserStatus]] =
+  def enableUser(userId: WorkbenchUserId, samRequestContext: SamRequestContext): Future[Option[UserStatus]] =
```
sagehen03, line 205: Thanks for removing the unusued param.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
sagehen03: I don't think I follow. I think you're talking about how the value for the ldap key is determined, but you might just be talking about this test.  I'm not sure :)

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
```diff
@@ -194,6 +194,63 @@ class UserServiceSpec extends AnyFlatSpec with Matchers with TestSupport with Mo
...
+    // create a user
+    val newUser = serviceTosEnabled.createUser(defaultUser, acceptTos = true, samRequestContext).futureValue
+    newUser shouldBe UserStatus(UserStatusDetails(defaultUserId, defaultUserEmail), Map("ldap" -> true, "allUsersGroup" -> true, "google" -> true, "tosAccepted" -> true))
+
+    // it should be enabled
+    dirDAO.isEnabled(defaultUserId, samRequestContext).unsafeRunSync() shouldBe true
+    registrationDAO.isEnabled(defaultUserId, samRequestContext).unsafeRunSync() shouldBe true
+  }
+
+  private def updateTosVersionThenEnableUser(tosEnabled: Boolean = true, userAcceptsNewTos: Boolean = false): Unit = {
```
sagehen03, line 238: I'm wondering if a better name for this might be setUpUserWithoutNewTosAcceptance() or something but maybe I'm missing what this method actually does.

sagehen03 (commented):
I'm not sure what you were getting at in your comment on UserRoutesV1Spec see my comment there, but otherwise looks like this will do the job.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
MatthewBemis: That's an interesting question. I think there is still merit to returning the value that's actually in LDAP, for account debugging purposes, even if it should be derivable from other pieces of information. Could we still return `ldap` as-is and return `postgres.isEnabled()` as a new additional value?

MatthewBemis (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1Spec.scala
jordansafer: Definitely, I'll add a value called "isEnabled" to the result map to capture this. @sagehen03, the first one! I'm interested in how the value for the ldap key is determined, because it used to be true when the user was enabled, but now it's only true if the user is enabled & accepts ToS (when ToS is enabled of course!).

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
jordansafer: I see, I want to keep the "enable user" part because this change is specific to the enable user API call, but I agree "WithoutNewTosAcceptance" is clearer than just saying "updateTosVersion", since the point of the test is that the user hasn't accepted the latest ToS when they are enabled. Updating!

sagehen03 (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
jordansafer: I decided it would be better not to update, because I realize this function is also used for cases where the user accepts the new tos, (hence the userAcceptsNewTos parameter), so it would be confusing to name it "withoutNewTosAcceptance" since that's not what happens in all cases-

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -124,10 +124,11 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
             allUsersStatus <- directoryDAO.isGroupMember(allUsersGroup.id, user.id, samRequestContext).unsafeToFuture() recover { case _: NameNotFoundException => false }
             tosAcceptedStatus <- tosService.getTosStatus(user.id).unsafeToFuture()
             ldapStatus <- registrationDAO.isEnabled(user.id, samRequestContext).unsafeToFuture()
+            adminEnabled <- directoryDAO.isEnabled(user.id, samRequestContext).unsafeToFuture()
           } yield {
             val enabledMap = Map("ldap" -> ldapStatus, "allUsersGroup" -> allUsersStatus, "google" -> googleStatus)
             val enabledStatuses = tosAcceptedStatus match {
-              case Some(status) => enabledMap + ("tosAccepted" -> status)
+              case Some(status) => enabledMap + ("tosAccepted" -> status) + ("adminEnabled" -> adminEnabled)
```
sagehen03, line 131: Why do we only include adminEnabled in the map if tosAcceptedStatus == true?  Am I reading that correctly?

jordansafer: Ah! The nuance is that tosAcceptedStatus can be Some(true) or Some(false), and adminEnabled will be added in both of these cases. If ToS is enabled, the status will be Some(_) and if it's disabled the status will be None. So with this change we're adding the adminEnabled field in cases where Tos is enabled.

sagehen03: Gotcha.  I should have realized that.  Thanks for your patience.

sagehen03 (approved):

MatthewBemis (approved):

PR #562, opened 2021-12-08
[AS-902] When user accepts ToS, add user to OpenDJ if they are enabled

jordansafer:
Ticket: https://broadworkbench.atlassian.net/browse/AS-902
- Added logic to enabled OpenDJ when ToS is accepted
- Updated tests to reflect the change

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -143,9 +143,8 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
       case Some(_) =>
         for {
           _ <- tosService.acceptTosStatus(userId)
-          _ <- directoryDAO.isEnabled(userId, samRequestContext).flatMap(enabled => {
-            if (enabled) registrationDAO.enableIdentity(userId, samRequestContext) else IO.none
-          })
+          enabled <- directoryDAO.isEnabled(userId, samRequestContext) 
+          _ <- if (enabled) registrationDAO.enableIdentity(userId, samRequestContext) else IO.none
           status <- IO.fromFuture(IO(getUserStatus(userId, false, samRequestContext)))
         } yield status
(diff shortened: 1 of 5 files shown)


sagehen03 (approved):
lgtm with the caveat that I don't know scala well enough to bless the UserService changes.  From the tests, seems pretty clear that they work, but would be good to get a scala veteran's take on whether they're idiomatic.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -143,6 +143,9 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
       case Some(_) =>
         for {
           _ <- tosService.acceptTosStatus(userId)
+          _ <- directoryDAO.isEnabled(userId, samRequestContext).flatMap(enabled => {
+            if (enabled) registrationDAO.enableIdentity(userId, samRequestContext) else IO.none
+          })
           status <- IO.fromFuture(IO(getUserStatus(userId, false, samRequestContext)))
         } yield status
```
davidangb, line 149: for-comprehensions are implemented under the covers as flatMaps, so you could write this as:
```
        for {
          _ <- tosService.acceptTosStatus(userId)
          directoryEnabled <- directoryDAO.isEnabled(userId, samRequestContext)
          _ <- if (directoryEnabled) registrationDAO.enableIdentity(userId, samRequestContext) else IO.none
          status <- IO.fromFuture(IO(getUserStatus(userId, false, samRequestContext)))
        } yield status
```
I think both approaches have their merits, I don't feel strongly about changing this.

davidangb (approved):
optional feedback inline, else :+1:

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
jordansafer: Ah! I prefer your approach given that we're already using for-iterations- updating-

PR #566, opened 2021-12-20
AS-1050: If ToS enforcement is enabled, show the relevant models in Swagger

MatthewBemis:
Tried and true, via https://github.com/broadinstitute/rawls/pull/1205

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SwaggerRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SwaggerRoutes.scala
@@ -1,18 +1,41 @@
 package org.broadinstitute.dsde.workbench.sam.api
 
+import akka.http.scaladsl.model.{ContentTypes, HttpEntity}
 import akka.http.scaladsl.server
 import akka.http.scaladsl.server.Directives._
 import akka.stream.scaladsl.Flow
 import akka.util.ByteString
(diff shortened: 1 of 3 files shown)


MatthewBemis:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SwaggerRoutes.scala
```diff
@@ -67,4 +90,34 @@ trait SwaggerRoutes {
     }
   }
 
+  private def loadResource(filename: String) = {
```
jordansafer, line 93: Since we're reusing these functions from Rawls, do you think it makes sense to put them in the common library and re-use the definition? Could be a future ticket-

jordansafer (approved):
LGTM :)

sagehen03 (approved):

MatthewBemis:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SwaggerRoutes.scala
MatthewBemis: That's a good idea! Rawls doesn't actually use this code anymore (we ripped it out when we removed the relevant feature flag), but this is code that I think will continue to be relevant so it would be nice to not have to dig through old commits to find it :)

PR #567, opened 2021-12-21
AS-905: Disable all human users when ToS version is updated

MatthewBemis:
https://broadworkbench.atlassian.net/browse/AS-905

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -600,6 +604,10 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
     })
   }
 
+  override def disableAllIdentities(samRequestContext: SamRequestContext): IO[Unit] = {
+    IO.unit //no-op for now. throw exception maybe?
```
marctalbott, line 608: I don't have a strong opinion on a no-op vs. throwing an exception here and in `createEnabledUsersGroup`. no-op seems fine to me

marctalbott (commented):
The LDAP stuff looks good

Is this going to work with pets? It seems like anytime we bump the terms of service version, we'll disable all existing pets and block them from accessing Terra APIs. Maybe `acceptTermsOfService` in the UserService needs to call `cloudExtensions.onUserEnable` to re-add the pets to the enabled users group? Although that does extra stuff wrt the user's proxy group

MatthewBemis:
@marctalbott that's a great call-out about the pet stuff. I'll put some more thought into that.

Since this only applies to human users, we don't need to actually disable pets in the first place. I'll see if I can alter that LDAP query to only deal with human users

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
```diff
@@ -93,6 +93,28 @@ class LdapRegistrationDAO(
...
+        unmarshalUser(result)
+      }
+    }).map { identityResults => identityResults.collect { case Right(user) => subjectDn(user.id) }}
+
+    val enabledIdentityDnsIO = executeLdap(IO(getAttributes(ldapConnectionPool.getEntry(directoryConfig.enabledUsersGroupDn, Attr.member), Attr.member)), "getAllIdentitiesEnabled", samRequestContext).map(_.toList)
+
+    for {
+      humanIdentityDns <- humanIdentityDnsIO //this retrieves all humans registered in the system (looking under the peopleOu)
+      enabledIdentityDns <- enabledIdentityDnsIO //this retrieves all enabled identities in the system (looking in enabledUsersGroupDn)
+      humanIdentityDnsToDisable = humanIdentityDns intersect enabledIdentityDns //intersecting them gives the list of all human identities who are enabled (AKA filtering out humans who are disabled in the system)
```
MatthewBemis, line 111: unfortunately this step is necessary because doing an LDAP delete of an attribute value that doesn't exist causes the entire batch to fail. so if there was a user who was already disabled in the system, we need to filter them out first

MatthewBemis:
@marctalbott this is ready for re-review. PTAL!

marctalbott (approved):

jordansafer (approved):
Looks great!!

PR #570, opened 2022-01-25
Fix typos in group policy error message

melissachang:


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -230,7 +230,7 @@ object ManagedGroupService {
       case `adminNotifierValue` => ManagedGroupService.adminNotifierPolicyName
       case _ =>
         throw new WorkbenchExceptionWithErrorReport(
-          ErrorReport(StatusCodes.NotFound, "Policy name for managed groups must be one of: [\"admins\", \"members\"]"))
+          ErrorReport(StatusCodes.NotFound, "Policy name for managed groups must be one of: [\"admin\", \"member\"]"))
```
jordansafer: Looks good! Consider using the variables, as on line 243 of this file to make it so it's not hardcoded for the future!

jordansafer (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
melissachang: I'm not familiar with Scala so this will have to do, lthanks. :)

jordansafer: It would be ` "Policy name for managed groups must be one of: ['$adminValue', '$memberValue']" `, if you're familiar with typescript/ES6+ or Kotlin, they have a similar thing!

melissachang: Done, thanks

jordansafer (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
melissachang: I get `possible missing interpolator: detected interpolated identifier `$adminValue``

https://github.com/broadinstitute/sam/runs/4956231910?check_suite_focus=true

jordansafer: My mistake!! It should be:
`s"Policy name for managed groups must be one of: ['$adminValue', '$memberValue']"`

... I was missing the s in my comment :/

PR #574, opened 2022-01-27
[draft] WOR-48 + WOR-58: Allow ToS to be enabled, but not enforced

MatthewBemis:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
@@ -3,8 +3,8 @@ package org.broadinstitute.dsde.workbench.sam.service
 import akka.http.scaladsl.model.StatusCodes
 import cats.effect.IO
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -19,19 +19,43 @@ class TosService (val directoryDao: DirectoryDAO, val registrationDao: Registrat
...
+                  case None => Set[WorkbenchSubject]()
+                } flatMap { exemptedUsers =>
+                  registrationDao.disableAllHumanIdentities(SamRequestContext(None), exemptedUsers).map { _ => Option(createdGroup) }
+                }
+              } else IO.pure(Option(createdGroup))
+
+            }
+          case group => IO.pure(group)
+        }
+      } else {
```
MatthewBemis, line 51: this `else` is essentially the previous implementation, with special logic to handle the `enforced` flag. some of the logic in this clause could be condensed with some of the logic in the `if` clause, but i feel like getting too cute with this stuff will make it more difficult to reason about. the `if` clause can also be removed after the first version of ToS, so this makes it easier to rip it out

PR #575, opened 2022-01-28
[CA-1746] Create disable user pub/sub topic/subscription, if necessary

melissachang:
Ticket: https://broadworkbench.atlassian.net/browse/PF-1292

While working on PF-1292, I realized that Verily SAM pub/sub was misconfigured. All 3 pub-subs -- group sync, user disable and google key cache -- were using the same topic/subscription name.

I fixed this by using topic/sub names `sam-disable-users`, `sam-google-key-cache`, and `sam-group-sync`. When I redeployed SAM, SAM automatically created the latter 2 and not `sam-disable-users`, which resulted in this SAM warning. This PR fixes that.

image

(As a workaround, I manually create the disable topic/subscription in GCP console. Then SAM started up fine.)

Regarding Security Risk Assessment -- PF-1292 is where I found that Verily's SAM pub/sub was incorrect. That ticket's Security Assessment is N/A, since it's mostly just configuring SAM groups. Should I create a separate ticket for this PR, and fill out a Security Assessment?

@jmczerk

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
@@ -58,6 +58,7 @@ class DisableUsersMonitorSupervisor(
 
   def init: Future[DisableUsersMonitorSupervisor.Start.type] =
     for {
+      _ <- pubSubDao.createTopic(pubSubTopicName)
       _ <- pubSubDao.createSubscription(pubSubTopicName, pubSubSubscriptionName)
     } yield Start
 


s-rubenstein (approved):
This LGTM, although could you test what happens if the topic is already in existence? I don't think we want to block subscribing to it if that errors. And then the other issue is what happens with local Sam instances in this case pointing at dev, although I think it also has the same 'what happens if the topic already exists issue'

Given this is about enabled users who should be disabled staying enabled, this feels maybe like it should be higher than N/A for security. Do we know why the original ticket was N/A?

melissachang:
> This LGTM, although could you test what happens if the topic is already in existence? I don't think we want to block subscribing to it if that errors. And then the other issue is what happens with local Sam instances in this case pointing at dev, although I think it also has the same 'what happens if the topic already exists issue'

Forgot to mention, [GoogleGroupSyncMonitor.scala](https://github.com/broadinstitute/sam/blob/e5192ddef7ea4189c22ca96e4007fd6cc94a1e8b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala#L66) and [GoogleKeyCacheMonitor.scala](https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCacheMonitor.scala#L76) already have this. It was probably an oversight that DisableUsersMonitor.scala doesn't have this.  So I don't think manual testing is necessary, WDYT?

> 
> Given this is about enabled users who should be disabled staying enabled, this feels maybe like it should be higher than N/A for security. Do we know why the original ticket was N/A?

I filed CA-1746. Here's what I put for Security Impact Description:

> According to https://broadinstitute.github.io/dsp-appsec-security-risk-assessment/ impact is MEDIUM.
> 
> However, to me it seems low to none. I'm just adding to DisableUsersMonitor.scala what [GoogleGroupSyncMonitor.scala](https://github.com/broadinstitute/sam/blob/e5192ddef7ea4189c22ca96e4007fd6cc94a1e8b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala#L66) and [GoogleKeyCacheMonitor.scala](https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCacheMonitor.scala#L76) already do.

s-rubenstein:
Seems fine then! Sounds good.

melissachang:
Jenkins run

melissachang:
jenkins retest

melissachang:
Jenkins doesn't like `/` in branch name. Sending #577.

PR #580, opened 2022-02-08
WOR-77: Update to latest Terms of Service text

MatthewBemis:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/resources/termsOfService.md
+++ b/src/main/resources/termsOfService.md
@@ -144,8 +144,8 @@ By accepting these Terms of Service, you acknowledge:
     datasets may result in disciplinary action, civil, and/or criminal
     penalties. Unauthorized users of these websites or datasets should
     have no expectation of privacy regarding any communications or
-    data processed by these websites. [A list of US
-    Government websites and datasets can be found here](https://support.terra.bio/hc/en-us/community/posts/360056900812-Government-owned-Websites-and-Datasets-Last-Revised-Feb-5-2020-)
+    data processed by these websites.
+    [A list of US Government websites and datasets can be found here](https://support.terra.bio/hc/en-us/community/posts/360056900812-Government-owned-Websites-and-Datasets-Last-Revised-Feb-5-2020-)
 
 -   By using any of these websites or otherwise accessing these datasets
     through Terra, you expressly consent to monitoring of your actions
--- a/src/main/resources/termsOfService.md
+++ b/src/main/resources/termsOfService.md
@@ -1,3 +1,5 @@
+# Terra Terms of Service
+
 Your access to systems and networks owned by The Broad Institute, Inc.
 ("Broad") is governed by, and subject to the following terms and
 conditions, as well as all Federal laws, including, but not limited to,
@@ -12,7 +14,7 @@ Privacy Act, and applicable State and Federal laws.
 Terra uses cookies to enable sign on and other essential features when
 signed in, and to provide statistics to our development team regarding
 how the site is used. For more information, see our privacy policy.
-[Privacy policy, including privacy policy of uploaded data, is here](/#privacy).
+[Privacy policy, including privacy policy of uploaded data, is here](https://app.terra.bio/#privacy).
 You agree to read and consider before uploading data.
 
 ## Modification of the Agreement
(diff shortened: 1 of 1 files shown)


src/main/resources/termsOfService.md
```diff
@@ -269,4 +271,4 @@ including attorneys' fees and costs, arising out of or relating to your
 uploading, storing, or transferring of PHI without having fully
 implemented the Clinical Features.
 
-Terms as of February 12, 2020.
\ No newline at end of file
+Terms as of February 25, 2022.
```
MatthewBemis, line 272: just guessing the date here, for now. i'll update when it's clearer

src/main/resources/termsOfService.md
```diff
@@ -160,20 +162,20 @@ legitimate tasks or assigned duties. If you believe you are being
...
-## Restricted Use of TCGA Controlled-Access Data
+## Restricted Use of Federal Controlled-Access Data
 
-To access TCGA controlled access data, you must first request to link
+To access federal controlled access data, you must first request to link
 your Terra account to the eRA Commons or NIH identity under which you
 are granted dbGaP authorized access. If you have dbGaP authorized
-access, Terra permits read access to the TCGA controlled-access data
-files.
+access, Terra permits read access to those the federal controlled-access
```
cahrens, line 168: There's something wrong here with the grammar. Perhaps `the` should be removed?

src/main/resources/termsOfService.md
```diff
@@ -160,20 +162,20 @@ legitimate tasks or assigned duties. If you believe you are being
...
+data files.
 
 If the NIH or eRA commons identity authentication fails or the
 authenticated identity no longer appears in the NIH-provided whitelist,
-you can no longer access TCGA controlled access data files. It is your
-responsibility to delete any TCGA controlled-access data in your
+you can no longer access federal controlled access data files. It is your
```
cahrens, line 173: Should be "controlled-access data files" for consistency with the other usages.

src/main/resources/termsOfService.md
```diff
@@ -160,20 +162,20 @@ legitimate tasks or assigned duties. If you believe you are being
...
+responsibility to delete any federal controlled-access data in your
 personal workspaces if you lose dbGaP authorized access. You must also
-refrain from distributing TCGA controlled access data to users unless
+refrain from distributing federal controlled access data to users unless
```
cahrens, line 176: Should be "controlled-access data", no?

cahrens: I see the text is pretty inconsistent with "controlled access" vs. "controlled-access". It seems that usually "controlled-access" is used to describe "data files", but "controlled access" for just data. that seems OK I guess.

src/main/resources/termsOfService.md
cahrens: I think we should change this one given the usage in the very next sentence…

MatthewBemis: I agree that it's inconsistent, but for context, this is what the legal team provided to us verbatim. I personally don't feel comfortable making even minor adjustments to a legal document that's been reviewed and agreed upon by the Broad legal team and the BVM partners.

cahrens: Is it possible to ask them (at least about the extra "the" above)? On the other hand, it's not like anyone will ever actually read it… I guess it doesn't matter.

cahrens (approved):
The grammarian in me is weeping.

src/main/resources/termsOfService.md
```diff
@@ -1,3 +1,5 @@
+# Terra Terms of Service
```
marctalbott: The UI currently adds this header to the text that Sam returns, if Sam also returns this, then it'll have the header twice. Do we need Sam to return this for other portals? Otherwise, I'd say we should remove this here and leave it to the UI to display the header

MatthewBemis: Yeah that may be the correct thing to do. I may have taken "verbatim" a bit too literally when copying over the entire doc. As long as the title is displayed when we present it to users, then we're good. ~Most portals will redirect to Terra UI, and those that won't (AoU) will be providing context for what it is~ All portals that I know of will be linking to Terra UI to display our terms

MatthewBemis: Also looking at the revision history for the source document, it doesn't look like this title was explicitly added in this version of the text either

marctalbott (approved):

PR #583, opened 2022-02-17
[WOR-70] add read_spend_report action to billing-project owner role

marctalbott:
Ticket: [WOR-70](https://broadworkbench.atlassian.net/browse/WOR-70)
Only billing project owners will be allowed to generate spend reports for their billing projects. We could use an existing action to decide whether a user should be able to do this, but why not just create a specific action?

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -652,6 +652,9 @@ resourceTypes = {
       read_spend_report_configuration = {
         description = "read the spend report config (i.e. the BigQuery dataset for spend reports) for this billing-project"
       }
+      read_spend_report = {
+        description = "read spend report which contains data pulled from BigQuery dataset for this billing-project"
+      }
       add_child = {
         description = "add children, e.g. google-project"
       }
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):
lgtm, and I think making a new action specific to this functionality is the best move 👍

aherbst-broad (approved):

src/main/resources/reference.conf
```diff
@@ -652,6 +652,9 @@ resourceTypes = {
       read_spend_report_configuration = {
         description = "read the spend report config (i.e. the BigQuery dataset for spend reports) for this billing-project"
       }
+      read_spend_report = {
```
cahrens, line 655: Why is it "read" instead "create"? It seems like potentially one could want both permissions (if we cached the result of running a report), but we plan to only allow billing project owners to both create a report and view it.

marctalbott: I went back and forth between `create` and `read`. I ended up choosing `read` over `create` because I felt that was what we were really checking -- whether the user could read data from the BQ dataset. Still, I'm not set on `read` so if you think `create` makes more sense than `read` then we can change it!

cahrens: I thought that we would be generating reports via GQ on the fly. Are you saying that the report data will already be present and we are just returning it? If so, "read" makes sense to me. But if we will be creating anything based on this permission I think that "create" is more descriptive. I'm conditioned to think that there will be both read and write permissions for things.

marctalbott: At the moment, we are just reading spend data from BigQuery that Google writes automatically if Billing Account spend export has been set up. That said, the data in BQ is just raw spend data, so the report generation that we're doing is gathering that data and transforming it into human readable spend reports which is all done on the fly

PR #584, opened 2022-02-18
WOR-119: Immediately enable new users if the grace period is active

MatthewBemis:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1145,3 +1145,7 @@ distributedLock {
 }
 
 blockedEmailDomains = ["qwiklabs-gsuite.net", "qwiklabs.net"]
+
+termsOfService {
+  enableGracePeriod = false
+}
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1147,5 +1147,5 @@ distributedLock {
 blockedEmailDomains = ["qwiklabs-gsuite.net", "qwiklabs.net"]
 
 termsOfService {
-  enableGracePeriod = false
+  isGracePeriodEnabled = false
 }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
@@ -136,7 +136,7 @@ object AppConfig {
   implicit val termsOfServiceConfigReader: ValueReader[TermsOfServiceConfig] = ValueReader.relative { config =>
     TermsOfServiceConfig(
       config.getBoolean("enabled"),
-      config.getBoolean("enableGracePeriod"),
+      config.getBoolean("isGracePeriodEnabled"),
       config.getInt("version"),
       config.getString("url")
     )
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
@@ -136,6 +136,7 @@ object AppConfig {
   implicit val termsOfServiceConfigReader: ValueReader[TermsOfServiceConfig] = ValueReader.relative { config =>
     TermsOfServiceConfig(
       config.getBoolean("enabled"),
+      config.getBoolean("enableGracePeriod"),
       config.getInt("version"),
       config.getString("url")
     )
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/TermsOfServiceConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/TermsOfServiceConfig.scala
@@ -3,9 +3,9 @@ package org.broadinstitute.dsde.workbench.sam.config
 /**
   * Terms of Service configuration.
   * @param enabled Set to false to disable Terms of Service enforcement
-  * @param enableGracePeriod Set to true if the grace period for ToS acceptance is active
+  * @param isGracePeriodEnabled Set to true if the grace period for ToS acceptance is active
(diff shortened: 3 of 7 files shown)


src/main/resources/reference.conf
```diff
@@ -1145,3 +1145,7 @@ distributedLock {
 }
 
 blockedEmailDomains = ["qwiklabs-gsuite.net", "qwiklabs.net"]
+
+termsOfService {
+  enableGracePeriod = false
```
MatthewBemis: to be enabled as part of https://github.com/broadinstitute/firecloud-develop/pull/2753 once the grace period goes live

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
```diff
@@ -136,6 +136,7 @@ object AppConfig {
   implicit val termsOfServiceConfigReader: ValueReader[TermsOfServiceConfig] = ValueReader.relative { config =>
     TermsOfServiceConfig(
       config.getBoolean("enabled"),
+      config.getBoolean("enableGracePeriod"),
```
ungwudik: Nit: would this be more readable as `isGracePeriodEnabled`? Assuming that's the question the boolean is asking.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -233,13 +233,15 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
   }
 
   private def enableIdentityIfTosAccepted(user: WorkbenchUser, samRequestContext: SamRequestContext): IO[Unit] = {
+    val gracePeriodEnabled = tosService.tosConfig.enableGracePeriod
+    println(gracePeriodEnabled)
```
MatthewBemis: oops forgot to remove this println from testing

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/TermsOfServiceConfig.scala
```diff
@@ -3,8 +3,9 @@ package org.broadinstitute.dsde.workbench.sam.config
 /**
   * Terms of Service configuration.
   * @param enabled Set to false to disable Terms of Service enforcement
+  * @param enableGracePeriod Set to true if the grace period for ToS acceptance is active
```
ungwudik: oh here's the definition. Same question as before naming.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -233,13 +233,14 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
     tosService.getTosStatus(user.id)
       .flatMap {
-        // If the user has accepted TOS or TOS is disabled, then enable the user in LDAP
-        // If the user is an SA, it's also acceptable to enable them in LDAP
+        // If the user has accepted TOS, the grace period is enabled, or TOS is disabled, then enable the user in LDAP
+        // Additionally, if the user is an SA, it's also acceptable to enable them in LDAP
         case Some(true) | None => registrationDAO.enableIdentity(user.id, samRequestContext)
         case _ =>
-          if(isServiceAccount(user.email.value)) registrationDAO.enableIdentity(user.id, samRequestContext)
+          if(isServiceAccount(user.email.value) || gracePeriodEnabled) registrationDAO.enableIdentity(user.id, samRequestContext)
```
cahrens, line 243: How long is the grace period? Where is the code that makes sure that the user must accept the ToS once the grace period has expired for them?

marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/TermsOfServiceConfig.scala
ungwudik: Ok, I think this is an 'action' variable, and is in line with other configs variables?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
MatthewBemis: It's an undefined amount of time. We're really only implementing this grace period to support a smoother rollout for AoU. They indicated that for this initial rollout of ToS, they need the APIs to be enabled in production but for it to not cause any breakage for users, so this is our solution to that. Once AoU gives us the go-ahead, we'll flip the boolean in sam.conf which will end the grace period. Later on, this grace period code can be removed.

ungwudik (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/TermsOfServiceConfig.scala
MatthewBemis: I renamed it to `isGracePeriodEnabled` per your suggestion. I defer strongly to the reviewer for this type of feedback, since the reviewer is more likely to know whether or not something is readable. I chose the name so I'm biased :)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
cahrens: Oh, I see. Is there a ticket to remove the grace period code when it is no longer needed?

MatthewBemis: Not yet! I will make one

PR #593, opened 2022-03-07
[CA-1687] Upgrade Client Java by upgrading workbench google and google2

s-rubenstein:
Ticket: [Link to Jira ticket](https://broadworkbench.atlassian.net/browse/CA-1687)
This upgrades our workbench libs dependencies. There are a few changes to the tests that need to be made in order to deal with increased inefficiencies or removed test functionality.

---

**PR checklist**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,4 +21,13 @@ sys.env.getOrElse("JAVA_OPTS", "").split(" ").toSeq.map { opt =>
   javaOptions in reStart += opt
 }
 
+inThisBuild(
+  List(
+    semanticdbEnabled := true, // enable SemanticDB
+    semanticdbVersion := scalafixSemanticdb.revision // only required for Scala 2.x
+  )
+)
+
+ThisBuild / scalacOptions += "-P:semanticdb:synthetics:on"
+
 addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1")
--- a/build.sbt
+++ b/build.sbt
@@ -21,13 +21,4 @@ sys.env.getOrElse("JAVA_OPTS", "").split(" ").toSeq.map { opt =>
   javaOptions in reStart += opt
 }
 
-inThisBuild(
-  List(
-    semanticdbEnabled := true, // enable SemanticDB
-    semanticdbVersion := scalafixSemanticdb.revision // only required for Scala 2.x
-  )
-)
-
-ThisBuild / scalacOptions += "-P:semanticdb:synthetics:on"
-
 addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1")
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.4.7
+sbt.version=1.5.0
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.5.0
+sbt.version=1.6.2
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -148,9 +148,8 @@ object Boot extends IOApp with LazyLogging {
           googleFire <- GoogleFirestoreInterpreter.firestore[IO](
             config.googleServicesConfig.serviceAccountCredentialJson.firestoreServiceAccountJsonPath.asString)
           googleStorage <- GoogleStorageInterpreter.storage[IO](
-            config.googleServicesConfig.serviceAccountCredentialJson.defaultServiceAccountJsonPath.asString,
-            Blocker.liftExecutionContext(blockingEc),
-            None)
+            config.googleServicesConfig.serviceAccountCredentialJson.defaultServiceAccountJsonPath.asString
+          )
           googleKmsClient <- GoogleKmsInterpreter.client[IO](config.googleServicesConfig.serviceAccountCredentialJson.defaultServiceAccountJsonPath.asString)
         } yield {
           implicit val loggerIO: StructuredLogger[IO] = Slf4jLogger.getLogger[IO]
@@ -160,8 +159,8 @@ object Boot extends IOApp with LazyLogging {
           // Use resourceNamePrefix to avoid collision between different fiab environments (we share same firestore for fiabs)
           val lock =
             DistributedLock[IO](s"sam-${config.googleServicesConfig.resourceNamePrefix.getOrElse("local")}", appConfig.distributedLockConfig, ioFireStore)
-          val newGoogleStorage = GoogleStorageInterpreter[IO](googleStorage, Blocker.liftExecutionContext(blockingEc), None)
-          val googleKmsInterpreter = GoogleKmsInterpreter[IO](googleKmsClient, blockingEc)
+          val newGoogleStorage = GoogleStorageInterpreter[IO](googleStorage, blockerBound = None)
+          val googleKmsInterpreter = GoogleKmsInterpreter[IO](googleKmsClient)
           val resourceTypeMap = appConfig.resourceTypes.map(rt => rt.name -> rt).toMap
           val cloudExtension = createGoogleCloudExt(
             foregroundAccessPolicyDAO,
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -6,7 +6,7 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import cats.data.NonEmptyList
 import cats.effect
-import cats.effect.{Blocker, ExitCode, IO, IOApp}
+import cats.effect.{ExitCode, IO, IOApp}
 import cats.implicits._
 import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
@@ -61,13 +61,11 @@ class GoogleGroupSyncMonitorSupervisor(
     case Status.Failure(t) => logger.error("error initializing google group sync monitor", t)
   }
 
-  def init = {
-    logger.info(s"Initializing GoogleGroupSyncMonitorActor")
+  def init =
     for {
       _ <- pubSubDao.createTopic(pubSubTopicName)
       _ <- pubSubDao.createSubscription(pubSubTopicName, pubSubSubscriptionName)
     } yield Start
-  }
 
 
 
@@ -128,6 +126,7 @@ class GoogleGroupSyncMonitorActor(
 
   override def receive = {
     case StartMonitorPass =>
+      logger.debug(s"Pulling messages off: $pubSubSubscriptionName")
       // start the process by pulling a message and sending it back to self
       pubSubDao.pullMessages(pubSubSubscriptionName, 1).map(_.headOption) pipeTo self
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
@@ -61,11 +61,15 @@ class GoogleGroupSyncMonitorSupervisor(
     case Status.Failure(t) => logger.error("error initializing google group sync monitor", t)
   }
 
-  def init =
+  def init = {
+    logger.info(s"Initializing GoogleGroupSyncMonitorActor")
     for {
       _ <- pubSubDao.createTopic(pubSubTopicName)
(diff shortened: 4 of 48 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -38,11 +38,36 @@ import org.scalatest.matchers.should.Matchers
 import org.scalatest.{BeforeAndAfterAll, PrivateMethodTester}
 import org.scalatestplus.mockito.MockitoSugar
 
-import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.ExecutionContext.Implicits.{global => globalEc}
 import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.util.{Success, Try}
 
+class SamMockGoogleDirectoryDAO extends MockGoogleDirectoryDAO {
```
s-rubenstein, line 47: This exists because at some point the MockGoogleDirectoryDAO lost the lower method. The strictly correct fix would be to go into workbench libs and make the change there. I want to follow up with that and then go back and change this, because I want to move as fast on the java upgrade as possible, since its chance to go stale is very high.

build.sbt
```diff
@@ -21,4 +21,13 @@ sys.env.getOrElse("JAVA_OPTS", "").split(" ").toSeq.map { opt =>
   javaOptions in reStart += opt
 }
 
+inThisBuild(
```
s-rubenstein: Used to run the scalafix. I can remove if we don't want it still in there.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitorSpec.scala
```diff
@@ -35,7 +35,7 @@ class GoogleGroupSyncMonitorSpec(_system: ActorSystem) extends TestKit(_system)
   }
 
   "GoogleGroupSyncMonitor" should "handle sync message" in {
-    implicit val patienceConfig = PatienceConfig(timeout = 1 second)
+    implicit val patienceConfig = PatienceConfig(timeout = 1 minute)
```
andy7i: why this change?

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
andy7i: are we tracking this anywhere?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -160,8 +159,8 @@ object Boot extends IOApp with LazyLogging {
           // Use resourceNamePrefix to avoid collision between different fiab environments (we share same firestore for fiabs)
           val lock =
             DistributedLock[IO](s"sam-${config.googleServicesConfig.resourceNamePrefix.getOrElse("local")}", appConfig.distributedLockConfig, ioFireStore)
-          val newGoogleStorage = GoogleStorageInterpreter[IO](googleStorage, Blocker.liftExecutionContext(blockingEc), None)
-          val googleKmsInterpreter = GoogleKmsInterpreter[IO](googleKmsClient, blockingEc)
+          val newGoogleStorage = GoogleStorageInterpreter[IO](googleStorage, blockerBound = None)
```
andy7i, line 162: is this going from blocking to non-blocking (`blockerBound = None`)?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
```diff
@@ -124,6 +126,7 @@ class GoogleGroupSyncMonitorActor(
 
   override def receive = {
     case StartMonitorPass =>
+      logger.debug(s"Pulling messages off: $pubSubSubscriptionName")
```
andy7i, line 129: what is this change / how does this change relate to the ticket?

src/test/scala/org/broadinstitute/dsde/workbench/sam/TestSupport.scala
```diff
@@ -49,8 +51,7 @@ trait TestSupport{
   def runAndWait[T](f: IO[T]): T = f.unsafeRunSync()
 
   implicit val futureTimeout = Timeout(Span(10, Seconds))
-  implicit val cs = IO.contextShift(scala.concurrent.ExecutionContext.global)
-  implicit val timer = IO.timer(scala.concurrent.ExecutionContext.global)
+//  implicit val timer = Temporal[IO]
```
andy7i: remove?
```suggestion

```

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -38,11 +38,36 @@ import org.scalatest.matchers.should.Matchers
...
+  override def lockedDownGroupSettings = new Groups()
+    .setWhoCanAdd("ALL_OWNERS_CAN_ADD")
+    .setWhoCanJoin("INVITED_CAN_JOIN")
+    .setWhoCanViewMembership("ALL_MANAGERS_CAN_VIEW")
+    .setWhoCanViewGroup("ALL_OWNERS_CAN_VIEW")
+    .setWhoCanInvite("NONE_CAN_INVITE")
+    .setArchiveOnly(
+      "true"
+    ) // .setWhoCanPostMessage("NONE_CAN_POST") setting archive only is the way to set it so no one can post
```
andy7i, line 56: should `.setWhoCanPostMessage("NONE_CAN_POST")` be in this comment, or should it be on the next line like:
```suggestion
    ) // setting archive only is the way to set it so no one can post
    .setWhoCanPostMessage("NONE_CAN_POST")
```

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.4.7
+sbt.version=1.5.0
```
andy7i: why 1.5.0? / did you try the latest (I think 1.6.2)?

build.sbt
andy7i: I think we should remove it once its usefulness has expired. Is it useful beyond the cats migration?

andy7i (commented):
overall, looks good. Thanks for taking on this monster!

(minor) - might be helpful to have signposts for areas we should review, especially when the PR has 48 files

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
gpolumbo-broad: I'm not sure what the right answer is here, but this might be helpful in figuring it out: https://typelevel.org/cats-effect/docs/migration-guide#blocker

gpolumbo-broad (approved):
All comments/questions I have, @andy7i already asked.  So long as those get addressed: 👍

gpolumbo-broad:
Leaving this link here since I didn't see it elsewhere:  https://typelevel.org/cats-effect/docs/migration-guide might be helpful for this PR and possibly if we look back at this in the future if any bugs crop up as a result.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitorSpec.scala
s-rubenstein: We think there's some inefficiencies that were introduced, and it is hard to tell what is going to be happening on a machine. On my machine it took between .75 and 2 seconds to run, so Edmund and I thought it was worth just bumping to a minute to give it sufficient time to run.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
s-rubenstein: No, its just naming the None which was the last arg on line 163. No change to the behavior, just naming the param

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
s-rubenstein: We added this to test the fix, and since its a debug I think it would be worth leaving there.

project/build.properties
s-rubenstein: I updated to the minimum possible. Because its SBT I would prefer to keep version changes as small as possible in order to reduce risk of changes to our commands - especially while we are already doing a big upgrade.

My general rule is one big upgrade at a time. If we want to upgrade this another major version, we can, but I wouldn't want to do that at the same time as this upgrade in order to keep things as small as possible

s-rubenstein:
@gpolumbo-broad that doc is what I actually followed to do this migration

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
s-rubenstein: https://broadworkbench.atlassian.net/browse/CA-1803

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
s-rubenstein: It should be in this comment - this is copied directly from the GoogleDirectoryDAO implementation file, without any changes.

s-rubenstein:
In general @andy7i I would agree with you on pointing out places to review, but when it comes to a java upgrade (and major library like Cats) I think it is worth taking a look at every little change because that's kind of the point of the upgrade.

If I had to pick places to take extra eyes, the places where I left individual comments are where I would take a look first. Those are where I think there's more trickiness/there was a decision point behind it.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
andy7i: looks like we don't need to name it if we don't want to. No strong preference there

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
andy7i: weird!

project/build.properties
andy7i: feels like we really **really** need something like scala steward

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitorSpec.scala
andy7i: did you consider the possibility that we might miss significant performance issues due to this change?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
s-rubenstein: We definitely don't need to, but both Edmund and I agreed that naming it was more clear - it makes it easier to see what the parameter is at a glance.

andy7i:
> In general @andy7i I would agree with you on pointing out places to review, but when it comes to a java upgrade (and major library like Cats) I think it is worth taking a look at every little change because that's kind of the point of the upgrade.

Were any of the changes made by scalafix? I think we could ignore most changes made by scalafix.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitorSpec.scala
s-rubenstein: Not sure - I don't think its significant because we didn't see other test failures, and the scale was low, but it is something we should keep an eye on for perf tests.

s-rubenstein:
The only changes made by scalafix were the Timer->Temporal and removing the contextshift stuff. The imports were done by hand, which is the majority of the file changes, and should probably be looked at, at least briefly.

project/build.properties
s-rubenstein: I agree

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitorSpec.scala
andy7i: I like the idea of making a change here so that we don't have to remember this / keep an eye on it. So I think I'd prefer a shorter timeout (maybe 5-10 seconds) rather than 1 minute.

andy7i (approved):
1 last comment but not blocking if you think it's not necessary.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitorSpec.scala
s-rubenstein: Makes sense. I'll drop it to 5 seconds, that should be fine

project/build.properties
s-rubenstein: Did you know `sbt build` isn't a command? I didn't, for some reason I thought that was the command to use - either way, the one sbt command that we use in Sam that is in our readme is `sbt test`, which worked.

PR #595, opened 2022-03-10
[DC-243] Add Data Catalog admin role

samanehsan:
Ticket: [DC-243](https://broadworkbench.atlassian.net/browse/DC-243)

In the Data Catalog, if user has access to a dataset in a storage system (such as the Data Repo) they can create/read/update/delete an entry in the catalog. This PR adds an admin role so that catalog team members can perform these actions for users if needed.

---
**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -869,6 +869,29 @@ resourceTypes = {
     }
     reuseIds = false
   }
+  catalog = {
+    actionPatterns = {
+      "create_metadata" = {
+        description = "Can create catalog metadata"
+      },
+      "read_metadata" = {
+        description = "Can read catalog metadata"
+      },
+      "update_metadata" = {
+        description = "Can update catalog metadata"
+      },
+      "delete_metadata" = {
+        description = "Can delete catalog metadata"
+      }
+    }
+    ownerRoleName = "admin"
+    roles = {
+      admin = {
+        roleActions = ["create_metadata", "read_metadata", "update_metadata", "delete_metadata"]
+      }
+    }
+    reuseIds = true
+  }
   datarepo = {
     actionPatterns = {
       create_dataset = {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -874,20 +874,20 @@ resourceTypes = {
       "create_metadata" = {
         description = "Can create catalog metadata"
       },
-      "read_metadata" = {
+      "read_any_metadata" = {
         description = "Can read catalog metadata"
       },
-      "update_metadata" = {
+      "update_any_metadata" = {
         description = "Can update catalog metadata"
       },
-      "delete_metadata" = {
+      "delete_any_metadata" = {
         description = "Can delete catalog metadata"
       }
     }
     ownerRoleName = "admin"
     roles = {
       admin = {
-        roleActions = ["create_metadata", "read_metadata", "update_metadata", "delete_metadata"]
+        roleActions = ["create_metadata", "read_any_metadata", "update_any_metadata", "delete_any_metadata"]
       }
     }
     reuseIds = true


src/main/resources/reference.conf
```diff
@@ -869,6 +869,29 @@ resourceTypes = {
     }
     reuseIds = false
   }
+  catalog = {
+    actionPatterns = {
+      "create_metadata" = {
+        description = "Can create catalog metadata"
+      },
+      "read_metadata" = {
```
pshapiro4broad: These action names aren't exactly accurate in that all users have permission to create/read/update metadata for any dataset based on their permission as reported by the dataset's storage system. These actions are like overrides that ignore the storage system permission setting. It's almost like `read_all_metadata` or `read_any_metadata`.

Note that these actions won't be set on an individual resource but are set on the application (`catalog`) resource.

src/main/resources/reference.conf
```diff
@@ -869,6 +869,29 @@ resourceTypes = {
...
+        description = "Can read catalog metadata"
+      },
+      "update_metadata" = {
+        description = "Can update catalog metadata"
+      },
+      "delete_metadata" = {
```
pshapiro4broad: I'm wondering if we need different permissions for create/update/delete separately. I can't think of a reason now why we would, but I guess there's no great harm in adding them all.

pshapiro4broad (approved):
The change looks OK to me although I have some questions.

src/main/resources/reference.conf
samanehsan: I thought it would be clearer this way, but since that admin role is at the catalog level it may not matter

src/main/resources/reference.conf
samanehsan: That's a good point on the naming!

gpolumbo-broad (approved):
Structurally this looks ok.  I have questions about naming, but without context of your app, I will stay out of that discussion, especially since it looks like @pshapiro4broad has already initiated that conversation with you.

My biggest question is if you meant to create a whole new Resource Type in this PR?  Totally fine if that's what you want/need to do, but just wanted to confirm.  Other than that, I think what you've done here looks fine.

samanehsan:
@gpolumbo-broad What is the alternative to creating a new resource type? I'm only familiar with adding roles to SAM resources. In this case, we will have one catalog resource per deployment environment and catalog team members will have the admin role.

src/main/resources/reference.conf
```diff
@@ -869,6 +869,29 @@ resourceTypes = {
     }
     reuseIds = false
   }
+  catalog = {
```
s-rubenstein, line 872: Currently all the actions have to do with catalog metadata. Could we name this `catalog_metadata`?

src/main/resources/reference.conf
s-rubenstein: In general the pattern for Sam is to make permissions as granular as possible, so from a pattern perspective this is correct for the repo as well.

s-rubenstein:
I think with the security restrictions you have (only ability to change metadata rather than data), this seems like the right approach. Other ways I can think of would be ensuring that the changes could be made by firecloud.org accounts (I am not sure where the metadata is stored, but an admin API might be an option). However, given that Sam is already used in this way by Datarepo, and it is extensible if you wanted to gate admin apis behind these permissions, this approach seems sound to me.

s-rubenstein (approved):
This generally seems fine to me!

src/main/resources/reference.conf
samanehsan: Hmmm since there will only be one of these resources per catalog instance, `catalog` seems clearer to me

samanehsan:
Yes exactly! The metadata is stored in the Data Catalog's Postgres database and is meant to only include public information about datasets, so I think this approach is okay (vs. implementing admin-only endpoints)

src/main/resources/reference.conf
pshapiro4broad: These are settings for the new service `catalog` which will most likely have more resources added in the future. So this is a resource of the same type as `datarepo`, which is why we're using the service name here.

pshapiro4broad:
> However, given that Sam is already used in this way by Datarepo, and it is extensible if you wanted to gate admin apis behind these permissions, this approach seems sound to me.

Note that this isn't a new resource to be used by data repo; it's a new resource that represents a new service (catalog).

PR #599, opened 2022-03-23
[QA-1759] Pragmatic fix for v1 billing project creation failure in tests

ehigham:
RR: https://broadworkbench.atlassian.net/browse/QA-1759
Bypass problematic deployment manager code for creating v1 billing
projects. Use GPAlloc instead. What could possibly go wrong?

The package path of the tests was also wrong and has been for quite a while. I'm not sure if this is intentional but the linter was giving me sad req squiggles so I fixed it + re-optimised imports.

**PR checklist**

### DIFF
--- a/automation/project/Dependencies.scala
+++ b/automation/project/Dependencies.scala
@@ -9,7 +9,7 @@ object Dependencies {
 
   val workbenchGoogleV = "0.21-ae11b9f"
   val workbenchGoogle2V = "0.23-7ddf186"
-  val workbenchServiceTestV = "1.0-01c83001-SNAP"
+  val workbenchServiceTestV = "1.0-d4fe49a1-SNAP"
 
   val excludeWorkbenchModel = ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-model_" + scalaV)
 
--- a/automation/project/Dependencies.scala
+++ b/automation/project/Dependencies.scala
@@ -9,7 +9,7 @@ object Dependencies {
 
   val workbenchGoogleV = "0.21-ae11b9f"
   val workbenchGoogle2V = "0.23-7ddf186"
-  val workbenchServiceTestV = "0.21-258d483"
+  val workbenchServiceTestV = "1.0-01c83001-SNAP"
 
   val excludeWorkbenchModel = ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-model_" + scalaV)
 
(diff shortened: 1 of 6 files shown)


gpolumbo-broad (approved):

automation/project/Dependencies.scala
```diff
@@ -9,7 +9,7 @@ object Dependencies {
 
   val workbenchGoogleV = "0.21-ae11b9f"
   val workbenchGoogle2V = "0.23-7ddf186"
-  val workbenchServiceTestV = "0.21-bdcdd45"
+  val workbenchServiceTestV = "1.0-d4fe49a1-SNAP"
```
andy7i: just checking -- will this be updated to an official version before merging?

automation/src/test/scala/org/broadinstitute/dsde/workbench/sam/api/SamApiSpec.scala
```diff
@@ -130,7 +128,7 @@ class SamApiSpec extends AnyFreeSpec with BillingFixtures with Matchers with Sca
       val userStatus = Sam.user.status()(userAuthToken).get
 
       // user a brand new billing project to ensure known state for pet (not present)
-      withBrandNewBillingProject("new-pet-test") { projectName =>
+      withCleanBillingProject(owner, userEmails = List(anyUser.email)) { projectName =>
```
andy7i: the comment above mentions why we were using a brand new billing project intentionally. Since we've changed that now, we can probably remove the comment

andy7i: since this change may cause other errors (which the comment alludes to), maybe we should leave a comment here as a breadcrumb in case it does cause issues?

On the other hand, I'm also ok with not leaving a comment since we can probably just look at the file history / git blame if/when we need to debug this test. 

On the other other hand, while leaving a comment would add some clutter, it would also save us time/effort in the future if we need to look at the test again and so I think I'm leaning toward adding a comment. 

Any thoughts?

andy7i (approved):
Approved after a couple updates -- probably should use an official version of wb-libs before merging this.

automation/project/Dependencies.scala
ehigham: Yep - waiting for https://github.com/broadinstitute/workbench-libs/pull/940

automation/src/test/scala/org/broadinstitute/dsde/workbench/sam/api/SamApiSpec.scala
ehigham: I'm not sure if this even passes yet! I'll remove the comment and update it when it passes!

PR #630, opened 2022-04-05
[WOR-226] Reject ToS endpoint

marctalbott:
Ticket: [WOR-226](https://broadworkbench.atlassian.net/browse/WOR-226)
* Add endpoint to reject the Terms of Service which will disable the user and remove them from the current Terms of Service group 

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -144,11 +144,23 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
       case Some(_) =>
         for {
           _ <- tosService.acceptTosStatus(userId)
-          enabled <- directoryDAO.isEnabled(userId, samRequestContext) 
+          enabled <- directoryDAO.isEnabled(userId, samRequestContext)
           _ <- if (enabled) registrationDAO.enableIdentity(userId, samRequestContext) else IO.none
           status <- IO.fromFuture(IO(getUserStatus(userId, false, samRequestContext)))
         } yield status
       case None => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Could not accept the Terms of Service. User not found.")))
+    }
+  }
+
+  def rejectTermsOfService(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Option[UserStatus]] = {
+    directoryDAO.loadUser(userId, samRequestContext).flatMap {
+      case Some(_) =>
+        for {
+          _ <- tosService.rejectTosStatus(userId)
+          _ <- registrationDAO.disableIdentity(userId, samRequestContext)
+          status <- IO.fromFuture(IO(getUserStatus(userId, false, samRequestContext)))
+        } yield status
+      case None => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Could not reject the Terms of Service. User not found.")))
     }
   }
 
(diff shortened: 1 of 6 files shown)


aherbst-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -144,14 +144,26 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
       case Some(_) =>
         for {
           _ <- tosService.acceptTosStatus(userId)
-          enabled <- directoryDAO.isEnabled(userId, samRequestContext) 
+          enabled <- directoryDAO.isEnabled(userId, samRequestContext)
```
cahrens, line 147: Why does this method check `directoryDAO.isEnabled`, but `rejectTermsOfService` does not?

cahrens, line 147: Why does this method check `directoryDAO.isEnabled`, but `rejectTermsOfService` does not?

cahrens: Note: I have no intuition for what `directoryDAO` means.

cahrens: Note: I have no intuition for what `directoryDAO` means.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -99,6 +99,24 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
     assert(getTosStatusResult.get, s"getTosStatus(${defaultUser.id}) should get the tos for the user")
   }
 
+  it should "allow a user to reject the ToS" in {
```
cahrens: Nit: The first half of this test is identical to the previous one. I would be tempted to either just delete have a single test point that both accepts and rejects (ie., this one), or share the duplicated between the 2 testpoints.

cahrens: Nit: The first half of this test is identical to the previous one. I would be tempted to either just delete have a single test point that both accepts and rejects (ie., this one), or share the duplicated between the 2 testpoints.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
marctalbott: Good question! The `directoryDAO` refers to Sam's Postgres instance and is the source of truth in Sam. The `registrationDAO` is LDAP/OpenDJ which is essentially a cache of enabled users that service Apache proxies check to determine if a user is enabled in Terra or not. 

We disable a user in the `registrationDAO` if they have not accepted the ToS so that the proxies will reject their requests, but we do not disable them in the `directoryDAO`. We reserve disabling users in the `directoryDAO` for cases where we need to quickly revoke a user's access to Terra, typically if they are a malicious user. In those cases, we _also_ disable them in the `registrationDAO`. 

When a user accepts the ToS, we only want to re-enable them in the `registrationDAO` if they are enabled in the `directoryDAO`. Without this check, a malicious user that had been disabled by a Terra admin would effectively be able to re-enable themselves by accepting the ToS. When rejecting the ToS, we don't need the same check as we want to disable the user in the `registrationDAO` regardless of their enabled/disabled status in the `directoryDAO` as this will prevent them from using Terra until they accept the ToS.

cahrens: Thank you so much-- this was very helpful!

cahrens (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -105,8 +105,6 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
     assert(rejectTosStatusResult.get, s"rejectTosStatus(${defaultUser.id}) should reject the tos for the user")
     val getTosStatusResultRejected = tosServiceEnabled.getTosStatus(defaultUser.id).unsafeRunSync()
     assertResult(expected = false, s"getTosStatus(${defaultUser.id}) should have returned false")(actual = getTosStatusResultRejected.get)
-    val isEnabledLdap = regDAO.isEnabled(defaultUser.id, samRequestContext).unsafeRunSync()
-    assertResult(expected = false, "regDAO.isEnabled should have returned false")(actual = isEnabledLdap)
```
cahrens: How did this pass before if it was at the wrong level?

marctalbott: Users are disabled by default so this was just asserting the default, `rejectTosStatus` didn't change the enabled status in the `registrationDAO`

PR #635, opened 2022-04-11
Better policy queries

dvoet:
Ticket: <Link to Jira ticket>

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
@@ -731,35 +731,42 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
                values (${policyGroupName}, ${policy.email}, ${Instant.now()})""".updateAndReturnGeneratedKey().apply())
   }
 
-  private def directMemberUserEmailsQuery(policy: FullyQualifiedPolicyId) = {
+  private def directMemberUserEmailsQuery(resource: FullyQualifiedResourceId, limitOnePolicy: Option[AccessPolicyName]) = {
     val u = UserTable.syntax("u")
     val p = PolicyTable.syntax("p")
     val gm = GroupMemberTable.syntax("gm")
 
-    samsql"""select ${u.resultAll} from ${PolicyTable as p}
+    samsql"""select ${p.result.name}, ${u.resultAll} from ${PolicyTable as p}
         join ${GroupMemberTable as gm} on ${gm.groupId} = ${p.groupId}
         join ${UserTable as u} on ${u.id} = ${gm.memberUserId}
-        where ${p.name} = ${policy.accessPolicyName}
-        and ${p.resourceId} = (${loadResourcePKSubQuery(policy.resource)})"""
-      .map(UserTable(u))
+        where ${p.resourceId} = (${loadResourcePKSubQuery(resource)})
+        ${limitOnePolicyClause(limitOnePolicy, p)}"""
+      .map(rs => (rs.get[AccessPolicyName](p.resultName.name), UserTable(u)(rs)))
   }
 
-  private def directMemberGroupEmailsQuery(policy: FullyQualifiedPolicyId) = {
+  private def directMemberGroupEmailsQuery(resource: FullyQualifiedResourceId, limitOnePolicy: Option[AccessPolicyName]) = {
     val g = GroupTable.syntax("g")
     val p = PolicyTable.syntax("p")
     val gm = GroupMemberTable.syntax("gm")
     val op = PolicyTable.syntax("op")
 
-    samsql"""select ${g.resultAll} from ${PolicyTable as p}
+    samsql"""select ${p.result.name}, ${g.resultAll} from ${PolicyTable as p}
         join ${GroupMemberTable as gm} on ${gm.groupId} = ${p.groupId}
         join ${GroupTable as g} on ${g.id} = ${gm.memberGroupId}
-        where ${p.name} = ${policy.accessPolicyName}
-        and ${p.resourceId} = (${loadResourcePKSubQuery(policy.resource)})
-        and not exists (select 1 from ${PolicyTable as op} where ${op.groupId} = ${g.id})"""
-      .map(GroupTable(g))
+        where ${p.resourceId} = (${loadResourcePKSubQuery(resource)})
+        and not exists (select 1 from ${PolicyTable as op} where ${op.groupId} = ${g.id})
+        ${limitOnePolicyClause(limitOnePolicy, p)}"""
+      .map(rs => (rs.get[AccessPolicyName](p.resultName.name), GroupTable(g)(rs)))
   }
 
-  private def directMemberPolicyIdAndEmailsQuery(policy: FullyQualifiedPolicyId) = {
+  private def limitOnePolicyClause(limitOnePolicy: Option[AccessPolicyName], p: QuerySQLSyntaxProvider[SQLSyntaxSupport[PolicyRecord], PolicyRecord]) = {
+    limitOnePolicy match {
+      case Some(policyName) => samsqls"and ${p.name} = ${policyName}"
+      case None => samsqls""
+    }
+  }
+
+  private def directMemberPolicyIdAndEmailsQuery(resource: FullyQualifiedResourceId, limitOnePolicy: Option[AccessPolicyName]) = {
     val mp = PolicyTable.syntax("mp") // member policy
     val mpr = ResourceTable.syntax("mpr") // member policy resource
     val mprt = ResourceTypeTable.syntax("mprt") // member policy resource type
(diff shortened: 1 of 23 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -731,6 +731,58 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+        join ${PolicyTable as mp} on ${mp.groupId} = ${gm.memberGroupId}
+        join ${GroupTable as mpg} on ${mpg.id} = ${mp.groupId}
+        join ${ResourceTable as mpr} on ${mpr.id} = ${mp.resourceId}
+        join ${ResourceTypeTable as mprt} on ${mprt.id} = ${mpr.resourceTypeId}
+        where ${p.name} = ${policy.accessPolicyName}
+        and ${p.resourceId} = (${loadResourcePKSubQuery(policy.resource)})"""
+        .map(rs => PolicyIdAndEmail(rs.get[AccessPolicyName](mp.resultName.name), rs.get[WorkbenchEmail](mpg.resultName.email), rs.get[ResourceTypeName](mprt.resultName.name), rs.get[ResourceId](mpr.resultName.name))).list().apply().to(LazyList)
+    })
+  }
+
```
dvoet: the new queries

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -468,28 +468,17 @@ class ResourceService(
...
-        accessPolicyDAO.loadPolicy(resourceAndPolicyName, samRequestContext).map { accessPolicyOption =>
-          accessPolicyOption.map { accessPolicy =>
-            PolicyIdAndEmail(resourceAndPolicyName.accessPolicyName, accessPolicy.email,
-              resourceAndPolicyName.resource.resourceTypeName, resourceAndPolicyName.resource.resourceId)
-          }
-        }
-      }
+      userEmails <- accessPolicyDAO.loadDirectMemberUserEmails(policy.id, samRequestContext)
+      groupEmails <- accessPolicyDAO.loadDirectMemberGroupEmails(policy.id, samRequestContext)
+      policiesAndEmails <- accessPolicyDAO.loadDirectMemberPolicyIdAndEmails(policy.id, samRequestContext)
```
dvoet: use the new queries

davidangb (approved):
lgtm - does the existing test coverage exercise this enough?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -932,75 +980,103 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+      val policyInfos = policyInfoQuery(resource, limitOnePolicy)
+
+      policyInfos.map { case (policyInfo, resultsByPolicy) =>
+        val (policyRoles, policyActions, policyDescendantPermissions) = unmarshalPolicyPermissions(resultsByPolicy.map(_._2))
+
+        val policyId = FullyQualifiedPolicyId(resource, policyInfo.name)
+        // queries to get all members as WorkbenchSubjects, there are 3 kinds, groups, users and policies
+        val memberGroups = directMemberGroupEmailsQuery(policyId).list().apply().map(_.name).toSet[WorkbenchSubject]
+        val memberUsers = directMemberUserEmailsQuery(policyId).list().apply().map(_.id).toSet[WorkbenchSubject]
+        val memberPolicies = directMemberPolicyIdAndEmailsQuery(policyId).list().apply().map(p => FullyQualifiedPolicyId(FullyQualifiedResourceId(p.resourceTypeName, p.resourceId), p.policyName)).toSet[WorkbenchSubject]
```
dvoet: The gist of this change is the highlighted code. The new code does a query for policy info that does not include membership info then does a follow on query for each of the 3 types of members. This means 1 query to get all the policies then 3 queries per policy. Now that I think about it though, if we like this, this could easily change to just 3 additional queries instead of 3n.

Contrast this to the previous state that did all this in 1 query with 5 more left joins and complicated handling of results. Complicated enough that I, who knows what the query is doing, was nervous touching it or adding to it to get the new information required by the use case.

From a performance perspective, the new code is probably worse but probably on the order of 10 to 20 ms (traces show the existing query takes ~3 ms loading a single policy). But I think we gain more maintainable code. And I think this approach is more stable as in I can think of some pathological examples that would cause some cartesian product discomfort in the existing query.

The other benefit is that the same queries to get policy info in one way are reused to get policy info in the form that the use case requires.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -932,75 +980,103 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+        // policies have extra information, users and groups are just emails
+        val memberGroups = directMemberGroupEmailsQuery(policyId).list().apply().map(_.email)
+        val memberUsers = directMemberUserEmailsQuery(policyId).list().apply().map(_.email)
+        val memberPolicies = directMemberPolicyIdAndEmailsQuery(policyId).list().apply().toSet
 
-        AccessPolicy(FullyQualifiedPolicyId(FullyQualifiedResourceId(policyInfo.resourceTypeName, policyInfo.resourceId), policyInfo.name), policyMembers, policyInfo.email, policyRoles, policyActions, policyDescendantPermissions, policyInfo.public)
+        AccessPolicyWithMembership(policyInfo.name, AccessPolicyMembership(memberPolicies.map(_.policyEmail) ++ memberUsers ++ memberGroups, policyActions, policyRoles, Option(policyDescendantPermissions), Option(memberPolicies)), policyInfo.email)
       }.to(LazyList)
     })
   }
```
davidangb, line 1116: `listPolicies` and `listPoliciesMemberships` are very similar - they differ only in how they map the `directMemberGroupEmailsQuery `, `directMemberUserEmailsQuery `, `directMemberUserEmailsQuery` results into their respective response objects. Is it worth factoring out a shared/common method here?

dvoet: yes, I have done so locally as well as fixing the 3n query problem, but I don't want to muddy my comment above with another push yet

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
dvoet: also note that the existing queries was often followed by queries to load member email addresses which is not required in the proposed code

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -224,6 +224,9 @@ consistent "has a" relationship is tracked by this ticket: https://broadworkbenc
                                                 memberPolicies: Option[Set[PolicyIdAndEmail]] = Option(Set.empty)) {
   def getDescendantPermissions: Set[AccessPolicyDescendantPermissions] = descendantPermissions.getOrElse(Set.empty)
 }
+// AccessPolicyWithMembership is practically the same as AccessPolicyResponseEntry but the latter is used in api responses
+// and the former is used at the DAO level so it seems better to keep them separate
+@Lenses final case class AccessPolicyWithMembership(policyName: AccessPolicyName, membership: AccessPolicyMembership, email: WorkbenchEmail)
```
gpolumbo-broad, line 229: Out of context, your decision here is fine by me.  In context though, we now have 5 (not counting AccessPolicyDescendantPermissions which is a different thing) different representations of an AccessPolicy.  It's confusing and if we can avoid adding more, that's my preference.  But if it's a can of worms, then maybe leave this as is and at least create a tech debt ticket to come back and try to harmonize all these representations and see if we can simplify.

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAOSpec.scala
```diff
@@ -1509,6 +1512,50 @@ class PostgresAccessPolicyDAOSpec extends AnyFreeSpec with Matchers with BeforeA
...
+        val reader = AccessPolicy(FullyQualifiedPolicyId(resource.fullyQualifiedId, AccessPolicyName("reader")), Set(owner.id, defaultUser.id), WorkbenchEmail("reader@policy.com"), resourceType.roles.map(_.roleName), Set(readAction, writeAction), Set.empty, false)
+
+        dao.createResourceType(resourceType, samRequestContext).unsafeRunSync()
+        dao.createResource(resource, samRequestContext).unsafeRunSync()
+        dao.createPolicy(owner, samRequestContext).unsafeRunSync()
+        dao.createPolicy(reader, samRequestContext).unsafeRunSync()
+
+        val ownerPolicyWithMembership = AccessPolicyWithMembership(owner.id.accessPolicyName, AccessPolicyMembership(Set(defaultUser.email, defaultGroup.email), owner.actions, owner.roles, Option(owner.descendantPermissions), Option(Set.empty)), owner.email)
+        val readerPolicyWithMembership = AccessPolicyWithMembership(reader.id.accessPolicyName, AccessPolicyMembership(Set(owner.email, defaultUser.email), reader.actions, reader.roles, Option(reader.descendantPermissions), Option(Set(PolicyIdAndEmail(owner.id.accessPolicyName, owner.email, owner.id.resource.resourceTypeName, owner.id.resource.resourceId)))), reader.email)
```
gpolumbo-broad, line 1530: Case in point of the different representations of Access Policies, needing to massage the `owner` and `reader` policies you already have into another shape is annoying and possibly error prone.  It's a ticket for another day to address this.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -1351,9 +1351,9 @@ class ResourceServiceSpec extends AnyFlatSpec with Matchers with ScalaFutures wi
       newPolicy <- policyDAO.createPolicy(AccessPolicy(
         FullyQualifiedPolicyId(res2.fullyQualifiedId, AccessPolicyName("foo")), Set(testGroup.id, dummyUserInfo.userId, FullyQualifiedPolicyId(res1.fullyQualifiedId, testPolicy.policyName)), WorkbenchEmail("a@b.c"), Set.empty, Set.empty, Set.empty, public = false), samRequestContext)
 
-      membership <- service.loadAccessPolicyWithEmails(newPolicy, samRequestContext)
+      membership <- policyDAO.loadPolicyMembership(newPolicy.id, samRequestContext)
     } yield {
-      membership.memberEmails should contain theSameElementsAs Set(
+      membership.get.memberEmails should contain theSameElementsAs Set(
```
gpolumbo-broad: nit: Using `OptionValues` might produce a slightly more helpful test failure message here?
```suggestion
      membership.value.memberEmails should contain theSameElementsAs Set(
```

gpolumbo-broad (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAOSpec.scala
dvoet: I agree... I tried but the roots were too deep and some things touch to api surface which I did not want to do.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
dvoet: today I learned

PR #675, opened 2022-04-26
POL-228 stop calling opendj on user info calls

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/POL-228

In `/register/user/v2/self/info`, instead of calling opendj to see if the user is in the enabled users group, return true if the user has accepted TOS and their record is enabled.

Also use a more efficient query in TOS service to check if the TOS group exists in postgres.

And some renames

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -172,11 +172,26 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
     directoryDAO.loadUser(userId, samRequestContext).flatMap {
       case Some(user) =>
         for {
-          tosStatus <- getTermsOfServiceStatus(user.id, samRequestContext)
+          tosStatus <- getTermsOfServiceStatusInternal(user, samRequestContext)
           adminEnabled <- directoryDAO.isEnabled(user.id, samRequestContext)
-        } yield Some(UserStatusInfo(user.id.value, user.email.value, tosStatus.getOrElse(true) && adminEnabled, adminEnabled))
+        } yield {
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -172,9 +172,9 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
     directoryDAO.loadUser(userId, samRequestContext).flatMap {
       case Some(user) =>
         for {
-          ldapStatus <- registrationDAO.isEnabled(user.id, samRequestContext)
+          tosStatus <- getTermsOfServiceStatus(user.id, samRequestContext)
           adminEnabled <- directoryDAO.isEnabled(user.id, samRequestContext)
-        } yield Some(UserStatusInfo(user.id.value, user.email.value, ldapStatus, adminEnabled))
+        } yield Some(UserStatusInfo(user.id.value, user.email.value, tosStatus.getOrElse(true) && adminEnabled, adminEnabled))
```
Qi77Qi: if `tosStatus` is None, we default it to `true`?

This seems to require all users to be `adminEnabled`?

dvoet: Yes to the first because if tosStatus is None that means tos is disabled and should be ignored

Yes to the second, `adminEnabled` means they have been enabled by an admin, or rather, have not been disabled by an admin. All new users start our with this true.

dvoet:
jenkins retest - unrelated fiab start failure

rtitle (approved):
Looks good to me - ToS is not enabled in all envs yet but it seems like this logic should handle that.

dvoet:
jenkins retest - weird NPE in sbt

marctalbott (approved):

PR #711, opened 2022-04-28
[IA-3326] add Service Account User role to PET only

gpcarr:
Tested in a fiab:

Screen Shot 2022-05-12 at 12 15 58 PM

Worked with devops to give the dev and QA sam service account the `roles/iam.serviceAccountAdmin` role to be able to update a service account (it only had permission to create service accounts before). Here's the terraform PR to do that: https://github.com/broadinstitute/terraform-ap-deployments/pull/646
Once this sam PR is merged, I plan on terraforming higher envs to give the sam SA that same role

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -322,6 +322,7 @@ class GoogleExtensions(
               // See: https://broadworkbench.atlassian.net/browse/CA-1005
               googleDirectoryDAO.addServiceAccountToGroup(proxyEmail, sa)
             }))
+            _ <- IO.fromFuture(IO(googleIamDAO.addServiceAccountUserRoleForUser(project, sa.email, sa.email)))
           } yield sa
         // SA already exists in google, use it
         case Some(sa) => IO.pure(sa)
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -322,6 +323,7 @@ class GoogleExtensions(
               // See: https://broadworkbench.atlassian.net/browse/CA-1005
               googleDirectoryDAO.addServiceAccountToGroup(proxyEmail, sa)
             }))
+            _ <- IO.fromFuture(IO(googleIamDAO.addIamRoles(project, sa.email, MemberType.ServiceAccount, Set("roles/iam.serviceAccountUser"), retryIfGroupDoesNotExist = true)))
```
rtitle: Posted in slack, but I'm wondering if `addIamRoles` is the right method. I think this will give the pet SA `Service Account User` at the _project_ level, which lets the pet SA impersonate any SA in the project (which is not what we want).

[This method](https://github.com/broadinstitute/workbench-libs/blob/develop/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala#L293) to me seems like what we want -- it applies the `Service Account User` role to a SA (not project) with a given target (which happens to be the same SA in this case).

I'm curious if the latter approach worked with NextFlow?

gpcarr: oh that makes sense why i've been seeing the 403! Yes I want to add the `Service Account User` role to only the SA not at the project level. I'll update and test out the `addServiceAccountUserRoleForUser` method!

gpcarr: I had originally used `addServiceAccountUserRoleForUser` and it didn't work but i didn't look at the sam logs. testing it out again while looking at sam logs to see what's going on. waiting on fiab-start

gpcarr:
jenkins retest

gpcarr:
unrelated leo test failure - jenkins retest

andy7i (approved):
In terms of order of operations, I think we should make sure that the terraform changes are in all envs _before_ this is merged. (I took a quick look at the [terraform PR](https://github.com/broadinstitute/terraform-ap-deployments/pull/646) and I _think_ it has already set up broad-dsde-dev, qa, and prod so we should be all set on that front)

Haven't tested these changes but assuming that the terraform changes are already on all envs 👍

gpcarr:
confirmed that terraform changes are on all envs: https://github.com/broadinstitute/terraform-ap-deployments/pull/648 going to go ahead and merge this pr

PR #721, opened 2022-05-06
refactor UserInfo and WorkbenchUser to SamUser

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/POL-235
This is the first step to handling the enabled user check in Sam. I want to handle the check in the `requireUserInfo` directive which already loads the user. The problem is that the `WorkbenchUser` does not include the `enabled` field. Adding it means changing workbench-libs which makes me want to jump out the window. And `WorkbenchUser` is not even exposed outside of Sam so why is it in a shared lib? So I made `SamUser`. The next problem is that most of the contents of `SamUser` is dropped on the floor when returning a `UserInfo` object. The special fields in `UserInfo` (access token and expiration) are not even used in Sam. So return `SamUser` instead. The rest is updating tests to excise all uses of `UserInfo` and `WorkbenchUser`. Also centralize construction of `SamUser`s in the Generator functions.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
@@ -13,6 +13,7 @@ import org.broadinstitute.dsde.workbench.model.google.ServiceAccountSubjectId
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.sam.api.StandardUserInfoDirectives._
 import org.broadinstitute.dsde.workbench.sam.dataAccess.{DirectoryDAO, RegistrationDAO}
+import org.broadinstitute.dsde.workbench.sam.model.SamUser
 import org.broadinstitute.dsde.workbench.sam.service.UserService._
 import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
 
@@ -23,24 +24,25 @@ import scala.util.matching.Regex
 trait StandardUserInfoDirectives extends UserInfoDirectives with LazyLogging with SamRequestContextDirectives {
   implicit val executionContext: ExecutionContext
 
-  def requireUserInfo(samRequestContext: SamRequestContext): Directive1[UserInfo] = requireOidcHeaders.flatMap { oidcHeaders =>
+  def requireUserInfo(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.flatMap { oidcHeaders =>
     onSuccess {
       getUserInfo(directoryDAO, registrationDAO, oidcHeaders, samRequestContext).unsafeToFuture()
     }
   }
 
-  def requireCreateUser(samRequestContext: SamRequestContext): Directive1[WorkbenchUser] = requireOidcHeaders.map(buildWorkbenchUser)
+  def requireCreateUser(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.map(buildWorkbenchUser)
 
-  private def buildWorkbenchUser(oidcHeaders: OIDCHeaders): WorkbenchUser = {
+  private def buildWorkbenchUser(oidcHeaders: OIDCHeaders): SamUser = {
     // google id can either be in the external id or google id from azure headers, favor the external id as the source
     val googleSubjectId = (oidcHeaders.externalId.left.toOption ++ oidcHeaders.googleSubjectIdFromAzure).headOption
     val azureB2CId = oidcHeaders.externalId.toOption // .right is missing (compared to .left above) since Either is Right biased
 
-    WorkbenchUser(
+    SamUser(
       genWorkbenchUserId(System.currentTimeMillis()),
       googleSubjectId,
       oidcHeaders.email,
-      azureB2CId)
+      azureB2CId,
+      false)
   }
 
   /**
@@ -49,22 +51,8 @@ trait StandardUserInfoDirectives extends UserInfoDirectives with LazyLogging wit
   private def requireOidcHeaders: Directive1[OIDCHeaders] = {
     (headerValueByName(accessTokenHeader).as(OAuth2BearerToken) &
       externalIdFromHeaders &
-      expiresInFromHeader &
       headerValueByName(emailHeader).as(WorkbenchEmail) &
       optionalHeaderValueByName(googleIdFromAzureHeader).map(_.map(GoogleSubjectId))).as(OIDCHeaders)
-  }
-
-  private def expiresInFromHeader: Directive1[Long] = {
-    // gets expiresInHeader as a string and converts it to Long raising an exception if it can't
-    optionalHeaderValueByName(expiresInHeader).flatMap {
-      case Some(expiresInString) =>
-        Try(expiresInString.toLong).fold(
-          t => failWith(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"expiresIn $expiresInString can't be converted to Long", t))).toDirective,
-          expiresInLong => provide(expiresInLong)
-        )
-
-      case None => provide(0)
-    }
   }
 
   private def externalIdFromHeaders: Directive1[Either[GoogleSubjectId, AzureB2CId]] = headerValueByName(userIdHeader).map { idString =>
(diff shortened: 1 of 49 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -242,6 +242,16 @@ consistent "has a" relationship is tracked by this ticket: https://broadworkbenc
 
 @Lenses final case class GroupSyncResponse(lastSyncDate: String, email: WorkbenchEmail)
 
+final case class SamUser(id: WorkbenchUserId,
+                         googleSubjectId: Option[GoogleSubjectId],
+                         email: WorkbenchEmail,
+                         azureB2CId: Option[AzureB2CId],
+//                         acceptedToS: Option[Boolean], // None means ToS acceptance not required (disabled or grace period)
```
dvoet, line 249: including `acceptedToS` in SamUser is an idea but I am not sure I like it

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -126,17 +113,19 @@ object StandardUserInfoDirectives {
...
-          IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden, s"Google Id $googleSubjectId not found in sam")))
+      maybeSamUser <- directoryDAO.loadUserByGoogleSubjectId(googleSubjectId, samRequestContext)
+      samUser <- maybeSamUser match {
+        case Some(samUser) => IO.pure(samUser)
+        case None => directoryDAO.loadSubjectFromGoogleSubjectId(googleSubjectId, samRequestContext).flatMap {
+          case Some(_) =>
+            IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"subjectId $googleSubjectId is not a WorkbenchUser")))
+          case None =>
+            IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden, s"Google Id $googleSubjectId not found in sam")))
+        }
```
dvoet, line 126: The old block of code called loadSubjectFromGoogleSubjectId, which queries both pet and user tables and could return either a pet id or user id, then throws an exception in all cases other than user id.

The new block of code looks up a user (because we want the full user object) and if it doesn't find one does a follow up query to see if there is a pet before decide which exception to throw.

dvoet: actually, the follow up query is unnecessary because the calling code already handles pets... I will remove it in the next iteration

dvoet:
jenkins retest - dependency issue

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
rtitle: +1, seems the pet case is already handled before this… so maybe we don't need the second `loadSubjectFromGoogleSubjectId` query and can throw a generic error if the user is not found.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -23,24 +24,25 @@ import scala.util.matching.Regex
 trait StandardUserInfoDirectives extends UserInfoDirectives with LazyLogging with SamRequestContextDirectives {
   implicit val executionContext: ExecutionContext
 
-  def requireUserInfo(samRequestContext: SamRequestContext): Directive1[UserInfo] = requireOidcHeaders.flatMap { oidcHeaders =>
+  def requireActiveUser(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.flatMap { oidcHeaders =>
```
rtitle, line 27: I didn't see where `SamUser.enabled` is actually checked -- am I missing it or is that for a subsequent PR?

rtitle (approved):
Went through this -- the refactoring + renaming looks good to me. +1 to having Sam control the `SamUser` object itself instead of wb-libs.

rtitle:
@zykonda FYI about this refactoring of `WorkbenchUser` -> `SamUser`

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: subsequent PR... focusing on 1 thing at a time.

dvoet:
jenkins retest - dependency issue

dvoet:
jenkins retest - dependency issue

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserInfoDirectives.scala
```diff
@@ -24,13 +24,13 @@ trait UserInfoDirectives {
   val cloudExtensions: CloudExtensions
   val termsOfServiceConfig: TermsOfServiceConfig
 
-  def requireUserInfo(samRequestContext: SamRequestContext): Directive1[UserInfo]
+  def requireActiveUser(samRequestContext: SamRequestContext): Directive1[SamUser]
```
gpolumbo-broad, line 27: I really like this renaming!

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -242,6 +242,16 @@ consistent "has a" relationship is tracked by this ticket: https://broadworkbenc
 
 @Lenses final case class GroupSyncResponse(lastSyncDate: String, email: WorkbenchEmail)
 
+final case class SamUser(id: WorkbenchUserId,
```
zykonda, line 245: Are you planning to move Pet Service Accounts to SamModel as well? What about Workbench Identities? Asking because I planned to add Azure managed identities to WB lib.

dvoet: I am not planning anything else at the moment

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesSpec.scala
```diff
@@ -25,15 +24,13 @@ import spray.json.{JsBoolean, JsValue}
   */
 class ResourceRoutesSpec extends AnyFlatSpec with Matchers with ScalatestRouteTest with TestSupport with AppendedClues {
 
-  val defaultUserInfo = UserInfo(OAuth2BearerToken("accessToken"), WorkbenchUserId("user1"), WorkbenchEmail("user1@example.com"), 0)
-  def defaultGoogleSubjectId = Generator.genGoogleSubjectId.sample
+  val defaultUserInfo = TestSamRoutes.defaultUserInfo
```
gpolumbo-broad, line 27: nit:

```suggestion
  val defaultSamUser = TestSamRoutes.defaultSamUser
```

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
```diff
@@ -459,82 +449,81 @@ class UserServiceSpec extends AnyFlatSpec with Matchers with TestSupport with Mo
...
     }.errorReport.statusCode shouldBe Some(StatusCodes.BadRequest)
   }
 
-  it should "return conflict when there's an existing subject for a given userId" in{
-    val user = genInviteUser.sample.get
-    val email = genNonPetEmail.sample.get
-    dirDAO.createUser(WorkbenchUser(user.inviteeId, None, email, None), samRequestContext).unsafeRunSync()
-    val res = service.inviteUser(user, samRequestContext).attempt.unsafeRunSync().swap.toOption.get.asInstanceOf[WorkbenchExceptionWithErrorReport]
-    Eq[WorkbenchExceptionWithErrorReport].eqv(res, new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"identity with id ${user.inviteeId} already exists"))) shouldBe true
-  }
```
andy7i, line 491: why don't we need a test like this anymore?

andy7i (commented):
Looks like a really useful refactor!

is there a corresponding wb-libs PR to remove WorkbenchUser?

(Commenting since I know Greg is also reviewing)

gpolumbo-broad (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
dvoet: I changed the `inviteUser` from taking an object that included both user id and email to just taking email so this case became impossible.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesSpec.scala
dvoet: yeah, it is tough to get all the renamings

dvoet:
@andy7i 
> is there a corresponding wb-libs PR to remove WorkbenchUser?

no, I am not certain it is unused

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesSpec.scala
dvoet: I will hit this with a different PR if that is ok

PR #722, opened 2022-05-09
POL-236 Enabled user check

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/POL-236
recommend ignoring whitespace

- refactored routes a little to centralize calls to `withSamRequestContext` and `requireActiveUser` in /api routes
- implemented terms of service check and enabled users tests
- a litte more renaming
- tests

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -26,11 +26,28 @@ trait StandardUserInfoDirectives extends UserInfoDirectives with LazyLogging wit
 
   def requireActiveUser(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.flatMap { oidcHeaders =>
     onSuccess {
-      getUserInfo(directoryDAO, registrationDAO, oidcHeaders, samRequestContext).unsafeToFuture()
+      val requireActiveUserIO = for {
+        user <- getSamUser(directoryDAO, registrationDAO, oidcHeaders, samRequestContext)
+        tosStatus <- tosService.getTosStatus(user.id, samRequestContext)
```
dvoet: Note that this is an extra db call per api call to sam but it should be a few millis. This can be wedged into an existing call but let's not optimize prematurely.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -114,18 +131,8 @@ object StandardUserInfoDirectives {
   }
 
   private def lookUpByGoogleSubjectId(googleSubjectId: GoogleSubjectId, directoryDAO: DirectoryDAO, samRequestContext: SamRequestContext): IO[SamUser] =
-    for {
-      maybeSamUser <- directoryDAO.loadUserByGoogleSubjectId(googleSubjectId, samRequestContext)
-      samUser <- maybeSamUser match {
-        case Some(samUser) => IO.pure(samUser)
-        case None => directoryDAO.loadSubjectFromGoogleSubjectId(googleSubjectId, samRequestContext).flatMap {
-          case Some(_) =>
-            IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"subjectId $googleSubjectId is not a WorkbenchUser")))
```
dvoet, line 123: this case will never happen because the calling code already checks for pets. So simplify by not having code that does an unnecessary check.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -77,10 +77,7 @@ class TosService (val directoryDao: DirectoryDAO, val registrationDao: Registrat
     */
   def getTosStatus(user: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Option[Boolean]] = {
     if (tosConfig.enabled) {
-      getTosGroupName(samRequestContext).flatMap {
-        case Some(groupName) => directoryDao.isGroupMember(groupName, user, samRequestContext).map(Option(_))
-        case None => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Terms of Service group ${getGroupNameString()} not found.")))
-      }
+      directoryDao.isGroupMember(getGroupName(), user, samRequestContext).map(Option(_))
```
dvoet: The changes in this file are to support this change. We don't need to see if the ToS group exists all the time. It is enough to check when Sam starts and whenever someone accepts or reject the terms. This saves an extra db round trip on every api call.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -140,59 +140,34 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
   def acceptTermsOfService(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Option[UserStatus]] = {
-    directoryDAO.loadUser(userId, samRequestContext).flatMap {
-      case Some(_) =>
-        for {
-          _ <- tosService.acceptTosStatus(userId, samRequestContext)
-          enabled <- directoryDAO.isEnabled(userId, samRequestContext)
-          _ <- if (enabled) registrationDAO.enableIdentity(userId, samRequestContext) else IO.none
-          status <- IO.fromFuture(IO(getUserStatus(userId, false, samRequestContext)))
-        } yield status
-      case None => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Could not accept the Terms of Service. User not found.")))
```
dvoet, line 151: this case cannot happen because the routes have already required an active user

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -140,59 +140,34 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
 
   def rejectTermsOfService(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Option[UserStatus]] = {
-    directoryDAO.loadUser(userId, samRequestContext).flatMap {
-      case Some(_) =>
-        for {
-          _ <- tosService.rejectTosStatus(userId, samRequestContext)
-          _ <- registrationDAO.disableIdentity(userId, samRequestContext)
-          status <- IO.fromFuture(IO(getUserStatus(userId, false, samRequestContext)))
-        } yield status
-      case None => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Could not reject the Terms of Service. User not found.")))
```
dvoet, line 163: this case cannot happen because the routes have already required an active user

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -140,59 +140,34 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
-    }
-  }
-
-  def getTermsOfServiceStatus(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Option[Boolean]] = {
-    tosService.getTosStatus(userId, samRequestContext)
```
dvoet, line 168: inlined this function

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -140,59 +140,34 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
         }
       case None => IO.pure(None)
     }
 
-  /**
-    * If grace period enabled, don't check ToS, return true
-    * If ToS disabled, return true
-    * Otherwise return true if user has accepted ToS
-    */
-  private def getTermsOfServiceStatusInternal(user: SamUser, samRequestContext: SamRequestContext) = {
```
dvoet, line 188: moved to TosService

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -114,18 +131,8 @@ object StandardUserInfoDirectives {
...
-          case None =>
-            IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden, s"Google Id $googleSubjectId not found in sam")))
-        }
-      }
-    } yield samUser
+    directoryDAO.loadUserByGoogleSubjectId(googleSubjectId, samRequestContext).map(
+      _.getOrElse(throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden, s"Google Id $googleSubjectId not found in sam"))))
```
rtitle: I think this works fine but I like to avoid `throw` inside IO map/flatMap because it's a side effect. Could do this instead:
```
    directoryDAO.loadUserByGoogleSubjectId(googleSubjectId, samRequestContext).flatMap { user =>
      IO.fromOption(user)(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden, s"Google Id $googleSubjectId not found in sam")))
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -75,15 +75,25 @@ class TosService (val directoryDao: DirectoryDAO, val registrationDao: Registrat
...
     } else IO.none
   }
 
+  /**
+    * If grace period enabled, don't check ToS, return true
+    * If ToS disabled, return true
+    * Otherwise return true if user has accepted ToS
+    */
+  def isTermsOfServiceStatusAcceptable(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Boolean] = {
+    if (tosConfig.isGracePeriodEnabled) {
```
rtitle: What is the grace period, and when would this be true?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -75,15 +75,25 @@ class TosService (val directoryDao: DirectoryDAO, val registrationDao: Registrat
     *         IO[Some(false)] if ToS is enabled and the user hasn't accepted
     *         IO[None] if ToS is disabled
     */
-  def getTosStatus(user: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Option[Boolean]] = {
+  def getTosStatus(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Option[Boolean]] = {
```
rtitle, line 37: This was preexisting so don't need to change anything, but the fact that these return `Option[Boolean]` is a bit strange to me. Eventually ToS will be enabled everywhere so will we then rip out the `Option`?

Generally I prefer clean types in the service layer and have the call site deal with whether a thing is enabled or not.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -26,11 +26,28 @@ trait StandardUserInfoDirectives extends UserInfoDirectives with LazyLogging wit
 
   def requireActiveUser(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.flatMap { oidcHeaders =>
     onSuccess {
-      getUserInfo(directoryDAO, registrationDAO, oidcHeaders, samRequestContext).unsafeToFuture()
+      val requireActiveUserIO = for {
+        user <- getSamUser(directoryDAO, registrationDAO, oidcHeaders, samRequestContext)
```
rtitle, line 30: This will generate 2 DB queries for every `/api` call, right? Was wondering if we could do it in 1 query like:
```
select * from sam_user u where exists (select * from sam_group where name = 'tos' and user_id = u.id)
```
Would save a DB round trip. Just an idea, I don't feel too strongly (the queries are probably quick).

rtitle (approved):
A few comments above, overall seems correct.

This probably needs a JIRA ticket + compliance stuff (security impact, etc).

Is removing the apache check next? Happy to review that too :)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
dvoet: It is a config setting. I think it means that users asked to accept the new ToS but Terra is not yet enforcing it.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
dvoet: I don't disagree. I think one use case may be the Verily deployment where ToS may never be used but I don't know.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: [see this prior comment](https://github.com/broadinstitute/sam/pull/722#discussion_r868471261)

I am also not super thrilled that ToS is implemented as a group and not a first class thing. Including ToS in the user lookup query would mean custom code paths to load users by pet, google id or azure id. And these would need to take in the name of the ToS group. It would probably be better if accepted ToS version were just a column on the sam_user table.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamRoutes.scala
```diff
@@ -55,8 +55,19 @@ abstract class SamRoutes(
...
       termsOfServiceRoutes ~
       withExecutionContext(ExecutionContext.global) {
-        pathPrefix("register") { userRoutes } ~
-        pathPrefix("api") { resourceRoutes ~ adminUserRoutes ~ extensionRoutes ~ groupRoutes ~ apiUserRoutes }
+        withSamRequestContext { samRequestContext =>
+          pathPrefix("register") { userRoutes(samRequestContext) } ~
+          pathPrefix("api") {
+            // IMPORTANT - all routes under /api must have an active user
+            requireActiveUser(samRequestContext) { samUser =>
+              resourceRoutes(samUser, samRequestContext) ~
```
gpolumbo-broad, line 63: super nit: indentation

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -26,11 +26,28 @@ trait StandardUserInfoDirectives extends UserInfoDirectives with LazyLogging wit
...
+        tosStatusAcceptable <- tosService.isTermsOfServiceStatusAcceptable(user.id, samRequestContext)
+      } yield {
+        val permittedToAccessTerra = tosStatusAcceptable && user.enabled
+        if (permittedToAccessTerra) {
+          user
+        } else {
+          throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Unauthorized, "User is disabled or has not accepted the terms of service."))
```
gpolumbo-broad: Can we send back the specific failure?  Would add a little more logic here, but I feel like it would be a better user experience to tell them specifically whether their user account is disabled or they need to accept the ToS.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -26,11 +26,28 @@ trait StandardUserInfoDirectives extends UserInfoDirectives with LazyLogging wit
...
+  def requireUserAllowInactive(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.flatMap { oidcHeaders =>
+    onSuccess {
+      getSamUser(directoryDAO, registrationDAO, oidcHeaders, samRequestContext).unsafeToFuture()
     }
   }
 
-  def requireCreateUser(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.map(buildWorkbenchUser)
+  def withNewUser(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.map(buildWorkbenchUser)
 
   private def buildWorkbenchUser(oidcHeaders: OIDCHeaders): SamUser = {
```
gpolumbo-broad:
```suggestion
  private def buildSamUser(oidcHeaders: OIDCHeaders): SamUser = {
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
```diff
@@ -26,11 +26,28 @@ trait StandardUserInfoDirectives extends UserInfoDirectives with LazyLogging wit
...
+          user
+        } else {
+          throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Unauthorized, "User is disabled or has not accepted the terms of service."))
+        }
+      }
+      requireActiveUserIO.unsafeToFuture()
+    }
+  }
+
+  def requireUserAllowInactive(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.flatMap { oidcHeaders =>
```
gpolumbo-broad: Naming nit - 
There are 3 directives in this file:

- `requireActiveUser`
- `requireUserAllowInactive`
- `withNewUser`

My preference would be that these be named with the same pattern, maybe:

- `withActiveUser`
- `withUserAllowInactive`
- `withNewUser`

The anomalous one feels like `requireActiveUser` which can throw an exception if the user doesn't satisfy our "active" requirements, but by naming these similarly, they trigger more clearly in my brain that they're interchangeable based on what state you want the SamUser to be in.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserInfoDirectives.scala
```diff
@@ -22,11 +22,30 @@ trait UserInfoDirectives {
   val directoryDAO: DirectoryDAO
```
gpolumbo-broad, line 22: Rename trait to `SamUserDirectives`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
```diff
@@ -62,15 +62,14 @@ trait UserRoutes extends UserInfoDirectives with SamRequestContextDirectives {
...
             }
-          }
-        } ~ withSamRequestContext { samRequestContext =>
+        } ~
           pathPrefix("termsofservice") {
             pathPrefix("status") {
               pathEndOrSingleSlash {
                 get {
-                  requireActiveUser(samRequestContext) { samUser =>
+                  requireUserAllowInactive(samRequestContext) { samUser =>
```
gpolumbo-broad: Seems to me like changing from `requireActiveUser` to `requireUserAllowInactive` makes sense here in that an inactive user should be able to interrogate their ToS status, just want to make sure this was intentional?  I'll also ask if this is tested?  (I'll come back and delete this if I find the test later in the review)

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserInfoDirectives.scala
```diff
@@ -22,11 +22,30 @@ trait UserInfoDirectives {
...
+    * Specifically named to be clear that inactive users are permitted.
+    * Throws 403 exception if user does not exist (not 404 because that would mean the requested URL does not exist).
+    * @param samRequestContext
+    * @return
+    */
+  def requireUserAllowInactive(samRequestContext: SamRequestContext): Directive1[SamUser]
+
+  def withNewUser(samRequestContext: SamRequestContext): Directive1[SamUser]
 
   def asWorkbenchAdmin(samUser: SamUser): Directive0 =
```
gpolumbo-broad: See above "renaming directives" comment.  This one would be included too.  We have 3 different prefixes on the Directives, `with`, `require`, `as`.  It doesn't really matter that much, but I feel like it would make this whole trait slightly more easy to understand overall if they followed a common naming convention.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardUserInfoDirectives.scala
dvoet: addressed in https://github.com/broadinstitute/sam/pull/723

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
```diff
@@ -155,125 +151,116 @@ trait UserRoutes extends UserInfoDirectives with SamRequestContextDirectives {
...
-          }
         }
       }
     }
 
-  def adminUserRoutes: server.Route =
+  def adminUserRoutes(samUser: SamUser, samRequestContext: SamRequestContext): server.Route =
     pathPrefix("admin") {
-      withSamRequestContext { samRequestContext =>
-        requireActiveUser(samRequestContext) { samUser =>
```
gpolumbo-broad, line 166: Crazy question… What if `admin` apis actually didn't require the user to be an active user?  It's for admin things and we already need to be in `fc-admins` to do things.  In another conversation, we were already talking about frustration in needing to register our @firecloud.org emails in prod sam eventually in order to do some backend admin type things.  

It's off-topic question, but maybe something we can/should think about for performing admin things.  Might provide an easier way for bootstrapping a Sam instance so you don't have to directly insert a first user into the db…

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/MockUserInfoDirectives.scala
```diff
@@ -35,7 +35,9 @@ trait MockUserInfoDirectives extends UserInfoDirectives {
     }.unsafeToFuture()
   }
 
-  override def requireCreateUser(samRequestContext: SamRequestContext): Directive1[SamUser] = workbenchUser match {
+  override def requireUserAllowInactive(samRequestContext: SamRequestContext): Directive1[SamUser] = requireActiveUser(samRequestContext)
+
+  override def withNewUser(samRequestContext: SamRequestContext): Directive1[SamUser] = workbenchUser match {
```
gpolumbo-broad: nit
```suggestion
  override def withNewUser(samRequestContext: SamRequestContext): Directive1[SamUser] = samUser match {
```
And the message in the `None` case.

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserInfoDirectives.scala
dvoet: I am going to delay this one until the next PR

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
dvoet: well, it is certainly more in the control fo sam code now. Registration never seemed like much of a burden to me. In any case, not this PR :)

gpolumbo-broad (approved):

dvoet:
jenkins retest

dvoet:
jenkins retest

PR #723, opened 2022-05-11
POL-237 terms of service first class

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/POL-237
This PR removes all the code that tried to make terms of service work with ldap. That hit a big road block anyway because of caching at the apache proxy layer. This also stops modeling ToS acceptance as a group within Sam and instead tracks the last version accepted by a user in the sam_user table. This makes it even easier to determine if a user has met all the criteria to make Terra api calls - now it requires only 1 database round trip.

---

**PR checklist**

### DIFF
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220511_tos_version.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220511_tos_version.xml
@@ -11,4 +11,10 @@
             </column>
         </addColumn>
     </changeSet>
+    <changeSet logicalFilePath="dummy" author="dvoet" id="migrate_tos">
+        <sql>
+            update sam_user set accepted_tos_version = '0' where id in
+            (select gmf.member_user_id from sam_group_member_flat gmf join sam_group g on g.id = gmf.group_id where g.name = 'tos_accepted_0' and gmf.member_user_id is not null)
+        </sql>
+    </changeSet>
 </databaseChangeLog>
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220511_tos_version.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220511_tos_version.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="add_tos_version">
+        <addColumn tableName="sam_user">
+            <column name="accepted_tos_version" type="VARCHAR(40)">
+            </column>
+        </addColumn>
+    </changeSet>
+</databaseChangeLog>
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
@@ -139,7 +139,7 @@ object AppConfig {
     TermsOfServiceConfig(
       config.getBoolean("enabled"),
       config.getBoolean("isGracePeriodEnabled"),
-      config.getInt("version"),
+      config.getString("version"),
       config.getString("url")
     )
   }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
@@ -332,7 +332,7 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val 
     serializableWriteTransaction("createUser", samRequestContext)({ implicit session =>
       val userColumn = UserTable.column
 
-      val insertUserQuery = samsql"insert into ${UserTable.table} (${userColumn.id}, ${userColumn.email}, ${userColumn.googleSubjectId}, ${userColumn.enabled}, ${userColumn.azureB2cId}) values (${user.id}, ${user.email}, ${user.googleSubjectId}, ${user.enabled}, ${user.azureB2CId})"
+      val insertUserQuery = samsql"insert into ${UserTable.table} (${userColumn.id}, ${userColumn.email}, ${userColumn.googleSubjectId}, ${userColumn.enabled}, ${userColumn.azureB2cId}, ${userColumn.acceptedTosVersion}) values (${user.id}, ${user.email}, ${user.googleSubjectId}, ${user.enabled}, ${user.azureB2CId}, ${user.acceptedTosVersion})"
 
       Try {
         insertUserQuery.update().apply()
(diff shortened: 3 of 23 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
```diff
@@ -139,7 +139,7 @@ object AppConfig {
     TermsOfServiceConfig(
       config.getBoolean("enabled"),
       config.getBoolean("isGracePeriodEnabled"),
-      config.getInt("version"),
+      config.getString("version"),
```
dvoet, line 142: I changed this to a string because I think we will want more descriptive version like a date or something

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -542,6 +542,25 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
     })
   }
 
+  override def acceptTermsOfService(userId: WorkbenchUserId, tosVersion: String, samRequestContext: SamRequestContext): IO[Boolean] = {
+    serializableWriteTransaction("acceptTermsOfService", samRequestContext)({ implicit session =>
+      val u = UserTable.column
+      samsql"""update ${UserTable.table} set ${u.acceptedTosVersion} = ${tosVersion}
+              where ${u.id} = ${userId}
+              and (${u.acceptedTosVersion} is null
```
marctalbott, line 550: Why are we checking this? Wouldn't it be safe to just set `acceptedTosVersion` to whatever this is called with? Is the thinking just that we might as well surface whether the ToS acceptance actually changed or not, even if it's not consumed upstream at the moment?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -542,6 +542,25 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
+                or ${u.acceptedTosVersion} != ${tosVersion})""".update().apply() > 0
+    })
+  }
+
+  override def rejectTermsOfService(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Boolean] = {
+    serializableWriteTransaction("rejectTermsOfService", samRequestContext)({ implicit session =>
+      val u = UserTable.column
+      samsql"""update ${UserTable.table} set ${u.acceptedTosVersion} = null
+              where ${u.id} = ${userId}
+              and ${u.acceptedTosVersion} is not null""".update().apply() > 0
```
marctalbott, line 560: same/similar question here I suppose, why check that the ToS has been accepted? Are we checking upstream that a user isn't double rejecting?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -224,7 +221,7 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
   private def enableUserInternal(user: SamUser, samRequestContext: SamRequestContext): Future[Unit] = {
     for {
       _ <- directoryDAO.enableIdentity(user.id, samRequestContext).unsafeToFuture()
-      _ <- enableIdentityIfTosAccepted(user, samRequestContext).unsafeToFuture()
+      _ <- registrationDAO.enableIdentity(user.id, samRequestContext).unsafeToFuture()
```
marctalbott, line 224: So this PR will restrict Sam APIs to users who have accepted the ToS, but not other Terra services (assuming grace period is over and tos is enabled)? ToS enforcement in other services will come when OpenDJ is fully retired?

marctalbott (approved):
A few clarifying questions but generally looks good to me

rtitle (approved):
I don't know all the ToS context but the change looks good to me. Having tos acceptance as a first class citizen of `SamUser` makes sense to me.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
dvoet: I was keeping somewhat to the original behavior. Tests were using it. I also prefer to return something other than Unit.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
dvoet: The plan is for services rely on sam's checks. But services must verify on their own that it is ok to do so and stop using opendj at their pace. However I believe services are mostly doing the right thing already. See the doc linked from the epic containing the jira associated to this PR

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -86,12 +45,8 @@ class TosService (val directoryDao: DirectoryDAO, val registrationDao: Registrat
...
     * Otherwise return true if user has accepted ToS
     */
-  def isTermsOfServiceStatusAcceptable(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Boolean] = {
-    if (tosConfig.isGracePeriodEnabled) {
-      IO.pure(true)
-    } else {
-      getTosStatus(userId, samRequestContext).map(_.getOrElse(true))
-    }
+  def isTermsOfServiceStatusAcceptable(user: SamUser): Boolean = {
+    tosConfig.isGracePeriodEnabled || !tosConfig.enabled || user.acceptedTosVersion.contains(tosConfig.version)
```
andy7i, line 49: maybe outside the scope of this PR, but is the grace period still necessary? or do we still use the cache https://docs.google.com/document/d/1XTSne7Yd3hUsjK3yG3bNwaq9LzuDxxjMvmVEsAyYEP8/edit?disco=AAAAYPeYiBY

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockDirectoryDAO.scala
```diff
@@ -290,4 +290,30 @@ class MockDirectoryDAO(val groups: mutable.Map[WorkbenchGroupIdentity, Workbench
 
   override def loadUserByGoogleSubjectId(userId: GoogleSubjectId, samRequestContext: SamRequestContext): IO[Option[SamUser]] =
     IO.pure(users.values.find(_.googleSubjectId.contains(userId)))
+
+  override def acceptTermsOfService(userId: WorkbenchUserId, tosVersion: String, samRequestContext: SamRequestContext): IO[Boolean] = {
+    loadUser(userId, samRequestContext).map {
+      case None => false
+      case Some(user) =>
+        if (user.acceptedTosVersion.contains(tosVersion)) {
```
andy7i, line 298: minor:
```suggestion
        if (user.acceptedTosVersion.equals(tosVersion)) {
```

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220511_tos_version.xml
```diff
@@ -0,0 +1,20 @@
...
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="add_tos_version">
+        <addColumn tableName="sam_user">
+            <column name="accepted_tos_version" type="VARCHAR(40)">
+            </column>
+        </addColumn>
+    </changeSet>
+    <changeSet logicalFilePath="dummy" author="dvoet" id="migrate_tos">
+        <sql>
+            update sam_user set accepted_tos_version = '0' where id in
```
andy7i, line 16: It looks like when we create a user, accepted_tos_version is is set to `None`. Any reason why we initialize to 0 here?

it's small potatoes but for consistency, should we set it to NULL here?
```suggestion
            update sam_user set accepted_tos_version = NULL where id in
```

andy7i (approved):
LGTM overall. Couple suggestions but not blocking.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
dvoet: Grace period has nothing to do with the cache. I think it is a time where we have started to ask users to accept the new ToS and record their answers but don't yet enforce that they have accepted.

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockDirectoryDAO.scala
dvoet: `acceptedTosVersion` is an `Option` so I think `contains` is correct

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220511_tos_version.xml
dvoet: no, this is a migration to set the version to '0' for anyone who has already accepted version 0.

andy7i: OH I see that now. Thanks!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
marctalbott: ➕ to Doug's understanding of the grace period

andy7i: gotcha, thanks

PR #729, opened 2022-05-18
[TOAZ-129] Fix concurrency issue when setting B2C ids for existing google users

rtitle:
Ticket: https://broadworkbench.atlassian.net/browse/TOAZ-129

I observed this while testing B2C in Terra UI. Upon auth the UI makes 2 Sam calls simultaneously:
* `GET /register/user/v2/self/info`
* `GET /register/user/v1/termsofservice/status`

This exposed a concurrency issue in Sam. If this is a Google user through B2C then both of these calls try to set the `azureB2CId` in the database+ldap and one of them fails (leading to an error in the UI).

To fix I relaxed the semantics to allow setting the azureB2CId if the existing value equals the value we're trying to set.

Confirmed in a fiab that this fixes the issue.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
@@ -214,4 +214,5 @@ class LdapRegistrationDAO(
 
   override def setUserAzureB2CId(userId: WorkbenchUserId, b2CId: AzureB2CId, samRequestContext: SamRequestContext): IO[Unit] =
     executeLdap(IO(ldapConnectionPool.modify(userDn(userId), new Modification(ModificationType.REPLACE, Attr.azureB2CId, b2CId.value))), "setUserAzureB2CId", samRequestContext)
+
 }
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
```diff
@@ -213,6 +213,6 @@ class LdapRegistrationDAO(
   }
 
   override def setUserAzureB2CId(userId: WorkbenchUserId, b2CId: AzureB2CId, samRequestContext: SamRequestContext): IO[Unit] =
-    executeLdap(IO(ldapConnectionPool.modify(userDn(userId), new Modification(ModificationType.ADD, Attr.azureB2CId, b2CId.value))), "setUserAzureB2CId", samRequestContext)
+    executeLdap(IO(ldapConnectionPool.modify(userDn(userId), new Modification(ModificationType.REPLACE, Attr.azureB2CId, b2CId.value))), "setUserAzureB2CId", samRequestContext)
```
rtitle, line 216: I figure `REPLACE` semantics is safe for LDAP because:
* this always happens after the postgres `setUserAzureB2CId`, which should fail if we're trying to change a pre-existing value.
* the ldap stuff is temporary anyway

dvoet (approved):

PR #730, opened 2022-05-19
implement audit logging

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/PPWM-13
from the doc linked in the ticket:
This section goes beyond admin apis and addresses audit logging in sam as a whole. Sam should have a new audit logging capability that emits messages for every access control change. Messages should contain who made the change, timestamp, client IP, and details about the change. Changes to capture are:

- Resource events:
  - Resource created (include parent if there is one)
  - Resource updated (new parent or removal of parent)
  - Resource deletion
- Access change events:
  - Access added - when a subjects or permissions are added to a policy (including policy creation)  log the net change - what subjects have gained which permissions (roles, actions, descendant roles and actions)
  - Access removed - when a subjects or permissions are removed from a policy (including policy deletion) log the net change - what subjects have lost which permissions (roles, actions, descendant roles and actions)
 

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/AuditLogger.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/AuditLogger.scala
@@ -1,0 +1,22 @@
+package org.broadinstitute.dsde.workbench.sam.audit
+
+import cats.effect.IO
+import com.typesafe.scalalogging.LazyLogging
+import net.logstash.logback.argument.StructuredArguments
+import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
+import spray.json.{JsonWriter, enrichAny}
+import SamAuditModelJsonSupport.AuditInfoFormat
+
+object AuditLogger extends LazyLogging {
+  def logAuditEventIO[T <: AuditEvent : JsonWriter](samRequestContext: SamRequestContext, auditEvents: T*): IO[Unit] = {
+    IO(auditEvents.foreach(logAuditEvent(_, samRequestContext)))
+  }
+
+  def logAuditEvent[T <: AuditEvent : JsonWriter](auditEvent: T, samRequestContext: SamRequestContext): Unit = {
+    logger.whenInfoEnabled {
+      logger.info("auditEvent",
+        StructuredArguments.raw("eventDetails", auditEvent.toJson.compactPrint),
+        StructuredArguments.raw("auditInfo", samRequestContext.createAuditInfo.toJson.compactPrint))
+    }
+  }
+}
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/AuditLogger.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/AuditLogger.scala
@@ -14,7 +14,7 @@ object AuditLogger extends LazyLogging {
 
   def logAuditEvent[T <: AuditEvent : JsonWriter](auditEvent: T, samRequestContext: SamRequestContext): Unit = {
     logger.whenInfoEnabled {
-      logger.info("auditEvent",
+      logger.info(auditEvent.eventType.toString,
         StructuredArguments.raw("eventDetails", auditEvent.toJson.compactPrint),
         StructuredArguments.raw("auditInfo", samRequestContext.createAuditInfo.toJson.compactPrint))
     }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/SamAuditModel.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/SamAuditModel.scala
@@ -1,0 +1,122 @@
+package org.broadinstitute.dsde.workbench.sam.audit
+
+import org.broadinstitute.dsde.workbench.model.{WorkbenchException, WorkbenchGroupName, WorkbenchSubject, WorkbenchUserId}
+import org.broadinstitute.dsde.workbench.sam.model.{FullyQualifiedPolicyId, FullyQualifiedResourceId, ResourceAction, ResourceRoleName, ResourceTypeName}
+import spray.json.{DefaultJsonProtocol, DeserializationException, JsObject, JsString, JsValue, RootJsonFormat, SerializationException}
+
+import java.net.InetAddress
+import java.time.Instant
+
+sealed trait AuditEvent
+
+final case class AuditInfo(userId: Option[WorkbenchUserId], timestamp: Instant, clientIp: Option[InetAddress])
+
+final case class ResourceChange(parent: Option[FullyQualifiedResourceId])
+
+sealed trait ResourceEventType
+case object ResourceCreated extends ResourceEventType
+case object ResourceUpdated extends ResourceEventType
+case object ResourceDeleted extends ResourceEventType
+
+final case class ResourceEvent(eventType: ResourceEventType,
+                               resource: FullyQualifiedResourceId,
+                               added: Option[ResourceChange] = None,
+                               removed: Option[ResourceChange] = None) extends AuditEvent
+
+case class AccessChange(member: WorkbenchSubject,
+                        roles: Option[Iterable[ResourceRoleName]],
+                        actions: Option[Iterable[ResourceAction]],
+                        descendantRoles: Option[Map[ResourceTypeName, Iterable[ResourceRoleName]]],
+                        descendantActions: Option[Map[ResourceTypeName, Iterable[ResourceAction]]])
+
+sealed trait AccessChangeEventType
+case object AccessAdded extends AccessChangeEventType
+case object AccessRemoved extends AccessChangeEventType
+
+final case class AccessChangeEvent(eventType: AccessChangeEventType,
+                                   resource: FullyQualifiedResourceId,
+                                   changeDetails: Set[AccessChange]) extends AuditEvent
+
+object SamAuditModelJsonSupport {
+  import DefaultJsonProtocol._
+  import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
+  import org.broadinstitute.dsde.workbench.model.google.GoogleModelJsonSupport.InstantFormat
+  import org.broadinstitute.dsde.workbench.sam.model.SamJsonSupport._
+
+  implicit object InetAddressFormat extends RootJsonFormat[InetAddress] {
+    def write(ip: InetAddress) = JsString(ip.getHostAddress)
+
+    def read(value: JsValue): InetAddress = value match {
+      case JsString(str) => InetAddress.getByAddress(str.getBytes)
+      case _             => throw new WorkbenchException(s"Unable to unmarshal InetAddress from $value")
+    }
+  }
+
+  implicit val AuditInfoFormat = jsonFormat3(AuditInfo)
+
+  implicit val ResourceChangeFormat = jsonFormat1(ResourceChange)
+  implicit val ResourceEventTypeFormat = new RootJsonFormat[ResourceEventType] {
+    def read(obj: JsValue): ResourceEventType = obj match {
+      case JsString("ResourceCreated") => ResourceCreated
+      case JsString("ResourceUpdated") => ResourceUpdated
+      case JsString("ResourceDeleted") => ResourceDeleted
+      case _                           => throw new DeserializationException(s"could not deserialize ResourceEventType: $obj")
+    }
+
+    def write(obj: ResourceEventType): JsValue = JsString(obj.getClass.getSimpleName.split("\\$")(0))
+  }
+
+  implicit val ResourceEventFormat = jsonFormat4(ResourceEvent)
+
+  implicit val WorkbenchSubjectFormat = new RootJsonFormat[WorkbenchSubject] {
+    val MEMBER_TYPE_FIELD = "memberType"
+    val USER_TYPE = "user"
+    val GROUP_TYPE = "group"
+    val POLICY_TYPE = "policy"
+    val USER_ID_FIELD = "userId"
+    val GROUP_NAME_FIELD = "groupName"
+    val POLICY_ID_FIELD = "policyId"
+
+    override def read(json: JsValue): WorkbenchSubject = {
+      json match {
+        case JsObject(fields) =>
+          val maybeSubject: Option[WorkbenchSubject] = fields.get(MEMBER_TYPE_FIELD) match {
+            case Some(JsString(USER_TYPE)) => fields.get(USER_ID_FIELD).map(WorkbenchUserIdFormat.read)
+            case Some(JsString(GROUP_TYPE)) => fields.get(GROUP_NAME_FIELD).map(WorkbenchGroupNameFormat.read)
+            case Some(JsString(POLICY_TYPE)) => fields.get(POLICY_ID_FIELD).map(FullyQualifiedPolicyIdFormat.read)
+            case _ => None
+          }
+          maybeSubject.getOrElse(throw new DeserializationException(s"could not deserialize WorkbenchSubject: $json"))
+
+        case _ =>
+          throw new DeserializationException(s"could not deserialize WorkbenchSubject: $json")
+      }
+    }
+
+    override def write(obj: WorkbenchSubject): JsValue = {
+      obj match {
+        case user: WorkbenchUserId =>
+          JsObject(Map(MEMBER_TYPE_FIELD -> JsString(USER_TYPE), USER_ID_FIELD -> WorkbenchUserIdFormat.write(user)))
+        case groupName: WorkbenchGroupName =>
+          JsObject(Map(MEMBER_TYPE_FIELD -> JsString(GROUP_TYPE), GROUP_NAME_FIELD -> WorkbenchGroupNameFormat.write(groupName)))
+        case policyId: FullyQualifiedPolicyId =>
+          JsObject(Map(MEMBER_TYPE_FIELD -> JsString(POLICY_TYPE), POLICY_ID_FIELD -> FullyQualifiedPolicyIdFormat.write(policyId)))
+
+        case _ => throw new SerializationException(s"could not serialize WorkbenchSubject: $obj")
+      }
+    }
+  }
+
+  implicit val AccessChangeFormat = jsonFormat5(AccessChange)
+  implicit val AccessChangeEventTypeFormat = new RootJsonFormat[AccessChangeEventType] {
+    def read(obj: JsValue): AccessChangeEventType = obj match {
+      case JsString("AccessAdded") => AccessAdded
+      case JsString("AccessRemoved") => AccessRemoved
+      case _                         => throw new DeserializationException(s"could not deserialize AccessChangeEventType: $obj")
+    }
+
+    def write(obj: AccessChangeEventType): JsValue = JsString(obj.getClass.getSimpleName.split("\\$")(0))
+  }
+
+  implicit val PolicyEventFormat = jsonFormat3(AccessChangeEvent)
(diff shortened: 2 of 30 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/SamAuditModel.scala
```diff
@@ -0,0 +1,120 @@
...
+import org.broadinstitute.dsde.workbench.sam.model.{FullyQualifiedPolicyId, FullyQualifiedResourceId, ResourceAction, ResourceRoleName, ResourceTypeName}
+import spray.json.{DefaultJsonProtocol, DeserializationException, JsObject, JsString, JsValue, RootJsonFormat, SerializationException}
+
+import java.net.InetAddress
+
+sealed trait AuditEvent
+
+final case class AuditInfo(userId: Option[WorkbenchUserId], clientIp: Option[InetAddress])
+
+final case class ResourceChange(parent: Option[FullyQualifiedResourceId])
```
ehigham: What kind of `ResourceChange` would not have a resource id? Its use below made me question that as it can be one of the following patterns and I'm not sure what's different between the first and middle pattern:

``` scala
None
Some(ResourceChange(None))
Some(ResourceChange(Some(resourceId)))
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/SamAuditModel.scala
```diff
@@ -0,0 +1,120 @@
...
+      case JsString(str) => InetAddress.getByAddress(str.getBytes)
+      case _             => throw new WorkbenchException(s"Unable to unmarshal InetAddress from $value")
+    }
+  }
+
+  implicit val AuditInfoFormat = jsonFormat2(AuditInfo)
+
+  implicit val ResourceChangeFormat = jsonFormat1(ResourceChange)
+  implicit val ResourceEventTypeFormat = new RootJsonFormat[ResourceEventType] {
+    def read(obj: JsValue): ResourceEventType = obj match {
```
ehigham, line 60: Do we ever need deserialisation?

src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/AuditLogger.scala
```diff
@@ -0,0 +1,22 @@
...
+import SamAuditModelJsonSupport.AuditInfoFormat
+
+object AuditLogger extends LazyLogging {
+  def logAuditEventIO[T <: AuditEvent : JsonWriter](samRequestContext: SamRequestContext, auditEvents: T*): IO[Unit] = {
+    IO(auditEvents.foreach(logAuditEvent(_, samRequestContext)))
+  }
+
+  def logAuditEvent[T <: AuditEvent : JsonWriter](auditEvent: T, samRequestContext: SamRequestContext): Unit = {
+    logger.whenInfoEnabled {
+      logger.info("auditEvent",
```
ehigham: doesn't `info` take care of `whenInfoEnabled`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -302,15 +322,16 @@ class ResourceService(
...
-              _ <- fireGroupUpdateNotification(policyId, samRequestContext)
-            } yield ()
+          accessPolicyDAO.listAccessPolicies(policyId.resource, samRequestContext).flatMap { originalPolicies =>
+            originalPolicies.find(_.id == policyId) match {
+              case None => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"policy $policyId does not exist")))
+              case Some(existingPolicy) if existingPolicy.members == newMembers => IO.unit
+              case _ => for {
+                _ <- accessPolicyDAO.overwritePolicyMembers(policyId, newMembers, samRequestContext)
+                _ <- onPolicyUpdate(policyId, originalPolicies, samRequestContext)
+              } yield ()
```
ehigham: I got a bit confused by more than one case in a pattern match on `Option` because it can only take two forms. This is where I think applicative syntax really adds value. Your choice, naturally.
```suggestion
              case Some(existingPolicy) => Applicative[IO].whenA(existingPolicy.members != newMembers) {
                for {
                  _ <- accessPolicyDAO.overwritePolicyMembers(policyId, newMembers, samRequestContext)
                  _ <- onPolicyUpdate(policyId, originalPolicies, samRequestContext)
                } yield ()
              }
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -562,10 +619,36 @@ class ResourceService(
...
+      maybeOldParent <- accessPolicyDAO.getResourceParent(resourceId, samRequestContext)
+      deleted <- maybeOldParent match {
+        case None => IO.pure(false)
+        case Some(_) =>
+          for {
+            _ <- accessPolicyDAO.deleteResourceParent(resourceId, samRequestContext)
+            _ <- AuditLogger.logAuditEventIO(
+              samRequestContext,
+              ResourceEvent(ResourceUpdated, resourceId, removed = Option(ResourceChange(maybeOldParent))))
+          } yield true
```
ehigham: I'll try to not make stylistic comments as they can be quite subjective and of course there are countless ways to write the same thing. That being said...
I don't think a pattern match here adds clarity over `if` as you're not extracting the value - it just adds syntax IMO. 
In this particular case, you could `traverse_` the `Option` and then yield `deleted.isDefined` as an alternative though obviously, it's your choice.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -562,10 +619,36 @@ class ResourceService(
...
 
   def listResourceChildren(resourceId: FullyQualifiedResourceId, samRequestContext: SamRequestContext): IO[Set[FullyQualifiedResourceId]] = {
     accessPolicyDAO.listResourceChildren(resourceId, samRequestContext)
   }
+
+  private def createAccessChangeEvents(resource: FullyQualifiedResourceId, beforePolicies: Iterable[AccessPolicy], afterPolicies: Iterable[AccessPolicy]): Iterable[AccessChangeEvent] = {
+    val before = new PermissionsByUsers(beforePolicies)
+    val after = new PermissionsByUsers(afterPolicies)
+
+    val removeEvent = AccessChangeEvent(AccessAdded, resource, before.removeAll(after))
```
ehigham: Is the event type correct?

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAccessPolicyDAO.scala
```diff
@@ -167,7 +167,12 @@ class MockAccessPolicyDAO(private val resourceTypes: mutable.Map[ResourceTypeNam
     maybePolicy match {
       case Some((_, policy: AccessPolicy)) =>
         val newPolicy = policy.copy(public = isPublic)
-        IO(policies.put(resourceAndPolicyName, newPolicy))
+        IO {
+          policies.put(resourceAndPolicyName, newPolicy) match {
+            case Some(AccessPolicy(_, _, _, _, _, _, public)) if public != isPublic => true
```
ehigham:
```suggestion
            case Some(AccessPolicy(_, _, _, _, _, _, public)) => public != isPublic
```

ehigham (commented):
i'll take another pass tomorrow. learning a lot from this. thanks.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/AccessPolicyDAO.scala
```diff
@@ -66,7 +66,7 @@ trait AccessPolicyDAO {
 
   def listFlattenedPolicyMembers(policyId: FullyQualifiedPolicyId, samRequestContext: SamRequestContext): IO[Set[SamUser]]
 
-  def setPolicyIsPublic(policyId: FullyQualifiedPolicyId, isPublic: Boolean, samRequestContext: SamRequestContext): IO[Unit]
+  def setPolicyIsPublic(policyId: FullyQualifiedPolicyId, isPublic: Boolean, samRequestContext: SamRequestContext): IO[Boolean]
```
ehigham, line 69: Why return a boolean if we're not using that information where this is called?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/PermissionsByUsers.scala
```diff
@@ -0,0 +1,83 @@
...
+        member,
+        memberRolesDiff.get(member),
+        memberActionsDiff.get(member),
+        memberDescendantRolesDiff.get(member).map(groupByFirstInPair),
+        memberDescendantActionsDiff.get(member).map(groupByFirstInPair)
+      )
+    }
+  }
+
+  private def flattenMemberRoles(policies: Iterable[AccessPolicy]): Set[(WorkbenchSubject, ResourceRoleName)] =
```
ehigham: These for-comps are nice though I think we could remove two of them by passing a selector in, something like (forgive the psuedo-types)
```scala
def flattenMember[A](f: Policy -> A, policies: Set[Policy]): Set[(WorkbenchSubject, A)]
def flattenMemberDependant[A](f: DependentPermission -> A, policies: Set[Policy]): Set[(WorkbenchSubject, (ResourceTypeName, A))]
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -326,21 +347,29 @@ class ResourceService(
...
           }
-        }
+        case Some(existingAccessPolicy) =>
+          val newAccessPolicy = AccessPolicy(policyIdentity, workbenchSubjects, existingAccessPolicy.email, policy.roles, policy.actions, policy.descendantPermissions, existingAccessPolicy.public)
+          if (newAccessPolicy == existingAccessPolicy) {
+            // short cut if access policy is unchanged
+            IO.pure(newAccessPolicy)
+          } else {
+            for {
+              result <- accessPolicyDAO.overwritePolicy(newAccessPolicy, samRequestContext)
```
ehigham, line 369: is this `result` different from `newAccessPolicy`?

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAccessPolicyDAO.scala
ehigham: or 
```scala
policies.put(resourceAndPolicyName, newPolicy).map(_.public != isPublic).getOrElse(false)
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: no, I left that to make sure I discovered it in tests

src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/AuditLogger.scala
dvoet: using  `whenInfoEnabled` saves building the json string... probably not a big deal here but I have certainly seen cases where it is

src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/SamAuditModel.scala
dvoet: no, but I could not figure out the implicit magic to allow only json writers

ehigham: ah yes. implicits.

src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/SamAuditModel.scala
dvoet: This class is capturing what changed about a resource. In this case the only changeable thing I ended up with is the resource's parent. Note that the id here is for the parent, not the resource that is changing.

The idea behind all the Options in this file is that I don't want things showing up in json if they were not changed. So I agree that it is confusing to make parent an Option since it is the only thing that can change. But I don't think that will always be the case. Making it an Option is also consistent with AccessChange.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: I always forget about traversing Options

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/AccessPolicyDAO.scala
dvoet: I think it is used

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: probably not

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/AccessPolicyDAO.scala
ehigham: My bad, github hiding large changes by default .

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -438,33 +469,52 @@ class ResourceService(
...
-    val maybeFireNotification = if (groupChanged) {
-      fireGroupUpdateNotification(policyIdentity, samRequestContext)
+  private def onPolicyUpdateIfChanged(policyIdentity: FullyQualifiedPolicyId, orginalPolicies: Iterable[AccessPolicy], samRequestContext: SamRequestContext)(policyChanged: Boolean) = {
+    val maybeFireNotification = if (policyChanged) {
+      onPolicyUpdate(policyIdentity, orginalPolicies, samRequestContext)
     } else {
       IO.unit
     }
-    maybeFireNotification.map(_ => groupChanged)
+    maybeFireNotification.map(_ => policyChanged)
```
ehigham: If you're interested in applicative syntax, you can reduce this to the expression: 
```scala
IO.pure(policyChanged) <* Applicative[IO].whenA(policyChanged) { 
  onPolicyUpdate(policyIdentity, orginalPolicies, samRequestContext)
}
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/SamAuditModel.scala
ehigham: I think I'm just confused by the naming here. Something called a "Change" that doesn't catpure what's changed feels a bit weird. In the context of your `ResourceEvent` (which could be named `ResourceChangedEvent` to be consistent with `AccessChangedEvent`s?), what does `added = Some(ResourceChange(None))` mean? Would it not be simpler to flatten the `Option` and just use `Option[FullyQualifiedResourceId]`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/SamAuditModel.scala
```diff
@@ -0,0 +1,120 @@
...
+                        descendantRoles: Option[Map[ResourceTypeName, Iterable[ResourceRoleName]]],
+                        descendantActions: Option[Map[ResourceTypeName, Iterable[ResourceAction]]])
+
+sealed trait AccessChangeEventType
+case object AccessAdded extends AccessChangeEventType
+case object AccessRemoved extends AccessChangeEventType
+
+final case class AccessChangeEvent(eventType: AccessChangeEventType,
+                                   resource: FullyQualifiedResourceId,
+                                   changeDetails: Set[AccessChange]) extends AuditEvent
```
ehigham, line 40: would `NonEmptyList` be more appropriate - what would an event with an empty change-set correspond to?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -1418,4 +1427,206 @@ class ResourceServiceSpec extends AnyFlatSpec with Matchers with ScalaFutures wi
       AccessPolicyDescendantPermissions(defaultResourceType.name, Set(ResourceAction("I don't exist either")), Set(ResourceRoleName("I don't exist"))),
     )).unsafeRunSync().size shouldBe 4
   }
+
+  "createAccessChangeEvents" should "show not changes for empty beforePolicies and afterPolicies" in {
```
ehigham: `not changes` or `no changes` (and similar in this file)

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -1418,4 +1427,206 @@ class ResourceServiceSpec extends AnyFlatSpec with Matchers with ScalaFutures wi
...
+    val resource = genResource.sample.get
+    val beforePolicies = List.empty
+    val afterPolicies = List.empty
+    val result = service.createAccessChangeEvents(resource.fullyQualifiedId, beforePolicies, afterPolicies)
```
ehigham, line 1440: This method doesn't seem to need any of the `service`'s state. Could you make it a free function so that one can call it without needing a `service` instance?

ehigham (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
dvoet: it is an internal utility function of the service. Is there a benefit to setting it free?

src/main/scala/org/broadinstitute/dsde/workbench/sam/audit/SamAuditModel.scala
dvoet: I find NELs often more trouble than they are worth and I would prefer to have the `Set` semantics.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: I am going to pass on this. It starts to get impenetrable to me. It is hard to internalize the meaning of symbols. The word "Applicative" also has no meaning to me. Then there is the "A" at the end of `whenA` that is confusing.

ehigham: Fair enough - Applicatives are a lot to take on board but are worth the investment in my very biased opinion! You're actually already using them when you `traverse`!

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
ehigham: I'm of the camp it's always better to set free. More often than not in this codebase I've wanted to re-use private utility methods to the point where I'm not sure what making something private achieves other than documenting "this is not unit tested". I also prefer free functions over methods as they're easier to re-use.

PR #714, opened 2022-04-29
POL-232 setup tracing from client

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/POL-232
this adds tracing capabilities in the sam java client

---

**PR checklist**

- [ ] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've filled out the [Security Risk Assessment](https://broadinstitute.github.io/dsp-appsec-security-risk-assessment/) and attached the result to the JIRA ticket

PR #737, opened 2022-05-24
[TOAZ-126] Add Sam API for pet managed identities

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/TOAZ-126

Design doc: https://docs.google.com/document/d/1cyrRTibpvMbu0DtAjDsXeQp6-U2RftO66hHNb3nw_h4/edit#heading=h.focnnsveczp4

Demo recording: https://drive.google.com/file/d/1HiiAxQDVteaznwlRQLuFrXfm9PkSJDyU/view?usp=drive_web

Friends with: https://github.com/broadinstitute/firecloud-develop/pull/2991 (but the PRs are independent)

Overview: 
* Adds `AzureRoutes` and `AzureService` classes to Sam
   * Contains single `POST /api/azure/v1/user/petManagedIdentity` API which gets or creates a pet-UAMI
* Adds CRL dependency to Sam
* Adds `SAM_PET_MANAGED_IDENTITY` SQL table and associated queries
* Updates `StandardUserInfoDirectives` to handle requests from pet UAMIs and switch back to the owning user
* Add unit + connected tests for the above

---

**PR checklist**

### DIFF
--- a/render-test-config.sh
+++ b/render-test-config.sh
@@ -22,6 +22,7 @@ cat > ${AZURE_PROPERTIES_OUTPUT_FILE_PATH} <<EOF
 azureServices.managedAppClientId=${AZURE_MANAGED_APP_CLIENT_ID}
 azureServices.managedAppClientSecret=${AZURE_MANAGED_APP_CLIENT_SECRET}
 azureServices.managedAppTenantId=${AZURE_MANAGED_APP_TENANT_ID}
+azureServices.managedAppPlanId=terra-workspace-dev-plan
 EOF
 
 rm $AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH
--- a/render-test-config.sh
+++ b/render-test-config.sh
@@ -11,13 +11,17 @@ AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH="$(dirname $0)"/src/test/resources/azu
 AZURE_PROPERTIES_OUTPUT_FILE_PATH="$(dirname $0)"/src/test/resources/application.conf
 
 docker run --rm --cap-add IPC_LOCK \
-            -e VAULT_TOKEN=$VAULT_TOKEN ${DSDE_TOOLBOX_DOCKER_IMAGE} \
+            -e VAULT_TOKEN=$VAULT_TOKEN \
+            ${DSDE_TOOLBOX_DOCKER_IMAGE} \
             vault read -format json ${VAULT_AZURE_MANAGED_APP_CLIENT_PATH} \
             | jq -r .data > ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH}
 
 AZURE_MANAGED_APP_CLIENT_ID=$(jq -r '."client-id"' ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH})
 AZURE_MANAGED_APP_CLIENT_SECRET=$(jq -r '."client-secret"' ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH})
 AZURE_MANAGED_APP_TENANT_ID=$(jq -r '."tenant-id"' ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH})
+
+# Note: the managed app plan id is hardcoded for now but should be updated once the Managed App
+# definition is in Terraform. See: https://broadworkbench.atlassian.net/browse/TOAZ-28
 cat > ${AZURE_PROPERTIES_OUTPUT_FILE_PATH} <<EOF
 azureServices.managedAppClientId=${AZURE_MANAGED_APP_CLIENT_ID}
 azureServices.managedAppClientSecret=${AZURE_MANAGED_APP_CLIENT_SECRET}
--- a/render-test-config.sh
+++ b/render-test-config.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# TODO: Add Janitor secrets here
+# https://broadworkbench.atlassian.net/browse/ID-96
+# Call from GitHub actions once Janitor is integrated
+
+VAULT_TOKEN=${1:-$(cat $HOME/.vault-token)}
+DSDE_TOOLBOX_DOCKER_IMAGE=broadinstitute/dsde-toolbox:consul-0.20.0
+VAULT_AZURE_MANAGED_APP_CLIENT_PATH=secret/dsde/terra/azure/common/managed-app-publisher
+AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH="$(dirname $0)"/src/test/resources/azure_managed_app_client.json
+AZURE_PROPERTIES_OUTPUT_FILE_PATH="$(dirname $0)"/src/test/resources/application.conf
+
+docker run --rm --cap-add IPC_LOCK \
+            -e VAULT_TOKEN=$VAULT_TOKEN ${DSDE_TOOLBOX_DOCKER_IMAGE} \
+            vault read -format json ${VAULT_AZURE_MANAGED_APP_CLIENT_PATH} \
+            | jq -r .data > ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH}
+
+AZURE_MANAGED_APP_CLIENT_ID=$(jq -r '."client-id"' ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH})
+AZURE_MANAGED_APP_CLIENT_SECRET=$(jq -r '."client-secret"' ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH})
+AZURE_MANAGED_APP_TENANT_ID=$(jq -r '."tenant-id"' ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH})
+cat > ${AZURE_PROPERTIES_OUTPUT_FILE_PATH} <<EOF
+azureServices.managedAppClientId=${AZURE_MANAGED_APP_CLIENT_ID}
+azureServices.managedAppClientSecret=${AZURE_MANAGED_APP_CLIENT_SECRET}
+azureServices.managedAppTenantId=${AZURE_MANAGED_APP_TENANT_ID}
+EOF
+
+rm $AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220524_pet_user_assigned_managed_identity.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220524_pet_user_assigned_managed_identity.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="rtitle" id="pet_user_assigned_managed_identity">
+        <createTable tableName="SAM_PET_MANAGED_IDENTITY">
+            <column name="user_id" type="VARCHAR">
+                <constraints nullable="false" foreignKeyName="FK_SPMI_USER" referencedTableName="SAM_USER" referencedColumnNames="id" primaryKey="true"/>
+            </column>
+            <column name="tenant_id" type="VARCHAR(40)">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="subscription_id" type="VARCHAR(40)">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="managed_resource_group_name" type="VARCHAR">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="object_id" type="VARCHAR">
+                <constraints nullable="false"/>
+            </column>
+            <column name="display_name" type="VARCHAR">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+    </changeSet>
+</databaseChangeLog>
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1118,10 +1118,16 @@ paths:
       responses:
         200:
           description: Successfully retrieved resource without creating it
-          content: { }
+          content:
+            application/json:
+              schema:
+                type: string
         201:
           description: Successfully created resource
-          content: { }
+          content:
+            application/json:
+              schema:
+                type: string
         400:
           description: Invalid or incomplete request body
           content: { }
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -73,6 +73,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
       responses:
         200:
           description: status of specified user
@@ -93,6 +98,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
       responses:
         200:
           description: status of specified user
@@ -230,6 +240,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
       responses:
         204:
           description: Successfully added a user to the policy
@@ -499,6 +514,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
       responses:
         201:
           description: Group created
@@ -551,6 +571,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
       responses:
         204:
           description: Request sent
@@ -610,6 +635,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
       responses:
         204:
           description: Successfully set access instructions
@@ -730,6 +760,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
       responses:
         204:
           description: Email successfully added
@@ -1089,6 +1124,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
       responses:
         200:
           description: Successfully synchronized membership
@@ -1238,6 +1278,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
       responses:
         204:
           description: Successfully created resource
@@ -2160,6 +2205,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
       responses:
         201:
           description: user successfully created
@@ -2911,13 +2961,16 @@ components:
       properties:
         tenantId:
           type: string
-          description: the Azure tenant id
+          description: the Azure tenant id (UUID)
+          example: 0cb7a640-45a2-4ed6-be9f-63519f86e04b
         subscriptionId:
           type: string
-          description: the Azure subscription id
+          description: the Azure subscription id (UUID)
+          example: 3efc5bdf-be0e-44e7-b1d7-c08931e3c16c
         managedResourceGroupName:
           type: string
           description: the name of the Azure managed resource group
+          example: my-managed-resource-group
       description: specifies a request for a pet managed identity
   securitySchemes:
     googleoauth:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1102,6 +1102,41 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/azure/v1/user/petManagedIdentity:
+    post:
+      tags:
+        - Azure
+      summary: gets or creates a pet managed identity for the specified user
+      operationId: getPetManagedIdentity
+      requestBody:
+        description: The details of the managed resource group in which to create the pet
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/GetOrCreatePetManagedIdentityRequest'
+        required: true
+      responses:
+        200:
+          description: Successfully retrieved resource without creating it
+          content: { }
+        201:
+          description: Successfully created resource
+          content: { }
+        400:
+          description: Invalid or incomplete request body
+          content: { }
+        403:
+          description: Caller does not have permission to create a pet managed identity in the given managed resource group
+          content: { }
+        404:
+          description: Managed resource group does not exist
+          content: { }
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
   /api/resources/v2/{resourceTypeName}:
     get:
       tags:
@@ -2856,6 +2891,23 @@ components:
           type: boolean
           description: true if the user is in their proxy group
       description: ""
+    GetOrCreatePetManagedIdentityRequest:
+      required:
+        - tenantId
+        - subscriptionId
+        - managedResourceGroupName
+      type: object
+      properties:
+        tenantId:
+          type: string
+          description: the Azure tenant id
+        subscriptionId:
+          type: string
+          description: the Azure subscription id
+        managedResourceGroupName:
+          type: string
+          description: the name of the Azure managed resource group
+      description: specifies a request for a pet managed identity
   securitySchemes:
     googleoauth:
       type: oauth2
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1666,6 +1666,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
       responses:
         204:
           description: Successfully added a user to the policy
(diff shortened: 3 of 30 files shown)


rtitle:
jenkins retest

render-test-config.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# TODO: Add Janitor secrets here
+# https://broadworkbench.atlassian.net/browse/ID-96
```
rtitle, line 4: I fleshed out the https://broadworkbench.atlassian.net/browse/ID-96 description a bit more

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -336,18 +337,22 @@ object Boot extends IOApp with LazyLogging {
     val managedGroupService =
       new ManagedGroupService(resourceService, policyEvaluatorService, resourceTypeMap, accessPolicyDAO, directoryDAO, cloudExtensionsInitializer.cloudExtensions, config.emailDomain)
     val samApplication = SamApplication(userService, resourceService, statusService, tosService)
-
+    val azureRoutes = config.azureServicesConfig.map { config =>
```
rtitle: Note the `azureServicesConfig` is optional. We won't add the AzureRoutes if it's not present.

This is NOT using the existing `CloudExtensionRoutes` pattern in Sam. `CloudExtensionRoutes` doesn't really fit this use case -- Sam code is structured to only have 1 cloud extension, and it mostly deals with Google group stuff which is not relevant on Azure.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1102,6 +1102,47 @@ paths:
...
+            schema:
+              $ref: '#/components/schemas/GetOrCreatePetManagedIdentityRequest'
+        required: true
+      responses:
+        200:
+          description: Successfully retrieved resource without creating it
+          content:
+            application/json:
+              schema:
+                type: string
```
rtitle, line 1164: Currently this just returns the managed identity `objectId` as a String. This fits the Leo/WSM use case. I debated returning more information in a JSON object but keeping it simple for now.

src/test/resources/reference.conf
```diff
@@ -116,6 +116,12 @@ testStuff = {
       reuseIds = false
     }
   }
+  # Test MRG in which to create managed identities
```
rtitle, line 119: We commonly use this MRG for testing in WSM, Leo, etc

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceSpec.scala
```diff
@@ -0,0 +1,116 @@
...
+import scala.jdk.CollectionConverters._
+
+class AzureServiceSpec extends AnyFlatSpec with Matchers with ScalaFutures {
+  implicit val ec = scala.concurrent.ExecutionContext.global
+  implicit val ioRuntime = cats.effect.unsafe.IORuntime.global
+
+  "AzureService" should "create a pet managed identity" taggedAs ConnectedTest in {
+    val azureServicesConfig = appConfig.azureServicesConfig
+
+    assume(azureServicesConfig.isDefined, "-- skipping Azure test")
```
rtitle, line 23: Note these tests only run if `render-test-config.sh` was run. This doesn't happen in CI/CD, so these tests are skipped. Part of https://broadworkbench.atlassian.net/browse/ID-96 should be enabling these tests in CI/CD.

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/MockCrlService.scala
```diff
@@ -0,0 +1,115 @@
...
+import com.azure.resourcemanager.resources.models.{GenericResource, GenericResources, Plan, ResourceGroup, ResourceGroups}
+import org.broadinstitute.dsde.workbench.sam.model.{FullyQualifiedResourceId, ResourceId, SamResourceTypes}
+import org.mockito.ArgumentMatchers
+import org.mockito.ArgumentMatchers.{any, anyString}
+import org.mockito.Mockito.when
+import org.scalatestplus.mockito.MockitoSugar
+
+import scala.jdk.CollectionConverters._
+
+object MockCrlService extends MockitoSugar {
```
rtitle, line 20: Hopefully this mockito usage isn't too horrible. 😂  Azure SDKs are mock-able, but create a lot of intermediate classes which need to be mocked. (FWIW WSM tests Azure SDK the same way with mocks.)

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+    * Validates a managed resource group. Algorithm:
+    * 1. Resolve the MRG as the publisher
+    * 2. Get the managed app id from the MRG
+    * 3. Resolve the managed app as the publisher
+    * 4. Get the managed app "plan" id
+    * 4. Validate the plan id matches the configured value
+    */
+  private def validateManagedResourceGroup(request: GetOrCreatePetManagedIdentityRequest): IO[Unit] = {
```
rtitle, line 89: This is a separate check than user access control -- it essentially validates the MRG was created from the Terra managed app.

rtitle:
@tlangs I cleaned this up and added tests -- I think it's working well. I moved out of draft mode and requested review. Feel free to assign others too as needed.

I tried to annotate the PR a bit above. Happy to answer any further questions. And I punted on Janitor integration as we discussed -- added some context to https://broadworkbench.atlassian.net/browse/ID-96 as a follow-on task.

Thanks!

(And I know it's Friday 4:45pm before a long weekend -- I don't expect you to see this until next week 😄 )

rtitle:
jenkins retest

rtitle:
jenkins retest

rtitle:
seems like a spurious rawls failure, jenkins retest

rtitle:
jenkins retest

render-test-config.sh
```diff
@@ -0,0 +1,28 @@
...
+# Call from GitHub actions once Janitor is integrated
+
+VAULT_TOKEN=${1:-$(cat $HOME/.vault-token)}
+DSDE_TOOLBOX_DOCKER_IMAGE=broadinstitute/dsde-toolbox:consul-0.20.0
+VAULT_AZURE_MANAGED_APP_CLIENT_PATH=secret/dsde/terra/azure/common/managed-app-publisher
+AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH="$(dirname $0)"/src/test/resources/azure_managed_app_client.json
+AZURE_PROPERTIES_OUTPUT_FILE_PATH="$(dirname $0)"/src/test/resources/application.conf
+
+docker run --rm --cap-add IPC_LOCK \
```
gpolumbo-broad, line 13: TIL about `--cap-add IPC_LOCK`.  This seems like something we should add anytime we're calling `docker run` to read values from vault.  https://hub.docker.com/_/vault  

Did you discover this on your own or did DevOps recommend this to you?  How can we more broadly spread this knowledge and, more annoyingly, update our existing scripts?

render-test-config.sh
```diff
@@ -0,0 +1,28 @@
...
+            -e VAULT_TOKEN=$VAULT_TOKEN ${DSDE_TOOLBOX_DOCKER_IMAGE} \
```
gpolumbo-broad: Nit:
```suggestion
            -e VAULT_TOKEN=$VAULT_TOKEN \
            ${DSDE_TOOLBOX_DOCKER_IMAGE} \
```
It honestly took me several rereadings of the command to find the image argument because my brain didn't expect it to be at the end of the `-e` line.

render-test-config.sh
```diff
@@ -0,0 +1,28 @@
...
+            | jq -r .data > ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH}
+
+AZURE_MANAGED_APP_CLIENT_ID=$(jq -r '."client-id"' ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH})
+AZURE_MANAGED_APP_CLIENT_SECRET=$(jq -r '."client-secret"' ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH})
+AZURE_MANAGED_APP_TENANT_ID=$(jq -r '."tenant-id"' ${AZURE_MANAGED_APP_CLIENT_OUTPUT_FILE_PATH})
+cat > ${AZURE_PROPERTIES_OUTPUT_FILE_PATH} <<EOF
+azureServices.managedAppClientId=${AZURE_MANAGED_APP_CLIENT_ID}
+azureServices.managedAppClientSecret=${AZURE_MANAGED_APP_CLIENT_SECRET}
+azureServices.managedAppTenantId=${AZURE_MANAGED_APP_TENANT_ID}
+azureServices.managedAppPlanId=terra-workspace-dev-plan
```
gpolumbo-broad, line 29: Not having any idea what this is or how it's used, just want to confirm that it shouldn't be something like:

```suggestion
azureServices.managedAppPlanId=terra-workspace-${ENV}-plan
```

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2856,6 +2897,23 @@ components:
           type: boolean
           description: true if the user is in their proxy group
       description: ""
+    GetOrCreatePetManagedIdentityRequest:
+      required:
+        - tenantId
+        - subscriptionId
+        - managedResourceGroupName
+      type: object
+      properties:
```
gpolumbo-broad, line 2961: For each of these properties, is the formatting obvious?  Like in Google land, when we specify billing accounts, sometimes it needs to be `billingAccount/1111-2222-3333` and sometimes just `1111-2222-3333` is OK.  If there are any formatting/style tips/help you can add to the descriptions, that would be good.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
gpolumbo-broad: Sounds like a renaming/refactoring opportunity to make sure other people don't get confused in the future.  

https://broadworkbench.atlassian.net/browse/ID-109

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamRoutes.scala
```diff
@@ -37,7 +38,8 @@ abstract class SamRoutes(
     val policyEvaluatorService: PolicyEvaluatorService,
     val tosService: TosService,
     val liquibaseConfig: LiquibaseConfig,
-    val oidcConfig: OpenIDConnectConfiguration)(
+    val oidcConfig: OpenIDConnectConfiguration,
+    val azureRoutes: Option[AzureRoutes])(
```
gpolumbo-broad: I'm trying to wrap my head around why the `AzureRoutes` seem to be following a separate pattern here.  All of the other "services" have their "services" passed in, `ResourceService`, `UserService`, etc. and their routing seems to be handled via traits.  For Azure though, we are not passing in `AzureService`, but `AzureRoutes` and this abstract class is not extending `AzureRoutes`.  

Maybe this will become more clear as I read the rest of the PR, but at the moment it is confusing, which makes me want to request one of the following changes, either:

1. Change this code to conform to the same pattern that already exists, OR
2. Leave some comments here explaining why `AzureRoutes` and `AzureServices` is getting treated differently

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -73,24 +75,33 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with
 
 object StandardSamUserDirectives {
   val SAdomain: Regex = "(\\S+@\\S+\\.iam\\.gserviceaccount\\.com$)".r
+  val UAMIPattern: Regex = "(^/subscriptions/\\S+/resourcegroups/\\S+/providers/Microsoft\\.ManagedIdentity/userAssignedIdentities/\\S+$)".r
```
gpolumbo-broad: "User Assigned Managed Identity"

nit - please add comment expanding acronym

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220524_pet_user_assigned_managed_identity.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="rtitle" id="pet_user_assigned_managed_identity">
+        <createTable tableName="SAM_PET_MANAGED_IDENTITY">
```
tlangs, line 9: Do we want to make this named `SAM_AZURE_MANAGED_IDENTITY_PET` for clarity?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
tlangs: Its disappointing that it's structured this way, but I think it makes sense to do it like this for the sake of expediency.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -73,24 +75,33 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with
...
+        // If it's a managed identity, treat it as its owner
+        directoryDAO.getUserFromPetManagedIdentity(objectId, samRequestContext).flatMap {
+          case Some(petsOwner) => IO.pure(petsOwner)
+          case None => loadUserMaybeUpdateAzureB2CId(azureB2CId, oidcHeaders.googleSubjectIdFromAzure, directoryDAO, registrationDAO, samRequestContext)
+        }
+
+      case OIDCHeaders(_, Right(azureB2CId), _, _, _) =>
         loadUserMaybeUpdateAzureB2CId(azureB2CId, oidcHeaders.googleSubjectIdFromAzure, directoryDAO, registrationDAO, samRequestContext)
     }
   }
```
tlangs: I feel like this could be restructured. There's a bit of duplicated logic when the headers denote it's a PET.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -0,0 +1,120 @@
...
+import org.broadinstitute.dsde.workbench.sam.dataAccess.DirectoryDAO
+import org.broadinstitute.dsde.workbench.sam.model._
+import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
+
+import scala.jdk.CollectionConverters._
+
+class AzureService(crlService: CrlService,
+                   directoryDAO: DirectoryDAO) {
+  // Static region in which to create all managed identities
+  private val managedIdentityRegion = Region.US_EAST
```
tlangs, line 25: Should this be a configuration value rather than hardcoded?

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -0,0 +1,120 @@
...
+import org.broadinstitute.dsde.workbench.sam._
+import org.broadinstitute.dsde.workbench.sam.dataAccess.DirectoryDAO
+import org.broadinstitute.dsde.workbench.sam.model._
+import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
+
+import scala.jdk.CollectionConverters._
+
+class AzureService(crlService: CrlService,
+                   directoryDAO: DirectoryDAO) {
+  // Static region in which to create all managed identities
```
gpolumbo-broad: Why static?  Any reason why we force them all into this region?  Any justification as to _why_ we force all into US East will be helpful in maintaining this code.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -0,0 +1,120 @@
...
+                                        samRequestContext: SamRequestContext): IO[(PetManagedIdentity, Boolean)] = {
+    val id = PetManagedIdentityId(user.id, request.tenantId, request.subscriptionId, request.managedResourceGroupName)
+    for {
+      existingPetOpt <- directoryDAO.loadPetManagedIdentity(id, samRequestContext)
+      pet <- existingPetOpt match {
+        // pet exists in Sam DB - return it
+        case Some(p) => IO.pure((p, false))
+        // pet does not exist in Sam DB - create it
+        case None =>
+          for {
```
gpolumbo-broad: my preference is to extract nested `for` comp into its own private function, `def createPetManagedIdentity(…)`

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -0,0 +1,120 @@
...
+    * 1. Resolve the MRG as the publisher
+    * 2. Get the managed app id from the MRG
+    * 3. Resolve the managed app as the publisher
+    * 4. Get the managed app "plan" id
+    * 4. Validate the plan id matches the configured value
+    */
+  private def validateManagedResourceGroup(request: GetOrCreatePetManagedIdentityRequest): IO[Unit] = {
+    for {
+      resourceManager <- crlService.buildResourceManager(request.tenantId, request.subscriptionId)
+      resolvedMrg <- IO(resourceManager.resourceGroups().getByName(request.managedResourceGroupName.value)).attempt
```
gpolumbo-broad: Lines 77 and 78 look like duplicates of 61 and 62 - consider extracting (don't need to do it if you don't want to, but looks like a pattern we might repeat frequently).

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -0,0 +1,120 @@
...
+        ma <- resolvedManagedApp.toOption
+        p <- Option(ma.plan())
+      } yield p.name()
+      planId <- IO.fromOption(planIdOpt)(
+        new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden, "Validation failed: could not retrieve plan for managed app"))
+      )
+      _ <- IO.raiseUnless(planId == crlService.getManagedAppPlanId)(
+        new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden, "Validation failed: wrong managed app plan"))
+      )
+    } yield ()
```
tlangs: Could this be simplified with helper functions? Nested for/yields aren't my favorite.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -0,0 +1,120 @@
...
+            petName = toManagedIdentityNameFromUser(user)
+            context = managedIdentityContext(request, petName)
+            azureUami <- IO(manager.identities().define(petName.value)
+              .withRegion(managedIdentityRegion)
+              .withExistingResourceGroup(request.managedResourceGroupName.value)
+              .withTags(managedIdentityTags(user).asJava)
+              .create(context))
+            petToCreate = PetManagedIdentity(id, ManagedIdentityObjectId(azureUami.id()), ManagedIdentityDisplayName(azureUami.name()))
+            createdPet <- directoryDAO.createPetManagedIdentity(petToCreate, samRequestContext)
+          } yield (createdPet, true)
```
tlangs: Could be pulled out into its own function

tlangs (changes requested):

render-test-config.sh
rtitle: Honestly I copied it from CRL: https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/render-test-config.sh

It does seem like we should always add it when using docker+vault. There are a probably a lot of scripts out there but I could drop a PSA in #dsp-devops-champions?

render-test-config.sh
rtitle: Right now there is only a `terra-workspace-dev` managed app so I think this is fine. There is a ticket to Terraform the managed app definitions per-environment, after which we could read the plan id from vault. I can add a comment here..

src/main/resources/swagger/api-docs.yaml
rtitle: They are generally UUIDs but yes, I can update the description.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamRoutes.scala
rtitle: I did it this way so I could make the `AzureRoutes` optional. By using composition instead of inheritance, I'm able to deal with the optionality at the `SamRoutes` layer, via:
```
azureRoutes.map(…).getOrElse(reject)
```

I suppose alternatively I could follow the existing pattern of using inheritance and deal with the optionality inside `AzureRoutes`. That felt a bit less clean to me but maybe it's better to follow existing patterns -- I can give it a try.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
rtitle: Denis and Qi like to say "umami"

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
rtitle: I can expand the comments. Basically the managed identity takes a region but it's not actually used -- the identity can still be assigned to resources across regions. So just using a static region instead of making the Sam API take a region, etc. Will include links to Azure docs.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
rtitle: I can break it out to a new method.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
rtitle: Good call -- and I think we're actually calling that Azure API twice in the Sam request handling, seems like we could be more efficient and call it once.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220524_pet_user_assigned_managed_identity.xml
rtitle: I followed the pattern of the pre-existing `SAM_PET_SERVICE_ACCOUNT` table.

We could throw "azure" in there: `SAM_PET_AZURE_MANAGED_IDENTITY`. wdyt?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
rtitle: Not quite sure how it could be simplified. They are separate cases and separate DAO calls.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
rtitle: Ah see comment above. I think this field is needed by Azure APIs but not actually used for managed identities. I'll update the comment and include Azure docs.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
rtitle: Will do

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
rtitle: Yeah maybe - will try. 

(To play devil's advocate I often like having related logic in 1 place even if it means nested for-yields. Sometimes I think adding single-purpose helper functions makes code harder overall to reason about.)

rtitle:
Thanks for the reviews @gpolumbo-broad and @tlangs!

Some good comments above -- will do another pass and ping you when ready for another look.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
tlangs: Ah, yeah, I see. Too much divergence to combined the PET/AMI account cases. Le sigh.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -73,6 +73,11 @@ paths:
           required: true
           schema:
             type: string
+      requestBody:
```
rtitle, line 76: These changes are to work around an OpenAPI codegen bug which surfaced in Workspace Manager: https://github.com/OpenAPITools/openapi-generator/issues/11833

Basically if PUT/POST requests don't have a request body and thus have a null content-type, the generated client throws a NPE. It used to default to `application/json` but this changed when we upgraded openapi in [this PR](https://github.com/broadinstitute/sam/pull/714).

To work around I simply set the content-type for each affected PUT/POST method in swagger. This doesn't affect the actual backend routes -- it's just a client fix.

An alternative fix would be to patch the codegen logic in the [mustache code](https://github.com/broadinstitute/sam/blob/develop/codegen_java/templates/libraries/okhttp-gson/api.mustache), but I don't know all the implications of that, and this change seems reasonable enough. 

Let me know if you have other thoughts..

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamRoutes.scala
rtitle: restructured to use existing pattern.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
rtitle: Factored these out into helper functions.

rtitle:
jenkins retest

rtitle:
Hey @tlangs @gpolumbo-broad, sorry for the delay -- I think I addressed all comments, can you take another look at this? Thanks.

tlangs (approved):

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220524_pet_user_assigned_managed_identity.xml
tlangs: Ehhh, I'll leave it. MANAGED_IDENTITY and SERVICE_ACCOUNT are Microsoft & Google concepts respectively, so it just requires that knowledge.

src/main/resources/swagger/api-docs.yaml
tlangs: I'm ok with this.

zykonda (approved):

PR #744, opened 2022-06-02
ID-70 Make Sam check for all users group existance before inserting

tlangs:
https://broadworkbench.atlassian.net/browse/ID-70
Sam sees a ton of the following errors in its logs:
```
SQL execution failed (Reason: ERROR: duplicate key value violates unique constraint "sam_group_name_key"
  Detail: Key (name)=(All_Users) already exists.):

   insert into SAM_GROUP (name, email, updated_date, synchronized_date) values (All_Users, GROUP_All_Users@firecloud.org, 2022-05-17T16:37:39.549503Z, null)
```

This is happening every time the “All Users Group” is being added to, because when Sam gets the all users group, it tries to create it each time, and if the create fails because the group already exists (which it always does), then the SQL error is logged and Sam continues knowing that the group exists. 

This is silly. It pollutes the log with red-herring error messages, and handling exceptions as successful is a pretty big code smell.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -77,21 +77,23 @@ class GoogleExtensions(
     s"${googleServicesConfig.resourceNamePrefix.getOrElse("")}GROUP_${CloudExtensions.allUsersGroupName.value}@$emailDomain")
 
   override def getOrCreateAllUsersGroup(directoryDAO: DirectoryDAO, samRequestContext: SamRequestContext)(implicit executionContext: ExecutionContext): Future[WorkbenchGroup] = {
-    val allUsersGroup = BasicWorkbenchGroup(CloudExtensions.allUsersGroupName, Set.empty, allUsersGroupEmail)
+    val allUsersGroupStub = BasicWorkbenchGroup(CloudExtensions.allUsersGroupName, Set.empty, allUsersGroupEmail)
     for {
-      createdGroup <- directoryDAO.createGroup(allUsersGroup, samRequestContext = samRequestContext).unsafeToFuture() recover {
-        case e: WorkbenchExceptionWithErrorReport if e.errorReport.statusCode == Option(StatusCodes.Conflict) => allUsersGroup
+      existingGroup <- directoryDAO.loadGroup(allUsersGroupStub.id, samRequestContext = samRequestContext).unsafeToFuture()
+      allUsersGroup <- existingGroup match {
+        case None => directoryDAO.createGroup(allUsersGroupStub, samRequestContext = samRequestContext).unsafeToFuture()
+        case Some(group) => Future(group)
       }
-      existingGoogleGroup <- googleDirectoryDAO.getGoogleGroup(createdGroup.email)
+      existingGoogleGroup <- googleDirectoryDAO.getGoogleGroup(allUsersGroup.email)
       _ <- existingGoogleGroup match {
         case None =>
-          googleDirectoryDAO.createGroup(createdGroup.id.toString, createdGroup.email, Option(googleDirectoryDAO.lockedDownGroupSettings)) recover {
+          googleDirectoryDAO.createGroup(allUsersGroup.id.toString, allUsersGroup.email, Option(googleDirectoryDAO.lockedDownGroupSettings)) recover {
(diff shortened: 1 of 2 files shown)


tlangs:
Jenkins tests passed on rebuild, but I don't know how to make GH kick them off again or get that reflected here...

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -77,21 +77,23 @@ class GoogleExtensions(
...
   override def getOrCreateAllUsersGroup(directoryDAO: DirectoryDAO, samRequestContext: SamRequestContext)(implicit executionContext: ExecutionContext): Future[WorkbenchGroup] = {
-    val allUsersGroup = BasicWorkbenchGroup(CloudExtensions.allUsersGroupName, Set.empty, allUsersGroupEmail)
+    val allUsersGroupStub = BasicWorkbenchGroup(CloudExtensions.allUsersGroupName, Set.empty, allUsersGroupEmail)
     for {
-      createdGroup <- directoryDAO.createGroup(allUsersGroup, samRequestContext = samRequestContext).unsafeToFuture() recover {
-        case e: WorkbenchExceptionWithErrorReport if e.errorReport.statusCode == Option(StatusCodes.Conflict) => allUsersGroup
+      existingGroup <- directoryDAO.loadGroup(allUsersGroupStub.id, samRequestContext = samRequestContext).unsafeToFuture()
+      allUsersGroup <- existingGroup match {
+        case None => directoryDAO.createGroup(allUsersGroupStub, samRequestContext = samRequestContext).unsafeToFuture()
+        case Some(group) => Future(group)
```
gpolumbo-broad:
```suggestion
        case Some(group) => Future.successful(group)
```

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
tlangs: Whoop. Gotta re-learn my scala skillz.

gpolumbo-broad (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -579,10 +579,10 @@ class GoogleExtensionSpec(_system: ActorSystem) extends TestKit(_system) with An
     val googleExtensions = new GoogleExtensions(TestSupport.fakeDistributedLock, mockDirectoryDAO, mock[RegistrationDAO](RETURNS_SMART_NULLS), null, mockGoogleDirectoryDAO, null, null, null, null, null, null, null, null, null, googleServicesConfig, null, configResourceTypes)
 
     val allUsersGroup = BasicWorkbenchGroup(CloudExtensions.allUsersGroupName, Set.empty, WorkbenchEmail(s"TEST_ALL_USERS_GROUP@test.firecloud.org"))
-    val allUsersGroupMatcher = new ArgumentMatcher[BasicWorkbenchGroup] {
-      override def matches(group: BasicWorkbenchGroup): Boolean = group.id == allUsersGroup.id
+    val allUsersGroupNameMatcher = new ArgumentMatcher[WorkbenchGroupName] {
```
andy7i, line 582: my intellij is taking forever to load up so I'm just going to ask here -- why the change from `WorkbenchGroupName` to `BasicWorkbenchGroup`?

andy7i (approved):
1 question but not blocking. overall seems reasonable to me. not sure if there's any historical reason for the old implementation. did not run this locally.

to rerun jenkins tests from GH, you can add a comment with the text `jenkins-retest` (but replace the hyphen with a single space)

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
tlangs: Since the ` getOrCreateAllUsersGroup` method now checks for the existence of the all-users group and only tries to create it if it doesn't exist, I needed to change the mocked method to provide a response for the `loadGroup` method instead of the `createGroup` method, and that has a different signature.

andy7i: oh right, that makes sense! thanks for the explanation!

PR #751, opened 2022-06-10
perf improvement for self/info

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/ID-92

This started off with me noting extra db calls on the sam traces and spiraled a little from there. First, the perf fix means `UserService.getUserStatusInfo` is not getting fresh state from the db and is just formatting data passed in. This affected some tests which led to changes in mocks, especially `MockSamUserDirectives`. `MockSamUserDirectives` uses more production code and pretty much just mocks the headers. This led to more tests caring about the enabled status of a user. And I removed a little unused code along the way.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
@@ -14,6 +14,7 @@ import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.sam.api.StandardSamUserDirectives._
 import org.broadinstitute.dsde.workbench.sam.dataAccess.{DirectoryDAO, RegistrationDAO}
 import org.broadinstitute.dsde.workbench.sam.model.SamUser
+import org.broadinstitute.dsde.workbench.sam.service.TosService
 import org.broadinstitute.dsde.workbench.sam.service.UserService._
 import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
 
@@ -26,27 +27,13 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with 
 
   def withActiveUser(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.flatMap { oidcHeaders =>
     onSuccess {
-      val requireActiveUserIO = for {
-        user <- getSamUser(directoryDAO, registrationDAO, oidcHeaders, samRequestContext)
-      } yield {
-        // service account users do not need to accept ToS
-        val tosStatusAcceptable = tosService.isTermsOfServiceStatusAcceptable(user) || SAdomain.matches(user.email.value)
-        if (!tosStatusAcceptable) {
-          throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Unauthorized, "User has not accepted the terms of service."))
-        }
-        if (!user.enabled) {
-          throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Unauthorized, "User is disabled."))
-        }
-
-        user
-      }
-      requireActiveUserIO.unsafeToFuture()
+      getActiveSamUser(oidcHeaders, directoryDAO, registrationDAO, tosService, samRequestContext).unsafeToFuture()
     }
   }
 
   def withUserAllowInactive(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.flatMap { oidcHeaders =>
     onSuccess {
-      getSamUser(directoryDAO, registrationDAO, oidcHeaders, samRequestContext).unsafeToFuture()
+      getSamUser(oidcHeaders, directoryDAO, registrationDAO, samRequestContext).unsafeToFuture()
     }
   }
 
@@ -91,7 +78,7 @@ object StandardSamUserDirectives {
   val userIdHeader = "OIDC_CLAIM_user_id"
   val googleIdFromAzureHeader = "OAUTH2_CLAIM_google_id"
 
-  def getSamUser(directoryDAO: DirectoryDAO, registrationDAO: RegistrationDAO, oidcHeaders: OIDCHeaders, samRequestContext: SamRequestContext): IO[SamUser] = {
+  def getSamUser(oidcHeaders: OIDCHeaders, directoryDAO: DirectoryDAO, registrationDAO: RegistrationDAO, samRequestContext: SamRequestContext): IO[SamUser] = {
     oidcHeaders match {
       case OIDCHeaders(_, Left(googleSubjectId), WorkbenchEmail(SAdomain(_)), _) =>
         // If it's a PET account, we treat it as its owner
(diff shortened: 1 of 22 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -26,27 +27,13 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with
...
-        if (!tosStatusAcceptable) {
-          throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Unauthorized, "User has not accepted the terms of service."))
-        }
-        if (!user.enabled) {
-          throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Unauthorized, "User is disabled."))
-        }
-
-        user
-      }
-      requireActiveUserIO.unsafeToFuture()
```
dvoet, line 43: this moved to the `StandardSamUserDirectives` object below so it could be called from tests

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -91,7 +78,7 @@ object StandardSamUserDirectives {
   val userIdHeader = "OIDC_CLAIM_user_id"
   val googleIdFromAzureHeader = "OAUTH2_CLAIM_google_id"
 
-  def getSamUser(directoryDAO: DirectoryDAO, registrationDAO: RegistrationDAO, oidcHeaders: OIDCHeaders, samRequestContext: SamRequestContext): IO[SamUser] = {
+  def getSamUser(oidcHeaders: OIDCHeaders, directoryDAO: DirectoryDAO, registrationDAO: RegistrationDAO, samRequestContext: SamRequestContext): IO[SamUser] = {
```
dvoet, line 81: parameters reordered to match `getActiveSamUser`

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutes.scala
```diff
@@ -126,13 +126,7 @@ trait UserRoutes extends SamUserDirectives with SamRequestContextDirectives {
               path("info") {
                 get {
                   complete {
-                    userService.getUserStatusInfo(user.id, samRequestContext).map { statusOption =>
-                      statusOption
-                        .map { status =>
-                          StatusCodes.OK -> Option(status)
-                        }
-                        .getOrElse(StatusCodes.NotFound -> None)
```
dvoet, line 134: this case could never happen because `withUserAllowInactive` would catch non-existing users first

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
```diff
@@ -98,28 +98,6 @@ class LdapRegistrationDAO(
     }
   }
 
-  //To be used only in the event of the ToS version being bumped
-  override def disableAllHumanIdentities(samRequestContext: SamRequestContext): IO[Unit] = {
```
dvoet, line 102: no longer required since ToS does not use opendj

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -158,17 +158,10 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
-        } yield {
-          val tosStatus = tosService.isTermsOfServiceStatusAcceptable(user)
-          Some(UserStatusInfo(user.id.value, user.email.value, tosStatus && adminEnabled, adminEnabled))
-        }
-      case None => IO.pure(None)
-    }
+  def getUserStatusInfo(user: SamUser, samRequestContext: SamRequestContext): IO[UserStatusInfo] = {
+    val tosStatus = tosService.isTermsOfServiceStatusAcceptable(user)
+    IO.pure(UserStatusInfo(user.id.value, user.email.value, tosStatus && user.enabled, user.enabled))
+  }
```
dvoet, line 164: This is the perf change. Instead of accepting an id then loading the user _again_ and calling `directoryDAO.isEnabled` which contains information already included in the user, just take in the full user object and reformat it to the expected type with the right logic around ToS.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/TestSamRoutes.scala
```diff
@@ -22,7 +22,7 @@ import scala.concurrent.ExecutionContext
 /**
   * Created by dvoet on 7/14/17.
   */
-class TestSamRoutes(resourceService: ResourceService, policyEvaluatorService: PolicyEvaluatorService, userService: UserService, statusService: StatusService, managedGroupService: ManagedGroupService, val user: SamUser, directoryDAO: DirectoryDAO, registrationDAO: RegistrationDAO, val cloudExtensions: CloudExtensions = NoExtensions, override val newSamUser: Option[SamUser] = None, tosService: TosService = null)(implicit override val system: ActorSystem, override val materializer: Materializer, override val executionContext: ExecutionContext)
```
dvoet, line 25: `tosService` was null which caused NPEs because the change to `MockSamUserDirectives` cares about tos

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/MockSamUserDirectives.scala
```diff
@@ -16,26 +15,16 @@ trait MockSamUserDirectives extends SamUserDirectives {
...
+  private lazy val fakeOidcHeaders =
+    OIDCHeaders(OAuth2BearerToken("dummy token"), user.googleSubjectId.toLeft(user.azureB2CId.get), user.email, user.googleSubjectId)
 
   override def withActiveUser(samRequestContext: SamRequestContext): Directive1[SamUser] = onSuccess {
-    directoryDAO.loadSubjectFromEmail(user.email, samRequestContext).map { maybeUser =>
-      maybeUser.map { _ =>
-        if (isPetSA(user.email.value)) {
-          user.copy(id = WorkbenchUserId("newuser"), email = WorkbenchEmail("newuser@new.com"))
-        } else {
-          user
```
dvoet, line 31: This was the main problem with how the mock worked. It returned the `user` as it was statically defined, not fresh information loaded from the directory dao.

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockDirectoryDAO.scala
```diff
@@ -21,7 +21,6 @@ class MockDirectoryDAO(val groups: mutable.Map[WorkbenchGroupIdentity, Workbench
   private val groupSynchronizedDates: mutable.Map[WorkbenchGroupIdentity, Date] = new TrieMap()
   private val users: mutable.Map[WorkbenchUserId, SamUser] = new TrieMap()
   private val userAttributes: mutable.Map[WorkbenchUserId, mutable.Map[String, Any]] = new TrieMap()
-  private val enabledUsers: mutable.Map[WorkbenchSubject, Unit] = new TrieMap()
```
dvoet, line 24: this mock changed to record enabled status in the objects stored in `users` instead of a separate collection

dvoet:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
gpolumbo-broad: Can we move this logic out of `UserService` then?  I'm thinking we do one of the following: 

- Create method in `SamUser` -> `def getUserStatusInfo(): UserStatusInfo = ???`
- Or create constructor in `UserStatusInfo` that knows how to construct itself from a `SamUser`

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockDirectoryDAO.scala
```diff
@@ -145,18 +144,34 @@ class MockDirectoryDAO(val groups: mutable.Map[WorkbenchGroupIdentity, Workbench
     listSubjectsGroups(groupName, Set.empty).map(_.id)
   }
 
-  override def enableIdentity(subject: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Unit] = IO.pure(enabledUsers += ((subject, ())))
-
-  override def disableIdentity(subject: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Unit] = IO {
-    enabledUsers -= subject
+  override def enableIdentity(subject: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Unit] = {
+    subject match {
```
gpolumbo-broad: Could be DRY'd up by extracting the wrapping `subject match` statement in this method and the next 2.

gpolumbo-broad (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockDirectoryDAO.scala
dvoet: done but the matching in `isEnabled` doesn't really lend itself to it

gpolumbo-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
dvoet: I couldn't figure out a good way to do this because it depends on biz logic in tosService.

Shakespeared (approved):
Thanks for leaving comments, that was really helpful!

PR #759, opened 2022-06-28
ID-97 Remove JSON service account configs

tlangs:
https://broadworkbench.atlassian.net/browse/ID-97

This removes the `adminSdkServiceAccounts` configuration variable, now that Sam reads the service accounts from local paths instead of JSON embedded in its application configuration.
---

**PR checklist**


gpolumbo-broad (approved):

andy7i (approved):
LGTM 👍 

I needed to wrap my head around why files were better than configs but it clicked when I saw the description in the prior ticket:
>They are creating a new service (Yale) which rotates SA keys for K8 resources. For Yale to rotate the keys automatically, Sam needs to load the keys from a file instead of embedding them directly in JSON.

PR #763, opened 2022-07-06
[AJ-455] New APIs to add/delete members to policy by resourceId + policyGroupName

jordansafer:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-455
In order to add users by policy group, we currently need the policy email. However, requesting policy information is sensitive. These APIs allows a caller to give/revoke permission to policy groups by name (reader, writer, owner, etc), instead of by policyEmail.

Changes:
- 2 new APIs for adding and deleting users by policy group name under `/api/resources/v2/{resourceTypeName}/{resourceId}/policies/{policyName}/memberPolicies/{memberResourceTypeName}/{memberResourceId}/{memberPolicyName}`
- Update to override API to add users from both the "memberEmail" and "memberPolicy" section of the override payload (previously the memberPolicy section was ignored)

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1598,7 +1598,7 @@ paths:
     put:
       tags:
         - Resources
-      summary: Add a user to a policy by policy email
+      summary: Add a user to a policy by user email
       operationId: addUserToPolicyV2
       parameters:
         - name: resourceTypeName
@@ -1699,8 +1699,8 @@ paths:
     put:
       tags:
         - Resources
-      summary: Add a user to a policy by role on another resource
-      operationId: addUserToPolicyByRoleV2
+      summary: Add a member policy to a parent policy
+      operationId: addMemberPolicyV2
       parameters:
         - name: resourceTypeName
           in: path
@@ -1742,9 +1742,6 @@ paths:
         204:
           description: Successfully added a user to the policy
           content: { }
-        400:
-          description: email is not found
-          content: { }
         403:
           description: You do not have permission to alter this policy
           content: { }
@@ -1761,8 +1758,8 @@ paths:
     delete:
       tags:
         - Resources
-      summary: Remove a user from a policy by role on another resource
-      operationId: removeUserFromPolicyByRoleV2
+      summary: Remove a member policy from a parent policy
+      operationId: removeMemberPolicyV2
       parameters:
         - name: resourceTypeName
           in: path
@@ -1803,9 +1800,6 @@ paths:
       responses:
         204:
           description: Successfully removed a user from the policy
-          content: { }
-        400:
-          description: email is not found
           content: { }
         403:
           description: You do not have permission to alter this policy
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1598,7 +1598,7 @@ paths:
     put:
       tags:
         - Resources
-      summary: Add a user to a policy
+      summary: Add a user to a policy by policy email
       operationId: addUserToPolicyV2
       parameters:
         - name: resourceTypeName
@@ -1672,6 +1672,131 @@ paths:
         - name: email
           in: path
           description: Email of user to be removed
+          required: true
+          schema:
+            type: string
+      responses:
+        204:
+          description: Successfully removed a user from the policy
+          content: { }
+        400:
+          description: email is not found
+          content: { }
+        403:
+          description: You do not have permission to alter this policy
+          content: { }
+        404:
+          description: Resource type does not exist, you are not a member of any policy
+            on the resource, or user was not found
+          content: { }
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+  /api/resources/v2/{resourceTypeName}/{resourceId}/policies/{policyName}/memberPolicies/{memberResourceTypeName}/{memberResourceId}/{memberPolicyName}:
+    put:
+      tags:
+        - Resources
+      summary: Add a user to a policy by role on another resource
+      operationId: addUserToPolicyByRoleV2
+      parameters:
+        - name: resourceTypeName
+          in: path
+          description: Type of resource
+          required: true
+          schema:
+            type: string
+        - name: resourceId
+          in: path
+          description: Id of resource
+          required: true
+          schema:
+            type: string
+        - name: policyName
+          in: path
+          description: Name of the policy
+          required: true
+          schema:
+            type: string
+        - name: memberResourceTypeName
+          in: path
+          description: Type of resource for the member policy
+          required: true
+          schema:
+            type: string
+        - name: memberResourceId
+          in: path
+          description: Resource id for the member policy
+          required: true
+          schema:
+            type: string
+        - name: memberPolicyName
+          in: path
+          description: Member policy name
+          required: true
+          schema:
+            type: string
+      responses:
+        204:
+          description: Successfully added a user to the policy
+          content: { }
+        400:
+          description: email is not found
+          content: { }
+        403:
+          description: You do not have permission to alter this policy
+          content: { }
+        404:
+          description: Resource type does not exist, you are not a member of any policy
+            on the resource, or user was not found
+          content: { }
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+    delete:
+      tags:
+        - Resources
+      summary: Remove a user from a policy by role on another resource
+      operationId: removeUserFromPolicyByRoleV2
+      parameters:
+        - name: resourceTypeName
+          in: path
+          description: Type of resource
+          required: true
+          schema:
+            type: string
+        - name: resourceId
+          in: path
+          description: Id of resource
+          required: true
+          schema:
+            type: string
+        - name: policyName
+          in: path
+          description: Name of the policy
+          required: true
+          schema:
+            type: string
+        - name: memberResourceTypeName
+          in: path
+          description: Type of resource for the member policy
+          required: true
+          schema:
+            type: string
+        - name: memberResourceId
+          in: path
+          description: Resource id for the member policy
+          required: true
+          schema:
+            type: string
+        - name: memberPolicyName
+          in: path
+          description: Member policy name
           required: true
           schema:
             type: string
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1740,14 +1740,14 @@ paths:
             type: string
       responses:
         204:
-          description: Successfully added a user to the policy
+          description: Successfully added a member policy to the parent policy
           content: { }
         403:
           description: You do not have permission to alter this policy
           content: { }
         404:
           description: Resource type does not exist, you are not a member of any policy
-            on the resource, or user was not found
+            on the resource, or member policy was not found
           content: { }
         500:
           description: Internal Server Error
@@ -1799,14 +1799,14 @@ paths:
             type: string
       responses:
         204:
-          description: Successfully removed a user from the policy
+          description: Successfully removed a member policy from the parent policy
           content: { }
         403:
           description: You do not have permission to alter this policy
           content: { }
         404:
           description: Resource type does not exist, you are not a member of any policy
-            on the resource, or user was not found
+            on the resource, or member policy was not found
           content: { }
         500:
           description: Internal Server Error
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -204,10 +204,12 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
                     path(Segment / Segment / Segment) { (memberResourceType, memberResourceId, memberPolicyName) =>
                       val memberResource = FullyQualifiedResourceId(ResourceTypeName(memberResourceType),
                         ResourceId(memberResourceId))
-                      val subject = FullyQualifiedPolicyId(memberResource, AccessPolicyName(memberPolicyName))
-                      pathEndOrSingleSlash {
-                        putUserInPolicy(policyId, subject, samRequestContext) ~
-                        deleteUserFromPolicy(policyId, subject, samRequestContext)
+                      val policySubject = FullyQualifiedPolicyId(memberResource, AccessPolicyName(memberPolicyName))
+                      withPolicy(policySubject, samRequestContext) { memberPolicy =>
+                        pathEndOrSingleSlash {
+                          putUserInPolicy(policyId, memberPolicy.id, samRequestContext) ~
+                          deleteUserFromPolicy(policyId, memberPolicy.id, samRequestContext)
+                        }
                       }
                     }
                   }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -199,6 +199,23 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
                     }
                   }
                 } ~
+                pathPrefix("memberPolicies") {
+                  requireActionsForSharePolicy(policyId, samUser, samRequestContext) {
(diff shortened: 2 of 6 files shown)


jordansafer:
jenkins retest

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1695,6 +1695,131 @@ paths:
...
+          description: Member policy name
+          required: true
+          schema:
+            type: string
+      responses:
+        204:
+          description: Successfully added a user to the policy
+          content: { }
+        400:
+          description: email is not found
```
calypsomatic: The nittiest of all nits, but for consistency/"professionalism" maybe capitalize email.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1598,7 +1598,7 @@ paths:
     put:
       tags:
         - Resources
-      summary: Add a user to a policy
+      summary: Add a user to a policy by policy email
```
davidangb: nit: this adds a user to a policy by _the user's email_, not by a _policy email_, isn't that correct?

src/main/resources/swagger/api-docs.yaml
davidangb: following on: which email is this even concerned with? There's no email address specified in the call.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1695,6 +1695,131 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/resources/v2/{resourceTypeName}/{resourceId}/policies/{policyName}/memberPolicies/{memberResourceTypeName}/{memberResourceId}/{memberPolicyName}:
+    put:
+      tags:
+        - Resources
+      summary: Add a user to a policy by role on another resource
```
davidangb: nit: this could add multiple users (in fact, that's our use case). Maybe "Add a member policy to a parent policy" or similar?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1695,6 +1695,131 @@ paths:
...
+      operationId: addUserToPolicyByRoleV2
```
davidangb: also update the operationId, maybe `addMemberPolicyV2`?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1695,6 +1695,131 @@ paths:
...
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+    delete:
+      tags:
+        - Resources
+      summary: Remove a user from a policy by role on another resource
+      operationId: removeUserFromPolicyByRoleV2
```
davidangb: same here, this isn't about removing a user, it's about removing a member policy

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -199,6 +199,23 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
...
+                          ResourceId(memberResourceId))
+                        pathPrefix(Segment) { memberPolicyName =>
+                          val subject = FullyQualifiedPolicyId(memberResource, AccessPolicyName(memberPolicyName))
+                          pathEndOrSingleSlash {
+                            putUserInPolicy(policyId, subject, samRequestContext) ~
+                            deleteUserFromPolicy(policyId, subject, samRequestContext)
+                          }
+                        }
+                      }
+                    }
```
davidangb: for code cleanliness, how about consolidating the three nested `pathPrefix(Segment)` and the final `pathEndOrSingleSlash` into one `path(…)` directive?

```suggestion
                  path(Segment / Segment / Segment) { (memberResourceType, memberResourceId, memberPolicyName) =>
                    val memberResource = FullyQualifiedResourceId(ResourceTypeName(memberResourceType),
                      ResourceId(memberResourceId))
                    val subject = FullyQualifiedPolicyId(memberResource, AccessPolicyName(memberPolicyName))
                    putUserInPolicy(policyId, subject, samRequestContext) ~
                      deleteUserFromPolicy(policyId, subject, samRequestContext)
                  }
```

note that this is slightly more strict - it does NOT match a slash at the end of the API, whereas the existing PR code using `pathEndOrSingleSlash` does allow a slash at the end. In my opinion this is ok; we don't need to allow a trailing slash.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -199,6 +199,23 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
                     }
                   }
                 } ~
+                pathPrefix("memberPolicies") {
+                  requireActionsForSharePolicy(policyId, samUser, samRequestContext) {
```
davidangb, line 203: I know there is some debate over this. I feel pretty strongly that this should assert the calling user has at least READ permission on the memberResource to-be-added.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
```diff
@@ -610,6 +610,118 @@ class ResourceRoutesV2Spec extends AnyFlatSpec with Matchers with TestSupport wi
...
+
+    runAndWait(samRoutes.resourceService.createResource(resourceType, ResourceId("bar"), defaultUserInfo, samRequestContext))
+    Put(s"/api/resources/v2/${resourceType.name}/foo/policies/${resourceType.ownerRoleName}/memberPolicies/${resourceType.name}/bar/reader") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.NoContent
+    }
+  }
+
+  it should "403 adding without permission" in {
+    // differs from happy case in that owner role does not have alter_policies
+    val resourceType = ResourceType(ResourceTypeName("rt"), Set(SamResourceActionPatterns.alterPolicies, SamResourceActionPatterns.sharePolicy), Set(ResourceRole(ResourceRoleName("owner"), Set(SamResourceActions.sharePolicy(AccessPolicyName("splat"))))), ResourceRoleName("owner"))
```
davidangb: extreeeeemely minor nit: format this `resourceType` using the same linebreaks/formatting as the happy path test above, to make it easier for readers to spot the differences (same goes for other tests)

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutesV2Spec.scala
```diff
@@ -610,6 +610,118 @@ class ResourceRoutesV2Spec extends AnyFlatSpec with Matchers with TestSupport wi
...
+      FullyQualifiedPolicyId(FullyQualifiedResourceId(resourceType.name, ResourceId("bar")),
+        AccessPolicyName("reader")),
+      samRequestContext
+    ))
+
+    Delete(s"/api/resources/v2/${resourceType.name}/foo/policies/${resourceType.ownerRoleName}/memberPolicies/${resourceType.name}/bar/reader") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.NotFound
+    }
+  }
+
```
davidangb, line 749: assuming you add a runtime assertion that the user must have >= read on the memberPolicy to-be-added, add unit tests to verify that

also, how about a unit test to verify behavior of memberPolicies in the overwrite API?

davidangb (commented):
initial review, see comments inline!

src/main/resources/swagger/api-docs.yaml
jordansafer: Ah! This is a mistake- there is no email validation and no 400 error case at all. I copied this from the "memberEmails" API case, but it doesn't apply here

src/main/resources/swagger/api-docs.yaml
jordansafer: correct, updating

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
jordansafer: pushing back on the `pathEndOrSingleSlash` for consistency and slight usability improvement (I know I'm someone who would have/not have a slash at the end of a path in my config, and watch a deployment break over it)

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1695,6 +1695,125 @@ paths:
...
+        204:
+          description: Successfully added a user to the policy
+          content: { }
+        403:
+          description: You do not have permission to alter this policy
+          content: { }
+        404:
+          description: Resource type does not exist, you are not a member of any policy
+            on the resource, or user was not found
+          content: { }
```
davidangb, line 1842: another nit: the 204 and 404 responses still talk about "user", not "member policy"

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1695,6 +1695,125 @@ paths:
...
+        204:
+          description: Successfully removed a user from the policy
+          content: { }
+        403:
+          description: You do not have permission to alter this policy
+          content: { }
+        404:
+          description: Resource type does not exist, you are not a member of any policy
+            on the resource, or user was not found
+          content: { }
```
davidangb, line 1901: same here - 204 and 404 talk about "user"

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -199,6 +199,21 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
...
+                    path(Segment / Segment / Segment) { (memberResourceType, memberResourceId, memberPolicyName) =>
+                      val memberResource = FullyQualifiedResourceId(ResourceTypeName(memberResourceType),
+                        ResourceId(memberResourceId))
+                      val subject = FullyQualifiedPolicyId(memberResource, AccessPolicyName(memberPolicyName))
+                      requireOneOfAction(memberResource, Set(SamResourceActions.readPolicies, SamResourceActions.readPolicy(subject.accessPolicyName)), samUser.id, samRequestContext) {
```
davidangb: I think this is not right. Here, you're requiring `read_policies` or `read_policy::{name}` … this will cause a problem in practice. Workspace writers, for instance, only have `read_policy::owner`, so this API would then fail if a workspace writer attempts to import a snapshot - since the import process will attempt to sync (reader, writer, owner, project-owner) roles.

https://github.com/broadinstitute/sam/blob/3081228a206fe4ad44911bb948af18bf506ab67c/src/main/resources/reference.conf#L158

I'm proposing we only require `read` permission on the memberResource here.

(this may cascade into requiring unit test changes, too)

jordansafer: Oops, I was thinking too much about alter_policies as I wrote this, thank you for catching it!

davidangb (commented):
These changes are exactly what I requested in my last review, and all the syntax looks good.

But - I hate to say it - now that I see it in code, I'm worried about hardcoding a `ResourceAction("read")` check right into the API. What happens if a user calls this API and specifies a member resource of `notebook-cluster` or `managed-group`? I don't know what the use case is for someone to do that, but as-is this API will never work for those calls, because those resources don't have a "read" action.

Maybe someone from identiteam would have ideas on how to handle this - I'm a little stuck. Maybe what we really want to check is if the calling user has at least one action on the member resource, no matter what that action is named?

jordansafer:
> These changes are exactly what I requested in my last review, and all the syntax looks good.
> 
> But - I hate to say it - now that I see it in code, I'm worried about hardcoding a `ResourceAction("read")` check right into the API. What happens if a user calls this API and specifies a member resource of `notebook-cluster` or `managed-group`? I don't know what the use case is for someone to do that, but as-is this API will never work for those calls, because those resources don't have a "read" action.
> 
> Maybe someone from identiteam would have ideas on how to handle this - I'm a little stuck. Maybe what we really want to check is if the calling user has at least one action on the member resource, no matter what that action is named?

Yeah! I agree and this isn't obvious to me. I could also see a usecase where a user is responsible for giving read access to a set of groups that they aren't part of. For example, if there's an ABC Workspace, and I need to give access to all ABC Workspace readers, even though I'm not a member of ABC workspace. I'm thinking of google docs/groups, we share docs/groups with people, and the main restriction on the sharer is that they have permission to share the resource. I could see a situation where you're only allowed to share with members of a certain group, but that feels like a different usecase than what we're talking about here.

I've made a ticket to track: https://broadworkbench.atlassian.net/browse/AJ-493

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -199,6 +199,21 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
                     }
                   }
                 } ~
+                pathPrefix("memberPolicies") {
+                  requireActionsForSharePolicy(policyId, samUser, samRequestContext) {
+                    path(Segment / Segment / Segment) { (memberResourceType, memberResourceId, memberPolicyName) =>
+                      val memberResource = FullyQualifiedResourceId(ResourceTypeName(memberResourceType),
+                        ResourceId(memberResourceId))
+                      val subject = FullyQualifiedPolicyId(memberResource, AccessPolicyName(memberPolicyName))
+                      requireOneOfAction(memberResource, Set(ResourceAction("read")), samUser.id, samRequestContext) {
```
dvoet: I think this is too restrictive. I am not sure what is being protected. In a google group analogy, I don't need to be a member of a group to share with it. The location of this check also effects removal of the member which is not correct. A user who may remove members from a policy should be able to do so regardless of their permissions on the member.

On the technical side, not all resource types have an action named `read` so this is not really generic enough if we want to keep it. It also will result in a 404 or 403 error, those errors probably should be mapped to a 400 error.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -199,6 +199,21 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
                     }
                   }
                 } ~
+                pathPrefix("memberPolicies") {
+                  requireActionsForSharePolicy(policyId, samUser, samRequestContext) {
+                    path(Segment / Segment / Segment) { (memberResourceType, memberResourceId, memberPolicyName) =>
+                      val memberResource = FullyQualifiedResourceId(ResourceTypeName(memberResourceType),
+                        ResourceId(memberResourceId))
+                      val subject = FullyQualifiedPolicyId(memberResource, AccessPolicyName(memberPolicyName))
```
dvoet: need something somewhere that checks that this subject exists

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
davidangb: This is me; I suggested checking a permission on the member resource here. It feels wrong at a gut level to be operating on memberResources that the user has no permissions on. However, I agree that checking `read` is also not correct (see https://github.com/broadinstitute/sam/pull/763#pullrequestreview-1037993317). 

What about - for adding the memberResource - checking if the user has a nonzero set of actions on the memberResource, i.e. has any permission, no matter what it is? Or seeing if the user is present in `listAllFlattenedResourceUsers`?

Agreed that we don't need any check on the memberResource for removing the memberResource

I hear you that you don't think the user needs any permission on the memberResource for adding, e.g. sharing with a group without being a member of it. I'm trying to get comfortable with that.

dvoet: Something that still would exist is sharing with a policy via email address which would not have a check. Why would this sharing mechanism be different?

tlangs (approved):
lgtm, once the other comments are addressed

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
jordansafer: updated-

davidangb (approved):
:+1: and thank you, once tests pass!

jordansafer:
jenkins retest

dvoet (approved):

PR #769, opened 2022-07-15
[TOAZ-177] Add route to administratively get a pet MI for a given user

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-177

**Problem statement:**
In PR https://github.com/broadinstitute/sam/pull/737 we added a [Sam route](https://sam.dsde-dev.broadinstitute.org/#/Azure/getPetManagedIdentity) to get a pet managed identity for the calling user. The WSM VM creation flight calls this Sam route with the calling user's credentials. However this doesn't actually work because Leo calls WSM as its own SA, not as the end user. So we ended up generating a pet for the Leo SA instead of the end user. 😞 

**Solution:**
The solution introduced in this PR is to add a new Sam route to request a pet managed identity on behalf of a user. This allows WSM to call Sam as its own SA, and request a pet for the user assigned to the VM. The new route follows the same pattern as the Google [getUserPetServiceAccountKey](https://sam.dsde-dev.broadinstitute.org/#/Google/getUserPetServiceAccountKey) route. The azure route checks the caller against the `cloud-extension/azure` policy in Sam. This policy will need to be created in a similar way to `cloud-extension/google` (which currently contains a handful of Terra service accounts). Aside from the new permission check, the route handling is identical to the original `getPetManagedIdentity` route.

Note: an alternative approach would be for to Leo request a pet MI up front with the calling user's token; and pass the objectId down to WSM. This approach wouldn't require Sam changes; however I thought it added a bit too much complexity to have VM creation logic spread across Leo and WSM -- it's nice for all elements of VM creation to be fully localized in the WSM stairway flight.

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1187,7 +1187,8 @@ paths:
     post:
       tags:
         - Azure
-      summary: administratively gets or creates a pet managed identity for the specified user
+      summary: gets or creates a pet managed identity for the specified user, get_pet_managed_identity
+        action on cloud-extension/azure required
       operationId: getPetManagedIdentityForUser
       parameters:
         - name: userEmail
@@ -1221,9 +1222,10 @@ paths:
           content: { }
         403:
           description: Caller does not have permission to create a pet managed identity in the given managed resource group
+            or caller does not have get_pet_managed_identity permission on cloud-extension/azure
           content: { }
         404:
-          description: Managed resource group does not exist
+          description: User or managed resource group does not exist
           content: { }
         500:
           description: Internal Server Error
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1146,8 +1146,56 @@ paths:
     post:
       tags:
         - Azure
-      summary: gets or creates a pet managed identity for the specified user
+      summary: gets or creates a pet managed identity for the calling user
       operationId: getPetManagedIdentity
+      requestBody:
+        description: The details of the managed resource group in which to create the pet
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/GetOrCreatePetManagedIdentityRequest'
+        required: true
+      responses:
+        200:
+          description: Successfully retrieved resource without creating it
+          content:
+            application/json:
+              schema:
+                type: string
+        201:
+          description: Successfully created resource
+          content:
+            application/json:
+              schema:
+                type: string
+        400:
+          description: Invalid or incomplete request body
+          content: { }
+        403:
+          description: Caller does not have permission to create a pet managed identity in the given managed resource group
+          content: { }
+        404:
+          description: Managed resource group does not exist
+          content: { }
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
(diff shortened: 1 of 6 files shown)


rtitle:
jenkins retest to get a sam-build

rtitle:
WSM PR to use this new Sam API: https://github.com/DataBiosphere/terra-workspace-manager/pull/737

Currently attempting to test on a BEE.

rtitle:
jenkins retest

rtitle:
This is working with Leo and WSM in my BEE. Requesting some reviews…

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -67,6 +67,18 @@ class AzureService(crlService: CrlService,
...
 
+  /**
+    * Loads a SamUser from the database by email.
+    */
+  def getSamUser(email: WorkbenchEmail, samRequestContext: SamRequestContext): IO[Option[SamUser]] =
+    for {
+      subjectOpt <- directoryDAO.loadSubjectFromEmail(email, samRequestContext)
+      samUserOpt <- subjectOpt match {
+        case Some(userId: WorkbenchUserId) => directoryDAO.loadUser(userId, samRequestContext)
+        case _ => IO.pure(None)
```
Qi77Qi: btw you could do `IO.none` :)

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutesSpec.scala
```diff
@@ -74,19 +72,113 @@ class AzureRoutesSpec extends AnyFlatSpec with Matchers with ScalatestRouteTest
...
+        status shouldEqual StatusCodes.Created
+      }
+    }
+
+    samRoutes
+  }
+
+  private def createSecondUser(samRoutes: TestSamRoutes, addToSpendProfile: Boolean = true): SamUser = {
+    val newUser = Generator.genWorkbenchUserGoogle.sample.get
+    IO.fromFuture(IO(samRoutes.userService.createUser(newUser, samRequestContext))).unsafeRunSync()
```
Qi77Qi, line 176: probably can just do `samRoutes.userService.createUser(newUser, samRequestContext)` here? looks like you want the side effect to happen here

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1183,6 +1183,54 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/azure/v1/petManagedIdentity/{userEmail}:
+    post:
+      tags:
+        - Azure
+      summary: administratively gets or creates a pet managed identity for the specified user
```
Qi77Qi: maybe some comment about "who" has permission to create PET managed identity on behalf of others?

rtitle: 👍 .. will update this text.

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutesSpec.scala
rtitle: `userService .createUser` returns a `Future`, I want to wait for it to complete… I guess I could use `Await(futre)` instead of converting to IO..

zykonda (approved):
Looks good

tlangs (approved):

PR #770, opened 2022-07-18
ID-155 remove usage of opendj isEnabled

andy7i:
https://broadworkbench.atlassian.net/browse/ID-155

some APIs on dev are failing because opendj responses are bigger than 20MB. It seems as though the registered users list is large enough to cause this issue right now. 

According to the Doge, opendj is not needed for checking if a user is enabled. This is a step towards progress in removing opendj and is also a fix for dev. 

Alternatives considered were optimizing the opendj call so we don't get the entire list every time. Another would be to increase the max message size limit. Both are not ideal as we are planning to get rid of opendj anyway.
Context: https://broadinstitute.slack.com/archives/C01NLH6SHJN/p1658171355005759?thread_ts=1658146736.680259&cid=C01NLH6SHJN

I think we can probably remove the implementation of registrationDAO's isEnabled, but can leave that in if folks feel it's useful to keep.

Ticket: <Link to Jira ticket> TODO - create one

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
@@ -99,15 +99,7 @@ class LdapRegistrationDAO(
   }
 
   override def isEnabled(subject: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Boolean] =
-    for {
-      entry <- executeLdap(IO(ldapConnectionPool.getEntry(directoryConfig.enabledUsersGroupDn, Attr.member)), "isEnabled", samRequestContext)
-    } yield {
-      val result = for {
-        e <- Option(entry)
-        members <- Option(e.getAttributeValues(Attr.member))
-      } yield members.contains(subjectDn(subject))
-      result.getOrElse(false)
-    }
+    IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, s"LdapRegistrationDAO is deprecated")))
 
   override def createEnabledUsersGroup(samRequestContext: SamRequestContext): IO[Unit] = {
     val objectClassAttr = new Attribute("objectclass", Seq("top", "groupofnames").asJava)
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
@@ -98,6 +98,7 @@ class LdapRegistrationDAO(
     }
   }
 
+  @deprecated
   override def isEnabled(subject: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Boolean] =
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -134,7 +134,7 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
             allUsersGroup <- cloudExtensions.getOrCreateAllUsersGroup(directoryDAO, samRequestContext)
             allUsersStatus <- directoryDAO.isGroupMember(allUsersGroup.id, user.id, samRequestContext).unsafeToFuture() recover { case _: NameNotFoundException => false }
             tosAcceptedStatus <- tosService.getTosStatus(user.id, samRequestContext).unsafeToFuture()
-            ldapStatus <- registrationDAO.isEnabled(user.id, samRequestContext).unsafeToFuture()
+            ldapStatus = directoryDAO.isEnabled(user.id, samRequestContext).unsafeToFuture() // calling postgres instead of opendj here as a temporary measure as we work toward eliminating opendj
```
ehigham: don't you want `<-`?

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
andy7i: yeah, whoops! I had hardcoded it to `true` and forgot to add the arrow back

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
```diff
@@ -99,15 +99,7 @@ class LdapRegistrationDAO(
...
-    for {
-      entry <- executeLdap(IO(ldapConnectionPool.getEntry(directoryConfig.enabledUsersGroupDn, Attr.member)), "isEnabled", samRequestContext)
-    } yield {
-      val result = for {
-        e <- Option(entry)
-        members <- Option(e.getAttributeValues(Attr.member))
-      } yield members.contains(subjectDn(subject))
-      result.getOrElse(false)
-    }
+    IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, s"LdapRegistrationDAO is deprecated")))
```
andy7i, line 103: not sure if this is the best way to deprecate this call -- let me know if you have ideas!

ehigham: I'd be tempted to add a `@deprecated` annotation to this class if it is indeed deprecated

ehigham (approved):

andy7i:
jenkins retest - seems like it could be the google issue

andy7i:
jenkins retest
- rawls failed with:
- The code passed to eventually never returned normally. Attempted 42 times over 16.395372288750004 minutes. Last failure message: "[Fail]ed" was not equal to "[Succeed]ed".

https://fc-jenkins.dsp-techops.broadinstitute.org/job/rawls-fiab-test-runner/46442/testReport/junit/org.broadinstitute.dsde.test.api/RawlsApiSpec/Rawls_should_support_running_workflows_with_private_docker_images/

Shakespeared (approved):

gpolumbo-broad (approved):

PR #772, opened 2022-07-21
ID-158 Make Sam return NOT_FOUND if user not authorized to modify group

tlangs:
https://broadworkbench.atlassian.net/browse/ID-158

Sam should not squash unauthorized group modification into a `NO_CONTENT` response. This gives users and downstream tools the wrong idea of what's going on. We should instead return a `NOT_FOUND` response.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
@@ -116,7 +116,8 @@ trait ManagedGroupRoutes extends SamUserDirectives with SecurityDirectives with 
       samRequestContext: SamRequestContext): Route =
     requireAction(managedGroup, SamResourceActions.sharePolicy(accessPolicyName), samUser.id, samRequestContext) {
       withSubject(WorkbenchEmail(email), samRequestContext) { subject =>
-        complete(managedGroupService.addSubjectToPolicy(managedGroup.resourceId, accessPolicyName, subject, samRequestContext).map(_ => StatusCodes.NoContent))
+        complete(managedGroupService.addSubjectToPolicy(managedGroup.resourceId, accessPolicyName, subject, samRequestContext)
+          .map(if (_) StatusCodes.NoContent else StatusCodes.NotFound))
       }
     }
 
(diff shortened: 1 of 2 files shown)


andy7i:
drive by question because I was curious about this one. did you find anything relating to this theory?:
> guessing the 204s might be changes to make the ui work a certain way, and docs not updated (that’s one ui dev strategy to allow calls w/ no page nav)? so am wagering that docs are just out of date. don’t have full history without looking at commits, but we can do a documentation update with team when we have time to fill things in.

Shakespeared (approved):

tlangs:
I'm not sure how this would be represented in the UI. The user wouldn't be able to even load a page for a resource they don't have access to, so there's no way they should be able to make this request from the UI.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -500,6 +500,15 @@ class ManagedGroupRoutesSpec extends AnyFlatSpec with Matchers with ScalatestRou
     }
   }
 
+  it should "respond with a 404 when the user does not have access to modify the group" in {
+    val samRoutes = TestSamRoutes(resourceTypes)
+    withUserNotInGroup(samRoutes){ nonMemberRoutes =>
+      Delete(s"/api/group/$groupId/admin/${samRoutes.user.email}") ~> nonMemberRoutes.route ~> check {
```
gpolumbo-broad, line 509: You're making a `DELETE` request here, but the code you modified is for `PUT`.  This test might be worth keeping too, but make sure the `PUT` is also tested.

gpolumbo-broad (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
tlangs: Ah shoot, you're totally right!

gpolumbo-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutes.scala
```diff
@@ -116,7 +116,8 @@ trait ManagedGroupRoutes extends SamUserDirectives with SecurityDirectives with
       samRequestContext: SamRequestContext): Route =
     requireAction(managedGroup, SamResourceActions.sharePolicy(accessPolicyName), samUser.id, samRequestContext) {
       withSubject(WorkbenchEmail(email), samRequestContext) { subject =>
-        complete(managedGroupService.addSubjectToPolicy(managedGroup.resourceId, accessPolicyName, subject, samRequestContext).map(_ => StatusCodes.NoContent))
+        complete(managedGroupService.addSubjectToPolicy(managedGroup.resourceId, accessPolicyName, subject, samRequestContext)
+          .map(if (_) StatusCodes.NoContent else StatusCodes.NotFound))
```
dvoet, line 120: I don't think this does what you want. The authorization check is the `requireAction` call above. `addSubjectToPolicy` returns false when no change was actually made (i.e. member was already in the groups), not when the caller lacks access.

PR #775, opened 2022-07-22
ID-171 Postgres Distributed Lock

tlangs:
https://broadworkbench.atlassian.net/browse/ID-171

Sam currently uses Firestore for distributed locking across its pods. This was done before Sam had a Postgres DB, and OpenDJ was definitely not a good candidate to implement a distributed lock in.

The problem with Firestore is, when its not being used, it spins down all of its resources so that their not sitting there idle. Unfortunately, there’s a startup cost once it starts getting used again, and that startup cost is reflected in delays or inability to secure a lock. Firestore is meant to be scalable for constant high-volume document operations, not sporadic document updates like we’re using it for. We only use the lock when creating PET service accounts, and sure enough, we’ve elongated response times when new PET service accounts are being created.

Firestore has a less-than-stellar track record of getting new updates and features, and other teams (like Jade) have warned against using it, as its caused a lot of trouble for them. Now that Sam has its own Postgres DB, we should implement the locking functionality there and remove the dependency on Firestore. This will lessen the complexity and footprint of Sam, increase response times, and reduce the number of un-actionable error messages we see in our [logs](https://cloudlogging.app.goo.gl/AsSKsGPqQ9P9o8fm9)

I've copied the logic of the lock from https://github.com/broadinstitute/workbench-libs/tree/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util, but ported the backend to use Sam's Postgres DB instead of Firestore.

---

**PR checklist**


automation/src/test/scala/org/broadinstitute/dsde/workbench/sam/SamConfig.scala
```diff
@@ -6,7 +6,6 @@ object SamConfig extends CommonConfig {
   // from common: qaEmail, pathToQAPem
   object GCS extends CommonGCS {
     val appsDomain = gcsConfig.getString("appsDomain")
-    val pathToSamTestFirestoreAccountPath = gcsConfig.getString("firestoreAccountPath")
```
gpolumbo-broad, line 9: Is there a corresponding change to firecloud-develop to remove this from the config?

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20220628_distributed_lock.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="tlangs" id="add_distributed_lock">
+        <createTable tableName="DISTRIBUTED_LOCK">
+            <column name="lock_name" type="VARCHAR">
```
gpolumbo-broad, line 10: TIL - https://dba.stackexchange.com/questions/20974/should-i-add-an-arbitrary-length-limit-to-varchar-columns

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDistributedLockDAO.scala
```diff
@@ -0,0 +1,128 @@
...
+import org.broadinstitute.dsde.workbench.sam.db._
+import org.broadinstitute.dsde.workbench.sam.db.tables._
+import org.broadinstitute.dsde.workbench.sam.util.lock.DistributedLockAlgebra
+import org.broadinstitute.dsde.workbench.sam.util.DatabaseSupport
+
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import scala.concurrent.duration.FiniteDuration
+
+class PostgresDistributedLockDAO[F[_]](
```
gpolumbo-broad, line 18: …what if we de-CATS-ified this while we're here?  I know that might cause a whole bunch of downstream refactors to change return types from this thing, but for what it's doing… this class is really hard to understand.  Is there a non-expired lock in the db?  yes or no.  There are always subtleties in this stuff, but man, I don't want to have to touch anything in this class ever as it's written now.

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDistributedLockDAO.scala
tlangs: I toyed with the idea of completely re-writing the logic, but decided against it. This is already a pretty large change in terms of how Sam works, so I wanted to keep the complexity of the change lower. With the DAO implemented, it wouldn't be hard to go in and refactor the actual lock functionality, but I see that as more of a follow-on task.

Essentially, the `DistributedLock` class presented a CATS-ified API, and I didn't want to change both that API and its backend in one swell foop. It would have required a complete rewrite of the tests and a re-wiring of the things that call the lock on top of getting it to use Postgres instead of Firestore. As it stands, this PR does what it sets out to do, which addresses the problems that have been reported by replacing Firestore with Postgres. I was nervous to re-engineer the logic, especially since this started as a side-project and got promoted to a "thing we need to do".

automation/src/test/scala/org/broadinstitute/dsde/workbench/sam/SamConfig.scala
tlangs: This is actually a value populated by workbench-libs, not firecloud-develop.

gpolumbo-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -168,8 +169,6 @@ object Boot extends IOApp with LazyLogging {
           val ioFireStore = GoogleFirestoreInterpreter[IO](googleFire)
```
andy7i: thanks to this PR, seems like we could delete this (and maybe `googleFire` and then clean up imports)

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -168,8 +169,6 @@ object Boot extends IOApp with LazyLogging {
...
           // googleServicesConfig.resourceNamePrefix is an environment specific variable passed in https://github.com/broadinstitute/firecloud-develop/blob/fade9286ff0aec8449121ed201ebc44c8a4d57dd/run-context/fiab/configs/sam/docker-compose.yaml.ctmpl#L24
           // Use resourceNamePrefix to avoid collision between different fiab environments (we share same firestore for fiabs)
```
andy7i: I think we need to update this line

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/DistributedLockTable.scala
```diff
@@ -0,0 +1,21 @@
...
+import java.time.Instant
+
+final case class DistributedLockRecord(lockName: String,
+                                       lockValue: String,
+                                       expiresAt: Instant)
+
+object DistributedLockTable extends SQLSyntaxSupportWithDefaultSamDB[DistributedLockRecord] {
+  override def tableName: String = "DISTRIBUTED_LOCK"
+
+  def apply(e: ResultName[DistributedLockRecord])(rs: WrappedResultSet): DistributedLockRecord = DistributedLockRecord(
```
andy7i: what does `e` mean?

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/DistributedLockTable.scala
```diff
@@ -0,0 +1,21 @@
...
+    rs.get(e.lockName),
+    rs.get(e.lockValue),
+    rs.get(e.expiresAt)
+  )
+
+  def apply(o: SyntaxProvider[DistributedLockRecord])(rs: WrappedResultSet): DistributedLockRecord = apply(o.resultName)(rs)
```
andy7i: what does the `o` mean?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDistributedLockDAO.scala
```diff
@@ -0,0 +1,128 @@
...
+final case class DistributedLockConfig(retryInterval: FiniteDuration, maxRetry: Int)
+final case class FailToObtainLock(lock: LockAccessor) extends RuntimeException {
+  override def getMessage: String = s"can't get lock: $lock"
+}
+
+sealed trait LockStatus extends Serializable with Product
+final case object Available extends LockStatus
+final case object Locked extends LockStatus
+
+final case class LockAccessor(lockName: String, lockValue: String, expiresIn: FiniteDuration)
```
andy7i: This name confused me when I saw it in a different file. Do you think `LockDetails` might be more accurate? Or any other ideas?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDistributedLockDAO.scala
andy7i: (no changes necessary here, just some questions)
another magic wish: refactoring this in wb-libs to be DB-agnostic and BYODatabase. 🧞 

actually, would it make sense for this to live in wb-libs for the sake of reusability? Or maybe that would be premature optimization and we could start with it here and move it to wb-libs in the future if an actual need arises.

andy7i (commented):
some comments are requested changes, some are suggestions, some are questions.

tlangs:
There's no way for a library to ensure that an app has the correct DB schema, the same flavor of SQL, or even the same way of connecting, so its very hard to put this functionality in a library. I took a look at the dev and prod google projects that house the firestore instances, and only saw Sam locks in there, so I'm fairly sure its only Sam using it at this point. 

Putting the code in Sam gives us much more flexibility in terms of logging, functionality, and future refactoring. I think it makes sense to live here.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDistributedLockDAO.scala
tlangs: I think that is a more accurate name! Naming things is hard...

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/DistributedLockTable.scala
tlangs: I copied the code from other `Table` classes, but totes agree that the names could be much clearer

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDistributedLockDAO.scala
```diff
@@ -0,0 +1,128 @@
...
+    } yield ()
+  }
+
+  private[dsde] def getLockStatus(lock: LockDetails): F[LockStatus] = {
+    val res = for {
+      lockRecord <- OptionT(F.delay(retrieveLock(lock)))
+      expiresAt <- OptionT.fromOption[F](Option(lockRecord.expiresAt))
+      statusF = F.realTimeInstant.map[LockStatus] { currentTime =>
+        if (currentTime.isAfter(expiresAt)) {
+          deleteLock(lock)
```
dvoet: I think there is a race condition here: 2 threads detect an expired lock, the first proceeds to delete it then re-establish it, then the second thread deletes it.

I think reading and writing state within the same serializable transactions would help

tlangs: Good point. Moving the transaction up one layer will fix this.

tlangs: For some reason, moving the transaction up a layer caused weird connection issues in the tests. Thank goodness for tests! I've moved the lock-expiry-check logic into the actual DAO method, so that the get/delete can happen within the same transaction, and `retrieveLock` will only return an unexpired lock, or nothing.

tlangs:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDistributedLockDAO.scala
```diff
@@ -49,17 +50,8 @@ class PostgresDistributedLockDAO[F[_]](
...
-        if (currentTime.isAfter(expiresAt)) {
-          deleteLock(lock)
-          Available
-        } else
-          Locked
-      }
-      status <- OptionT.liftF[F, LockStatus](statusF)
-    } yield status
+      _ <- OptionT(F.delay(retrieveLock(lock)))
+    } yield Locked
```
andy7i: the way I'm reading this, it will always yield Locked (unless something went wrong in the for comp) which doesn't seem right. Am I missing something here?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDistributedLockDAO.scala
```diff
@@ -0,0 +1,131 @@
...
+    samsql"""delete from ${DistributedLockTable.table}
+            where ${dl.lockName} = ${lock.lockName} and ${dl.lockValue} = ${lock.lockValue}"""
+  }
+
+  private[dsde] def deleteLock(lock: LockDetails): Unit = {
+    writeDbRef.inLocalTransaction(implicit session => {
+      deleteLockSamSql(lock).update().apply()
+    })
+  }
+}
```
andy7i, line 107: it looks like `deleteLock` is now only used by `releaseLock`. Is it still useful to abstract this out like this or could we merge the two? I think these two functions are small enough (and not too complex) that we could merge them, but I also can see the value in small, simple functions. Which approach is better and why?

tlangs: I'm doing my best to separate the weird Cats lock logic from the more straightforward DAO logic. They could totally get schmunched into one, but I think preserving the logic from the data access makes sense. Plus, it makes it easier to de-catsify in the future!

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDistributedLockDAO.scala
tlangs: I've modified the `retrieveLock` function to only return a `Some(Lock)` if one is available and unexpired. If it is expired, `retrieveLock` deletes it and returns `None`. If `retrieveLock` returns `None`, then the `OptionT` will not  reach the `yield` clause, as it short-circuits and `res` is `None`. The `fold` operation on res` then provides a default `Available`. If `res` is a `Some(Locked)`, then the `fold` operation will return the `Locked`.

This could be re-written as `OptionT(F.delay(retrieveLock(lock))).map(_ => Locked).fold[LockStatus](Available)(identity)`, or simplified even further to `F.delay(retrieveLock(lock).fold[LockStatus](Available)(_ => Locked))`.

In fact, why don't I just make it the simplified version and get rid of `OptionT`...

andy7i (approved):
thanks for all your awesome patience and explanations and changes! LGTM 👍

PR #803, opened 2022-09-12
QA-2005: Add GHA in broadinstitute/sam that runs Sam-Test (in terra-github-workflows) on PR

ichengchang:
Ticket: 
[QA-2005](https://broadworkbench.atlassian.net/browse/QA-2005)
[ID-238](https://broadworkbench.atlassian.net/browse/ID-238)
---

The `build_jar.sh` and `clean_install.sh` are new scripts used by IRWG pipeline.
They are essentially just `sbt` assembly stripping away the connected and smoke test tasks.

These tests, if they are still needed, will be moved to separate workflows.

**PR checklist**


.github/workflows/sam-automation-tests.yaml
```diff
@@ -0,0 +1,118 @@
...
+      - name: dispatch to terra-github-workflows
+        uses: aurelien-baudet/workflow-dispatch@93e95b157d791ae7f42aef8f8a0d3d723eba1c31 #commit sha for v2.1.1
+        with:
+          workflow: sam-tests
+          repo: broadinstitute/terra-github-workflows
+          ref: refs/heads/main
+          token: ${{ secrets.BROADBOT_TOKEN}} # github token for access to kick off a job in the private repo
+          # manually recalculate b/c env context is broken https://github.com/actions/runner/issues/480
+          inputs: '{ "bee-name": "${{ github.workflow }}-${{ github.run_id }}", "ENV": "qa" }'
+          # inputs: '{ "bee-name": "fiab-ichang-comic-lizard", "ENV": "qa" }'
```
jyang-broad: remove old comments.

.github/workflows/sam-automation-tests.yaml
```diff
@@ -0,0 +1,118 @@
...
+
+      - name: Extract branch from PR event
+        if: github.event_name == 'pull_request'
+        id: extract-branch-pr
+        env:
+          GITHUB_CONTEXT: ${{ toJSON(github) }}
+        run: |
+          echo ::set-output name=ref::refs/heads/${{ github.head_ref }}
+
+      - name: dispatch to terra-github-workflows
```
jyang-broad: recomment `dispatch build to terra-github-workflows` to be more description from the high level.

docker/build_jar.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
jyang-broad, line 1: add a file-level comment for what this does and where it might be called. I'm assuming it's called in terra-github-workflows.

docker/clean_install.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
jyang-broad, line 1: add a file-level comment for what this does and where it might be called. I'm assuming it's called in terra-github-workflows.

jyang-broad (commented):
Looks like this is on the right path! Looks like this does a
build, tag, upload
bee create
sam test
bee destroy

I'm hesitant to thumb this b/c it looks like the build for sam is still failing (before it gets to bee-create)

.github/workflows/sam-automation-tests.yaml
```diff
@@ -0,0 +1,114 @@
...
+          GITHUB_EVENT_NAME=${{ github.event_name }}
+          if [[ "$GITHUB_EVENT_NAME" == "push" ]]; then
+            GITHUB_REF=${{ github.ref }}
+            GITHUB_SHA=${{ github.sha }}
+          elif [[ "$GITHUB_EVENT_NAME" == "pull_request" ]]; then
+            GITHUB_REF=refs/heads/${{ github.head_ref }}
+            GITHUB_SHA=${{ github.event.pull_request.head.sha }}
+          else
+            GITHUB_REF=
+            GITHUB_SHA=
```
gpolumbo-broad: Should we exit or throw some kind of error?

.github/workflows/sam-automation-tests.yaml
```diff
@@ -0,0 +1,114 @@
...
+          fi
+          echo $GITHUB_REF
+          echo $GITHUB_SHA
+          echo ::set-output name=ref::$GITHUB_REF
+          echo ::set-output name=sha::$GITHUB_SHA
+
+      - name: dispatch to terra-github-workflows
+        uses: aurelien-baudet/workflow-dispatch@93e95b157d791ae7f42aef8f8a0d3d723eba1c31 #commit sha for v2.1.1
```
gpolumbo-broad, line 39: Assume this is a docker image?  Do we need to run this by appsec to make sure it is an approved/safe image?

.github/workflows/sam-automation-tests.yaml
```diff
@@ -0,0 +1,114 @@
...
+        run: |
+          echo "$GITHUB_CONTEXT"
+          GIT_SHA=${{ steps.extract-branch.outputs.sha }}
+          HASH_TAG=${GIT_SHA:0:12}
+          # Need to escape twice because custom-version-json is passed like an cli argument string
+          echo "HASH_TAG=$HASH_TAG"
+          echo ::set-output name=custom-version-json::"{\\\"sam\\\":{\\\"appVersion\\\":\\\"$HASH_TAG\\\"}}"
+
+  create-bee-workflow:
+    runs-on: ubuntu-latest
```
gpolumbo-broad, line 60: might want to be careful using `ubuntu-latest` as it will constantly float on us.  This _might_ be what we want… But for something like this, I think we might want to use a specific LTS version to avoid the image changing/causing problems.

.github/workflows/sam-automation-tests.yaml
```diff
@@ -0,0 +1,114 @@
...
+    needs: [create-bee-workflow]
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    steps:
+      - name: dispatch to terra-github-workflows
+        uses: aurelien-baudet/workflow-dispatch@93e95b157d791ae7f42aef8f8a0d3d723eba1c31 #commit sha for v2.1.1
+        with:
+          workflow: sam-tests
```
gpolumbo-broad, line 90: Sorry if I'm missing it, but what is kicking off the tests?  What tests are running?

docker/build_jar.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -e
+
+# Get the last commit hash of the model directory and set it as an environment variable
+GIT_MODEL_HASH=$(git log -n 1 --pretty=format:%h)
+
+docker run --rm -v $PWD:/working -v jar-cache:/root/.ivy -v jar-cache:/root/.ivy2 broadinstitute/scala-baseimage:jdk11-2.13.5-1.4.7 /working/docker/clean_install.sh /working
```
gpolumbo-broad: Note Sam JDK I think was just updated to java 17, so we might want to use a new base image now.

.github/workflows/sam-automation-tests.yaml
jyang-broad: this is a public github action, which is basically a library we use. This doesn't touch anything terra, it just helps CI/CD pass data to/from other GHA

.github/workflows/sam-automation-tests.yaml
jyang-broad: When the action runs, the dispatch will send you a link the private repo the job is running from. The actual code for the running is: https://github.com/broadinstitute/terra-github-workflows/tree/main/.github/workflows

docker/build_jar.sh
jyang-broad: is there a way we can dynamically pull the "correct" java version in the sam repo?

.github/workflows/sam-automation-tests.yaml
jyang-broad: We use `ubuntu-latest` for a number of reasons, one of the most critical is that it gets patched for any security/maintenance issues automatically, unfortunately there's no floating lts version so the alternative is to pin which creates a whole slew of maintenance issues.

.github/workflows/sam-automation-tests.yaml
ichengchang: All set. Removed old comments.

.github/workflows/sam-automation-tests.yaml
ichengchang: Updated comment.

docker/clean_install.sh
ichengchang: Updated comment.

.github/workflows/sam-automation-tests.yaml
ichengchang: I updated the code to `exit 1` and echo an error message.

docker/build_jar.sh
ichengchang: Updated the base image to `hseeberger/scala-sbt:eclipse-temurin-17.0.2_1.6.2_2.13.8` so this should be OK for now.

docker/build_jar.sh
ichengchang: Updated comment.

jyang-broad (approved):
looks good! I think CI is running multiple times, should limit push trigger to mainline (develop?)

https://github.com/broadinstitute/sherlock/blob/1f43e79f1b26bdb76e21c701569d95ee567783c7/.github/workflows/test.yaml#L4

.github/workflows/sam-automation-tests.yaml
```diff
@@ -0,0 +1,113 @@
...
+          echo ::set-output name=custom-version-json::"{\\\"sam\\\":{\\\"appVersion\\\":\\\"$HASH_TAG\\\"}}"
+
+  create-bee-workflow:
+    runs-on: ubuntu-latest
+    needs: [sam-build-tag-publish-job]
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    steps:
+      - name: Check hash tag
```
gpolumbo-broad: nit:
```suggestion
      - name: Print hash tag
```
Since it's not actually doing any checking/verifying

gpolumbo-broad (approved):

.github/workflows/sam-automation-tests.yaml
ichengchang: Updated comment. Rerun tests. All 12 checks passed.

PR #804, opened 2022-09-12
ID-222: Tl logging improvements -- User Registration

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-222
Improving info/debug logging around user registration and enabling/disabling users.

Do info logs show up in prod? 

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -232,7 +232,7 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
     for {
       _ <- directoryDAO.enableIdentity(user.id, samRequestContext).unsafeToFuture()
       _ <- cloudExtensions.onUserEnable(user, samRequestContext)
-    } yield logger.info(s"Enabled new user ${user.toUserIdInfo}")
+    } yield logger.info(s"Enabled user ${user.toUserIdInfo}")
 
   val serviceAccountDomain = "\\S+@\\S+\\.gserviceaccount\\.com".r
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
@@ -14,6 +14,7 @@ import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.sam.dataAccess.DirectoryDAO
 import org.broadinstitute.dsde.workbench.sam.model._
 import org.broadinstitute.dsde.workbench.sam.service.UserService.genWorkbenchUserId
+import org.broadinstitute.dsde.workbench.sam.util.AsyncLogging.{FutureWithLogging, IOWithLogging}
 import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -32,7 +33,10 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
       _ <- enableUserInternal(createdUser, samRequestContext)
       _ <- addToAllUsersGroup(createdUser.id, samRequestContext)
       userStatus <- getUserStatus(createdUser.id, samRequestContext = samRequestContext)
-      res <- userStatus.toRight(new WorkbenchException("getUserStatus returned None after user was created")).fold(Future.failed, Future.successful)
+      res <- userStatus
+        .toRight(new WorkbenchException("getUserStatus returned None after user was created"))
+        .fold(Future.failed, Future.successful)
+        .withInfoLogMessage(s"New user ${createdUser.toUserIdInfo} was successfully created")
     } yield res
 
   def addToAllUsersGroup(uid: WorkbenchUserId, samRequestContext: SamRequestContext): Future[Unit] =
@@ -40,7 +44,7 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
 
       allUsersGroup <- cloudExtensions.getOrCreateAllUsersGroup(directoryDAO, samRequestContext)
       _ <- directoryDAO.addGroupMember(allUsersGroup.id, uid, samRequestContext).unsafeToFuture()
-    } yield ()
+    } yield logger.info(s"Added user uid ${uid.value} to the All Users group")
 
   def inviteUser(inviteeEmail: WorkbenchEmail, samRequestContext: SamRequestContext): IO[UserStatusDetails] =
     for {
@@ -65,9 +69,11 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
       updated <- subjectWithEmail match {
         case Some(uid: WorkbenchUserId) =>
           acceptInvitedUser(user, samRequestContext, uid)
+            .withInfoLogMessage(s"Accepted invited user ${user.email} with uid ${uid.value}")
 
         case None =>
           createUserInternal(user, samRequestContext)
+            .withComputedInfoLogMessage(u => s"Created user ${u.email} with uid ${u.id}")
 
         case Some(_) =>
           // We don't support inviting a group account or pet service account
@@ -226,7 +232,7 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
     for {
       _ <- directoryDAO.enableIdentity(user.id, samRequestContext).unsafeToFuture()
       _ <- cloudExtensions.onUserEnable(user, samRequestContext)
-    } yield ()
+    } yield logger.info(s"Enabled new user ${user.toUserIdInfo}")
 
   val serviceAccountDomain = "\\S+@\\S+\\.gserviceaccount\\.com".r
 
@@ -247,10 +253,13 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
   def deleteUser(userId: WorkbenchUserId, samRequestContext: SamRequestContext): Future[Unit] =
     for {
       allUsersGroup <- cloudExtensions.getOrCreateAllUsersGroup(directoryDAO, samRequestContext)
-      _ <- directoryDAO.removeGroupMember(allUsersGroup.id, userId, samRequestContext).unsafeToFuture()
+      _ <- directoryDAO
+        .removeGroupMember(allUsersGroup.id, userId, samRequestContext)
+        .unsafeToFuture()
+        .withInfoLogMessage(s"Removed $userId from the All Users group")
       _ <- cloudExtensions.onUserDelete(userId, samRequestContext)
-      deleteResult <- directoryDAO.deleteUser(userId, samRequestContext).unsafeToFuture()
-    } yield deleteResult
+      _ <- directoryDAO.deleteUser(userId, samRequestContext).unsafeToFuture()
+    } yield logger.info(s"Deleted user $userId")
 }
 
 object UserService {
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/util/AsyncLogging.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/util/AsyncLogging.scala
@@ -1,0 +1,53 @@
+package org.broadinstitute.dsde.workbench.sam.util
+
+import cats.effect.IO
+import com.typesafe.scalalogging.LazyLogging
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object AsyncLogging extends LazyLogging {
+
+  implicit class IOWithLogging[A](io: IO[A]) {
+
+    def withDebugLogMessage(message: String): IO[A] = io.<*(IO(logger.debug(message)))
+
+    def withInfoLogMessage(message: String): IO[A] = io.<*(IO(logger.info(message)))
+
+    def withComputedInfoLogMessage(messageMapper: A => String): IO[A] = io.map { a =>
+      logger.info(messageMapper.apply(a))
+      a
+    }
+
+    def withWarnLogMessage(message: String): IO[A] = io.<*(IO(logger.warn(message)))
+
+    def withErrorLogMessage(message: String): IO[A] = io.<*(IO(logger.error(message)))
+  }
+
+  implicit class FutureWithLogging[A](future: Future[A])(implicit ec: ExecutionContext) {
+
+    def withDebugLogMessage(message: String): Future[A] = future.map { a =>
+      logger.debug(message)
+      a
+    }
+
+    def withInfoLogMessage(message: String): Future[A] = future.map { a =>
+      logger.info(message)
+      a
+    }
+
+    def withComputedInfoLogMessage(messageMapper: A => String): Future[A] = future.map { a =>
+      logger.info(messageMapper.apply(a))
+      a
+    }
+
+    def withWarnLogMessage(message: String): Future[A] = future.map { a =>
+      logger.warn(message)
+      a
+    }
+
+    def withErrorLogMessage(message: String): Future[A] = future.map { a =>
+      logger.error(message)
+      a
+    }
+  }
+}


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
```diff
@@ -185,5 +194,7 @@ class LdapRegistrationDAO(
 
   override def setUserAzureB2CId(userId: WorkbenchUserId, b2CId: AzureB2CId, samRequestContext: SamRequestContext): IO[Unit] =
     executeLdap(IO(ldapConnectionPool.modify(userDn(userId), new Modification(ModificationType.REPLACE, Attr.azureB2CId, b2CId.value))), "setUserAzureB2CId", samRequestContext)
+      .withInfoLogMessage(s"Set Azure Subject ID for ${userId.value} to ${b2CId.value}")
+      .map(_ => ())
```
Ghost-in-a-Jar: The method expects an IO[Unit] back

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
```diff
@@ -89,13 +92,13 @@ class LdapRegistrationDAO(
...
       }
-    }
+    }.withInfoLogMessage(s"Enabled $subject in LDAP")
 
   override def disableIdentity(subject: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Unit] = {
     executeLdap(IO(ldapConnectionPool.modify(directoryConfig.enabledUsersGroupDn, new Modification(ModificationType.DELETE, Attr.member, subjectDn(subject)))).void, "disableIdentity", samRequestContext).recover {
       case ldape: LDAPException if ldape.getResultCode == ResultCode.NO_SUCH_ATTRIBUTE
         || ldape.getResultCode == ResultCode.NO_SUCH_OBJECT =>
-    }
+    }.withInfoLogMessage(s"Disabled $subject in LDAP")
```
Ghost-in-a-Jar: Do these WorkbenchSubject objects risk leaking any pii into logs?

Ghost-in-a-Jar:
I am thinking we don't have a ton of users registering every day i.e. these new info logs wont be generating too many new log messages.

Shakespeared (approved):
LGTM

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamModelDirectives.scala
```diff
@@ -22,7 +22,7 @@ trait SamModelDirectives {
   def withSubject(email: WorkbenchEmail, samRequestContext: SamRequestContext): Directive1[WorkbenchSubject] =
     onSuccess(userService.getSubjectFromEmail(email, samRequestContext)).map {
       case Some(subject) => subject
-      case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"${email} not found"))
+      case None => throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"Email not found"))
```
gpolumbo-broad: As we're adding new logging, I can see trying to avoid using messages with email, but I'm not sure if we want to change this one.  I'd be curious to know appsec's thoughts on this.  For this particular message, it's actually really valuable to know which email address couldn't be found.  What if we just logged part of the email?  Would that be any better?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
```diff
@@ -62,6 +63,7 @@ class LdapRegistrationDAO(
       case ldape: LDAPException if ldape.getResultCode == ResultCode.ENTRY_ALREADY_EXISTS =>
         new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Conflict, s"identity with id ${user.id} already exists"))
     } *> IO.pure(user)
+      .withInfoLogMessage(s"Created user ${user.id.value} in LDAP")
```
gpolumbo-broad: So, hopefully `LdapRegistrationDAO` is going away real soon.   These changes are fine, but `PostgresDirectoryDAO` extends `DirectoryDAO` which extends `RegistrationDAO` as well.  I don't know what is going on in `PostgresDirectoryDAO` re: logging improvements, but I think anything you're adding here for more info we probably will also want there.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -85,13 +89,13 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
       groups <- directoryDAO.listUserDirectMemberships(uid, samRequestContext)
       _ <- user.googleSubjectId.traverse { googleSubjectId =>
         for {
-          _ <- directoryDAO.setGoogleSubjectId(uid, googleSubjectId, samRequestContext)
+          _ <- directoryDAO.setGoogleSubjectId(uid, googleSubjectId, samRequestContext).withDebugLogMessage(s"Adding user $uid to directory with Google Subject ID: $googleSubjectId")
```
gpolumbo-broad:
```suggestion
          _ <- directoryDAO.setGoogleSubjectId(uid, googleSubjectId, samRequestContext).withDebugLogMessage(s"Updating User: $uid - setting Google Subject ID: $googleSubjectId")
```

Feel free to wordsmith this differently.  The point we want to convey is that we are setting the Google Subject ID for an already existing user.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -85,13 +89,13 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
           // Note: we need to store the user's Google Subject Id in LDAP
-          _ <- registrationDAO.setGoogleSubjectId(uid, googleSubjectId, samRequestContext)
+          _ <- registrationDAO.setGoogleSubjectId(uid, googleSubjectId, samRequestContext).withDebugLogMessage(s"Adding user $uid to LDAP with Google Subject ID: $googleSubjectId")
```
gpolumbo-broad: See comment above.  This should be reworded to indicate that we are setting the Google Subject ID for an already existing User record.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -85,13 +89,13 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
         } yield ()
       }
       _ <- user.azureB2CId.traverse { azureB2CId =>
-        directoryDAO.setUserAzureB2CId(uid, azureB2CId, samRequestContext)
+        directoryDAO.setUserAzureB2CId(uid, azureB2CId, samRequestContext).withDebugLogMessage(s"Adding user $uid to Azure B2C with ID: $azureB2CId")
```
gpolumbo-broad: Again, see comments above.  This should be reworded to indicate that we are setting the Azure B2C ID for an already existing User record.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -103,7 +107,9 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
     for {
       existingUser <- newWorkbenchUser match {
         case SamUser(_, Some(googleSubjectId), _, _, _, _) => directoryDAO.loadSubjectFromGoogleSubjectId(googleSubjectId, samRequestContext)
+          .withComputedDebugLogMessage(subject => s"New workbench user with id ${newWorkbenchUser.toUserIdInfo.userSubjectId} has Google subject: $subject")
```
gpolumbo-broad: Not sure what we mean when we say `Google subject` here.  Maybe just:

```suggestion
          .withComputedDebugLogMessage(subject => s"New workbench user with id ${newWorkbenchUser.toUserIdInfo.userSubjectId} WorkbenchSubject: $subject")
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -103,7 +107,9 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
         case SamUser(_, _, _, Some(azureB2CId), _, _) => directoryDAO.loadUserByAzureB2CId(azureB2CId, samRequestContext).map(_.map(_.id))
+          .withComputedDebugLogMessage(subject => s"New workbench user with id ${newWorkbenchUser.toUserIdInfo.userSubjectId} has B2C subject: $subject")
```
gpolumbo-broad: Same here.  I don't know what the best message is, but I think we can make it more clear.  Happy to pair on this if you can't think of a better message or aren't sure what I mean.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -116,8 +122,8 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
 
   private def createUserInternal(user: SamUser, samRequestContext: SamRequestContext): IO[SamUser] = {
     for {
-      createdUser <- directoryDAO.createUser(user, samRequestContext)
-      _ <- registrationDAO.createUser(user, samRequestContext)
+      createdUser <- directoryDAO.createUser(user, samRequestContext).withDebugLogMessage(s"Creating user with id ${user.toUserIdInfo.userSubjectId} in directory")
```
gpolumbo-broad:
```suggestion
      createdUser <- directoryDAO.createUser(user, samRequestContext).withDebugLogMessage(s"Creating user with id ${user.toUserIdInfo.userSubjectId}")
```

I think the simpler message is OK here.  Postgres is the source of record and so shouldn't need the additional location specified.  Subsequent line stating LDAP is good though, that makes sense and will hopefully go away soon.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -126,6 +132,7 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
   def getUserStatus(userId: WorkbenchUserId, userDetailsOnly: Boolean = false, samRequestContext: SamRequestContext): Future[Option[UserStatus]] = {
     directoryDAO.loadUser(userId, samRequestContext).unsafeToFuture().flatMap {
       case Some(user) =>
+        logger.debug(s"Found user with id ${user.id}")
```
gpolumbo-broad: Would prefer a little more context here.  I could imagine this message popping up all over the place.  
```suggestion
        logger.debug(s"getUserStatus - Found user with id ${user.id}")
```

Something like this.  Again, word it however you prefer.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -244,19 +254,21 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
           _ <- cloudExtensions.onUserDisable(user, samRequestContext)
           userStatus <- getUserStatus(user.id, samRequestContext = samRequestContext)
         } yield {
+          logger.info(s"Disabled user ${user.toUserIdInfo.userSubjectId}")
           userStatus
         }
-      case None => Future.successful(None)
+      case None => Future.successful(None).withDebugLogMessage(s"User $userId was not disabled, not found")
```
gpolumbo-broad:
```suggestion
      case None => Future.successful(None).withDebugLogMessage(s"User $userId was not found and therefore could not be disabled")
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -287,9 +299,9 @@ object UserService {
     email.value match {
       case emailString if blockedEmailDomains.exists(domain => {
         emailString.endsWith("@" + domain) || emailString.endsWith("." + domain)
-      }) => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"email domain not permitted [${email.value}]")))
+      }) => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"email domain not permitted")))
```
gpolumbo-broad: Again, appreciate the effort to try to reduce email addresses in logs, but I am worried about the usefulness of this message without the email.  Can we at least log the `domain`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/AsyncLogging.scala
```diff
@@ -0,0 +1,83 @@
+package org.broadinstitute.dsde.workbench.sam.util
+
+import cats.effect.IO
+import com.typesafe.scalalogging.LazyLogging
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object AsyncLogging extends LazyLogging {
```
gpolumbo-broad, line 8: I know logging is all side effects, but can we find a way to have some tests for this?

gpolumbo-broad (changes requested):
Overall I think we should talk to AppSec about the pros/cons of logging email addresses.  I know their default answer will be, "don't do it", but in some cases, it will make a lot of sense.  Furthermore, our Apache logs log email addresses all the time for every service, so fixing the email logging for Sam doesn't really solve the logging of emails problem.  On the contrary, I think it will make our debugging harder if we remove it while not really adding much protection for our users.  If it's something we want to fix, it's an initiative we will need to take on across the apps.  In the meantime, I'd rather not make our lives more difficult.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamModelDirectives.scala
tlangs: I'm in favor of keeping email addresses in logs. It makes it so much easier to debug. Plus, if someone has access to our logs, we have a bigger problem than the infiltrator being able to see email addresses

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
tlangs: Subject IDs are not PII

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/AsyncLogging.scala
tlangs: That's really hard to do. There's no real mechanism for monitoring the stdout or stderr without a lot of setup. Given that these methods just call the logger, and that we can rely on Scala's type system to make sure they're not affecting whatever stream they're a part of, I'm ok with not specifically testing them.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
tlangs: I would say just leave email addresses in.

tlangs:
Can we make sure we're keeping emails and subject IDs and such in the logs? Not having them will harm our debugging ability.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamModelDirectives.scala
gpolumbo-broad: Agreed, I'd rather keep the email address in, which affects a lot of this PR.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
gpolumbo-broad: Looking like we'll probably merge https://github.com/broadinstitute/sam/pull/810 before this

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -294,12 +294,12 @@ class GoogleExtensions(
       _ <- withProxyEmail(user.id) { proxyEmail =>
         googleDirectoryDAO.addMemberToGroup(proxyEmail, WorkbenchEmail(user.email.value))
       }
-      _ <- forAllPets(user.id, samRequestContext) { (petServiceAccount: PetServiceAccount) => enablePetServiceAccount(petServiceAccount, samRequestContext) }
+      _ <- forAllPets(user.id, samRequestContext)((petServiceAccount: PetServiceAccount) => enablePetServiceAccount(petServiceAccount, samRequestContext))
```
gpolumbo-broad: :squints: Looking at the history for this line is confusing… I guess I just want to make sure this is all still functionally correct.  OR if this is now functioning correctly and fixing a previously unidentified bug, then that's also worth noting because it may be indicative of something else we need to go cleanup.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -37,14 +38,17 @@ class UserService(
       _ <- enableUserInternal(createdUser, samRequestContext)
       _ <- addToAllUsersGroup(createdUser.id, samRequestContext)
       userStatus <- getUserStatus(createdUser.id, samRequestContext = samRequestContext)
-      res <- userStatus.toRight(new WorkbenchException("getUserStatus returned None after user was created")).fold(Future.failed, Future.successful)
+      res <- userStatus
+        .toRight(new WorkbenchException("getUserStatus returned None after user was created"))
+        .fold(Future.failed, Future.successful)
+        .withInfoLogMessage(s"New user with id ${createdUser.toUserIdInfo.userSubjectId} was successfully created")
```
gpolumbo-broad: Again, maybe just log the user's email?  Don't hold up the PR for this.  If we need to have further discussion in person, that's fine.  I guess I'd rather just be consistent until we're told we must do things differently.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -87,13 +95,17 @@ class UserService(
       groups <- directoryDAO.listUserDirectMemberships(uid, samRequestContext)
       _ <- user.googleSubjectId.traverse { googleSubjectId =>
         for {
-          _ <- directoryDAO.setGoogleSubjectId(uid, googleSubjectId, samRequestContext)
+          _ <- directoryDAO
+            .setGoogleSubjectId(uid, googleSubjectId, samRequestContext)
+            .withDebugLogMessage(s"Adding user $uid to directory with Google Subject ID: $googleSubjectId")
```
gpolumbo-broad:
```suggestion
            .withDebugLogMessage(s"Accepting invited user $uid - Setting Google Subject ID: $googleSubjectId")
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -87,13 +95,17 @@ class UserService(
...
           // Note: we need to store the user's Google Subject Id in LDAP
-          _ <- registrationDAO.setGoogleSubjectId(uid, googleSubjectId, samRequestContext)
+          _ <- registrationDAO
+            .setGoogleSubjectId(uid, googleSubjectId, samRequestContext)
+            .withDebugLogMessage(s"Adding user $uid to LDAP with Google Subject ID: $googleSubjectId")
```
gpolumbo-broad: bye bye ldap!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
gpolumbo-broad:
```suggestion
        directoryDAO.setUserAzureB2CId(uid, azureB2CId, samRequestContext).withDebugLogMessage(s"Accepting invited user $uid - Setting Azure B2C ID: $azureB2CId")
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -121,8 +140,10 @@ class UserService(
 
   private def createUserInternal(user: SamUser, samRequestContext: SamRequestContext): IO[SamUser] =
     for {
-      createdUser <- directoryDAO.createUser(user, samRequestContext)
-      _ <- registrationDAO.createUser(user, samRequestContext)
+      createdUser <- directoryDAO
+        .createUser(user, samRequestContext)
+        .withDebugLogMessage(s"Creating user with id ${user.toUserIdInfo.userSubjectId} in directory")
```
gpolumbo-broad: this would be a good message to include the email in.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -234,7 +260,7 @@ class UserService(
       _ <- directoryDAO.enableIdentity(user.id, samRequestContext).unsafeToFuture()
       _ <- registrationDAO.enableIdentity(user.id, samRequestContext).unsafeToFuture()
       _ <- cloudExtensions.onUserEnable(user, samRequestContext)
-    } yield ()
+    } yield logger.info(s"Enabled new user ${user.toUserIdInfo}")
```
gpolumbo-broad:
```suggestion
    } yield logger.info(s"Enabled user ${user.toUserIdInfo}")
```
Not sure if this is only called for "new" users or if we can call it to reenable a previously disabled user.

gpolumbo-broad (changes requested):

gpolumbo-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamModelDirectives.scala
tlangs: Can we get the email address added back in?

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutes.scala
```diff
@@ -87,6 +87,6 @@ trait AzureRoutes extends SecurityDirectives {
     onSuccess(azureService.flatTraverse(_.getSamUser(email, samRequestContext))).flatMap {
       case Some(samUser) => provide(samUser)
       case None =>
-        failWith(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"User ${email.value} not found")))
+        failWith(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"User not found")))
```
tlangs: Add email back in

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
tlangs: Is this no longer needed?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/LdapRegistrationDAO.scala
```diff
@@ -103,7 +107,7 @@ class LdapRegistrationDAO(
           ).void
         case ldape: LDAPException if ldape.getResultCode == ResultCode.ATTRIBUTE_OR_VALUE_EXISTS => IO.unit
       }
-    }
+    }.withInfoLogMessage(s"Enabled $subject in LDAP")
```
tlangs: No longer needed. Maybe needs a rebase?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -290,8 +322,8 @@ object UserService {
   def validateEmailAddress(email: WorkbenchEmail, blockedEmailDomains: Seq[String]): IO[Unit] =
     email.value match {
       case emailString if blockedEmailDomains.exists(domain => emailString.endsWith("@" + domain) || emailString.endsWith("." + domain)) =>
-        IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, s"email domain not permitted [${email.value}]")))
```
tlangs: Leave the emails!

tlangs (approved):

PR #807, opened 2022-09-14
[SU-213] Support leaving managed groups

MatthewBemis:
Ticket:  https://broadworkbench.atlassian.net/browse/SU-213

We recently added a leaveResource endpoint to Sam, but only turned it on for the `workspace` resource type. This adds support for the `managed-group` resource type and fixes some logic that was somewhat specific to managed groups.

---

**PR checklist**


MatthewBemis:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -503,17 +503,17 @@ class ResourceService(
       samRequestContext: SamRequestContext
   ): IO[List[Boolean]] =
     accessPolicyDAO.listAccessPolicies(resourceId, samRequestContext) flatMap { policiesForResource =>
```
MatthewBemis, line 505: some minor logic changes below. technically, every managed group has a public policy (the admin notifiers policy). so with the previous implementation, it was impossible to leave a managed group because it looked for the presence of a single public policy and rejected the request if it found one. the new logic checks to see if the user has additional policies on the resource beyond the public policy, and allows the user to leave them if so.

nawatts (approved):

src/main/resources/reference.conf
```diff
@@ -480,6 +480,7 @@ resourceTypes = {
         roleActions = ["notify_admins"]
       }
     }
+    allowLeaving = true
```
Ghost-in-a-Jar, line 483: does this mean that all resources will allow leaving by default? Adn I suppose we can set it differently for different envs?

Ghost-in-a-Jar (approved):

src/main/resources/reference.conf
MatthewBemis: This just enables it for the `managed-group` type (the default value for allowLeaving if not specified for a type is false). We decided to implement this at the resource type level because we didn't want to introduce the functionality to other resource types that might not be expecting this behavior to be possible. We started with workspaces, now groups, and our third and final target will probably be billing projects.

And yeah- we could totally do this on a per-env basis as well if we wanted to have some more granularity

PR #818, opened 2022-09-20
skip tests - ID-243 portable smoke tests

gpolumbo-broad:
Ticket: https://broadworkbench.atlassian.net/browse/ID-243
Added python based smoke tests that can be executed anywhere and point at any Sam host to do basic/minimal checking that a Sam instance is up and running.

---

**PR checklist**


smoke_test/.python-version
```diff
@@ -0,0 +1 @@
+venv-3.10.3-sam
```
andy7i: the readme instructions run the command from the sam root folder. My pyenv does not find this `.python-version` file unless I am in the `smoke_test` folder. I think we need to either change the instructions or move this file to `sam/`

smoke_test/tests/sam_smoke_tests.py
```diff
@@ -0,0 +1,23 @@
+from functools import cache
+from unittest import TestCase
+from urllib.parse import urljoin
+
+import requests
+from requests import Response
+
+
+class SamSmokeTests(TestCase):
```
andy7i: It's confusing having a smoke_test.py and sam_smoke_tests.py, and it's unclear what the purpose of the file is without reading it.

I think we could rename the former to something like `test_runner.py`. Or rename this file to something like `test_helpers.py`. Thoughts?

README.md
```diff
@@ -252,5 +252,17 @@ https://local.broadinstitute.org:50443/status](https://local.broadinstitute.org:
 [Swagger page:
 https://local.broadinstitute.org:50443/#/](https://local.broadinstitute.org:50443/#/)
 
+#### Verify that Sam is running anywhere
+Sam includes smoke tests that can be run anywhere and test a running Sam instance running anywhere.  These tests run
+quickly and in 2 modes: authenticated and unauthenticated.
```
andy7i: Out of curiosity, could we have implemented a detailed status check as an API? I think there would be a few benefits to do it that way, but I guess it wouldn't simulate a user making the same calls.

smoke_test/.python-version
andy7i: I reran the command from this folder and now I get:
```
python smoke_test.py https://sam.dsde-dev.broadinstitute.org/
pyenv: version `venv-3.10.3-sam' is not installed (set by /Users/ajang/Firecloud/sam/smoke_test/.python-version)
```

I wonder if we could save a python virtualenv to the project so that anyone/everyone is using the same version of python.

smoke_test/smoke_test.py
```diff
@@ -0,0 +1,99 @@
+import argparse
+import sys
+import unittest
+from unittest import TestSuite
+
+import requests
```
andy7i, line 6: import error
```
▶ python smoke_test.py https://sam.dsde-dev.broadinstitute.org/
Traceback (most recent call last):
  File "/Users/ajang/Firecloud/sam/smoke_test/smoke_test.py", line 6, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
(venv-3.10.3-sam)
```

we might need a requirements.txt?

andy7i (changes requested):
partial review since I'm unable to run the tests

smoke_test/tests/authenticated/sam_resource_types_tests.py
```diff
@@ -0,0 +1,20 @@
...
+    def test_status_code_is_200(self):
+        response = SamSmokeTests.call_sam(self.resource_types_url(), SamSmokeTests.USER_TOKEN)
+        self.assertEqual(response.status_code, 200, f"Resource Types HTTP Status is not 200: {response.text}")
+
+    # def test_content_contains_correct_keys(self):
+    #     response = SamSmokeTests.call_sam(self.user_info_url(), SamSmokeTests.USER_TOKEN)
+    #     user_info = json.loads(response.text)
+    #     self.assertListEqual(list(user_info.keys()),
+    #                          ["adminEnabled", "enabled", "userEmail", "userSubjectId"],
+    #                          "User Info object does not match expected fields")
```
andy7i: looks like this was moved to `smoke_test/tests/authenticated/sam_user_info_tests.py`

smoke_test/.python-version
gpolumbo-broad: Just made a commit with several improvements for this.  With the way `pyenv virtualenv` works, I think it makes sense to _not_ commit a `.python-version` and instead just state the required version in the `README`.  It will be up to the user to make sure that they install the right python version using whatever means they want.   I also added a `requirements.txt` because that was missing and was needed.

smoke_test/tests/sam_smoke_tests.py
gpolumbo-broad: Yeah, I admit that it is possibly a little confusing.  My logic is that most normal python `unittest` suites extend `TestCase`.  In this case, there's a little extra configuration I want each Sam Smoke Test Case to share so I want each test to extend this.  I just renamed the class and pushed up the change.  Let me know if this is a little bit better.  I think it still leaves the possibility for some confusion, but I think it is also the most specific name I can give it.  Happy to keep trying though if you think it's still not right.

README.md
gpolumbo-broad: Yeah, 100% think we should make the `/status` endpoint better but I think that's also just another ticket.  I don't disagree that it would be great if the `/status` endpoint just tells us everything, but the reality is, sometimes, we still don't trust it and go in and manually poke at things.  These tests are another alternative that should make the "go click around" test trivially easy.

smoke_test/smoke_test.py
```diff
@@ -0,0 +1,99 @@
...
+        # Need to pop off sys.argv values to avoid messing with args passed to unittest.main()
+        for _ in range(len(sys.argv[1:])):
+            sys.argv.pop()
+
+        main(args)
+        sys.exit(0)
+
+    except Exception as e:
+        print(e)
+        sys.exit(1)
```
Ghost-in-a-Jar, line 99: This looks great, just FYI I have been using https://pypi.org/project/pydantic-cli/ instead of argparse recently and I like it a lot better.

smoke_test/tests/authenticated/sam_resource_types_tests.py
```diff
@@ -0,0 +1,20 @@
...
+    def test_status_code_is_200(self):
+        response = SamSmokeTestCase.call_sam(self.resource_types_url(), SamSmokeTestCase.USER_TOKEN)
+        self.assertEqual(response.status_code, 200, f"Resource Types HTTP Status is not 200: {response.text}")
+
+    # def test_content_contains_correct_keys(self):
+    #     response = SamSmokeTests.call_sam(self.user_info_url(), SamSmokeTests.USER_TOKEN)
+    #     user_info = json.loads(response.text)
+    #     self.assertListEqual(list(user_info.keys()),
+    #                          ["adminEnabled", "enabled", "userEmail", "userSubjectId"],
+    #                          "User Info object does not match expected fields")
```
Ghost-in-a-Jar: why is this one commented?

smoke_test/tests/authenticated/sam_user_info_tests.py
```diff
@@ -0,0 +1,20 @@
...
+    def test_status_code_is_200(self):
+        response = SamSmokeTestCase.call_sam(self.user_info_url(), SamSmokeTestCase.USER_TOKEN)
+        self.assertEqual(response.status_code, 200, f"User info HTTP Status is not 200: {response.text}")
+
+    def test_content_contains_correct_keys(self):
+        response = SamSmokeTestCase.call_sam(self.user_info_url(), SamSmokeTestCase.USER_TOKEN)
+        user_info = json.loads(response.text)
+        self.assertListEqual(list(user_info.keys()),
+                             ["adminEnabled", "enabled", "userEmail", "userSubjectId"],
+                             "User Info object does not match expected fields")
```
Ghost-in-a-Jar, line 20: is it worth maybe creating/deleting a user, something like that?

another thought is we create a test user and auth them successfully for a test

Ghost-in-a-Jar (approved):
some comments

smoke_test/tests/sam_smoke_tests.py
andy7i: I think it does help! thanks for the update

README.md
andy7i: thanks for the explanation!

smoke_test/tests/sam_smoke_test_case.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class SamSmokeTestCase(TestCase):
+    SAM_HOST = None
+    USER_TOKEN = None
+
+    @staticmethod
+    def build_sam_url(path: str) -> str:
+        assert SamSmokeTestCase.SAM_HOST, "ERROR - SamSmokeTests.SAM_HOST not properly set"
+        return urljoin(f"https://{SamSmokeTestCase.SAM_HOST}", path)
```
andy7i: tested this out with command:
```
▶ python smoke_test.py https://sam.dsde-dev.broadinstitute.org
```

and got this unhelpful error:
```
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='https', port=443): Max retries exceeded with url: /version (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x103557160>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))
```

turns out it is doubling up on `https://`.

I think this could be personal preference, but since I'm usually going to copy the URL from the browser (which automatically copies the scheme), I think it would be helpful to remove the `https://` prefix here

```suggestion
        return urljoin(f"{SamSmokeTestCase.SAM_HOST}", path)
```
(tested the change locally and it works)

another reason to make this change is that the error message is much better if a user leaves out the scheme prefix:
```
requests.exceptions.MissingSchema: Invalid URL '/version': No scheme supplied. Perhaps you meant http:///version?
```

(Also, running against localhost might use http instead of https but I'm not 100% sure about that.)

smoke_test/.python-version
andy7i: looks great / works great! thanks!

smoke_test/README.md
```diff
@@ -0,0 +1,38 @@
...
+```pip install -r requirements.txt```
+
+## Run
+
+The smoke tests have 2 different modes that they can run in: authenticated or unauthenticated.  The mode will be 
+automatically selected based on the arguments you pass to `smoke_test.py`.
+
+To run the _unauthenticated_ smoke tests:
+
+```python smoke_test.py {SAM_HOST}```
```
Shakespeared, line 27: sorry if this is a dumb question but… what are some examples of values acceptable as `SAM_HOST`? I'm trying to run the tests but I don't quite know what to put as an arg here

smoke_test/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+To run the _authenticated_ smoke tests:
```
andy7i: nit

someone (definitely not me!) might think they have to run `smoke_test.py` twice, with and without a token, in order to run all tests

I think we can make it more immediately clear. Here's a suggested edit:

```suggestion
To run all (authenticated and unauthenticated) smoke tests:
```

smoke_test/smoke_test.py
```diff
@@ -0,0 +1,99 @@
...
+            "-v",
+            "--verbosity",
+            type=int,
+            choices=[0, 1, 2],
+            default=1,
+            help="""Python unittest verbosity setting: 
+0: Quiet - Prints only number of tests executed
+1: Minimal - (default) Prints number of tests executed plus a dot for each success and an F for each failure
+2: Verbose - Help string and its result will be printed for each test"""
+        )
```
andy7i, line 76: cool!

smoke_test/tests/authenticated/sam_user_info_tests.py
andy7i: We don't have APIs for deleting users.

I like the idea of adding a test for creating a user but feel conflicted about it. I think the tests were intended to cover common test scenarios. Creating a user could be useful but it's typically not something we test. But it is an essential part of the system and is a bit annoying to test, which makes it a good candidate for automating. Maybe we start with these tests for now and add that later?

smoke_test/tests/sam_smoke_test_case.py
```diff
@@ -0,0 +1,23 @@
...
+
+    @staticmethod
+    @cache
+    def call_sam(url: str, user_token: str = None) -> Response:
+        """Function is memoized so that we only make the call once"""
```
andy7i, line 25: interesting! so this means that we can have multiple test cases for a single api call?

smoke_test/tests/unauthenticated/sam_version_tests.py
```diff
@@ -0,0 +1,18 @@
+import json
+
+from ..sam_smoke_test_case import SamSmokeTestCase
+
+
+class SamVersionTests(SamSmokeTestCase):
```
andy7i, line 6: what do you see as the benefit in checking the version endpoint?

andy7i (commented):
I LOVE how fast this runs. Setting up takes 10-100x longer. 
```
Ran 7 tests in 0.731s
```

tested with 
`https://sam.dsde-dev.broadinstitute.org/` and `https://sam.dsde-dev.broadinstitute.org:443`

did not test locally

smoke_test/README.md
andy7i: great question!
try: `python smoke_test.py sam.dsde-dev.broadinstitute.org`

I like the idea of adding a usable example pointing to dev or prod here (to save folks a step or two)!

smoke_test/smoke_test.py
gpolumbo-broad: TIL - thanks

smoke_test/tests/authenticated/sam_resource_types_tests.py
gpolumbo-broad: Because copy pasta.  Removed.

smoke_test/tests/authenticated/sam_user_info_tests.py
gpolumbo-broad: My take:

1.  I don't want the smoke tests to affect the state of the data or the service in any way.  This will avoid conflicts of multiple people running these tests at the same time or the potential of a test failing to clean up after itself and leaving the database in a partially cleaned up state, etc.  Additionally, by not worrying about modifying state in any way, these tests become much more portable.  They require nothing to run (beyond python) and can target any environment.  The only requirement for the test runner is that they _may optionally_ choose to provide a user access token.  
2.  Smoke tests are not end-to-end (E2E) or full functional tests of the application.  These are intended to be small, quick idempotent tests that just "sniff" the app to see that it generally is up and can perform the most basic functionality.  If we want to write more complex Sam E2E tests, we can do that, but I would prefer those to be a separate suite/script because these will be much more involved and how/when we run them will be very different.

smoke_test/tests/authenticated/sam_resource_types_tests.py
gpolumbo-broad: fixed

smoke_test/tests/sam_smoke_test_case.py
gpolumbo-broad: fixed to optionally allow the protocol to be specified

smoke_test/README.md
gpolumbo-broad: good nit.  fixed.

smoke_test/tests/sam_smoke_test_case.py
gpolumbo-broad: indeed, saves making repetitive calls to reduce load on Sam and improve performance.

smoke_test/tests/unauthenticated/sam_version_tests.py
gpolumbo-broad: It's just something we can hit without authenticating.  Doesn't add a _ton_ of value, but also doesn't cost much either.  Happy to remove it if we think it's silly.

smoke_test/README.md
gpolumbo-broad: Added dev hosts to the README

smoke_test/tests/authenticated/sam_user_info_tests.py
Ghost-in-a-Jar: Agreed sounds good.

andy7i (approved):
tested the same testcases again and LGTM! thanks for addressing all of the feedback!

Shakespeared (commented):
LGTM!

PR #823, opened 2022-09-25
[TOAZ-163] Add Sam checks to Landing Zone Service 

rtitle:
Ticket: https://broadworkbench.atlassian.net/browse/TOAZ-163

Per agreed upon design, adds a `landing-zone` resource as a child of `spend-profile`. Sam checks are implemented in the LZ service: https://github.com/DataBiosphere/terra-landing-zone-service/pull/22

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1233,11 +1233,14 @@ resourceTypes = {
       read_policies = {
         description = "view all policies and policy details for this landing zone"
       }
+      set_parent = {
+        description = "set the parent spend-profile of this landing zone"
+      }
     }
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["delete", "list_resources", "read_policies"]
+        roleActions = ["delete", "list_resources", "read_policies", "set_parent"]
       }
       user = {
         roleActions = ["list_resources"]
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1001,9 +1001,6 @@ resourceTypes = {
       "add_child" = {
         description = "add a child resource"
       }
-      "remove_child" = {
-        description = "remove a child resource"
-      }
       "list_children" = {
         description = "list child resources"
       }
@@ -1014,7 +1011,7 @@ resourceTypes = {
         descendantRoles = {
           landing-zone = ["owner"]
         }
-        roleActions = ["update_billing_account", "update_metadata", "delete", "link", "share_policy::owner", "share_policy::user", "read_policies", "add_child", "remove_child", "list_children"]
+        roleActions = ["update_billing_account", "update_metadata", "delete", "link", "share_policy::owner", "share_policy::user", "read_policies", "add_child", "list_children"]
       }
       user = {
         descendantRoles = {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -998,6 +998,15 @@ resourceTypes = {
       "alter_policies" = {
         description = "Can alter policies"
       }
+      "add_child" = {
+        description = "add a child resource"
+      }
+      "remove_child" = {
+        description = "remove a child resource"
+      }
+      "list_children" = {
+        description = "list child resources"
+      }
     }
     ownerRoleName = "owner"
     roles = {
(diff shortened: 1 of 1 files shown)


rtitle:
I tested this end to end on a BEE, it seems to work. I was able to confirm that owner and user permissions propagate down from `spend-profile` to `landing-zone`.

src/main/resources/reference.conf
```diff
@@ -1207,6 +1222,32 @@ resourceTypes = {
...
+      delete = {
+        description = "delete this landing zone"
+      }
+      list_resources = {
+        description = "list resources in this landing zone"
+      }
+      read_policies = {
+        description = "view all policies and policy details for this landing zone"
+      }
+      set_parent = {
```
rtitle, line 1233: I got a little tripped up that I apparently need the magic `set_parent` and `add_child` actions in order for the hierarchical resource to work.

Ghost-in-a-Jar (approved):

rtitle:
`create-bee-workflow` failed after a couple retries, this seems new, not sure why..

src/main/resources/reference.conf
```diff
@@ -998,14 +998,29 @@ resourceTypes = {
       "alter_policies" = {
         description = "Can alter policies"
       }
+      "add_child" = {
+        description = "add a child resource"
+      }
+      "remove_child" = {
```
dvoet: do we want to allow landing zones to be removed?

src/main/resources/reference.conf
```diff
@@ -998,14 +998,29 @@ resourceTypes = {
...
+          landing-zone = ["owner"]
+        }
+        roleActions = ["update_billing_account", "update_metadata", "delete", "link", "share_policy::owner", "share_policy::user", "read_policies", "add_child", "remove_child", "list_children"]
       }
       user = {
-        roleActions = ["link", "share_policy::user"]
+        descendantRoles = {
+          landing-zone = ["user"]
+        }
+        roleActions = ["link", "share_policy::user", "add_child"]
```
dvoet, line 1020: I know this has not changed, but it seems weird that users can add other users

src/main/resources/reference.conf
dvoet: `set_parent` also allows changing the parent. If we want "set once" semantics you need to set the parent in the resource creation call. Then you don't need this action.

src/main/resources/reference.conf
rtitle: Landing zones can be deleted by the owner. Not sure if there is a need for `remove_child` at the parent level though, since billing profile owners inherit owner on the landing zone. I can remove this..

src/main/resources/reference.conf
rtitle: Hmm I agree, and it goes against the spec: https://docs.google.com/document/d/1AZq0_BQLRx68skieQmdHha8AGY_hcEO2WnKyy8qMmzw/edit#heading=h.tpa5qq630pjp

I can try to convince myself nothing will break then remove this..

src/main/resources/reference.conf
rtitle: I am setting the parent as part of the resource creation call, I think I still got an exception without `set_parent` though. Not 100% sure, will double check..

src/main/resources/reference.conf
dvoet: deleting a LZ is different than removing its parent

src/main/resources/reference.conf
rtitle: @ddietterich just checking, does anything on the Verily side break if billing-profile users can't add other users? I don't think anything in Terra CWB relies on this..

src/main/resources/reference.conf
dvoet: ok, [I think you are correct](https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala#L77), `set_parent` is required. But I still think you should set the parent on creation. If you allow `remove_child` from the parent then the LZ can be orphaned or moved which I don't think we want.

rtitle: Ok cool, I will keep `set_parent`, remove `remove_child`, and set the parent at resource creation. 👍

src/main/resources/reference.conf
dvoet: is this possibly a remnant of how `billing-project`s work? Workspace owner who may only be `billing-project` users can add workspace writers which grants them compute access or something?

dvoet (approved):

src/main/resources/reference.conf
rtitle: On second thought I'm going to ticket & look into this separately: https://broadworkbench.atlassian.net/browse/TOAZ-220

PR #831, opened 2022-09-30
QA-2028: GHA workflow for triggering Sam smoke tests. 

ichengchang:
Ticket: <Link to Jira ticket> [QA-2028](https://broadworkbench.atlassian.net/browse/QA-2028)

Implemented Sam workflow to dispatch smoke tests to private repo `terra-github-workflows`.

---

**PR checklist**


gpolumbo-broad (commented):
I think I need a little more clarification on where and how this is going to run.  The smoke tests themselves, by design, run very quickly, <1 sec.  My hope was that this would make them easy to to add to some other job or process.  As for needing Python to run, I can appreciate where that might complicate things.  However, this job seems to be creating and destroying a BEE as part of the workflow, which I do not understand.  The smoke tests do not need a BEE to run and they should be executable against any environment (live or BEE or local).  

So, my apologies if I'm just misunderstanding, but I just feel like this job is doing more than I'd expect and running at different times than I'd expect.

ichengchang:
> I think I need a little more clarification on where and how this is going to run. The smoke tests themselves, by design, run very quickly, <1 sec. 

My hope was that this would make them easy to to add to some other job or process. 

How about we make this workflow callable from another job by using the following trigger?
```
on:
  workflow_call:
  ...
```

As for needing Python to run, I can appreciate where that might complicate things. However, this job seems to be creating and destroying a BEE as part of the workflow, which I do not understand. The smoke tests do not need a BEE to run and they should be executable against any environment (live or BEE or local).
> 
> So, my apologies if I'm just misunderstanding, but I just feel like this job is doing more than I'd expect and running at different times than I'd expect.

@gpolumbo-broad How about using a `workflow_call` trigger (see my inline comments above)?

gpolumbo-broad:
@ichengchang I'm not terribly familiar with GHA and Workflows, but I think this is more in line with what I was thinking.  It should be a simple, minimal job that we can easily tack onto other workflows.  If someone just wants to run a "Sam Smoke Test" job against Prod or Staging or something, GHA may not be the place where we want that to happen.  However, for deployments, it seems reasonable.  The catch is that for a deployment, for the smoke tests to pass, you have to wait for the service to fully start.  Maybe that's what we want, maybe not, I haven't thought about it that much yet.  Maybe the better place to run these tests is part of the `k8s` liveness/readiness checks?  I don't know.

For now though, adding a job that can optionally call if we want to seems perfectly reasonable seeing as you've already done the work.  If we decide we never really want to use the Smoke Tests from GHA, that's fine, easier to just delete it in the future than to recreate this job in the future if we decide that we need it.

ichengchang:
> @ichengchang I'm not terribly familiar with GHA and Workflows, but I think this is more in line with what I was thinking. It should be a simple, minimal job that we can easily tack onto other workflows. If someone just wants to run a "Sam Smoke Test" job against Prod or Staging or something, GHA may not be the place where we want that to happen. However, for deployments, it seems reasonable. The catch is that for a deployment, for the smoke tests to pass, you have to wait for the service to fully start. Maybe that's what we want, maybe not, I haven't thought about it that much yet. Maybe the better place to run these tests is part of the `k8s` liveness/readiness checks? I don't know.
> 
> For now though, adding a job that can optionally call if we want to seems perfectly reasonable seeing as you've already done the work. If we decide we never really want to use the Smoke Tests from GHA, that's fine, easier to just delete it in the future than to recreate this job in the future if we decide that we need it.

Understood! I'll remove the BEE code from the smoke test workflow and instead make it callable from another workflow with Sam host as a parameter.

ichengchang:
@gpolumbo-broad @jyang-broad Revamped the smoke test workflow using `workflow_call` trigger. Added smoke test as part of `sam-automation-tests` workflow.

See workflow run [https://github.com/broadinstitute/sam/actions/runs/3191740842](https://github.com/broadinstitute/sam/actions/runs/3191740842)

jyang-broad (approved):
Looks good!
Recommend renaming this workflow as it does the actual build + upload of the artifact in addition to all the testing.

Other misc changes:
- adds `bumper` to this job as the action runs against the repo it's located in, previously it was in `terra-github-workflows` which meant it was bumping against terra-github-workflows code.
- adds matrix workflow to create-bee-workflow, sam-automation-test-job, sam-smoke-test-job, bee-destroy-workflow
- adds sam-smoke-test-job to run in parallel to sam-automation-test-job

jyang-broad:
on second look, is there a reason to maintain a separate sam-smoke-test file in `broadinstitute/sam` since it's just a dispatch to `terra-github-workflows`?

.github/workflows/sam-automation-tests.yaml
```diff
@@ -42,7 +53,7 @@ jobs:
           repo: broadinstitute/terra-github-workflows
           ref: refs/heads/main
           token: ${{ secrets.BROADBOT_TOKEN}} # github token for access to kick off a job in the private repo
-          inputs: '{ "repository": "${{ github.event.repository.full_name }}", "ref": "${{ steps.extract-branch.outputs.ref }}" }'
+          inputs: '{ "repository": "${{ github.event.repository.full_name }}", "ref": "${{ steps.extract-branch.outputs.ref }}", "sam-release-tag": "${{ steps.tag.outputs.tag }}" }'
```
gpolumbo-broad, line 56: I'm assuming that this workflow builds AND releases Sam?  My initial reaction is that these should be 2 separate steps.  But if they're one step, that seems fine too I guess, I don't know.  If it's easy to separate them, I guess I would.  If it's a pain, then don't worry about it.

.github/workflows/sam-automation-tests.yaml
```diff
@@ -54,9 +65,12 @@ jobs:
           HASH_TAG=${GIT_SHA:0:12}
           # Need to escape twice because custom-version-json is passed like an cli argument string
           echo "HASH_TAG=$HASH_TAG"
-          echo ::set-output name=custom-version-json::"{\\\"sam\\\":{\\\"appVersion\\\":\\\"$HASH_TAG\\\"}}"
+          echo ::set-output name=custom-version-json::"{\\\"sam\\\":{\\\"appVersion\\\":\\\"${{ steps.tag.outputs.tag }}\\\"}}"
```
gpolumbo-broad, line 64: With this change, do we still need lines 65 and 67?

.github/workflows/sam-automation-tests.yaml
```diff
@@ -92,9 +110,26 @@ jobs:
           ref: refs/heads/main
           token: ${{ secrets.BROADBOT_TOKEN}} # github token for access to kick off a job in the private repo
           # manually recalculate b/c env context is broken https://github.com/actions/runner/issues/480
-          inputs: '{ "bee-name": "${{ github.workflow }}-${{ github.run_id }}", "ENV": "qa" }'
+          inputs: '{ "bee-name": "${{ github.event.repository.name }}-${{ github.run_id }}-${{ matrix.terra-env }}", "ENV": "${{ matrix.testing-env }}" }'
+
+  sam-smoke-test-job:
```
gpolumbo-broad: `sam-smoke-test-job` should be run before`sam-automation-test-job` and should not run automation if the smoke test fails.  That's the whole point of the smoke tests, they're a way to to super quickly do the simplest possible, "Is Sam up and running" check possible.  No sense in asking Sam if it is ready to handle requests _after_ we just ran the whole test suite that sends thousands of requests 😄

.github/workflows/sam-smoke-tests.yaml
```diff
@@ -1,94 +1,27 @@
...
-    branches:
-      - develop
-    paths-ignore: ['**.md']
+  workflow_call:
+    inputs:
+      sam-hostname:
+        description: 'a generic hostname for Sam'
+        required: true
+        type: string
+      ENV:
```
gpolumbo-broad: FWIW, the smoke tests do not need an `ENV` variable.  All it needs is the `sam-hostname`.  That being said, I see that you're passing in a SA email here too, so if the `ENV` is needed in order to render the correct access token for the `smoke_test.py` script, then that's fine.

.github/workflows/sam-smoke-tests.yaml
```diff
@@ -1,94 +1,27 @@
...
+        description: 'test environment (qa, alpha)'
```
gpolumbo-broad:
```suggestion
        description: 'test environment (dev, staging, qa, alpha)'
```
I think we should either list all possible environments here, or none.

.github/workflows/sam-smoke-tests.yaml
```diff
@@ -1,94 +1,27 @@
...
+        required: true
+        type: string
+      auth-service-account-email:
+        description: 'email of auth service account to run smoke test'
```
gpolumbo-broad:
```suggestion
        description: 'email address of Sam user that will be use for testing authenticated endpoints'
```
My first reaction here is to not call this a `service-account` email, but just an `email`.  The reality for the smoke tests is that they just need an access token for an enabled user on the Sam host that is being tested.  It does not need to be an SA.  Now, in this particular run-context on GHA, you may have particular criteria in the downstream job(s) that require this to be an SA, and that's fine with me.  Whatever you think makes the most sense.

.github/workflows/sam-smoke-tests.yaml
```diff
@@ -99,29 +32,7 @@ jobs:
...
-      - name: dispatch to terra-github-workflows
-        uses: aurelien-baudet/workflow-dispatch@93e95b157d791ae7f42aef8f8a0d3d723eba1c31 #commit sha for v2.1.1
-        with:
-          workflow: bee-destroy
-          repo: broadinstitute/terra-github-workflows
-          ref: refs/heads/main
-          token: ${{ secrets.BROADBOT_TOKEN}} # github token for access to kick off a job in the private repo
-          # manually recalculate b/c env context is broken https://github.com/actions/runner/issues/480
-          inputs: '{ "bee-name": "${{ github.event.repository.name }}-${{ matrix.terra-env }}-${{ github.run_id }}" }'
+          inputs: '{ "sam-hostname": "${{ inputs.sam-hostname }}", "ENV": "qa", "auth-service-account-email": "${{ inputs.auth-service-account-email }}" }'
```
gpolumbo-broad: Double check the `ENV` var being passed in here, I think you want to use the variable that was passed in, not the hardcoded `qa` value.

gpolumbo-broad (changes requested):

ichengchang:
> on second look, is there a reason to maintain a separate sam-smoke-test file in `broadinstitute/sam` since it's just a dispatch to `terra-github-workflows`?

Makes sense! I just moved the smoke test code inside `sam-automation-tests` and dispatch from there.

.github/workflows/sam-automation-tests.yaml
ichengchang: Cleaned up the unused code.

.github/workflows/sam-automation-tests.yaml
ichengchang: Exactly! I have changed the order of the tests to run smoke test first.

.github/workflows/sam-smoke-tests.yaml
ichengchang: @jyang-broad suggested dropping this file since we already have `sam-smoke-tests.yaml` [here](https://github.com/broadinstitute/terra-github-workflows/blob/main/.github/workflows/sam-smoke-tests.yaml) so will change the description there instead.

.github/workflows/sam-smoke-tests.yaml
ichengchang: Yes, this hardcode value has been replaced by the matrix variable `${{ ${{ matrix.testing-env }}` in the refactored workflow `sam-automation-tests.yaml`.

.github/workflows/sam-automation-tests.yaml
ichengchang: Great suggestion! We can make build and release as separate steps within the same job. Will create a separate ticket for this.

.github/workflows/sam-smoke-tests.yaml
ichengchang: Yes that's correct. For now, let's keep the `ENV` variable just incase we need it.

.github/workflows/sam-smoke-tests.yaml
ichengchang: Updated the variable name to `smoke-test-user-email` to clearly state its purpose.

.github/workflows/sam-build-tag-publish-and-run-tests.yaml
```diff
@@ -105,33 +101,42 @@ jobs:
...
-          workflow: sam-tests
+          workflow: sam-smoke-tests
           repo: broadinstitute/terra-github-workflows
           ref: refs/heads/main
-          token: ${{ secrets.BROADBOT_TOKEN}} # github token for access to kick off a job in the private repo
+          token: ${{ secrets.BROADBOT_TOKEN }} # github token for access to kick off a job in the private repo
           # manually recalculate b/c env context is broken https://github.com/actions/runner/issues/480
-          inputs: '{ "bee-name": "${{ github.event.repository.name }}-${{ github.run_id }}-${{ matrix.terra-env }}", "ENV": "${{ matrix.testing-env }}" }'
+          # The smoke tests just need an access token for an enabled user on the Sam host that is being tested.
+          # inputs: '{ "sam-hostname": "sam.${{ github.event.repository.name }}-${{ github.run_id }}.bee.envs-terra.bio", "ENV": "qa", "auth-service-account-email": "firecloud-qa@broad-dsde-qa.iam.gserviceaccount.com" }'
```
gpolumbo-broad: delete old line

gpolumbo-broad (approved):

.github/workflows/sam-build-tag-publish-and-run-tests.yaml
ichengchang: All set. Thanks!

PR #849, opened 2022-10-13
[ID-258] add grafana metrics

andy7i:
https://broadworkbench.atlassian.net/browse/ID-258

Initial setup of grafana metrics for Sam. This sets us up to add other metrics going forward. In this initial pass, we are saving the results of the status endpoint.

Changes include some scalafmt changes.

This will be tested on dev. It is difficult to test locally due to the fact that we're using docker. Discussed with Qi about this and the conclusion was to move forward with merging in the changes since they are low risk. 

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -97,7 +97,7 @@ object Boot extends IOApp with LazyLogging {
       // This is for sending custom metrics to stackdriver. all custom metrics starts with `OpenCensus/sam/`.
       // Typing in `sam` in metrics explorer will show all sam custom metrics.
       // As best practice, we should have all related metrics under same prefix separated by `/`
-      implicit0(openTelemetry: OpenTelemetryMetrics[IO]) <- OpenTelemetryMetrics.resource[IO] ("sam", appConfig.prometheusConfig.endpointPort)
+      implicit0(openTelemetry: OpenTelemetryMetrics[IO]) <- OpenTelemetryMetrics.resource[IO]("sam", appConfig.prometheusConfig.endpointPort)
 
       cloudExtensionsInitializer <- cloudExtensionsInitializerResource(
         appConfig,
@@ -118,7 +118,10 @@ object Boot extends IOApp with LazyLogging {
           extraAuthParams = Some("prompt=login")
         )
       )
-    } yield createAppDependenciesWithSamRoutes(appConfig, cloudExtensionsInitializer, foregroundAccessPolicyDAO, foregroundDirectoryDAO, oauth2Config)(actorSystem, openTelemetry)
+    } yield createAppDependenciesWithSamRoutes(appConfig, cloudExtensionsInitializer, foregroundAccessPolicyDAO, foregroundDirectoryDAO, oauth2Config)(
+      actorSystem,
+      openTelemetry
+    )
 
   private def cloudExtensionsInitializerResource(
       appConfig: AppConfig,
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -3,8 +3,7 @@ package org.broadinstitute.dsde.workbench.sam
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import cats.data.NonEmptyList
-import cats.effect
-import cats.effect.{ExitCode, IO, IOApp}
+import cats.effect._
 import cats.implicits._
 import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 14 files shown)


andy7i:
jenkins retest bee failure

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -94,6 +94,11 @@ object Boot extends IOApp with LazyLogging {
 
       googleSynchronizerExecutionContext <- ExecutionContexts.fixedThreadPool[IO](24)
 
+      // This is for sending custom metrics to stackdriver. all custom metrics starts with `OpenCensus/sam/`.
+      // Typing in `sam` in metrics explorer will show all sam custom metrics.
+      // As best practice, we should have all related metrics under same prefix separated by `/`
+      implicit0(openTelemetry: OpenTelemetryMetrics[IO]) <- OpenTelemetryMetrics.resource[IO]("sam", appConfig.prometheusConfig.endpointPort)
+
```
andy7i, line 101: this bit was copied over from Leonardo: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/Boot.scala#L306

I'm not entirely sure if this comment is correct but unfortunately I can't test locally so I won't know until it's on dev.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StatusRoutes.scala
```diff
@@ -17,15 +20,23 @@ object BuildTimeVersion {
...
           complete {
             statusService.getStatus().map { statusResponse =>
-              val httpStatus = if (statusResponse.ok) StatusCodes.OK else StatusCodes.InternalServerError
+              val httpStatus = if (statusResponse.ok) {
+                openTelemetry.incrementCounter("checkStatus-success", tags = openTelemetryTags).unsafeToFuture()
+                StatusCodes.OK
+              } else {
+                openTelemetry.incrementCounter("checkStatus-failure", tags = openTelemetryTags).unsafeToFuture()
+                StatusCodes.InternalServerError
+              }
```
andy7i, line 39: this is where the magic happens

This is using the global IORuntime, but I'm not entirely sure if it's ok to use that here.

andy7i:
jenkins retest postgres failure
```
12:20:18 [ERROR] [10/14/2022 16:20:18.861] [org-broadinstitute-dsde-workbench-sam-api-StatusRouteSpec-akka.actor.default-dispatcher-6] [akka.actor.ActorSystemImpl(org-broadinstitute-dsde-workbench-sam-api-StatusRouteSpec)] HttpMethod(GET) http://example.com/status: 500 Internal Server Error entity: {"ok":false,"systems":{"Database":{"messages":["Postgres database connection invalid or timed out checking"],"ok":false}}}
```

gpolumbo-broad (approved):
I don't love the idea of adding more implicits, but I also don't know how hard or gross this would be without it.

Ghost-in-a-Jar (approved):

andy7i:
team discussion re: implicits - tl;dr we want to use them here. 

Reasons include:
* the code is already using implicits and we'll follow the code standard (for now) rather than rock the boat
* if we were **starting** to use implicits we would consider that a faux pax. But we're merely adding to an existing list of implicits
* using implicits follows the pattern for how to use opentelemetry

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -747,6 +751,7 @@ object ManagedGroupRoutesSpec {
   ): TestSamRoutes = {
     val directoryDAO = new MockDirectoryDAO()
     val policyDao = new MockAccessPolicyDAO(resourceTypeMap, directoryDAO)
+    implicit val openTelemetry: OpenTelemetryMetricsInterpreter[IO] = mock[OpenTelemetryMetricsInterpreter[IO]](RETURNS_SMART_NULLS)
```
Qi77Qi, line 754: could've used https://github.com/broadinstitute/workbench-libs/blob/develop/openTelemetry/src/test/scala/org/broadinstitute/dsde/workbench/openTelemetry/FakeOpenTelemetryMetricsInterpreter.scala#L8 so that u don't have to mock everywhere

andy7i: I did see that but when I tried to drop that in, the types were mismatched. So then I went with mockito because it was simple to mock out and reduces the reliance on workbench-libs. I could see both approaches working though! Maybe we'll switch to using `FakeOpenTelemetryMetricsInterpreter` as we add more metrics!

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -283,7 +291,7 @@ object Boot extends IOApp with LazyLogging {
       accessPolicyDAO: AccessPolicyDAO,
       directoryDAO: PostgresDirectoryDAO,
       oauth2Config: OpenIDConnectConfiguration
-  )(implicit actorSystem: ActorSystem): AppDependencies = {
+  )(implicit actorSystem: ActorSystem, openTelemetry: OpenTelemetryMetricsInterpreter[IO]): AppDependencies = {
```
Qi77Qi, line 294: this should be expecting `OpenTelemetryMetrics`

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
Qi77Qi: so at call site, you should always use `OpenTelemetryMetrics` instead of the interpreter..interpreter is just concrete implementation of the `trait`, `trait` is the interface

andy7i: aha! that makes sense. thanks for the info!

PR #863, opened 2022-10-20
ID-249 test with no dependencies

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-249

As a part of our IndependiSam efforts, we want to make all of our unit tests not rely on an external database. Relying on an external database slows things down and introduces more overhead to test running. The future of Sam's unit tests are to be able to be run without a database at all.

In the meantime, we need a way to run tests without a database to be able to verify that they work or not. The JVM option `-Dpostgres.enabled=false` can be supplied to test runs, which ignores all tests that reach out to the database. About 580 tests will pass without a database at this time. 

The approach I took was to shim in a `TestDbReference` which throws exceptions when any query is attempted when `-Dpostgres.enabled=false` is supplied. `TestSupport` has also been updated to only do things to the database if the database is enabled. The rest of the changes are just going through each spec and marking the tests that talk to the db as cancelled if `-Dpostgres.enabled=false` is supplied.

I've left this as an option, and not the default, because our tests still cover a lot of use-cases and business logic. I'm unsure about outright disabling them outright at this time, but the option is there and easy to do if we want to. 

---

**PR checklist**


src/test/resources/logback-test.xml
```diff
@@ -15,13 +15,9 @@
...
-    <appender name="Console-Stackdriver" class="ch.qos.logback.core.ConsoleAppender">
-        <encoder class="ch.qos.logback.core.encoder.LayoutWrappingEncoder">
-            <layout class="org.springframework.cloud.gcp.logging.StackdriverJsonLayout">
-                <includeTraceId>true</includeTraceId>
-                <includeSpanId>true</includeSpanId>
-            </layout>
-        </encoder>
+    <appender name="Console-Stackdriver" class="com.google.cloud.logging.logback.LoggingAppender">
+        <enhancer>com.google.cloud.logging.logback.TraceLoggingEventEnhancer</enhancer>
+        <redirectToStdout>true</redirectToStdout>
```
tlangs, line 20: We don't use spring anymore!

INDEPENDENT_TESTS.md
```diff
@@ -0,0 +1,39 @@
...
+AdminResourcesRoutesSpec
+TermsOfServiceRouteSpec
+UserRoutesSpec
+AdminResourceTypesRoutesSpec
+AdminUserRoutesSpec
+GoogleExtensionRoutesV1Spec
+ResourceRoutesV1Spec
+```
+
+As we convert tests to not need a running database, we should update this list.
```
Ghost-in-a-Jar, line 39: lets link this in the readme.md

Ghost-in-a-Jar: and/or more into test folders

src/test/scala/org/broadinstitute/dsde/workbench/sam/db/TestDbReference.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import java.security.cert.CertPathBuilderException
+import java.sql.SQLTimeoutException
+import scala.concurrent.ExecutionContext
+
+object TestDbReference extends LazyLogging {
+  private val config = ConfigFactory.load()
+  private val databaseEnabled = config.getBoolean("db.enabled")
+
+  // This code is copied over from DbReference. We didnt want to have to refactor DbReference to be able to use it in tests.
```
andy7i, line 27: We added this line in the mob. Does this sound right to you?

Ghost-in-a-Jar (approved):
@tlangs do you think it would be low effort to refactor DbReference so that we dont have to duplicate the code in the companion object? If it would take more than 30 min i think we should just clean it up later. I pushed up an inline comment about it as well.

andy7i (approved):
1 question but not blocking. 👍 otherwise

Shakespeared (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/db/TestDbReference.scala
tlangs: Yes!

tlangs:
Jenkins retest

PR #874, opened 2022-10-26
ID-321 Add ability to retry flaky tests

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-321

I've added the ability to make a unit test suite retry failing tests up to 3 times. This is done by making the test suite extend the new `RetryableAnyFlatSpec` instead of `AnyFlatSpec`.

---

**PR checklist**

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/RetryableAnyFlatSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/RetryableAnyFlatSpec.scala
@@ -1,6 +1,6 @@
 package org.broadinstitute.dsde.workbench.sam
 
-import org.scalatest.{Canceled, Failed, Outcome, Retries}
+import org.scalatest.{Exceptional, Outcome, Retries}
 import org.scalatest.flatspec.AnyFlatSpec
 
 trait RetryableAnyFlatSpec extends AnyFlatSpec with Retries {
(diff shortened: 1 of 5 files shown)


tlangs:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/sam/RetryableAnyFlatSpec.scala
```diff
@@ -0,0 +1,22 @@
...
+  val retries = 3
+
+  override def withFixture(test: NoArgTest): Outcome =
+    withFixture(test, 1)
+
+  def withFixture(test: NoArgTest, count: Int): Outcome = {
+    val outcome = super.withFixture(test)
+    outcome match {
+      case Failed(_) | Canceled(_) =>
+        println(s"Test '${test.name}' failed on try $count.")
```
gpolumbo-broad:
```suggestion
        println(s"Test '${test.name}' ${outcome} on try $count.")
```

src/test/scala/org/broadinstitute/dsde/workbench/sam/RetryableAnyFlatSpec.scala
```diff
@@ -0,0 +1,22 @@
...
+        if (count >= retries) super.withFixture(test) else withFixture(test, count + 1)
```
gpolumbo-broad, line 19: I am reading this that after we hit max retries, this will run the test 1 more time?  Is that right?  

Would the alternative be:
```suggestion
        if (count >= retries) outcome else withFixture(test, count + 1)
```
I don't care either way, other than I think this suggestion is ever so slightly more intuitive.  "If we've retried max times, return the last outcome and exit" vs. "If we've retried max times, give it ONE more try just in case."

gpolumbo-broad (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/RetryableAnyFlatSpec.scala
tlangs: This will still only run the test 3 times. If it fails on `1`st try, and fails on `2`nd try, then `super.withFixture` will be the `3`rd and final try, which is returned.

src/test/scala/org/broadinstitute/dsde/workbench/sam/RetryableAnyFlatSpec.scala
tlangs: I've updated the code to give more information about the failed test, including the type of failure. I hope that satisfies your change request. Lemme know if you had something else in mind. `${outcome}` stuck the whole object in there instead of just the name.

Shakespeared (approved):

gpolumbo-broad (approved):

PR #886, opened 2022-11-01
new resource for wds-instance [AJ-598]

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/AJ-598

Sets up a new resource type `wds-instance` to capture permissions for WDS instances, which run as apps behind Leo in Azure. `wds-instance` will inherit reader/writer/owner from a parent workspace.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -153,6 +153,7 @@ resourceTypes = {
           controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
           controlled-user-private-workspace-resource = ["deleter"]
           controlled-application-shared-workspace-resource = ["writer", "reader"]
+          wds-instance = ["owner"]
         }
       }
       application = {
@@ -167,6 +168,7 @@ resourceTypes = {
           controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
           controlled-application-shared-workspace-resource = ["writer", "reader"]
           google-project = ["pet-creator"]
+          wds-instance = ["writer"]
         }
       }
       reader = {
@@ -175,6 +177,7 @@ resourceTypes = {
           controlled-user-shared-workspace-resource = ["reader"]
           controlled-application-shared-workspace-resource = ["reader"]
           google-project = ["pet-creator"]
+          wds-instance = ["reader"]
         }
       }
       discoverer = {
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -167,6 +168,7 @@ resourceTypes = {
           controlled-user-shared-workspace-resource = ["editor", "writer", "reader"]
           controlled-application-shared-workspace-resource = ["writer", "reader"]
           google-project = ["pet-creator"]
+          wds-instance = ["writer"]
```
jordansafer, line 171: This made me think about naming. I'm trying to decide if it makes more sense to have something specific & traceable like "wds-instance", or something generic with an angle towards supporting multiple dataservices (for which we have no plans). I lean towards wds-instance, since we avoid creating new names & more confusion.

I'm also thinking about control plane vs dataplane permissions. Seeing permissions for a wds-instance makes me think: "Does this give permission to create and delete WDS instances?" Right now we definitely don't have permissions at that granularity, so this is a bit of a future question.

jordansafer (approved):

src/main/resources/reference.conf
davidangb: Agreed on future permissions. I could imagine a lot of granularity here, such as permissions to change schema vs. write to the existing schema, permission to delete data vs. insert/update, permission to clone, etc. etc.

For now I'm trying to keep it as simple as possible, just the permissions I can see us needing for MVP. The "delete" permission may be necessary if/when the parent workspace is deleted and we need to delete (from Sam) the child `wds-instance` resources.

davidangb:
automation tests failed with what looks like a transient error connecting to Thurloe. Jenkins retest.

tlangs (approved):

rtitle (approved):
LGTM. For context, the relay listener will check the `kubernetes-app` resource type to check whether (or not) the user can "connect" to the app. Makes sense to then check a `wds-instance` resource for specific WDS actions.

davidangb:
@rtitle thanks for the context. We'll definitely need to check the `write` permission for certain actions in WDS - I'm waiting until the implementation phase to see if there's any reason we need to check for `read`; we can probably just rely on the listener for that. But we'll see!

davidangb:
*sigh* automation had one unrelated error in Orch:
`org.broadinstitute.dsde.test.api.orch.AsyncImportSpec.Orchestration should import a PFB file via import service for the owner of a workspace`

Jenkins retest one more time

davidangb:
Jenkins retest.

davidangb:
Jenkins retest. again.

PR #888, opened 2022-11-03
ID-334 Updating regex used to check if an email address belongs to a Pet SA

gpolumbo-broad:
https://broadworkbench.atlassian.net/browse/ID-334

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
@@ -66,7 +66,7 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with 
 }
 
 object StandardSamUserDirectives {
-  val SAdomain: Regex = "(\\S+@\\S+\\.iam\\.gserviceaccount\\.com$)".r
+  val SAdomain: Regex = "\\S+@((\\S+\\.)|)gserviceaccount.com".r
   // UAMI == "User Assigned Managed Identity" in Azure
   val UamiPattern: Regex = "(^/subscriptions/\\S+/resourcegroups/\\S+/providers/Microsoft\\.ManagedIdentity/userAssignedIdentities/\\S+$)".r
   val accessTokenHeader = "OIDC_access_token"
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -66,7 +66,7 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with
 }
 
 object StandardSamUserDirectives {
-  val SAdomain: Regex = "(\\S+@\\S+\\.iam\\.gserviceaccount\\.com$)".r
+  val SAdomain: Regex = "\\S+@((\\S+\\.)|)gserviceaccount.com".r
```
tlangs: Why regex? Why not just `String.endsWith`?

gpolumbo-broad: I guess we can make a call.  I know the tests I wrote really just check `endswith`.  However, this thing was checking, very roughly, that this string also sorta looks like an email with a name part, an `@`, and then some string afterward.  

So we could just say, meh, and check the `endswith`.  And/or we could say, check the `endswith` AND make sure it's got an `@` in there someplace.  Or we could try to do all of that in one check using a regex.  🤷 at the moment I don't feel strongly one way or the other.

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
tlangs: If @dvoet thinks the regex is fine, I won't argue and delay the fix.

dvoet: because the scala code uses pattern matching with regex, but you could rewrite that too

dvoet: and it can't just be `endswith` because I can register `foobargserviceaccount.com`

gpolumbo-broad:
jenkins retest

tlangs:
jenkins retest

tlangs:
jenkins retest

yonghaoy (approved):

tlangs:
Even though Sam swatomation tests are failing, I'm going to merge this. The Sam swat tests are failing for a different reason, addressed in [this PR](https://github.com/broadinstitute/terra-helmfile/pull/3536). In an effort to unblock our dev users, I'm making the call to merge this rather than waiting for another round of swat tests to tell us info we already know.

PR #889, opened 2022-11-04
TOAZ-258: Add admin getArbitraryPet endpoint

rtitle:
Ticket: https://broadworkbench.atlassian.net/browse/TOAZ-258

Leo (and potentially other services) have a need for the `arbitrary pet` feature of Sam. However, the [existing API](https://sam.dsde-dev.broadinstitute.org/#/Google/getArbitraryPetServiceAccountKey) only operates on the calling user. We want the ability to get an arbitrary pet for a _provided_ user email, similar to [this API](https://sam.dsde-dev.broadinstitute.org/#/Google/getUserPetServiceAccountKey).

This PR adds that API. I tested it with Leonardo on a BEE. Any feedback appreciated.

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -875,6 +875,46 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/google/v1/petServiceAccount/{userEmail}/key:
+    get:
+      tags:
+        - Google
+      summary: gets a key for the user's arbitrary pet service account, get_pet_private_key action on cloud-extension/google required
+      operationId: getUserArbitraryPetServiceAccountKey
+      parameters:
+        - name: userEmail
+          in: path
+          description: User's email address
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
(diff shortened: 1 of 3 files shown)


rtitle:
Tested on a BEE, I think this is working :)

tlangs (approved):

gpolumbo-broad (approved):

dvoet:
It is a little more work, but I don't think Sam should return keys. Sam should return access tokens from [this api](https://cloud.google.com/iam/docs/create-short-lived-credentials-direct#sa-credentials-oauth).

At some point we should discontinue the use of keys. Extending the use of keys here will increase later pain. But I leave that tradeoff up to you.

rtitle:
> Sam should return access tokens from [this api](https://cloud.google.com/iam/docs/create-short-lived-credentials-direct#sa-credentials-oauth).

That's a good point. I'll look to see what that involves in Sam/Leo..

rtitle:
I made [this ticket](https://broadworkbench.atlassian.net/browse/ID-337). I think this probably goes outside the scope of this TOAZ ticket so I'd like to do it separately. Leo already uses the Sam `/key` endpoints, so there will be work regardless to phase it out.

PR #890, opened 2022-11-04
ID-336 Service Accounts do not need to accept ToS

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-336
Enabling ToS all-of-a-sudden meant that Service Accounts needed to accept the ToS! This causes tests to fail, and causes all sorts of issues. 

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
@@ -41,7 +41,9 @@ class TosService(val directoryDao: DirectoryDAO, val appsDomain: String, val tos
   /** If grace period enabled, don't check ToS, return true If ToS disabled, return true Otherwise return true if user has accepted ToS, or is a service account
     */
   def isTermsOfServiceStatusAcceptable(user: SamUser): Boolean =
-    tosConfig.isGracePeriodEnabled || !tosConfig.enabled || user.acceptedTosVersion.contains(tosConfig.version) || StandardSamUserDirectives.SAdomain.matches(user.email.value)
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -37,10 +38,10 @@ class TosService(val directoryDao: DirectoryDAO, val appsDomain: String, val tos
       directoryDao.loadUser(userId, samRequestContext).map(_.map(_.acceptedTosVersion.contains(tosConfig.version)))
     } else IO.none
 
-  /** If grace period enabled, don't check ToS, return true If ToS disabled, return true Otherwise return true if user has accepted ToS
+  /** If grace period enabled, don't check ToS, return true If ToS disabled, return true Otherwise return true if user has accepted ToS, or is a service account
```
Shakespeared, line 41: Do you think you could just reword this comment while you're in there? I think I know what it's getting at but I'm having trouble parsing it

Shakespeared (approved):

dvoet (approved):

tlangs:
jenkins retest

andy7i (commented):
This PR is missing some context around why SAs don't need to accept the ToS (and I checked in the ticket as well). You already have 2 thumbs so I won't continue reviewing.

tlangs:
@andy7i  Good point. Service Accounts don't need to accept the ToS because they're not users. We don't expect service accounts to be acting on behalf of users, and probably don't even have a way to accept ToS. I'm not sure about how we ensure a user doesn't register with a Service Account email and use that to bypass the ToS, but I'm assuming other people thought about that in the implementation of ToS.

tlangs:
jenkins retest

PR #892, opened 2022-11-07
ID-338 Don't fail on missing Azure config

tlangs:
Ticket: [<Link to Jira ticket>](https://broadworkbench.atlassian.net/browse/ID-338)

Looks like we didn't test if Sam would boot if the Azure configs are missing. This happens because Prod doesn't have Azure enabled. I've fixed the issue, and also removed the `azureEnabled` variable from the code, since it's not used for anything.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
@@ -153,14 +153,23 @@ object AppConfig {
     )
   }
 
-  implicit val azureServicesConfigReader: ValueReader[AzureServicesConfig] = ValueReader.relative { config =>
-    AzureServicesConfig(
-      config.as[Option[Boolean]]("azureEnabled"),
-      config.getString("managedAppClientId"),
-      config.getString("managedAppClientSecret"),
-      config.getString("managedAppTenantId"),
-      config.as[Seq[String]]("managedAppPlanIds")
-    )
+  implicit val azureServicesConfigReader: ValueReader[Option[AzureServicesConfig]] = ValueReader.relative { config =>
+    config
+      .getAs[Boolean]("azureEnabled")
+      .flatMap(azureEnabled =>
+        if (azureEnabled) {
+          Option(
+            AzureServicesConfig(
+              config.getString("managedAppClientId"),
+              config.getString("managedAppClientSecret"),
+              config.getString("managedAppTenantId"),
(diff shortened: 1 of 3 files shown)


jack-r-warren (approved):
Not sure my vote counts for much but LGTM!

Shakespeared (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
```diff
@@ -198,7 +207,7 @@ object AppConfig {
       termsOfServiceConfig = config.as[TermsOfServiceConfig]("termsOfService"),
       oidcConfig = config.as[OidcConfig]("oidc"),
       adminConfig = config.as[AdminConfig]("admin"),
-      azureServicesConfig = config.getAs[AzureServicesConfig]("azureServices"),
+      azureServicesConfig = config.as[Option[AzureServicesConfig]]("azureServices"),
```
rtitle: Just trying to understand -- isn't `getAs[A]` and `as[Option[A]]` the same thing? Both return an `Option[A]`. Not sure I see why this has any effect (e.g. either should handle the config not being there)..

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
```diff
@@ -153,14 +153,23 @@ object AppConfig {
...
-      config.getString("managedAppClientId"),
-      config.getString("managedAppClientSecret"),
-      config.getString("managedAppTenantId"),
-      config.as[Seq[String]]("managedAppPlanIds")
-    )
+  implicit val azureServicesConfigReader: ValueReader[Option[AzureServicesConfig]] = ValueReader.relative { config =>
+    config
+      .getAs[Boolean]("azureEnabled")
+      .flatMap(azureEnabled =>
+        if (azureEnabled) {
```
rtitle, line 160: Since it wasn't being read before, is `azureEnabled` set to true in dev? I don't see it here: https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/sam/sam.conf.ctmpl#L284

(Want to make sure we don't turn off azure in dev)

tlangs:
jenkins retest

rtitle (approved):
👍 contingent on https://github.com/broadinstitute/terra-helmfile/pull/3546 getting in

PR #900, opened 2022-11-17
ID-242 Updating comments that describe how configs are loaded

gpolumbo-broad:
Ticket: https://broadworkbench.atlassian.net/browse/ID-242

What:
The comments on this method before were outdated and not really all that helpful.  

Why:
Other projects will be referencing this code when they break away from firecloud-develop and we want them to understand the how and why of why we load configs the way we do now.

How:
I rewrote the comments and moved some comments around to hopefully make it easier to understand what the code is doing and how to use it.  

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
@@ -176,15 +176,12 @@ object AppConfig {
     PrometheusConfig(config.getInt("endpointPort"))
   }
 
-  /**
-    * Loads all the configs for the Sam App.  All values defined in `src/main/resources/sam.conf` will take precedence
-    * over any other configs.  In this way, we can still use configs rendered by `firecloud-develop` that render to
-    * `config/sam.conf` if we want.  To do so, you must render `config/sam.conf` and then do not populate ENV variables
-    * for `src/main/resources/sam.conf`.
+  /** Loads all the configs for the Sam App. All values defined in `src/main/resources/sam.conf` will take precedence over any other configs. In this way, we
+    * can still use configs rendered by `firecloud-develop` that render to `config/sam.conf` if we want. To do so, you must render `config/sam.conf` and then do
+    * not populate ENV variables for `src/main/resources/sam.conf`.
     *
-    * The ENV variables that you need to populate can be found in `src/main/resources/sam.conf` variable substitutions.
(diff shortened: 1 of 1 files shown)


Ghost-in-a-Jar (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
```diff
@@ -175,13 +175,25 @@ object AppConfig {
   implicit val prometheusConfig: ValueReader[PrometheusConfig] = ValueReader.relative { config =>
     PrometheusConfig(config.getInt("endpointPort"))
   }
+
+  /**
+    * Loads all the configs for the Sam App.  All values defined in `src/main/resources/sam.conf` will take precedence
+    * over any other configs.  In this way, we can still use configs rendered by `firecloud-develop` that render to
+    * `config/sam.conf` if we want.  To do so, you must render `config/sam.conf` and then do not populate ENV variables
```
Shakespeared: is it obvious from other documentation how to render sam.conf? I might also make this extra explicit and say "to use configs from firecloud-develop…"

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
```diff
@@ -175,13 +175,25 @@ object AppConfig {
...
-    // ConfigFactory.load automatically pulls in the default reference.conf,
-    // which then ends up overriding any conf files provided as java options.
-    // We need to get _just_ the contents of env.conf so that normal overriding can occur.
+    // Start by parsing sam config files (like 'sam.conf') from wherever they live on the classpath
     val samConfig = ConfigFactory.parseResourcesAnySyntax("sam").resolve()
+    // Load any other configs on the classpath following: https://github.com/lightbend/config#standard-behavior
+    // This is where things like `src/main/resources/reference.conf` will get loaded
     val config = ConfigFactory.load()
+    // Combine the two sets of configs.  If a value is defined in both sets of configs, the ones in `samConfig` will
+    // take precedence over those in `config`
```
Shakespeared, line 193: I still find the naming confusing… am I understanding correctly that basically what's defined here as `config` is the same as `config/sam.conf` which is what's being rendered from firecloud-develop, and `samconfig` is the same as `src/main/resources/sam.conf` rendered from env vars? If that's the case, could you add a comment or reference that somewhere? For me that's still the main point of confusion here.

Shakespeared (approved):
thumbed with comments

PR #908, opened 2022-11-21
ID-317 Added new ToS and made ToS version configurable

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-317

What:
The Terms of Service is now versioned from within Sam's codebase. Files in `resource/tos/` shall be named `termsOfService-VERSION.md`, where `VERSION` is an integer corresponding to the ToS config in sam.conf. The new Terms of Service for Azure Public Preview has been added, and the original ToS is now in this directory format as well.

This PR does not actually increment the ToS version being seen by users.

Why:

We need to be able to have Sam serve the correct terms of service for a given environment. For example, the pre-prod Sam instances need to serve the Azure Public Preview ToS, but Prod can't be updated until later. Instead of having all environments have their own deployed git hash, we can instead just make the ToS Version control which file is served. This makes it super easy to control which ToS is served in each environment.

How:

The ToSService now looks at the config and uses the ToS Version to resolve the file contain ToS text.

---

**PR checklist**


tlangs:
jenkins retest

tlangs:
jenkins retest

Ghost-in-a-Jar (approved):

Shakespeared (approved):

PR #911, opened 2022-11-22
TOAZ-224 store MRG coordinates in sam

dvoet:
Ticket: [TOAZ-224](https://broadworkbench.atlassian.net/browse/TOAZ-224)

See the ticket for details pls. Note that the old way of using tags on the MRG is remaining intact for backwards compatibility. Look to a future PR to remove that and migrate pre-existing pet identities. 

Also note that I did not implement a new API to get a pet in the interest of not having to change external dependencies. I think the new api in the ticket looks nicer but the billing profile id might not be readily available and we know that the mrg coordinates are.

---

**PR checklist**

### DIFF
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20221115_azure_managed_resource_group.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20221115_azure_managed_resource_group.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="azure_managed_resource_group">
+        <createTable tableName="SAM_AZURE_MANAGED_RESOURCE_GROUP">
+            <column autoIncrement="true" name="id" type="BIGINT">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="tenant_id" type="VARCHAR">
+                <constraints nullable="false"/>
+            </column>
+            <column name="subscription_id" type="VARCHAR">
+                <constraints nullable="false"/>
+            </column>
+            <column name="managed_resource_group_name" type="VARCHAR">
+                <constraints nullable="false"/>
+            </column>
+            <column name="billing_profile_id" type="VARCHAR">
+                <constraints nullable="false" unique="true"/>
+            </column>
+        </createTable>
+        <addUniqueConstraint tableName="SAM_AZURE_MANAGED_RESOURCE_GROUP"
+                             columnNames="tenant_id,subscription_id,managed_resource_group_name"/>
+    </changeSet>
+</databaseChangeLog>
--- a/src/main/resources/sam.conf
+++ b/src/main/resources/sam.conf
@@ -177,6 +177,17 @@ azureServices {
   managedAppClientId = ${?AZURE_MANAGED_APP_CLIENT_ID}
   managedAppClientSecret = ${?AZURE_MANAGED_APP_CLIENT_SECRET}
   managedAppTenantId = ${?AZURE_MANAGED_APP_TENANT_ID}
-  managedAppPlanIds = ["terra-dev", "terra-workspace-dev-plan"]
+  managedAppPlans = [
+    {
+        name = "terra-dev"
+        publisher = "thebroadinstituteinc1615909626976"
+        authorizedUserKey = authorizedTerraUser
+    }
+    {
+        name = "terra-workspace-dev-plan"
+        publisher = "thebroadinstituteinc1615909626976"
+        authorizedUserKey = authorizedTerraUser
+    }
+  ]
 }
 
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1207,8 +1207,168 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/azure/v1/billingProfile/{billingProfileId}/managedResourceGroup:
+    post:
+      tags:
+        - Azure
+      summary: associates billing profile id to managed resource group coordinates
+      operationId: createManagedResourceGroup
+      parameters:
+        - name: billingProfileId
+          in: path
+          description: billing profile id
+          required: true
+          schema:
+            type: string
+      requestBody:
+        description: The details of the managed resource group in which to create the pet
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/ManagedResourceGroupCoordinates'
+        required: true
+      responses:
+        201:
+          description: Successfully created managed resource group
+          content: { }
+        400:
+          description: Invalid or incomplete request body
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        403:
+          description: Caller does not have permission to the managed resource group or it cannot be found or is otherwise invalid
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        404:
+          description: Caller does not have permission to the billing profile or it does not exist
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        409:
+          description: Managed resource group and/or billing profile already has an association
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+  /api/azure/v1/user/petManagedIdentity/{billingProfileId}:
+    post:
+      tags:
+        - Azure
+      summary: gets or creates a pet managed identity for the calling user
+      operationId: getPetManagedIdentityInBillingProfile
+      parameters:
+        - name: billingProfileId
+          in: path
+          description: billing profile id
+          required: true
+          schema:
+            type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: object
+      responses:
+        200:
+          description: Successfully retrieved resource without creating it
+          content:
+            application/json:
+              schema:
+                type: string
+        201:
+          description: Successfully created resource
+          content:
+            application/json:
+              schema:
+                type: string
+        400:
+          description: billing profile is not associated to managed resource group, call createManagedResourceGroup
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        404:
+          description: Caller does not have access to billing profile or it does not exist
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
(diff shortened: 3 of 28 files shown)


src/main/resources/sam.conf
```diff
@@ -177,6 +177,17 @@ azureServices {
   managedAppClientId = ${?AZURE_MANAGED_APP_CLIENT_ID}
   managedAppClientSecret = ${?AZURE_MANAGED_APP_CLIENT_SECRET}
   managedAppTenantId = ${?AZURE_MANAGED_APP_TENANT_ID}
-  managedAppPlanIds = ["terra-dev", "terra-workspace-dev-plan"]
+  managedAppPlans = [
+    {
+        name = "terra-dev"
+        publisher = "thebroadinstituteinc1615909626976"
```
dvoet, line 183: `publisher` is added because checking the name alone is too broadly scoped (anyone can make a plan with any name). The value came from the azure portal looking at the json for a deployed app.

src/main/resources/sam.conf
```diff
@@ -177,6 +177,17 @@ azureServices {
...
+        authorizedUserKey = authorizedTerraUser
```
dvoet, line 184: `authorizedUserKey` was added mimicking how BPM checks if users are authorized to use a MRG. Sam needs to do this check as well to keep users from doing mean things by calling Sam directly.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1238,10 +1292,10 @@ paths:
           content: { }
         403:
           description: Caller does not have permission to create a pet managed identity in the given managed resource group
-          content: { }
-        404:
```
dvoet, line 1242: removed 404 as a response because it does not really make sense. 404 should only be used when something in the resource path cannot be found. Plus it leaks information about what exists and what doesn't

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutes.scala
```diff
@@ -53,25 +54,53 @@ trait AzureRoutes extends SecurityDirectives {
...
+
+    onSuccess(azureService.flatTraverse(_.getBillingProfileId(request, samRequestContext))).flatMap {
       case Some(resourceId) =>
-        requireAction(FullyQualifiedResourceId(SamResourceTypes.spendProfile, resourceId), SamResourceActions.link, samUser.id, samRequestContext)
-      case None =>
-        failWith(
-          new WorkbenchExceptionWithErrorReport(
-            ErrorReport(StatusCodes.NotFound, s"Managed resource group ${request.managedResourceGroupName.value} not found")
-          )
+        maskNotFound & requireAction(
```
dvoet, line 94: `requireAction` throws Not Found, need to convert that to 403

src/test/resources/reference.conf
```diff
@@ -101,7 +101,7 @@ testStuff = {
   azure {
     tenantId = "0cb7a640-45a2-4ed6-be9f-63519f86e04b"
     subscriptionId = "3efc5bdf-be0e-44e7-b1d7-c08931e3c16c"
-    managedResourceGroupName = "mrg-terra-integration-test-20211118"
+    managedResourceGroupName = "mrg-terra-dev-previ-20220930130036"
```
dvoet, line 104: this one has an authorized user

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -50,23 +93,25 @@ class AzureService(crlService: CrlService, directoryDAO: DirectoryDAO) {
   private def createUserPetManagedIdentity(
       id: PetManagedIdentityId,
       user: SamUser,
-      request: GetOrCreatePetManagedIdentityRequest,
+      mrgCoords: ManagedResourceGroupCoordinates,
       samRequestContext: SamRequestContext
   ): IO[(PetManagedIdentity, Boolean)] =
     for {
-      _ <- validateManagedResourceGroup(request)
```
dvoet, line 57: this validate call was moved to `getBillingProfileIdFromAzureTag` for 2 reasons, 
1. it only needs to be called when getting the billing profile id from a tag because we need to verify we trust the MRG, this check is not required when getting the billing profile id from the db because it is done before saving it
2. checking here is a little late in the game since we have already read and used the tag before verifying trust.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamRoutes.scala
```diff
@@ -72,7 +72,7 @@ abstract class SamRoutes(
                   extensionRoutes(samUser, samRequestContextWithUser) ~
                   groupRoutes(samUser, samRequestContextWithUser) ~
                   apiUserRoutes(samUser, samRequestContextWithUser) ~
-                  azureRoutes(samUser, samRequestContext)
+                  azureRoutes(samUser, samRequestContextWithUser)
```
dvoet, line 75: finding this took an hour of my life

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3233,6 +3287,25 @@ components:
...
+      required:
+        - tenantId
+        - subscriptionId
+        - managedResourceGroupName
+      type: object
+      properties:
+        tenantId:
+          type: string
+          description: the Azure tenant id (UUID)
+          example: 0cb7a640-45a2-4ed6-be9f-63519f86e04b
```
aherbst-broad, line 3300: Nitpick: it might be good to use truly fake UUIDs in these examples, I believe these are the UUIDs for the analysis playground sub.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20221115_azure_managed_resource_group.xml
```diff
@@ -0,0 +1,29 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="billing_profile_id" type="VARCHAR">
+                <constraints nullable="false" unique="true"/>
+            </column>
+        </createTable>
+        <addUniqueConstraint tableName="SAM_AZURE_MANAGED_RESOURCE_GROUP"
+                             columnNames="tenant_id,subscription_id,managed_resource_group_name"/>
+    </changeSet>
+</databaseChangeLog>
```
Ghost-in-a-Jar, line 29: Not a fan of xml migrations is there any way we can make them yaml or something more readable just a thought

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAzureManagedResourceGroupDAO.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  override def getManagedResourceGroupByCoordinates(
+      managedResourceGroupCoordinates: ManagedResourceGroupCoordinates,
+      samRequestContext: SamRequestContext
+  ): IO[Option[ManagedResourceGroup]] =
+    IO(mrgs.find(_.managedResourceGroupCoordinates == managedResourceGroupCoordinates))
+
+  override def deleteManagedResourceGroup(billingProfileId: BillingProfileId, samRequestContext: SamRequestContext): IO[Int] =
+    IO(mrgs.filterInPlace(_.billingProfileId != billingProfileId)).map(_ => 1)
+}
```
Ghost-in-a-Jar, line 28: We have been trying to move away from storing state in tests and using mocks instead, that being said this is super simple so I don't feel strongly about it.

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceSpec.scala
```diff
@@ -106,10 +110,193 @@ class AzureServiceSpec extends AnyFlatSpec with Matchers with ScalaFutures {
...
+    val svc =
+      new AzureService(MockCrlService(user, managedResourceGroup.managedResourceGroupCoordinates.managedResourceGroupName), new MockDirectoryDAO(), mockMrgDAO)
+
+    val err = intercept[WorkbenchExceptionWithErrorReport] {
+      svc
+        .createManagedResourceGroup(managedResourceGroup, samRequestContext.copy(samUser = user.map(_.copy(email = WorkbenchEmail("other@email.com")))))
+        .unsafeRunSync()
+    }
+    err.errorReport.statusCode shouldBe Some(StatusCodes.Forbidden)
+    mockMrgDAO.mrgs should not contain managedResourceGroup
```
Ghost-in-a-Jar, line 327: Are there any other sanity checks we could do in this Spec for all the null safe parsing stuff you have in place?

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20221115_azure_managed_resource_group.xml
dvoet: I don't really care but I try to stay consistent unless you want to make a decree that from this point forth, yada yada

Ghost-in-a-Jar (approved):
Left minor comments regarding the tests

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockAzureManagedResourceGroupDAO.scala
dvoet: I think a balance is important. Mocking function calls can be cumbersome and brittle at times. I think it is worth it if the thing being mocked is complicated and only the return value matters for the test. But when the thing being mocked is simple as in this case, the mock behaves almost like the real thing and is probably easier to deal with down the road.

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceSpec.scala
dvoet: I added some

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAzureManagedResourceGroupDAO.scala
```diff
@@ -0,0 +1,93 @@
...
+
+class PostgresAzureManagedResourceGroupDAO(protected val writeDbRef: DbReference, protected val readDbRef: DbReference)
+    extends AzureManagedResourceGroupDAO
+    with DatabaseSupport {
+
+  override def insertManagedResourceGroup(managedResourceGroup: ManagedResourceGroup, samRequestContext: SamRequestContext): IO[Int] =
+    serializableWriteTransaction("insertManagedResourceGroup", samRequestContext) { implicit session =>
+      val mrgTableColumn = AzureManagedResourceGroupTable.column
+      val insertGroupQuery =
+        samsql"""insert into ${AzureManagedResourceGroupTable.table} (${mrgTableColumn.tenantId}, ${mrgTableColumn.subscriptionId}, ${mrgTableColumn.managedResourceGroupName}, ${mrgTableColumn.billingProfileId})
```
aherbst-broad, line 24: TIL about the samsql fn, very nice.

aherbst-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -84,64 +129,103 @@ class AzureService(crlService: CrlService, directoryDAO: DirectoryDAO) {
...
+  ): IO[ResourceGroup] =
+    traceIOWithContext("validateManagedResourceGroup", samRequestContext) { _ =>
+      for {
+        resourceManager <- crlService.buildResourceManager(mrgCoords.tenantId, mrgCoords.subscriptionId)
+        mrg <- lookupMrg(mrgCoords, resourceManager)
+        appManager <- crlService.buildApplicationManager(mrgCoords.tenantId, mrgCoords.subscriptionId)
+        appsInSubscription <- IO(appManager.applications().list().asScala)
+        managedApp <- IO.fromOption(appsInSubscription.find(_.managedResourceGroupId() == mrg.id()))(managedAppValidationFailure)
+        plan <- validatePlan(managedApp, crlService.getManagedAppPlans)
+        _ <- if (validateUser) validateAuthorizedAppUser(managedApp, plan, samRequestContext) else IO.unit
```
rtitle, line 170: It probably doesn't matter if BPM is the only thing calling this API at billing-profile creation time -- but do we want to check `authorizedTerraUser` in Sam, or just rely on the `set_managed_resource_group` action?

rtitle (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
dvoet: as it stands, anyone can make a spend-profile in sam then call this new api so I think this check is necessary

PR #919, opened 2022-12-02
[WOR-637] Add MRG DELETE endpoint and associated tests

aherbst-broad:
Ticket: [Integrate BPM with new Sam API to store MRG coordinates](https://broadworkbench.atlassian.net/browse/WOR-637)

What:

Adds the ability to delete a managed resource group from a billing profile. This is a follow-on to #911 .

Why:

When an Azure  billing profile is deleted, we should remove the associated managed resource group record so it can be reused later. If we don't do this, a subsequent billing profile that attempts to use the same MRG will get a 409 and fail out. 

 
How:

* Adds a DELETE endpoint at `/api/azure/v1/billingProfile/{billingProfileId}/managedResourceGroup`,
---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutes.scala
@@ -71,7 +71,19 @@ trait AzureRoutes extends SecurityDirectives with LazyLogging {
                     }
                   }
                 }
-              }
+              } ~
+                delete {
+                  requireAction(
+                    FullyQualifiedResourceId(SamResourceTypes.spendProfile, ResourceId(billingProfileId)),
+                    SamResourceActions.delete,
+                    samUser.id,
+                    samRequestContext
+                  ) {
+                    complete {
+                      service.deleteManagedResourceGroup(BillingProfileId(billingProfileId), samRequestContext).map(_ => StatusCodes.NoContent)
+                    }
+                  }
+                }
             }
         }
       }
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceSpec.scala
@@ -364,4 +364,29 @@ class AzureServiceSpec extends AnyFlatSpec with Matchers with ScalaFutures {
       mockMrgDAO.mrgs should not contain managedResourceGroup
     }
   }
+
+  "deleteManagedResourceGroup" should "delete a managed resource group" in {
+    val user = Generator.genWorkbenchUserAzure.sample
+    val managedResourceGroup = Generator.genManagedResourceGroup.sample.get
+    val mockMrgDAO = new MockAzureManagedResourceGroupDAO
+    val svc =
+      new AzureService(MockCrlService(user, managedResourceGroup.managedResourceGroupCoordinates.managedResourceGroupName), new MockDirectoryDAO(), mockMrgDAO)
+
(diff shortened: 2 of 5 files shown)


aherbst-broad:
jenkins retest

aherbst-broad:
jenkins retest

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutes.scala
```diff
@@ -71,7 +71,19 @@ trait AzureRoutes extends SecurityDirectives with LazyLogging {
                     }
                   }
                 }
-              }
+              } ~
+                delete {
+                  requireAction(
+                    FullyQualifiedResourceId(SamResourceTypes.spendProfile, ResourceId(billingProfileId)),
+                    SamResourceActions.delete,
```
Ghost-in-a-Jar, line 78: I am not super familiar with how our auth middleware is setup here, just to double check is there any additional authorization we can/should check on delete? On create it looks like there is a `requireCloudExtensionCreatePetAction` maybe we need a similar deletion requirement?

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceSpec.scala
```diff
@@ -364,4 +364,29 @@ class AzureServiceSpec extends AnyFlatSpec with Matchers with ScalaFutures {
...
+    val mockMrgDAO = new MockAzureManagedResourceGroupDAO
+    val managedResourceGroup = Generator.genManagedResourceGroup.sample.get
+    val svc =
+      new AzureService(MockCrlService(user, managedResourceGroup.managedResourceGroupCoordinates.managedResourceGroupName), new MockDirectoryDAO(), mockMrgDAO)
+
+    val err = intercept[WorkbenchExceptionWithErrorReport] {
+      svc.deleteManagedResourceGroup(managedResourceGroup.billingProfileId, samRequestContext.copy(samUser = user)).unsafeRunSync()
+    }
+    err.errorReport.statusCode shouldBe Some(StatusCodes.NotFound)
+  }
```
Ghost-in-a-Jar, line 391: Are there any tests that make sense to add around auth or handling failure of auth?

Ghost-in-a-Jar (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutes.scala
aherbst-broad: I believe that `requireCloudExtensionCreatePetAction` is required on creation of a pet managed identity, which has a hard requirement on the azure cloud extension. For creation of an MRG, I don't think we have any such [requirement,](https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutes.scala#L61) so this is symmetrical.

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceSpec.scala
aherbst-broad: The assumption is that the assertion in the route for DELETE on the owning spend profile should be sufficient. The create methods in this service assert that the creating user is the one baked into the properties of the MRG -- I don't think we want to make a similar assertion at the time of deletion. That is, if I have DELETE permissions on the billing profile, the deletion should succeed regardless of the properties on the MRG.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutes.scala
dvoet: I don't think there is any other check required

dvoet (approved):

PR #920, opened 2022-12-02
ID-363 let Sam host privacy policy in tandem with ToS

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-363

What:

Makes Sam host the Terra Privacy Policy as well as the Terms of Service.

Why:

The Terms of Service and Privacy are updated at the same time, and users need to accept the new bundle of documents before continuing to use Terra. Having Sam host both the Terms of Service and Privacy Policy means both can be updated at once with a single config change, meaning we don't need to coordinate UI and Sam PRs. It also means the Privacy Policy version is configurable per-environment, which is really useful for keeping prod and pre-prod on different versions.

How:

Just like the Terms of Service, the Privacy Policy is now kept in Sam's resources folder, and versioned. Sam now has the ability to serve up the text of the Privacy Policy, and the text served is based on Sam's configured Terms of Service version.

---

**PR checklist**

### DIFF
--- a/src/main/resources/tos/privacyPolicy-0.md
+++ b/src/main/resources/tos/privacyPolicy-0.md
@@ -1,0 +1,95 @@
+If there are any questions about privacy, please email [privacy@broadinstitute.org](mailto:privacy@broadinstitute.org).
+
+We keep your name, email and company so you can use our system.
+
+We place cookies on your browser to make our system faster.
+
+We may use your name and email to send you information about our system.
+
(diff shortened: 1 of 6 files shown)


gpolumbo-broad (changes requested):
Should probably add some tests

tlangs:
@gpolumbo-broad added a test

Ghost-in-a-Jar (approved):

gpolumbo-broad (approved):

PR #922, opened 2022-12-06
WOR-595: add missing actions to roles

blakery:
Ticket: [WOR-595](https://broadworkbench.atlassian.net/browse/WOR-595)

What:
Adding list_children to the google-project notebook-user role
~~Adding list_children to the workflow-collection owner role~~

Why:
Allows for validation before deleting a workspace that the workspace does not contain any resources that will cause deletion to fail

---

**PR checklist**

Link returns 404

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -514,11 +514,14 @@ resourceTypes = {
       read_policies = {
         description = "view all policies and policy details for this workflow-collection"
       }
+      list_children = {
+        description = "list child resources"
+      }
     }
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["delete", "add", "view", "abort", "get_cost", "update", "alter_policies", "read_policies"]
+        roleActions = ["delete", "add", "view", "abort", "get_cost", "update", "alter_policies", "read_policies", "list_children"]
       }
       reader = {
         roleActions = ["view", "get_cost"]
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -514,14 +514,11 @@ resourceTypes = {
       read_policies = {
         description = "view all policies and policy details for this workflow-collection"
       }
-      list_children = {
-        description = "list child resources"
-      }
     }
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["delete", "add", "view", "abort", "get_cost", "update", "alter_policies", "read_policies", "list_children"]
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -607,7 +613,7 @@ resourceTypes = {
         }
       }
       notebook-user = {
-        roleActions = ["launch_notebook_cluster", "create_persistent_disk", "create_kubernetes_app", "add_child"]
+        roleActions = ["launch_notebook_cluster", "create_persistent_disk", "create_kubernetes_app", "add_child", "list_children"]
```
marctalbott: This should be added to the `owner` role above instead

src/main/resources/reference.conf
```diff
@@ -514,11 +514,14 @@ resourceTypes = {
...
       }
+      list_children = {
+        description = "list child resources"
+      }
     }
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["delete", "add", "view", "abort", "get_cost", "update", "alter_policies", "read_policies"]
+        roleActions = ["delete", "add", "view", "abort", "get_cost", "update", "alter_policies", "read_policies", "list_children"]
```
marctalbott: I'm having second thoughts on adding this action to the `workflow-collection` resource type. None of the roles currently allow users to set children on a `workflow-collection`, so I don't know that we really need the ability to list out children since there can't be any

src/main/resources/reference.conf
blakery: My thinking was that `add_child` > `list_children` (and `notebook-user` is included in the `owner` role), so it made sense for it to be here.

marctalbott: `add_child` is greater than `list_children`, but in this case I'd advocate to keep the `notebook-user` role scoped to actions that are required to run a notebook in a Google project, and isolate all child-related actions in the `owner` role. If there's a need for `notebook-user`s to see the children in the future, then we could always add that action at that time

src/main/resources/reference.conf
blakery: Is it just that the ticket was being overly cautious, then?
I'll remove requirement to check the workflow-collection from the ticket as well

marctalbott: Yeah, I think I included the note on the `workflow-collection` because if it did actually have a child then that would cause issues, but that feels overly cautious after seeing the existing role actions

marctalbott (approved):

aherbst-broad (approved):

PR #928, opened 2022-12-09
IA-3814 Pact provider test

Ghost-in-a-Jar:
Ticket: [IA-3814](https://broadworkbench.atlassian.net/browse/IA-3814)

What:

- Set up project structure and dependencies for Pact.IO using `pact4s`(https://github.com/jbwheatley/pact4s)
- Develop provider tests to verify contract `leo-consumer-sam-provider` from [Pact Broker](https://pact-broker.dsp-eng-tools.broadinstitute.org/)
- Develop the CI/CD pipeline (GHA workflow) to run provider tests and publish the results to Pact Broker

Why:

1. Give a jump start to [Contract Testing Pod](https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2536505347/Contract+Testing+Pod)
2. This work provides a blueprint on contract tests development in Scala

How:

This PR demonstrates the lay of the land on what we have learnt from writing contract tests using `pact4s`(https://github.com/jbwheatley/pact4s) for both consumer and provider.

---

**PR checklist**

[IA-3814]: https://broadworkbench.atlassian.net/browse/IA-3814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/verify_consumer_pacts.yml
```diff
@@ -0,0 +1,64 @@
...
+          docker run --rm -v $PWD:/working \
+                          -v jar-cache:/root/.ivy \
+                          -v jar-cache:/root/.ivy2 \
+                          -w /working \
+                          -e BRANCH=${{ steps.extract-branch.outputs.branch }} \
+                          -e GIT_SHA_SHORT=${{ steps.extract-branch.outputs.sha_short }} \
+                          -e PACT_BROKER_URL=${{ env.PACT_BROKER_URL }} \
+                          -e PACT_BROKER_USERNAME=${{ secrets.PACT_BROKER_USERNAME }} \
+                          -e PACT_BROKER_PASSWORD=${{ secrets.PACT_BROKER_PASSWORD }} \
+                          sbtscala/scala-sbt:openjdk-17.0.2_1.7.2_2.13.10 \
```
mspector, line 63: just curious - does this image need to be downloaded each time the workflow runs? I wonder if it will make sense (in the future) for us to build a docker image to use instead of `ubuntu:latest` that comes preloaded with tools useful to DSP teams using Pact...

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
```diff
@@ -0,0 +1,155 @@
...
+import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
+import org.broadinstitute.dsde.workbench.sam.{Generator, MockSamDependencies, MockTestSupport}
+import org.broadinstitute.dsde.workbench.util.health.{StatusCheckResponse, SubsystemStatus, Subsystems}
+import org.mockito.ArgumentMatchers.any
+import org.mockito.Mockito.when
+import org.scalatest.BeforeAndAfterAll
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatestplus.mockito.MockitoSugar.mock
+import pact4s.provider.Authentication.BasicAuth
+import pact4s.provider._
```
mspector, line 25: I'm a Scala novice, so if this is bad advice, feel free to ignore...

I think this imports everything in `pact4s.provider` into the local namespace, right? Would it be better to explicitly import objects by name, so that future readers of this code can more easily see which parts of this spec are Pact-specific?

It won't stop me from approving, just thought it was worth mentioning.

mspector (approved):
Looks good to me -- take my approval with a grain of salt, though, as I am not too familiar with Sam!

.github/workflows/verify_consumer_pacts.yml
ichengchang: [DSP Toolbox](https://github.com/broadinstitute/dsp-toolbox) used to be the place to go for this type of tools you just mentioned. However, we are trying not to do this now due to security concerns. Also release version can change quite often, so it's best to leave it to developers to decide which image has the required dependencies. It also allows developers with a different jdk version to run the test locally using the correct image.

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
ichengchang: In this case, a handful of scala classes are being utilized:

`ProviderInfoBuilder, PactSource, Branch, PublishVerificationResults, ProviderVerificationOption`

They are all from `pact4s.provider._` , it could have been written as

`pact4s.provider.{ProviderInfoBuilder, PactSource, Branch, PublishVerificationResults, ProviderVerificationOption}`

but here we just use the shorthand `pact4s.provider._` 

We will soon have another ticket to refactor boilerplate code, this is when pact-specific code will be lumped together.

ichengchang (approved):
Approved on behalf of Tristan.

PR #939, opened 2022-12-21
ID-360 Add timing metrics user service.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-360

What:

Adding openapi timing metrics to sam user endpoints so that we can get some visibility into which endpoints are slow and when.

Why:

To get more visibility into how sam behaves in production, should help with debugging etc.

How:

I wrapped each user service call in a openTelemetry.time call, which requires an IO so I also had to do a little refactoring. 

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.workbench.sam.google
 import akka.actor.SupervisorStrategy.{Escalate, Stop}
 import akka.actor._
 import akka.pattern._
+import cats.effect.unsafe.implicits.global
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.google.GooglePubSubDAO
 import org.broadinstitute.dsde.workbench.google.GooglePubSubDAO.PubSubMessage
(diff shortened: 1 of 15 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
```diff
@@ -164,12 +165,13 @@ class DisableUsersMonitorActor(
...
     }
 
-  private def attemptToDisableUser(message: PubSubMessage) = {
+  private def attemptToDisableUser(message: PubSubMessage): Future[Product] = {
     logger.debug(s"received disable user message: $message")
     val userId = WorkbenchUserId(message.contents)
     Tracing.trace("DisableUsersMonitor-PubSubMessage") { span =>
       userService
         .disableUser(userId, samRequestContext = SamRequestContext(Option(span)))
+        .unsafeToFuture()
```
Qi77Qi, line 174: might as well just change it to IO?

Qi77Qi (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUsersMonitor.scala
Ghost-in-a-Jar: I tried and it was a bit of a rabbit hole, I'd like to leave that for another pr.

PR #940, opened 2022-12-22
ID-360 Add timing metrics to managed groups service.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-360

 What:

Adding openapi timing metrics to sam user endpoints so that we can get some visibility into which endpoints are slow and when.

Why:

To get more visibility into how sam behaves in production, should help with debugging etc.

How:

I wrapped each user service call in a openTelemetry.time call, which requires an IO so I also had to do a little refactoring.
---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
@@ -4,11 +4,12 @@ import akka.http.scaladsl.model.StatusCodes
 import cats.effect.IO
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.model._
+import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
 import org.broadinstitute.dsde.workbench.sam._
 import org.broadinstitute.dsde.workbench.sam.dataAccess.{AccessPolicyDAO, DirectoryDAO}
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ManagedGroupService.scala
```diff
@@ -33,7 +35,7 @@ class ManagedGroupService(
       samUser: SamUser,
       accessInstructionsOpt: Option[String] = None,
       samRequestContext: SamRequestContext
-  ): IO[Resource] = {
+  ): IO[Resource] = openTelemetry.time("api.v1.managedGroup.create.time", API_TIMING_DURATION_BUCKET) {
```
Qi77Qi, line 38: do u need a new duration bucket? I'm guessing group sync will take much longer than api response time (assuming `API_TIMING_DURATION_BUCKET` was for API response time)

Ghost-in-a-Jar: I was sort of hoping one bucket could be used for all the calls, I tried to do a pretty large range

Qi77Qi: feels weird to me that one bucket fits all....but if things look good once u chart them, I'm fine with it

Qi77Qi (approved):

PR #955, opened 2023-01-17
Add `export`s to turn shell variables into environment variables

breilly2:
Maybe bash handles this differently, but zsh seems to only set these as environment variables if they're exported. `test.env` and `src/main/resources/rendered/secrets.env` already use `export` so this change brings `local.env` in line with those.


Ghost-in-a-Jar (approved):

Ghost-in-a-Jar:
Seems reasonable, the jenkens integration tests are a bit flakey right now so I would ignore them as far as merging this since it is just for local dev.

Shakespeared (approved):

PR #961, opened 2023-01-23
ID-407 Don't allow new users to use API

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-407

What:

As a new Terra user, I should not be able to use Sam's APIs without accepting the Terms of Service, even during a grace period. If I'm an existing user, I should be able to use Sam's APIs with a previously-accepted version of the Terms of Service during a grace period. If I'm an existing user and have not accepted the current Terms of Service version and there is no grace period, I should not be able to use Sam's APIs.

Why:

We've never actually enabled a grace period in prod, so we missed the "As a new Terra user, I should not be able to use Sam's APIs without accepting the Terms of Service, even during a grace period" case. This PR fixes that.

How:

Simple making the `isTermsOfServiceStatusAcceptable` method check that a user has accepted some ToS version if there's a grace period, instead of just checking that there's a grace period.

---

**PR checklist**

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
@@ -19,6 +19,7 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
   private val tosServiceEnabledV0 = new TosService(dirDAO, "example.com", TestSupport.tosConfig.copy(enabled = true, version = "0"))
   private val tosServiceEnabled = new TosService(dirDAO, "example.com", TestSupport.tosConfig.copy(enabled = true))
   private val tosServiceEnabledV2 = new TosService(dirDAO, "example.com", TestSupport.tosConfig.copy(enabled = true, version = "2"))
+  private val tosServiceGracePeriodEnabled = new TosService(dirDAO, "example.com", TestSupport.tosConfig.copy(enabled = true, version = "2", isGracePeriodEnabled = true))
   private val tosServiceDisabled = new TosService(dirDAO, "example.com", TestSupport.tosConfig.copy(enabled = false))
 
   override protected def beforeAll(): Unit = {
@@ -52,15 +53,7 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
     assertResult(expected = false, s"getTosStatus(${defaultUser.id}) should have returned false")(actual = getTosStatusResultRejected.get)
   }
 
-  it should "exclude service accounts from ToS checks" in {
-    assume(databaseEnabled, databaseEnabledClue)
-
-    dirDAO.createUser(serviceAccountUser, samRequestContext).unsafeRunSync()
-
-    tosServiceEnabledV0.isTermsOfServiceStatusAcceptable(serviceAccountUser) should be(true)
-  }
-
-  it should "accept new version of ToS" in {
+  it should "allow users to accept new version of ToS" in {
     assume(databaseEnabled, databaseEnabledClue)
 
     dirDAO.createUser(defaultUser, samRequestContext).unsafeRunSync()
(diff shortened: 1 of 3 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -73,4 +67,38 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
     tosServiceEnabledV2.acceptTosStatus(defaultUser.id, samRequestContext).unsafeRunSync() shouldBe Option(true)
     tosServiceEnabledV2.getTosStatus(defaultUser.id, samRequestContext).unsafeRunSync() shouldBe Option(true)
   }
+
+  "TosService.isTermsOfServiceStatusAcceptable" should "allow all requests to the API if TOS is disabled" in {
+    assume(databaseEnabled, databaseEnabledClue)
+    dirDAO.createUser(defaultUser, samRequestContext).unsafeRunSync()
```
Ghost-in-a-Jar, line 69: nit: maybe instead of defaultUser it could be named more specifically to its tos state?

Ghost-in-a-Jar (approved):
minor nitpick otherwise looks good

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
tlangs: `defaultUser` is used because that's the one generated for each test case on line 16

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -73,4 +67,38 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
+    val userAcceptedPreviousVersion = dirDAO.loadUser(defaultUser.id, samRequestContext).unsafeRunSync().orNull
+    tosServiceGracePeriodEnabled.isTermsOfServiceStatusAcceptable(userAcceptedPreviousVersion) should be(true)
+  }
+
+  it should "not allow users that have accepted a previous ToS to use the API without a grace period" in {
+    assume(databaseEnabled, databaseEnabledClue)
+    dirDAO.createUser(defaultUser, samRequestContext).unsafeRunSync()
+    dirDAO.acceptTermsOfService(defaultUser.id, "1", samRequestContext).unsafeRunSync()
+    val userAcceptedPreviousVersion = dirDAO.loadUser(defaultUser.id, samRequestContext).unsafeRunSync().orNull
+    tosServiceEnabledV2.isTermsOfServiceStatusAcceptable(userAcceptedPreviousVersion) should be(false)
```
gpolumbo-broad, line 100: Might want to try being more explicit with these tests overall.  I know there are default settings coming in, and that's not bad, but these tests would be even better if you can set up the ToS config.  For example, in all of these tests, I don't know if ToS is enabled or Disabled or what the starting state of the grace period is, I need to go look that up in the config.  I think to truly test all the scenarios here, we need to be able to set the config into all sorts of different states.  I don't want to necessarily drag this fix out too long, but I think just taking a second look at this to see if you can be more explicit within each test, these could be better and we can do a better job making sure we have all the possible scenarios covered.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
gpolumbo-broad: I have a slight preference towards generating a new test user within the test too, since that is what we are testing here, but I would not hold up the PR for this fix, especially since the way you are doing it here is the pattern used by all the other existing tests.  However, going forward, I do think it makes the test a little more resilient to refactoring and less likely to conflict with other tests if we (within reason) can keep the tests well isolated and sharing (subjectively) appropriate setup/fixture logic.  This is one of the skills we need to sharpen, when is or is not the right time to do this.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
tlangs: I've gone through the new (and existing) tests and made it more explicit about the state of `TosService`

tlangs:
jenkins retest

gpolumbo-broad (approved):

PR #966, opened 2023-01-25
ID-398 Terms of Service Code Improvements

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-398

What:

When we released the original implementation of Terms of Service grace period functionality a couple weeks ago, it was very much an MVP tacked on-top of the existing code. Now that the Terms of Service version has been incremented, we can take some time to improve the code and make the Terms of Service a first-class feature.

Why:

Better code, better Sam, ~Papa Johns~.

How:

Some notable changes:

- Terms of Service cannot be disabled. We will never be operating without a terms of service, so the option to disable it has been removed
- The new TermsOfServiceAdherence endpoint will be used by the UI going forward, reducing the amount of logic needed in Terra UI. 
- Terms of Service "Acceptability" is now consolidated into one function, removing the disjointed way ToS status was being retrieved.
- Code quality improvements
- Lots of test changes

The majority of lines in this PR are simply fallout from changes in `TosService`. Start there for reviewing.

---

**PR checklist**


tlangs:
jenkins retest

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2694,6 +2694,19 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/TermsOfServiceDetails'
+  /register/user/v2/self/termsOfServiceAdherenceStatus:
```
gpolumbo-broad: Maybe a nit - I think of "Adherence" as the things users do or don't do while using the system.  I am wondering if "Agreement" may be a better term?  I think we ruled out "Acceptance" already, right?  Naming is hard.  If we want to get change this, now is the time, so I think it might be worth spending time with the team to get it as good as we can.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3392,6 +3405,22 @@ components:
         userAcceptedVersion:
           type: string
           description: the version of the terms of service the user has accepted
+    TermsOfServiceAdherenceStatus:
+      required:
+        - userId
+        - userHasAcceptedLatestTos
+        - acceptedTosAllowsUsage
```
gpolumbo-broad:
```suggestion
        - permitSystemUsage
```
More naming nit

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2694,6 +2694,19 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/TermsOfServiceDetails'
+  /register/user/v2/self/termsOfServiceAdherenceStatus:
+    get:
+      tags:
+        - Terms of Service
+      summary: gets info on whether the user has accepted the latest terms of service and if they can use Terra based on that
```
gpolumbo-broad: Just thinking about this, we probably say "Terra" in plenty of places in the swagger UI and in the code, but I think we kind of technically shouldn't.  I think the more we think of Terra as a generic open source tool that could be used by anybody, we will be more likely to design and build better code for Terra even.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3392,6 +3405,22 @@ components:
...
+      properties:
+        userId:
+          type: string
+          description: user id terms of service adherence pertains to
+        userHasAcceptedLatestTos:
+          type: boolean
+          description: has the user accepted the latest terms of service?
+        acceptedTosAllowsUsage:
+          type: boolean
+          description: based on terms of service status, can the user use Terra?
```
gpolumbo-broad: Just following up on my comment from above:
```suggestion
          description: based on the user's currently accepted terms of service version and when the terms of service were last changed, should the user still be permitted to use the system?
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -101,10 +101,10 @@ object StandardSamUserDirectives {
   def getActiveSamUser(oidcHeaders: OIDCHeaders, directoryDAO: DirectoryDAO, tosService: TosService, samRequestContext: SamRequestContext): IO[SamUser] =
     for {
       user <- getSamUser(oidcHeaders, directoryDAO, samRequestContext)
+      tosAcceptanceDetails <- tosService.getTosAdherenceStatus(user)
     } yield {
       // service account users do not need to accept ToS
-      val tosStatusAcceptable = tosService.isTermsOfServiceStatusAcceptable(user)
-      if (!tosStatusAcceptable) {
+      if (!tosAcceptanceDetails.acceptedTosAllowsUsage) {
         throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Unauthorized, "User has not accepted the terms of service."))
```
gpolumbo-broad:
```suggestion
        throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.Unauthorized, "User must accept the latest terms of service."))
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -101,10 +101,10 @@ object StandardSamUserDirectives {
   def getActiveSamUser(oidcHeaders: OIDCHeaders, directoryDAO: DirectoryDAO, tosService: TosService, samRequestContext: SamRequestContext): IO[SamUser] =
     for {
       user <- getSamUser(oidcHeaders, directoryDAO, samRequestContext)
+      tosAcceptanceDetails <- tosService.getTosAdherenceStatus(user)
     } yield {
       // service account users do not need to accept ToS
```
gpolumbo-broad: It appears to me like any logic relating to this comment is gone or moved?  This statement seems like an important one none the less.  Somewhere we must be checking if the request is coming in for an SA and we're letting them pass ToS.  Either way, if it was here before, I would not be surprised if someone is depending on this and I hope we are still behaving this way and have tests to confirm it is still working.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -123,11 +125,13 @@ object SamResourceTypes {
...
     inGoogleProxyGroup: Boolean,
-    tosAccepted: Option[Boolean],
+    tosAccepted: Boolean,
     adminEnabled: Boolean
 )
 @Lenses final case class TermsOfServiceAcceptance(value: String) extends ValueObject
 
+@Lenses final case class TermsOfServiceAdherenceStatus(userId: WorkbenchUserId, userHasAcceptedLatestTos: Boolean, acceptedTosAllowsUsage: Boolean)
+
 @Lenses final case class TermsOfServiceDetails(isEnabled: Boolean, isGracePeriodEnabled: Boolean, currentVersion: String, userAcceptedVersion: Option[String])
```
gpolumbo-broad, line 141: Is this deprecated as of this PR?  Should we mark the corresponding pieces throughout the code as deprecated too?  Typically for stuff like this I like to make sure there is a ticket in our backlog for getting rid of the deprecated stuff and I like to link to that ticket in the comments with the deprecation tag

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -3,58 +3,53 @@ package org.broadinstitute.dsde.workbench.sam.service
...
-  def getTosStatus(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Option[Boolean]] =
-    if (tosConfig.enabled) {
-      getUserAcceptedVersion(userId, samRequestContext).map(_.map(_.contains(tosConfig.version)))
-    } else IO.none
-
-  def getUserAcceptedVersion(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Option[Option[String]]] =
-    directoryDao.loadUser(userId, samRequestContext).map(_.map(_.acceptedTosVersion))
+  def getTosAdherenceStatus(samUser: SamUser): IO[TermsOfServiceAdherenceStatus] = {
+    val userHasAcceptedLatestVersion = userHasAcceptedLatestTosVersion(samUser)
+    val canUseTerra = tosAcceptanceAllowsTerraUsage(samUser)
```
gpolumbo-broad: Again, maybe `canUseSystem` or something like that just so it is not explicit to Terra, technically Sam could be used for anything.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -3,58 +3,53 @@ package org.broadinstitute.dsde.workbench.sam.service
...
+    IO.pure(TermsOfServiceAdherenceStatus(samUser.id, userHasAcceptedLatestVersion, canUseTerra))
+  }
 
   /** If grace period enabled, don't check ToS, return true If ToS disabled, return true Otherwise return true if user has accepted ToS, or is a service account
```
gpolumbo-broad, line 43:
```suggestion
  /** IF user has accepted Current ToS, THEN return TRUE 
     * ELSE IF grace period enabled AND user has accepted ANY ToS, THEN return TRUE
     * ELSE IF user is a service account, THEN return TRUE
     * ELSE return FALSE
     */
```
I wonder if it might be easier to just code it like this comment is written out, then I think you can get ride of all of the inline comments making the logic easier to understand (in my opinion) and making the code self documenting.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -206,10 +208,10 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
             // ticket: https://broadworkbench.atlassian.net/browse/ID-266
             ldap <- adminEnabledStatus
             allUsers <- allUsersStatus
-            tosAccepted <- tosAcceptedStatus
+            tosAccepted <- tosAcceptanceStatus
             google <- googleStatus
             adminEnabled <- adminEnabledStatus
-          } yield Option(UserStatusDiagnostics(ldap, allUsers, google, tosAccepted, adminEnabled))
+          } yield Option(UserStatusDiagnostics(ldap, allUsers, google, tosAccepted.acceptedTosAllowsUsage, adminEnabled))
```
gpolumbo-broad: hmmm...(scratches head) I guess this is out of scope for this ticket… but maybe not… I think the reason we were declaring the `val`s on 197-202 and then using the `for` comp here was because all of those `val`s used to be `Future`s that would start immediately.  Now I think all of those return `IO`.  I am not an IO expert by any means, but I think we don't need to have the declaration block outside of the `for` comp anymore.  Either that, or we need to to actually kick off the IOs or something in the declaration block.  I'm not sure.  Again, maybe this is totally out of scope for this ticket, but this also seems like a minor enough little simplification that I'd be OK doing it as part of this PR (because otherwise we'll NEVER do it) so long as you feel confident in how IO works vs. Future here.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
```diff
@@ -68,7 +68,7 @@ class ManagedGroupRoutesSpec extends AnyFlatSpec with Matchers with ScalatestRou
 
   // Makes an anonymous object for a user acting on the same data as the user specified in samRoutes
   def makeOtherUser(samRoutes: TestSamRoutes, samUser: SamUser = defaultNewUser) = new {
-    runAndWait(samRoutes.userService.createUser(samUser, samRequestContext))
+    samRoutes.createUserAndAcceptTos(samUser, samRequestContext)
```
gpolumbo-broad, line 71: why remove the `runAndWait` here?

tlangs:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
```diff
@@ -381,7 +381,16 @@ trait GoogleExtensionRoutesSpecHelper extends AnyFlatSpec with Matchers with Sca
       status shouldEqual StatusCodes.Created
       val res = responseAs[UserStatus]
       res.userInfo.userEmail shouldBe user.email
-      res.enabled shouldBe Map("ldap" -> true, "allUsersGroup" -> true, "google" -> true)
+      res.enabled shouldBe TestSupport.enabledMapNoTosAccepted
+    }
+
+    // accept the tos
+    Post("/register/user/v1/termsofservice", TermsOfServiceAcceptance("app.terra.bio/#terms-of-service")) ~> createRoutes.route ~> check {
```
gpolumbo-broad, line 388: Why is this test in this file?  I would expect it to be in the UserRoutesSpecs

tlangs:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
```diff
@@ -171,14 +188,22 @@ class UserServiceSpec
...
     // it should be enabled
     dirDAO.isEnabled(defaultUser.id, samRequestContext).unsafeRunSync() shouldBe true
 
     // disable the user
     val response = service.disableUser(defaultUser.id, samRequestContext).unsafeRunSync()
-    response shouldBe Some(UserStatus(UserStatusDetails(defaultUser.id, defaultUser.email), Map("ldap" -> false, "allUsersGroup" -> true, "google" -> true)))
+    response shouldBe Some(
+      UserStatus(
+        UserStatusDetails(defaultUser.id, defaultUser.email),
+        Map("ldap" -> false, "allUsersGroup" -> true, "google" -> true, "tosAccepted" -> false, "adminEnabled" -> false)
```
gpolumbo-broad: Think we need a test for when we disable a user who had accepted the ToS?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
   protected def clearDatabase(): Unit =
     TestSupport.truncateAll
 
-  "TosService" should "accept, get, and reject the ToS for a user" in {
+  "TosService.acceptTosStatus" should "accept the ToS for a user" in {
```
gpolumbo-broad:
```suggestion
  "TosService" should "accept the ToS for a user" in {
```
Something I just read in https://learning.oreilly.com/library/view/unit-testing-principles/9781617296277/ , putting the method name in the test is not recommended since it ties the test name to the implementation.  The name of the method is irrelevant, what we want to assert is the behavior that a user can accept the ToS.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
     assume(databaseEnabled, databaseEnabledClue)
 
     dirDAO.createUser(defaultUser, samRequestContext).unsafeRunSync()
-    val tosServiceEnabled = new TosService(dirDAO, "example.com", TestSupport.tosConfig.copy(enabled = true))
+    val tosService = new TosService(dirDAO, TestSupport.tosConfig)
```
gpolumbo-broad: This will make the test a little weird because this isn't really doing much, but I think you want to mock `DirectoryDao.acceptTermsOfService` here to just return true.  This test then doesn't do that much, but that's ok, it's not about how meety the test is, it's about verifying the functionality/behavior.  

Maybe for this test all we want to do is verify that the mocked method was called.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
-    tosServiceEnabledV0.getTosStatus(defaultUser.id, samRequestContext).unsafeRunSync() shouldBe Option(false)
-    tosServiceEnabledV0.acceptTosStatus(defaultUser.id, samRequestContext).unsafeRunSync() shouldBe Option(true)
-    tosServiceEnabledV0.getTosStatus(defaultUser.id, samRequestContext).unsafeRunSync() shouldBe Option(true)
-
-    val tosServiceEnabledV2 = new TosService(dirDAO, "example.com", TestSupport.tosConfig.copy(enabled = true, version = "2"))
-    tosServiceEnabledV2.getTosStatus(defaultUser.id, samRequestContext).unsafeRunSync() shouldBe Option(false)
-    tosServiceEnabledV2.acceptTosStatus(defaultUser.id, samRequestContext).unsafeRunSync() shouldBe Option(true)
-    tosServiceEnabledV2.getTosStatus(defaultUser.id, samRequestContext).unsafeRunSync() shouldBe Option(true)
+    // reject and get ToS status
+    val tosServiceV2 = new TosService(dirDAO, TestSupport.tosConfig.copy(version = "0"))
```
gpolumbo-broad: Looks identical to line 55?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
+    val rejectTosStatusResult = tosServiceV2.rejectTosStatus(defaultUser.id, samRequestContext).unsafeRunSync()
+    assert(rejectTosStatusResult, s"rejectTosStatus(${defaultUser.id}) should reject the tos for the user")
+    val userRejectedToS = dirDAO.loadUser(defaultUser.id, samRequestContext).unsafeRunSync().orNull
+    val adherenceStatus = tosServiceV2.getTosAdherenceStatus(userRejectedToS).unsafeRunSync()
+    assertResult(expected = false, s"getTosStatus(${defaultUser.id}) should have returned false")(actual = adherenceStatus.userHasAcceptedLatestTos)
```
gpolumbo-broad: Again, think about what you can test here if we mock the `dirDAO`, let's just assume that that is going to work correctly.  For our functionality here, we should not care about the database actually getting updated.  All we need to care about is that when I call the dao I get the answer back that I want.  Again, that means this test might just be making sure we are making the right calls to the dao or just that the `TosService` has a method on it that we can call to reject.  The goal is to try to isolate testing of "We need to be able to reject ToS" for a user, from the implementation of the dao.  The tests here should not care about how the dao is implemented.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
+    tosServiceV0.acceptTosStatus(defaultUser.id, samRequestContext).unsafeRunSync() shouldBe true
+    val userAcceptedV0 = dirDAO.loadUser(defaultUser.id, samRequestContext).unsafeRunSync().orNull
+    val adherenceStatusV0 = tosServiceV0.getTosAdherenceStatus(userAcceptedV0).unsafeRunSync()
+    adherenceStatusV0.userHasAcceptedLatestTos shouldBe true
+
+    // Accept V2
+    val tosServiceV2 = new TosService(dirDAO, TestSupport.tosConfig.copy(version = "2"))
+    val adherenceStatusNotAcceptedLatest = tosServiceV2.getTosAdherenceStatus(userAcceptedV0).unsafeRunSync()
+    adherenceStatusNotAcceptedLatest.userHasAcceptedLatestTos shouldBe false
+    adherenceStatusNotAcceptedLatest.acceptedTosAllowsUsage shouldBe false
```
gpolumbo-broad: For this test, I don't think I should be asserting on `.acceptedTosAllowsUsage`  Whether that is true or false is not part of what's under test here.  Yes, this coverage is more thorough or complete, but it adds complexity to read and maintain and introduces ways for this test to fail that are not the intended focus per the name of the test.  Surely we have other tests that assert all the various permutations and calculations for `.acceptedTosAllowsUsage` and those should cover the scenario constructed here and should do so more thoroughly and precisely, thus negating the need for the additional/duplicative testing here.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
+
+    tosServiceV2.acceptTosStatus(userAcceptedV0.id, samRequestContext).unsafeRunSync() shouldBe true
+    val userAcceptedV2 = dirDAO.loadUser(userAcceptedV0.id, samRequestContext).unsafeRunSync().orNull
+    val adherenceStatusV2 = tosServiceV2.getTosAdherenceStatus(userAcceptedV2).unsafeRunSync()
+    adherenceStatusV2.userHasAcceptedLatestTos shouldBe true
+    adherenceStatusV2.acceptedTosAllowsUsage shouldBe true
```
gpolumbo-broad: Again, think about how you can test this with a mocked dao.  If there is logic for rules pertaining to which versions the user can or cannot accept based on the currently accepted version, those rules should live in the TosService and we should have tests for them here.  If this test is just, "A user who previously accepted the ToS can accept a new version of the ToS", then maybe the solution is to just mock dao calls that get the ToS version so that they return a specific version and then run your test trying to accept a new/different version.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
   }
 
-  it should "not allow users who have never accepted a ToS version to use the API, even with a grace period enabled" in {
```
gpolumbo-broad, line 74: I'm interpreting this old test as being covered by Use Case 4?  Do you agree?  Just want to make sure we didn't loose this check.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
+  it should "always allow service account users to use Terra" in {
     assume(databaseEnabled, databaseEnabledClue)
```
gpolumbo-broad: assumption not needed if using `mockDirDAO`?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
-    dirDAO.createUser(defaultUser, samRequestContext).unsafeRunSync()
     val tosServiceGracePeriodEnabled =
-      new TosService(dirDAO, "example.com", TestSupport.tosConfig.copy(enabled = true, version = "2", isGracePeriodEnabled = true))
-    tosServiceGracePeriodEnabled.isTermsOfServiceStatusAcceptable(defaultUser) should be(false)
+      new TosService(mockDirDAO, TestSupport.tosConfig.copy(version = "2"))
+    val adherenceStatus = tosServiceGracePeriodEnabled.getTosAdherenceStatus(serviceAccountUser).unsafeRunSync()
+    adherenceStatus.userHasAcceptedLatestTos shouldBe false
```
gpolumbo-broad: I don't think we need this assertion.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
+    * | 1    | false                | null             | "2"             | false                | false              |
+    * | 2    | false                | "0"              | "2"             | false                | false              |
+    * | 3    | false                | "2"              | "2"             | true                 | true               |
+    * | 4    | true                 | null             | "2"             | false                | false              |
+    * | 5    | true                 | "0"              | "2"             | false                | true               |
+    * | 6    | true                 | "2"              | "2"             | true                 | true               |
+    */
+
+  // CASE 1
+  "TosService.getAdherenceStatus without a grace period" should "say a new user cannot use Terra and hasn't accepted ToS" in {
```
gpolumbo-broad:
```suggestion
  "Adherence Status userHasAcceptedLatestTos" should " be false if a user has never accepted any ToS" in {
```

Again, think of it as close to the behavior you're trying to describe as possible, "If a user has _never_ accepted ToS, they should not be allowed to do anything."

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
```diff
@@ -29,82 +33,137 @@ class TosServiceSpec extends AnyFlatSpec with TestSupport with BeforeAndAfterAll
...
-    val userAcceptedPreviousVersion = dirDAO.loadUser(defaultUser.id, samRequestContext).unsafeRunSync().orNull
+  /** | Case | Grace Period Enabled | Accepted Version | Current Version | User accepted latest | User Can Use Terra |
+    * |:-----|:---------------------|:-----------------|:----------------|:---------------------|:-------------------|
+    * | 1    | false                | null             | "2"             | false                | false              |
+    * | 2    | false                | "0"              | "2"             | false                | false              |
+    * | 3    | false                | "2"              | "2"             | true                 | true               |
+    * | 4    | true                 | null             | "2"             | false                | false              |
+    * | 5    | true                 | "0"              | "2"             | false                | true               |
+    * | 6    | true                 | "2"              | "2"             | true                 | true               |
+    */
```
gpolumbo-broad: Might want to consider breaking out the "Adherence Status" checks into their own close just for readability and such.  I do not feel strongly about this though, just as files get bigger, they can cause their own challenges and problems if not careful.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
gpolumbo-broad: Then the question becomes, should we have _separate_ tests for `.userHasAcceptedLatestTos` and `.acceptedTosAllowsUsage`?  I kind of think it might be good to do so.  When is field A true/false is different from When is field B true/false.

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -151,19 +151,21 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
               // leaving that logic unchanged for now.
               // ticket: https://broadworkbench.atlassian.net/browse/ID-266
-              val enabledMap = Map("ldap" -> adminEnabled, "allUsersGroup" -> allUsersStatus, "google" -> googleStatus)
-              val enabledStatuses = tosAcceptedStatus match {
-                case Some(status) => enabledMap + ("tosAccepted" -> status) + ("adminEnabled" -> adminEnabled)
-                case None => enabledMap
-              }
-              val res = Option(UserStatus(UserStatusDetails(user.id, user.email), enabledStatuses))
+              val enabledMap = Map(
+                "ldap" -> adminEnabled,
```
Shakespeared, line 163: can we get rid of this ldap yet?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -182,17 +184,17 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+      tosAcceptanceDetails <- tosService.getTosAdherenceStatus(user)
+    } yield UserStatusInfo(user.id.value, user.email.value, tosAcceptanceDetails.acceptedTosAllowsUsage && user.enabled, user.enabled)
 
   def getUserStatusDiagnostics(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Option[UserStatusDiagnostics]] =
     openTelemetry.time("api.v1.user.statusDiagnostics.time", API_TIMING_DURATION_BUCKET) {
       directoryDAO.loadUser(userId, samRequestContext).flatMap {
         case Some(user) =>
           // pulled out of for comprehension to allow concurrent execution
-          val tosAcceptedStatus = tosService.getTosStatus(user.id, samRequestContext)
+          val tosAcceptanceStatus = tosService.getTosAdherenceStatus(user)
```
Shakespeared: is there a difference between `tosAcceptanceStatus` and `tosAdherenceStatus`? It looks like you've used both in this file

src/main/resources/swagger/api-docs.yaml
Shakespeared: What about "compliance"?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3392,6 +3405,22 @@ components:
         userAcceptedVersion:
```
Shakespeared, line 3405: might be slightly redundant but `userAcceptedTosVersion` would be more explicit

Shakespeared (changes requested):
Generally LGTM, I think there's value in doing one more renaming pass just to make sure it's clear for future us!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
tlangs: Nope! Just renaming things as a go and forgetting to change all references!

src/main/resources/swagger/api-docs.yaml
tlangs: I like "compliance"!

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
tlangs: This check happens in `TosService.tosAcceptanceAllowsTerraUsage`, which is called while gathering the compliance status. I'll move the comment so it doesn't cause confusion.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
tlangs: I don't consider the code cleanup "done" until these deprecated features are removed. I'm gonna mark them as Deprecated for easy finding though!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
tlangs: returning `true/false` in `if/else` is a code smell and will get flagged by SonarCloud.

tlangs: re-written to be clearer

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
tlangs: You're right, we don't need to declare these `val`s outside of the `for` comprehension, but I do think that's out-of-scope for this ticket. In an argument for the separate declarations, it does make the `for` comprehension a bit cleaner.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/ManagedGroupRoutesSpec.scala
tlangs: `samRoutes.createUserAndAcceptTos` creates the user and accepts the ToS. These are two different calls, and both are called with `runAndWait`. The two actions were pulled into one method because we need to create user that has accepted the ToS all over our tests.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
tlangs: Thats... a great question...

tlangs: OH! Because this is creating a test user for all the `GoogleExtensionRoutesSpec`. The user needs to accept the terms of service. This block is following the paradigm of the block above it, which creates the user, and then makes sure it was successfully created.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpec.scala
tlangs: Good point. I'll do that.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
tlangs: TIL!

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
tlangs: Whoops! Copypasta

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
tlangs: yes!

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TosServiceSpec.scala
tlangs: Since we're testing the results matrix, I think its good to make sure all outputs are as expected

tlangs:
jenkins retest

Shakespeared (approved):
LGTM, and the tests have been so engoodened!

tlangs:
jenkins retest

src/main/resources/swagger/api-docs.yaml
gpolumbo-broad: I don't love compliance, sorry, not trying to be difficult.  But at the same time, I don't feel strongly that I know of something better.  If you want to stick with Compliance, that's fine with me I guess, but if you want to try to find another word, I propose we talk face-to-face with the team to brainstorm what we want.  But like I said, if this is what we want to stick with, that's fine.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2694,6 +2694,19 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/TermsOfServiceDetails'
+  /register/user/v2/self/termsOfServiceComplianceStatus:
+    get:
+      tags:
+        - Terms of Service
+      summary: gets info on whether the user has accepted the latest terms of service and if they can use the system based on that
```
gpolumbo-broad, line 2701: I wonder if we should consider getting much for formal and descriptive with our swagger API summaries?  No need to do so on this PR if you don't want to, but if you wanted to give it a shot too, I don't think it would be inappropriate.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
gpolumbo-broad: NICE!  I like this a lot!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
gpolumbo-broad: Yeah, comprehension is important, but also understanding the behavior difference when declaring `Future`s inside a `for` comp vs. outside is very important.  I believe that with `IO`s, this is not a problem (because unlike `Future`, an `IO` doesn't start running immediately when declared).  We would just want to make sure this behaves as we want.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
gpolumbo-broad: Would love to, but we will need to bump the endpoint version when doing so because we are removing a field from the response object.  This is something we SHOULD do, just not in this PR.

gpolumbo-broad (approved):

tlangs:
jenkins retest

PR #968, opened 2023-01-26
ID-402 Log SamUser object from Directives

gpolumbo-broad:
Ticket: https://broadworkbench.atlassian.net/browse/ID-402

What:

For tracking and troubleshooting reasons, it is useful to log information from the request headers and match them to the corresponding user found in Sam

Why:

To help with troubleshooting and production support

How:

There are varied ways for pretty printing a case class.  This seemed the most straight forward for what we are trying to accomplish here.  

As for where to log, that was a little tricky because we want to make sure we always log the OIDCHeaders on pretty much every request that comes in.  This solution should do that and additionally log the corresponding user we load from the DB if one is found.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
@@ -9,6 +9,7 @@ import akka.http.scaladsl.server.directives.OnSuccessMagnet._
 import cats.effect.IO
 import cats.effect.unsafe.implicits.global
 import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.workbench.sam.util.AsyncLogging.IOWithLogging
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.model.google.ServiceAccountSubjectId
 import org.broadinstitute.dsde.workbench.sam.api.StandardSamUserDirectives._
@@ -65,7 +66,7 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with 
   }
 }
 
-object StandardSamUserDirectives {
+object StandardSamUserDirectives extends LazyLogging {
   val SAdomain: Regex = "(\\S+@\\S*gserviceaccount\\.com$)".r
   // UAMI == "User Assigned Managed Identity" in Azure
   val UamiPattern: Regex = "(^/subscriptions/\\S+/resourcegroups/\\S+/providers/Microsoft\\.ManagedIdentity/userAssignedIdentities/\\S+$)".r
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -28,13 +28,19 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with
 
   def withActiveUser(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.flatMap { oidcHeaders =>
     onSuccess {
-      getActiveSamUser(oidcHeaders, directoryDAO, tosService, samRequestContext).unsafeToFuture()
+      getActiveSamUser(oidcHeaders, directoryDAO, tosService, samRequestContext).unsafeToFuture().map(samUser => {
+        logger.info(s"Handling request for active Sam User: $samUser")
```
tlangs: Do we need all the info from SamUser, or just the userId?

Ghost-in-a-Jar (approved):
The only thing i worry about here is logging secrets, but we keep the prod logs under tight lock and key, and we already log sensitive info in sam prod. Just wanted to call it out.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
gpolumbo-broad: I think it is better to log the full user for debugging azure/google or whatever

gpolumbo-broad:
@Ghost-in-a-Jar - no secrets are being logged.  Sensitive information being logged in the users googleSubjectId and Azure B2C ID and email, all stuff that is already logged and the logs have protected access.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -29,12 +29,18 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with
   def withActiveUser(samRequestContext: SamRequestContext): Directive1[SamUser] = requireOidcHeaders.flatMap { oidcHeaders =>
     onSuccess {
       getActiveSamUser(oidcHeaders, directoryDAO, tosService, samRequestContext).unsafeToFuture()
+    }.tmap { samUser =>
+      logger.info(s"Handling request for active Sam User: $samUser")
```
tlangs, line 33: Logging the entire `SamUser` is going to create a lot of chatter. Do we need to know the accepted ToS version for each user request? Their enabled status? Their Google Subject ID or Azure B2C ID? I think it might make sense to log their email and userId, but more that that is just more log volume.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -152,4 +163,18 @@ final case class OIDCHeaders(
...
+  override def toString: String = {
+    val extId = externalId match {
+      case Left(googleSubjectId) => s"GoogleSubjectId($googleSubjectId)"
+      case Right(azureB2CId) => s"AzureB2CId($azureB2CId)"
+    }
+    s"OIDCHeaders(" +
+      s"externalId: $extId, " +
+      s"email: $email, " +
+      s"googleSubjectIdFromAzure: $googleSubjectIdFromAzure, " +
+      s"managedIdentityObjectId: $managedIdentityObjectId)"
```
tlangs, line 178: Do we want to be logging all of these fields for every request?

tlangs (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
gpolumbo-broad: Yes, that is the goal of the ticket.

tlangs (approved):

PR #981, opened 2023-02-08
ZEBRA-54: Add Zebrafish MCT Resource Types

walarue:
Ticket: [ZEBRA-54](https://broadworkbench.atlassian.net/browse/ZEBRA-54)

What:

Zebrafish is a service that is being developed by a consultant (Manifold) in partnership with the Broad Institute. For now, code is in a private repo run by Manifold, and JIRA is a Manifold account (linked anyway above).

Part of Zebrafish is the Mapping Curation Tool (MCT). This is a user interface that helps define mappings between data sets and a target schema. The point is ingesting data into TDR with the given target schema (project led by the DSP Data Ingest team).

Why:

In order to run within the Terra security perimeter and have SAM be the source of truth for authorization, we need to represent MCT resources in SAM. There are two resource types to protect:

- `mct_mapping`
    - this is the configuration for transforming the source dataset to the target schema
- `mct_target`
    - this is the target schema itself
    - over time, the tool may be used for different target schemas 
   
These resources are very simple. They are essentially documents in the application's database. These documents should be publicly readable. We just need to protect owner and editor access to these documents. The documents aren't subdivided, so it's as simple as being able to write or not write, etc.

The point of this PR is defining those resource types. The definitions are copied from `controlled-user-shared-workspace-resource` since that resource type had the required actions and roles.

I prefixed the resource types with `mct_` for clarity/specificity, but that was just a guess at a naming convention, I can use whatever naming scheme is preferred.

How:

MCT resource types defined in `reference.conf`

---

**PR checklist**

[ZEBRA-54]: https://broadworkbench.atlassian.net/browse/ZEBRA-54?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1358,6 +1358,70 @@ resourceTypes = {
     }
     reuseIds = true
   }
+  mct_mapping = {
+    actionPatterns = {
+      own = {
+        description = "perform owner actions on the resource"
+      }
+      delete = {
+        description = "delete this resource"
+      }
+      read_policies = {
+        description = "view all policies and policy details for this resource"
+      }
+      write = {
+        description = "perform writer actions on the resource"
+      }
+      "share_policy::owner" = {
+        description = "change the membership of the owner policy for this resource"
+      }
+      "share_policy::editor" = {
+        description = "change the membership of the editor policy for this resource"
+      }
+    }
+    ownerRoleName = "owner"
+    roles = {
+      owner = {
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "own", "write"]
+      }
+      editor = {
+        roleActions = ["write", "share_policy::editor"]
+      }
+    }
+    reuseIds = false
+  }
+  mct_target = {
+    actionPatterns = {
+      own = {
+        description = "perform owner actions on the resource"
+      }
+      delete = {
+        description = "delete this resource"
+      }
+      read_policies = {
+        description = "view all policies and policy details for this resource"
+      }
+      write = {
+        description = "perform writer actions on the resource"
+      }
+      "share_policy::owner" = {
+        description = "change the membership of the owner policy for this resource"
+      }
+      "share_policy::editor" = {
+        description = "change the membership of the editor policy for this resource"
+      }
+    }
+    ownerRoleName = "owner"
+    roles = {
+      owner = {
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "own", "write"]
+      }
+      editor = {
+        roleActions = ["write", "share_policy::editor"]
+      }
+    }
+    reuseIds = false
+  }
 }
 
 
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1385,7 +1385,7 @@ resourceTypes = {
         roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "own", "write"]
       }
       editor = {
-        roleActions = ["write", "share_policy::editor"]
+        roleActions = ["write"]
       }
     }
     reuseIds = false
@@ -1417,7 +1417,7 @@ resourceTypes = {
         roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "own", "write"]
       }
       editor = {
-        roleActions = ["write", "share_policy::editor"]
+        roleActions = ["write"]
       }
     }
     reuseIds = false


src/main/resources/reference.conf
```diff
@@ -1354,6 +1354,106 @@ resourceTypes = {
...
   }
+  mct_mapping = {
+    actionPatterns = {
+        delete = {
+            description = "delete this resource"
+        }
+        read_policies = {
+            description = "view all policies and policy details for this resource"
+          }
+          set_parent = {
```
gpolumbo-broad: Will this resource need a parent/child relationship?  If not, you can just get rid of this action

src/main/resources/reference.conf
```diff
@@ -1354,6 +1354,106 @@ resourceTypes = {
...
+          "share_policy::reader" = {
+            description = "change the membership of the reader policy for this resource"
+          }
+          "read" = {
+            description = "perform reader actions on the resource"
+          }
+          "write" = {
+            description = "perform writer actions on the resource"
+          }
+          "edit" = {
```
gpolumbo-broad: again, not sure what actions you need here and I'm not sure if you need both `write` and `edit` actions.  These actions can be whatever you want, there are no particular actionPatterns that must be present here, so just make sure you're defining actions you actually care about.  So maybe you do want `write` and `edit`, I have no idea what the product needs are here, just wanted to clarify.

src/main/resources/reference.conf
```diff
@@ -1354,6 +1354,106 @@ resourceTypes = {
...
+            description = "perform editor actions on the resource"
+          }
+    }
+    ownerRoleName = "owner"
+    roles = {
+      owner = {
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "read", "write", "set_parent"]
```
gpolumbo-broad: I don't see an `own` action defined in the `actionPatterns`

src/main/resources/reference.conf
```diff
@@ -1354,6 +1354,106 @@ resourceTypes = {
...
+    reuseIds = false
+  }
+  mct_target = {
+    actionPatterns = {
+        delete = {
+            description = "delete this resource"
+        }
+        read_policies = {
+            description = "view all policies and policy details for this resource"
+          }
```
gpolumbo-broad: nit - formatting - note the indentation for this block from here down

src/main/resources/reference.conf
gpolumbo-broad: same goes for `mct_target`

gpolumbo-broad (changes requested):

src/main/resources/reference.conf
walarue: added, thanks

src/main/resources/reference.conf
walarue: at least initially we don't need it, thanks

src/main/resources/reference.conf
walarue: removed edit because it doesn't have a clear use for this resource

src/main/resources/reference.conf
walarue: my bad, fixed

src/main/resources/reference.conf
```diff
@@ -1358,6 +1358,103 @@ resourceTypes = {
...
+      "share_policy::reader" = {
+        description = "change the membership of the reader policy for this resource"
+      }
+    }
+    ownerRoleName = "owner"
+    roles = {
+      owner = {
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::editor", "share_policy::writer", "share_policy::reader", "own", "edit", "read", "write", "set_parent"]
+      }
+      editor = {
```
nmalfroy, line 1419: Is it expected that only the owner is allowed to share?  I guess I'm not familiar with the use case but I would imagine that, for instance, an editor might want to share with readers?  Just as an example.

walarue: that's a good call, I added that level of sharing (read-only) to the editor role

src/main/resources/reference.conf
walarue: note: removed the `edit` action because as you point out there's no real distinction here, but left the editor role which is kind of a super-writer with more sharing

ncalvanese1 (approved):
LGTM!

src/main/resources/reference.conf
dvoet: editors being able to share with other editors is a viral sharing model, it is probably not what you want

dvoet: also, being able to share a policy without being able to read policies is a tough position to put users in

walarue: yup, that was dumb, thanks for calling it out. removed sharing for the editor role

gpolumbo-broad (approved):

gpolumbo-broad:
jenkins retest this.  Again.  Please… sheesh

gpolumbo-broad:
jenkins retest

PR #982, opened 2023-02-08
ID-455 improve google group synchronizer

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/ID-455
see the ticket for details about what and why

much of the refactoring is extracting functions for better readability. A split view might be better

---

**PR checklist**

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -14,7 +14,7 @@ object Dependencies {
   val workbenchUtilV = "0.6-74c9fc2"
   val workbenchUtil2V = "0.2-447afa5"
   val workbenchModelV = "0.15-f9f0d4c"
-  val workbenchGoogleV = "0.21-8ce5b9b"
+  val workbenchGoogleV = "0.22-abd44a6"
   val workbenchGoogle2V = "0.24-447afa5"
   val workbenchNotificationsV = "0.3-d74ff96"
   val workbenchOauth2V = "0.2-20f9225"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -11,14 +11,15 @@ object Dependencies {
   val postgresDriverVersion = "42.5.0"
   val sentryVersion = "6.6.0"
 
-  val workbenchUtilV = "0.6-74c9fc2"
-  val workbenchUtil2V = "0.2-447afa5"
-  val workbenchModelV = "0.15-f9f0d4c"
-  val workbenchGoogleV = "0.22-abd44a6"
-  val workbenchGoogle2V = "0.24-447afa5"
-  val workbenchNotificationsV = "0.3-d74ff96"
-  val workbenchOauth2V = "0.2-20f9225"
-  val workbenchOpenTelemetryV = "0.3-0096bac"
+  val workbenchLibV = "abd44a6"
+  val workbenchUtilV = s"0.6-$workbenchLibV"
+  val workbenchUtil2V = s"0.2-$workbenchLibV"
+  val workbenchModelV = s"0.15-$workbenchLibV"
+  val workbenchGoogleV = s"0.22-$workbenchLibV"
+  val workbenchGoogle2V = s"0.25-$workbenchLibV"
+  val workbenchNotificationsV = s"0.3-$workbenchLibV"
+  val workbenchOauth2V = s"0.2-$workbenchLibV"
+  val workbenchOpenTelemetryV = s"0.3-$workbenchLibV"
   val monocleVersion = "2.0.5"
   val crlVersion = "1.2.4-SNAPSHOT"
   val slf4jVersion = "2.0.6"
@@ -31,6 +32,7 @@ object Dependencies {
   val excludeWorkbenchMetrics = ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-metrics_2.12")
   val excludeWorkbenchGoogle = ExclusionRule(organization = "org.broadinstitute.dsde.workbench", name = "workbench-google_2.12")
   val excludeGoogleCloudResourceManager = ExclusionRule(organization = "com.google.apis", name = "google-api-services-cloudresourcemanager")
+  val excludeGoogleServiceUsage = ExclusionRule(organization = "com.google.apis", name = "google-api-services-serviceusage")
   val excludeSLF4J = ExclusionRule(organization = "org.slf4j")
   val excludeJerseyCore = ExclusionRule(organization = "org.glassfish.jersey.core", name = "*")
   val excludeJerseyMedia = ExclusionRule(organization = "org.glassfish.jersey.media", name = "*")
@@ -90,8 +92,9 @@ object Dependencies {
   // leading to merge errors during sbt assembly. At this time the old version of auto-value is pulled in through the google2
   // workbench-libs dependency so exclude auto-value from there
   val excludGoogleAutoValue = ExclusionRule(organization = "com.google.auto.value", name = "auto-value")
+  val excludeSlf4j = ExclusionRule("org.slf4j")
   val workbenchGoogle2: ModuleID =
-    "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V excludeAll (excludeWorkbenchModel, excludeWorkbenchUtil, excludGoogleAutoValue)
+    "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V excludeAll (excludeWorkbenchModel, excludeWorkbenchUtil, excludGoogleAutoValue, excludeSlf4j)
   val workbenchNotifications: ModuleID =
     "org.broadinstitute.dsde.workbench" %% "workbench-notifications" % workbenchNotificationsV excludeAll (excludeWorkbenchGoogle, excludeWorkbenchModel)
   val workbenchGoogleTests: ModuleID =
@@ -142,7 +145,7 @@ object Dependencies {
   )
 
   val cloudResourceLib: ModuleID =
-    "bio.terra" % "terra-cloud-resource-lib" % crlVersion excludeAll (excludeGoogleCloudResourceManager, excludeJerseyCore, excludeJerseyMedia, excludeSLF4J)
+    "bio.terra" % "terra-cloud-resource-lib" % crlVersion excludeAll (excludeGoogleServiceUsage, excludeGoogleCloudResourceManager, excludeJerseyCore, excludeJerseyMedia, excludeSLF4J)
   val azureManagedApplications: ModuleID =
     "com.azure.resourcemanager" % "azure-resourcemanager-managedapplications" % "1.0.0-beta.1"
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -44,7 +44,7 @@ object Dependencies {
   val jacksonCore: ModuleID = "com.fasterxml.jackson.core" % "jackson-core" % jacksonV
 
   val logstashLogback: ModuleID = "net.logstash.logback" % "logstash-logback-encoder" % "6.6"
-  val logbackClassic: ModuleID = "ch.qos.logback" % "logback-classic" % "1.2.11"
+  val logbackClassic: ModuleID = "ch.qos.logback" % "logback-classic" % "1.4.5"
   val ravenLogback: ModuleID = "com.getsentry.raven" % "raven-logback" % "7.8.6"
   val scalaLogging: ModuleID = "com.typesafe.scala-logging" %% "scala-logging" % scalaLoggingV
   val ficus: ModuleID = "com.iheart" %% "ficus" % "1.5.2"
@@ -92,10 +92,9 @@ object Dependencies {
   // leading to merge errors during sbt assembly. At this time the old version of auto-value is pulled in through the google2
   // workbench-libs dependency so exclude auto-value from there
   val excludGoogleAutoValue = ExclusionRule(organization = "com.google.auto.value", name = "auto-value")
-  val excludeSlf4j = ExclusionRule("org.slf4j")
   val excludeBouncyCastle = ExclusionRule("org.bouncycastle")
   val workbenchGoogle2: ModuleID =
-    "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V excludeAll (excludeWorkbenchModel, excludeWorkbenchUtil, excludGoogleAutoValue, excludeSlf4j, excludeBouncyCastle)
+    "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V excludeAll (excludeWorkbenchModel, excludeWorkbenchUtil, excludGoogleAutoValue, excludeBouncyCastle)
   val workbenchNotifications: ModuleID =
     "org.broadinstitute.dsde.workbench" %% "workbench-notifications" % workbenchNotificationsV excludeAll (excludeWorkbenchGoogle, excludeWorkbenchModel)
   val workbenchGoogleTests: ModuleID =
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -93,8 +93,9 @@ object Dependencies {
   // workbench-libs dependency so exclude auto-value from there
   val excludGoogleAutoValue = ExclusionRule(organization = "com.google.auto.value", name = "auto-value")
   val excludeSlf4j = ExclusionRule("org.slf4j")
+  val excludeBouncyCastle = ExclusionRule("org.bouncycastle")
   val workbenchGoogle2: ModuleID =
-    "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V excludeAll (excludeWorkbenchModel, excludeWorkbenchUtil, excludGoogleAutoValue, excludeSlf4j)
+    "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2V excludeAll (excludeWorkbenchModel, excludeWorkbenchUtil, excludGoogleAutoValue, excludeSlf4j, excludeBouncyCastle)
   val workbenchNotifications: ModuleID =
     "org.broadinstitute.dsde.workbench" %% "workbench-notifications" % workbenchNotificationsV excludeAll (excludeWorkbenchGoogle, excludeWorkbenchModel)
   val workbenchGoogleTests: ModuleID =
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -7,3 +7,5 @@ addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.0.6")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.5.0")
 
 addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.10.0")
+
+addDependencyTreePlugin
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -33,13 +33,11 @@ import org.broadinstitute.dsde.workbench.sam.google._
 import org.broadinstitute.dsde.workbench.sam.model._
 import org.broadinstitute.dsde.workbench.sam.service._
 import org.broadinstitute.dsde.workbench.util.DelegatePool
-import org.broadinstitute.dsde.workbench.util2.ExecutionContexts
 import org.typelevel.log4cats.StructuredLogger
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 
 import java.io.File
 import java.nio.file.{Files, Paths}
-import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.jdk.CollectionConverters._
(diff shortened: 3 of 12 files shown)


dvoet:
tracing is better too
image

project/Dependencies.scala
```diff
@@ -11,14 +11,15 @@ object Dependencies {
...
-  val workbenchOpenTelemetryV = "0.3-0096bac"
+  val workbenchLibV = "abd44a6"
+  val workbenchUtilV = s"0.6-$workbenchLibV"
+  val workbenchUtil2V = s"0.2-$workbenchLibV"
+  val workbenchModelV = s"0.15-$workbenchLibV"
+  val workbenchGoogleV = s"0.22-$workbenchLibV"
+  val workbenchGoogle2V = s"0.25-$workbenchLibV"
+  val workbenchNotificationsV = s"0.3-$workbenchLibV"
+  val workbenchOauth2V = s"0.2-$workbenchLibV"
+  val workbenchOpenTelemetryV = s"0.3-$workbenchLibV"
```
dvoet, line 22: I made all these the latest and consistent

project/plugins.sbt
```diff
@@ -7,3 +7,5 @@ addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.0.6")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.5.0")
 
 addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.10.0")
+
+addDependencyTreePlugin
```
tlangs, line 11: Do we want this plugin in the codebase? I'm not against it, and find it incredibly useful for development, but do we want it in the committed source?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -127,8 +125,6 @@ object Boot extends IOApp with LazyLogging {
         appConfig.samDatabaseConfig.samBackground
       )
 
-      googleSynchronizerExecutionContext <- ExecutionContexts.fixedThreadPool[IO](24)
```
dvoet, line 130: since running things serially via `IO` this execution context is no longer needed

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
```diff
@@ -132,28 +138,45 @@ class GoogleGroupSyncMonitorActor(
 
   override def receive = {
     case StartMonitorPass =>
-      logger.debug(s"Pulling messages off: $pubSubSubscriptionName")
+      logger.info(s"Pulling messages off: $pubSubSubscriptionName")
```
tlangs, line 141: Thanks for this. Its annoying that we use DEBUG logs everywhere in prod...

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
```diff
@@ -94,9 +99,10 @@ class GoogleGroupSyncMonitorSupervisor(
 object GoogleGroupSyncMonitor {
   case object StartMonitorPass
 
-  sealed abstract class SynchronizGroupMembersResult(ackId: String)
```
dvoet, line 97: fixed spelling

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
```diff
@@ -132,28 +138,45 @@ class GoogleGroupSyncMonitorActor(
...
     case StartMonitorPass =>
-      logger.debug(s"Pulling messages off: $pubSubSubscriptionName")
+      logger.info(s"Pulling messages off: $pubSubSubscriptionName")
       // start the process by pulling a message and sending it back to self
       pubSubDao.pullMessages(pubSubSubscriptionName, 1).map(_.headOption) pipeTo self
 
     case Some(message: PubSubMessage) =>
-      logger.debug(s"received sync message: $message")
-      import Tracing._
-      trace("GoogleGroupSyncMonitor-PubSubMessage") { span =>
```
dvoet, line 142: the original tracing here just traced the creation of the future

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
gpolumbo-broad: wow, even WITH this comment, it still took me 30 seconds to find the error.  Human brains are crazy…or at least mine is.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
```diff
@@ -132,28 +138,45 @@ class GoogleGroupSyncMonitorActor(
...
             groupId,
-            samRequestContext = SamRequestContext(Option(span))
-          ) // Since this is an internal pub/sub call, we have to start a new SamRequestContext.
-          .toTry
-          .map(sr => sr.fold(t => FailToSynchronize(t, message.ackId), x => ReportMessage(x, message.ackId))) pipeTo self
-      }
+            samRequestContext = samRequestContext
+          )
+          .redeem(t => FailToSynchronize(t, message.ackId), x => ReportMessage(x, message.ackId))
+          .guarantee(IO(deadlineExtender.cancel()))
```
tlangs: Some scala IO explanation for other reviewers:
`redeem` takes two functions, one to execute in the failure case and one to execute in the success case. 
`guarantee` will always run, no matter if the IO is successful, failed, or cancelled. This way, the `deadlineExtender` is always cancelled once the IO is finished.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
```diff
@@ -132,28 +138,45 @@ class GoogleGroupSyncMonitorActor(
...
+      }.unsafeToFuture() pipeTo self
+
+    case ExtendDeadline(ackId, deadlineSeconds, samRequestContext) =>
+      traceIOWithContext("extendDeadline", samRequestContext) { _ =>
+        IO.fromFuture(IO(pubSubDao.extendDeadlineById(pubSubSubscriptionName, Seq(ackId), deadlineSeconds))).timeout(10 seconds).handleError { t =>
```
tlangs, line 170: `IO.fromFuture, but being called on something wrapped in an `IO`? My brain hurts... Care to explain?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
```diff
@@ -49,90 +47,132 @@ class GoogleGroupSynchronizer(
...
-                } else {
-                  loadedPolicy
-                }
-              })
-        }
+        group <- loadGroup(groupId, samRequestContext)
+        members <- maybeCalculateIntersection(group, samRequestContext)
+        subGroupSyncs <- maybeSyncSubGroups(group, visitedGroups, samRequestContext)
+        googleMemberEmails <- loadGoogleGroupMemberEmailsMaybeCreateGroup(group, samRequestContext)
+        samMemberEmails <- loadSamMemberEmails(members, samRequestContext)
```
tlangs: I _LOVE_ the cleanup of this for comprehension!

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
dvoet: https://typelevel.org/cats-effect/docs/2.x/datatypes/io#fromfuture

project/plugins.sbt
dvoet: idc, up to you

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
tlangs: Ah, `extendDeadlineById` returns a `Future`. Missed that. 

Side-grump: Its annoying that we mix `IO` and `Future` everywhere...

dvoet: I agree. The google part of workbench-libs we use is all `Future`y. The google2 part is more compatible with `IO` but then requires swallowing `fs2.Stream`s and porting the google directory dao (and maybe others)

dvoet:
jenkins retest

dvoet:
jenkins retest after devops did a thing

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
```diff
@@ -132,28 +138,45 @@ class GoogleGroupSyncMonitorActor(
...
-          .toTry
-          .map(sr => sr.fold(t => FailToSynchronize(t, message.ackId), x => ReportMessage(x, message.ackId))) pipeTo self
-      }
+            samRequestContext = samRequestContext
+          )
+          .redeem(t => FailToSynchronize(t, message.ackId), x => ReportMessage(x, message.ackId))
+          .guarantee(IO(deadlineExtender.cancel()))
+      }.unsafeToFuture() pipeTo self
+
+    case ExtendDeadline(ackId, deadlineSeconds, samRequestContext) =>
```
gpolumbo-broad, line 168: 👃  -  This This `match` statement:
* This `match` statement is up to 9 `case` statements now
* 80 lines long
* The parent types for what we're matching on is varying: `SynchronizeGroupMembersResult`, `Option`, etc.
* Nested `match`es
* This `case` separates the `Some()` from the `None` case

Since you're in here and we're trying to make more sense of how the group syncing works and retries, etc, I propose that we keep the same `case` statements, but reorder them so the same parent type of what we're matching on is grouped together.  Additionally, extract everything on the right side of the `case` statements into their own methods that are specifically named for what they're doing.  

This all might be overkill for this PR, but I think that it might be helpful as we dig into this more.

tlangs (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
```diff
@@ -132,28 +138,45 @@ class GoogleGroupSyncMonitorActor(
...
     case Some(message: PubSubMessage) =>
-      logger.debug(s"received sync message: $message")
-      import Tracing._
-      trace("GoogleGroupSyncMonitor-PubSubMessage") { span =>
+      logger.info(s"received sync message: $message")
+
+      traceIO("GoogleGroupSyncMonitor-PubSubMessage", SamRequestContext()) { samRequestContext =>
+        // every 15s reset the deadline to 60s in the future, canceled when IO is done
+        // 15s so we get a few tries at extending before deadline occurs
+        val deadlineExtender = system.scheduler.scheduleAtFixedRate(Duration.Zero, 15 seconds, self, ExtendDeadline(message.ackId, 60, samRequestContext))
```
gpolumbo-broad, line 151: I have a lot of questions about what this is and what this is doing.  All of these questions stem from a lack of understanding of how the Actors actually work.  At first glance, it seems weird to me to continually reset the timeout to 60 seconds in the future every 15 seconds…how is that different from just letting the thing run until it completes?  If that _is_ what we want and Actors just _require_ a timeout, then let's just say that.  

Also - https://doc.akka.io/api/akka/current/akka/actor/Scheduler.html#scheduleAtFixedRate(initialDelay:java.time.Duration,interval:java.time.Duration,receiver:akka.actor.ActorRef,message:Any,executor:scala.concurrent.ExecutionContext,sender:akka.actor.ActorRef):akka.actor.Cancellable

> Warning: scheduleAtFixedRate can result in bursts of scheduled messages after long garbage collection pauses, which may in worst case cause undesired load on the system. Therefore scheduleWithFixedDelay is often preferred.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
```diff
@@ -30,15 +29,14 @@ import scala.util.{Failure, Success, Try}
...
-  * @param executionContext
   */
 class GoogleGroupSynchronizer(
     directoryDAO: DirectoryDAO,
     accessPolicyDAO: AccessPolicyDAO,
     googleDirectoryDAO: GoogleDirectoryDAO,
     googleExtensions: GoogleExtensions,
     resourceTypes: Map[ResourceTypeName, ResourceType]
-)(implicit executionContext: ExecutionContext)
+) //(implicit executionContext: ExecutionContext)
```
gpolumbo-broad: Is there a reason this should be kept just commented out?  If so, a comment is in order to explain why.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
```diff
@@ -49,90 +47,132 @@ class GoogleGroupSynchronizer(
...
+        }
+      case group: BasicWorkbenchGroup => IO.pure(group.members)
+    }
+
+  /** Loads the group whether a policy or basic group. If it is a public policy add the all users group to members.
+    * @param groupId
+    * @param samRequestContext
+    * @return
+    */
+  private def loadGroup(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[WorkbenchGroup] =
```
gpolumbo-broad: How about `loadSamGroup` to help make it explicit when we are loading from Sam and when we are loading from Google.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
```diff
@@ -49,90 +47,132 @@ class GoogleGroupSynchronizer(
...
+        }
+      case _ => IO.pure(Map.empty[WorkbenchEmail, Seq[SyncReportItem]])
+    }
+
+  /** If this is a policy group constrained by an auth domain calculate the intersection group membership, otherwise just return the group membership
+    * @param group
+    * @param samRequestContext
+    * @return
+    */
+  private def maybeCalculateIntersection(group: WorkbenchGroup, samRequestContext: SamRequestContext): IO[Set[WorkbenchSubject]] =
```
gpolumbo-broad: Suggest renaming to something like, `calculateAuthDomainIntersection` or `calculateIntersectionWithAuthDomain`.  

My general thought here is that we don't need the `maybe`, which sort of implies to me that I am getting an `Option` back.  And then since this only takes 1 group argument and intersects with its Auth Domain, let's make that clear in the name.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
```diff
@@ -49,90 +47,132 @@ class GoogleGroupSynchronizer(
...
     }
   }
 
+  /** synchronize any sub groups that have never been synchronized otherwise they don't exist in google and can't be added
+    * @param group
+    * @param visitedGroups
+    * @param samRequestContext
+    * @return
+    */
+  private def maybeSyncSubGroups(group: WorkbenchGroup, visitedGroups: Set[WorkbenchGroupIdentity], samRequestContext: SamRequestContext) =
```
gpolumbo-broad: Once again, I suggest we forget about the "maybe" here and just call it `syncSubGroups`.  

In addition, I'm wondering if we should put some sort of warning or flag in the comment too, to let people know that this is the thing that cascades to cause additional Group Syncs.  Yes, it's there in the body if you read the code, but to me, this is the sort of functionality that should get a red flag on it.  It is something we need to do of course, so putting some other warning or something here is probably also a good idea to get people's attention and warn them to be careful in here.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
```diff
@@ -49,90 +47,132 @@ class GoogleGroupSynchronizer(
...
 
-        toAdd = samMemberEmails -- googleMemberEmails
-        toRemove = googleMemberEmails -- samMemberEmails
+  /** convert each subject to an email address
+    *
+    * @param members
+    * @param samRequestContext
+    * @return
+    */
+  private def loadSamMemberEmails(members: Set[WorkbenchSubject], samRequestContext: SamRequestContext) =
```
gpolumbo-broad, line 86: This functionality seems like something we may want/need/do elsewhere, converting a `WorkbenchSubject` to `WorkbenchEmail` or `String`.  If you know of an easy way to pull this logic out into `WorkbenchSubject` or something, that seems useful, but looks to me like the DAOs make this messy.  Just wanted to mention it in case you know of an easy way to extract this and make it reusable.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
```diff
@@ -49,90 +47,132 @@ class GoogleGroupSynchronizer(
...
-        googleMemberEmails <- googleDirectoryDAO.listGroupMembers(group.email) flatMap {
-          case None =>
-            googleDirectoryDAO.createGroup(groupId.toString, group.email, Option(googleDirectoryDAO.lockedDownGroupSettings)) map (_ => Set.empty[String])
-          case Some(members) => Future.successful(members.map(_.toLowerCase).toSet)
-        }
-        samMemberEmails <- Future
-          .traverse(members) {
-            case group: WorkbenchGroupIdentity => directoryDAO.loadSubjectEmail(group, samRequestContext).unsafeToFuture()
+        addTrials <- toAdd.toList.traverse(addMemberToGoogleGroup(group, samRequestContext))
+        removeTrials <- toRemove.toList.traverse(removeMemberFromGoogleGroup(group, samRequestContext))
```
gpolumbo-broad: Maybe rename to `addedUserSyncReports` and `removedUserSyncReports` or `syncReportsForAddedUsers` and `syncReportsForRemovedUsers`.  I would rather not use the word "trials" here.  And I like the idea of indicating that these are the SyncReports or SyncResults for the various sets of users.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleModel.scala
```diff
@@ -1,13 +1,19 @@
...
   import org.broadinstitute.dsde.workbench.model.ErrorReportJsonSupport._
 
-  implicit val SyncReportItemFormat = jsonFormat3(SyncReportItem)
+  implicit val SyncReportItemFormat = jsonFormat3(SyncReportItem.apply)
 }
 
-case class SyncReportItem(operation: String, email: String, errorReport: Option[ErrorReport])
+final case class SyncReportItem(operation: String, email: String, errorReport: Option[ErrorReport])
+
+object SyncReportItem {
```
gpolumbo-broad, line 25: I don't love the name `SyncReportItem`, or maybe it needs a comment or something to explain what it is.  Perhaps out of scope for this ticket, but if our goal is to improve the synching stuff, maybe it isn't?  Up to you.

gpolumbo-broad (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMonitor.scala
dvoet: first, I have another PR coming that will tear all this actor stuff out. But it is too much to include in this PR so I want to do it next.

> At first glance, it seems weird to me to continually reset the timeout to 60 seconds in the future every 15 seconds…how is that different from just letting the thing run until it completes?

It is not every 15s reset to 60s from _now_. It is every 15s reset to 60s from _then_. It is just perpetually keeping the deadline between 45s to 60s in the future. This is also not about the actor timeout, it is about the pubsub message handling deadline, this is preventing redeliver of messages that are in process.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
dvoet: no, will remove

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
dvoet: I personally prefer the maybe to be clear that it is conditional but I am happy to do what you want

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
dvoet: in my head, `maybe` in a variable name is an `Option`, but `maybe` in a function name means it there is some conditionality but I understand overloading is not optimal.

how about instead of prefix with `maybe`, suffix with `IfRequired`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
dvoet: this is actually slightly special because it uses the user's proxy group email address instead of their actual email address

gpolumbo-broad (approved):

PR #965, opened 2023-01-24
[IA-4013] adding a kubernetes-app-shared resource

LizBaldo:
Ticket: https://broadworkbench.atlassian.net/browse/IA-3911

What:

Adds a kubernetes-app-shared sam resources to enable leonardo to create shared apps. See design doc: https://docs.google.com/document/d/1Qqljdgnzojca-8xVdpYZaEWBZs0htf59JA5Yfbeg8po/edit#heading=h.nwloytsh75t7

Why:

 The new user role will be able to connect to the URL of the shared app

How:

A new resource has been added to the `reference.conf` file with the new owner and user roles that inherit from the workspace roles

---

**PR checklist**

- [ ] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've filled out the [Security Risk Assessment](https://sdarq.dsp-appsec.broadinstitute.org/jira-ticket-risk-assesment) (requires Broad Internal network access) and attached the result to the JIRA ticket

PR #989, opened 2023-02-14
[IA-4012] give kubernetes-app resource workspace inheritance

LizBaldo:
Ticket: https://broadworkbench.atlassian.net/browse/IA-4012

What:

Give app permission inheritance from Workspace in addition to the Google Project. This will enable a better user experience with V2 apps.

Why:

 This way the behavior of the `kubernetes-app` resource will be similar to the new `kubernetes-app-shared` resource. 
https://github.com/broadinstitute/sam/pull/965

We want to keep the inheritance from the Google project for now to avoid any breaking change or need for a migration quite yet on legacy apps.

How:

Creator and manager roles both inherit from the workspace owner, while the manager role also inherit from the workspace reader and writer roles.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -619,7 +619,7 @@ resourceTypes = {
         roleActions = ["launch_notebook_cluster", "create_persistent_disk", "create_kubernetes_app", "add_child"]
         includedRoles = ["pet-creator"]
       }
-      pet-creator = {x
+      pet-creator = {
         roleActions = ["create-pet"]
       }
     }
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -156,6 +156,7 @@ resourceTypes = {
           controlled-user-private-workspace-resource = ["deleter"]
           controlled-application-shared-workspace-resource = ["writer", "reader"]
           wds-instance = ["owner"]
+          kubernetes-app = ["creator", "manager"]
           kubernetes-app-shared = ["owner", "user"]
         }
       }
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -156,6 +156,7 @@ resourceTypes = {
           controlled-user-private-workspace-resource = ["deleter"]
           controlled-application-shared-workspace-resource = ["writer", "reader"]
           wds-instance = ["owner"]
+          kubernetes-app = ["creator", "manager"]
```
rtitle: I think we only want to give workspace owners `manager` role on kubernetes-app. Similar to `google-project` inheritance: https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L608-L614

So here:
```
kubernetes-app = ["manager"]
```

src/main/resources/reference.conf
```diff
@@ -172,6 +173,7 @@ resourceTypes = {
           controlled-application-shared-workspace-resource = ["writer", "reader"]
           google-project = ["pet-creator"]
           wds-instance = ["writer"]
+          kubernetes-app = ["manager"]
```
rtitle: I don't think workspace writers or readers should inherit any permissions to `kubernetes-app`. Only workspace owners should be able to manage kubernetes-apps in the workspace.

So remove this line, and the below.

src/main/resources/reference.conf
```diff
@@ -182,6 +184,7 @@ resourceTypes = {
           controlled-application-shared-workspace-resource = ["reader"]
           google-project = ["pet-creator"]
           wds-instance = ["reader"]
+          kubernetes-app = ["manager"]
```
rtitle: Remove as well

rtitle (changes requested):
See above -- I think this PR will end up being a 1-liner :)

LizBaldo:
Indeed, I had it backwards 😵‍💫 
How about now?

rtitle (approved):
:shipit: !

jdcanas (approved):

PR #990, opened 2023-02-14
ID-458 switch to Subscriber from google for pubsub

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/ID-458

What:

The main gist of this PR is to stop using akka actors to monitor pubsub and instead do something [like this](https://github.com/googleapis/java-pubsub/blob/HEAD/samples/snippets/src/main/java/pubsub/SubscribeAsyncExample.java) where we supply a `MessageReceiver` implementation to a `Subscriber` class (maintained by Google).

A secondary thing came about which was an update to mockito because I wanted to use something new. The new version defaults to requiring that you mock only the required function calls, no more and no less. There are a few places where we use mockito mocks in a generic way. I had to a `lenient()` before the call to `when()` to allow these to keep working. In other places I removed unnecessary mocks. I am actually quite happy with this change to the library, I was alway worried that while I mocked a thing I though was right, my test would pass anyway.

Why:

Remove custom pubsub polling code that no one understands. Replace with something maintained and documented by Google.

How:

`GoogleKeyCacheMonitor`, `GoogleGroupSyncMonitor` and `DisableUserMonitor` have been removed. In their place there is a single `GooglePubSubMonitor` that is responsible for creating topics and subscriptions then setting up and starting the `Subscriber`. It also hooks into the actor system to catch termination events for graceful shutdown.

3 `MessageReceiver` implementations were created from the guts of the removed Monitor classes (their logic should be unchanged). The `MessageReceivers` encapsulate the logic for parsing messages, calling the right internal services to do the action, then handle success or failure and ack or nack the messages as appropriate. There are new tests for these classes, replacing the prior tests.

All the changes to existing test classes are related to mockito update.

---

**PR checklist**


src/test/scala/org/broadinstitute/dsde/workbench/sam/RetryableAnyFlatSpec.scala
```diff
@@ -7,7 +7,7 @@ trait RetryableAnyFlatSpec extends AnyFlatSpec with Retries {
 
   val retries = 3
 
-  override def withFixture(test: NoArgTest): Outcome =
+  override protected def withFixture(test: NoArgTest): Outcome =
```
dvoet, line 10: this is `protected` in the parent class, making it public here causes conflicts when mixing in other traits that have a similar definition

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/GoogleServicesConfig.scala
```diff
@@ -39,8 +39,6 @@ object GoogleServicesConfig {
   implicit val googlePubSubConfigReader: ValueReader[GooglePubSubConfig] = ValueReader.relative { config =>
     GooglePubSubConfig(
       config.getString("pubSubProject"),
-      org.broadinstitute.dsde.workbench.util.toScalaDuration(config.getDuration("pollInterval")),
-      org.broadinstitute.dsde.workbench.util.toScalaDuration(config.getDuration("pollJitter")),
```
dvoet, line 43: the google `Subscriber` takes care of polling for us now.

dvoet:
jenkins retest

dvoet:
jenkins retest after orch fix

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUserMessageReceiver.scala
```diff
@@ -0,0 +1,41 @@
...
+  override def receiveMessage(message: PubsubMessage, consumer: AckReplyConsumer): Unit =
+    OpenCensusIOUtils
+      .traceIO("DisableUsersMonitor-PubSubMessage", SamRequestContext()) { samRequestContext =>
+        val userId = WorkbenchUserId(message.getData.toStringUtf8)
+        logger.info(s"received disable user message: $userId")
+        userService
+          .disableUser(userId, samRequestContext)
+          .redeem(handleFailure(_, consumer), handleDisableUserResponse(userId, _, consumer))
+      }
+      .unsafeRunSync()
```
Ghost-in-a-Jar, line 24: This unsafeRunSync is a little smelly, just want to make sure we want to block receiving these messages. I could see a situation where we get backed up and get dead letters.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMessageReceiver.scala
```diff
@@ -0,0 +1,88 @@
...
+        _.putAttribute("groupId", AttributeValue.stringAttributeValue(groupId.toString))
+      )
+      groupSynchronizer
+        .synchronizeGroupMembers(
+          groupId,
+          Set.empty,
+          samRequestContext
+        )
+        .redeem(syncFailure(_, consumer), syncComplete(_, consumer))
+    }.unsafeRunSync()
```
Ghost-in-a-Jar, line 36: Same here, having unsafeRunSync in the receive message could end up with dead letters

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCacheMessageReceiver.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  override def receiveMessage(message: PubsubMessage, consumer: AckReplyConsumer): Unit =
+    OpenCensusIOUtils
+      .traceIO("GoogleKeyCacheMonitor--PubSubMessage", SamRequestContext()) { samRequestContext =>
+        logger.info(s"received key deletion message: ${message.getData.toStringUtf8}")
+        val (project, serviceAccountEmail, keyId) = parseMessage(message)
+        removeServiceAccountKey(project, serviceAccountEmail, keyId)
+          .redeem(handleError(consumer, _), handleSuccess(consumer, _))
+      }
+      .unsafeRunSync()
```
Ghost-in-a-Jar, line 28: same comment as other unsafeRunSync

Ghost-in-a-Jar (approved):
Looks good to me, the only thing I am worried about is unsafeRunSync in the receive method of the actors. I could be wrong but feels like it could cause dead letters.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/DisableUserMessageReceiver.scala
dvoet: It is intended. This subscriber stuff is running within a Java framework and is expected to be blocking. Managing the back pressure is part of the google library.

dvoet:
> unsafeRunSync in the receive method of the actors

These are no longer actors

> it could cause dead letters

I am not sure what you mean by dead letters. Dead letter to me means redelivery continually remains unacknowledged up to the retry limit (which is 7 days). If you mean unacknowledged, then they will be redelivered with an exponential backoff. If you are concerned they throw an exception, the `redeem` is meant to be sure that all exceptions are handled.

Shakespeared (approved):

PR #994, opened 2023-02-17
ID-428 Engoodened Invite User Specs

gpolumbo-broad:
Ticket: https://broadworkbench.atlassian.net/browse/ID-428

What:

Better described behavior for Invite User

Why:

To better describe and test the expected _behavior_ (not implementation) of `inviteUser`

How:

  \<For your reviewers' sake, please describe in ~1 paragraph how this PR accomplishes its goal.\>

  \<If the PR is big, please indicate where a reviewer should start reading it (i.e. which file or function).\>

---

**PR checklist**

- [ ] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've filled out the [Security Risk Assessment](https://sdarq.dsp-appsec.broadinstitute.org/jira-ticket-risk-assesment) (requires Broad Internal network access) and attached the result to the JIRA ticket

PR #998, opened 2023-02-21
MOB ID-440 Refactor/engooden CreateUserSpec and InviteUserspec.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-440

branches off of https://github.com/broadinstitute/sam/pull/994


src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpecs/CreateUserSpec.scala
```diff
@@ -27,71 +20,42 @@ class CreateUserSpec extends UserServiceTestTraits {
...
-
-        // Assert
-        inside(userStatus) { status =>
-          status should beForUser(userWithOnlyGoogleId)
-          "google" should beEnabledIn(status)
-          "ldap" should beEnabledIn(status)
-          "allUsersGroup" should beEnabledIn(status)
-          "adminEnabled" should beEnabledIn(status)
-          "tosAccepted" shouldNot beEnabledIn(status)
+  val userWithBoth = (genWorkbenchUserBoth.sample.get, "user with google and azure ids")
```
Shakespeared, line 23: I like the tuples, my only concern is that it makes it a little harder to read it in the code - although the test output is still clear. Maybe we could add a comment or a readme somewhere suggesting that as an easy way to debug/check expected behavior?  Maybe everyone would think to do that, but I'm not used to the test results printing out in such a helpful manner so I'm not sure if it would occur to me if I were a dev looking at this for the first time.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpecs/CreateUserSpec.scala
```diff
@@ -27,71 +20,42 @@ class CreateUserSpec extends UserServiceTestTraits {
...
+  val userWithGoogle = (genWorkbenchUserGoogle.sample.get, "user with google id")
+  val userWithAzure = (genWorkbenchUserAzure.sample.get, "user with azure id")
+  val newUsers = Vector(userWithBoth, userWithGoogle, userWithAzure)
+
+
+  describe("when a new user is created") {
+    describe("that does not already exist") {
+      newUsers.foreach { userTuple =>
+        val (newUser, description) = userTuple
+        describe(s"UserService.createUser with a $description") {
```
Shakespeared, line 33: Nit: can we reword this message a tiny bit? I find the phrasing of "when a new user… that does not already exist… Userservice.createUser with a user with google and azure ids" more confusing than it was before

Shakespeared: I think it needs one additional verb or something

Shakespeared (approved):

PR #1008, opened 2023-03-02
Refactor CreateUser

gpolumbo-broad:
Ticket: https://broadworkbench.atlassian.net/browse/ID-486 

What:

Refactors CreateUser to make it easier to read, hopefully easier to maintain, easier to test, and less reliant on side effects.

Why:

Test engoodening - our attempt to engooden tests was underwhelming because of design issues in the application code itself.  Now it is better.  Not fantastic, but better.  

How:

Start in UserService.createUser

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -26,20 +25,116 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+    }
+
+  // TODO: Add a simple "updateUser" method to directoryDAO so we can do all this in one call and return the updated user record
+  private def updateUser(existingUserId: WorkbenchUserId, user: SamUser, samRequestContext: SamRequestContext): IO[SamUser] =
+    openTelemetry.time("api.v1.user.updateUser.time", API_TIMING_DURATION_BUCKET) {
+      for {
+        _ <- if (user.googleSubjectId.nonEmpty) directoryDAO.setGoogleSubjectId(existingUserId, user.googleSubjectId.get, samRequestContext) else IO.unit
+        _ <- if (user.azureB2CId.nonEmpty) directoryDAO.setUserAzureB2CId(existingUserId, user.azureB2CId.get, samRequestContext) else IO.unit
+      } yield user.copy(id = existingUserId)
+    }
```
Ghost-in-a-Jar, line 135: If we want one of these to succeed regardless of whether the other succeeds we should add a unit test for that. I think if one fails the for yield will short circuit.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -26,20 +25,116 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
-        _ <- enableUserInternal(createdUser, samRequestContext)
-        _ <- addToAllUsersGroup(createdUser.id, samRequestContext)
-        userStatus <- getUserStatus(createdUser.id, samRequestContext = samRequestContext)
-        res <- IO
-          .fromOption(userStatus)(new WorkbenchException("getUserStatus returned None after user was created"))
-          .withInfoLogMessage(s"New user ${createdUser.toUserIdInfo} was successfully created")
-      } yield res
+        updatedUser <- updateUser(invitedUserId, invitedUser, samRequestContext)
+        groups <- directoryDAO.listUserDirectMemberships(updatedUser.id, samRequestContext)
+        _ <- IO.fromFuture(IO(cloudExtensions.onGroupUpdate(groups, samRequestContext)))
```
Ghost-in-a-Jar: I think we are converting from IO -> Future -> IO here

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
Ghost-in-a-Jar: Also if possible we should override the user id higher up in the call stack.

Ghost-in-a-Jar (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -26,20 +25,116 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+      val validationErrors = validateUser(possibleNewUser)
+      if (validationErrors.nonEmpty) {
+        return IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "invalid user", validationErrors.get)))
+      }
+
+      for {
+        newUser <- assertUserIsNotAlreadyRegistered(possibleNewUser, samRequestContext)
+        maybeWorkbenchSubject <- directoryDAO.loadSubjectFromEmail(newUser.email, samRequestContext)
+        registeredUser <- attemptToRegisterSubjectAsAUser(maybeWorkbenchSubject, newUser, samRequestContext)
+        registeredAndEnabledUser <- makeUserEnabled(registeredUser, samRequestContext)
```
tlangs, line 42: nitpick, but we can probably use `registeredUser` on line 47 instead of `registeredAndEnabledUser`, and use `_` on line 40 here, since the `id` and `email` aren't going to change through enablement.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -26,20 +25,116 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+      case None => IO(user)
+    }
+
+  // TODO: Add a simple "updateUser" method to directoryDAO so we can do all this in one call and return the updated user record
+  private def updateUser(existingUserId: WorkbenchUserId, user: SamUser, samRequestContext: SamRequestContext): IO[SamUser] =
+    openTelemetry.time("api.v1.user.updateUser.time", API_TIMING_DURATION_BUCKET) {
+      for {
+        _ <- if (user.googleSubjectId.nonEmpty) directoryDAO.setGoogleSubjectId(existingUserId, user.googleSubjectId.get, samRequestContext) else IO.unit
+        _ <- if (user.azureB2CId.nonEmpty) directoryDAO.setUserAzureB2CId(existingUserId, user.azureB2CId.get, samRequestContext) else IO.unit
+      } yield user.copy(id = existingUserId)
```
tlangs: The `if/else` within a `for` comprehension reads weirdly for me. Maybe a `match` would be easier to comprehend?

```scala
    val setId = (user.googleSubjectId, user.azureB2CId) match {
      case (Some(googleId), None) => directoryDAO.setGoogleSubjectId(existingUserId, googleId, samRequestContext)
      case (None, Some(azureId)) => directoryDAO.setUserAzureB2CId(existingUserId, azureId, samRequestContext)
    }
    setId.map(_ => user.copy(id = existingUserId))
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -26,20 +25,116 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+  // Try to find user by GoogleSubject, AzureB2CId
+  // A registered user is one that has a record in the database and has a Cloud Identifier specified
+  private def tryToFindUserByCloudId(user: SamUser, samRequestContext: SamRequestContext): IO[Option[SamUser]] =
+    openTelemetry.time("api.v1.user.tryToFindUserByCloudId.time", API_TIMING_DURATION_BUCKET) {
+      if (user.googleSubjectId.nonEmpty) {
+        directoryDAO.loadUserByGoogleSubjectId(user.googleSubjectId.get, samRequestContext)
+      } else if (user.azureB2CId.nonEmpty) {
+        directoryDAO.loadUserByAzureB2CId(user.azureB2CId.get, samRequestContext)
+      } else {
+        IO(None)
```
tlangs: `match` feels more concise than `if/else if/else`.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -26,20 +25,116 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+        _ <- IO.fromFuture(IO(cloudExtensions.onGroupUpdate(groups, samRequestContext)))
+      } yield {
+        updatedUser
+      }
     }
 
+  // For now, it looks like createUserInternal does what we need here, but added a new alias method here for naming
+  // consistency and just in case things change more as we are refactoring
+  private def registerBrandNewUser(possibleSamUser: SamUser, samRequestContext: SamRequestContext): IO[SamUser] =
+    createUserInternal(possibleSamUser, samRequestContext)
```
tlangs: It confuses me a little bit to wrap `createUserInternal` for one call, but leave it bare for another call. Would it be clearer instead to rename `createUserInternal` to something more descriptive, like `createUserInDatabase`?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -26,20 +25,116 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+      if (validationErrors.nonEmpty) {
+        return IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "invalid user", validationErrors.get)))
+      }
+
+      for {
+        newUser <- assertUserIsNotAlreadyRegistered(possibleNewUser, samRequestContext)
+        maybeWorkbenchSubject <- directoryDAO.loadSubjectFromEmail(newUser.email, samRequestContext)
+        registeredUser <- attemptToRegisterSubjectAsAUser(maybeWorkbenchSubject, newUser, samRequestContext)
+        registeredAndEnabledUser <- makeUserEnabled(registeredUser, samRequestContext)
+        _ <- addToAllUsersGroup(registeredAndEnabledUser.id, samRequestContext)
```
tlangs, line 43: I _think_ this call is unnecessary. When ` createUserInternal` is called, which happens when a user is invited or created, their proxy group is created and added to the all users group.

tlangs (changes requested):
I'm really happy to see us take a good hard look at how users are created. As the first interaction of a user with Sam, its crucial that the functionality is clear and easy to comprehend!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
tlangs: Also, this gets around the short-circuiting that @Ghost-in-a-Jar pointed out.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
gpolumbo-broad: `registeredAndEnabledUser.isEnabled` is set to `true` and that is how it differs from `registeredUser`

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
gpolumbo-broad: I think it is ok if this short circuits if one of the updates fails.  Per the comment, ideally, I would like to have a vanilla `updateUser` method that just updates both of these fields (and the whole user object) in one sql statement, but we don't have that right now.  If we had that though, it would be a "both succeed or nothing gets updated" scenario.  The catch there is that expanding this refactor into the `DirectoryDAO` I think blows up the scope a bit.  I'm  happy to do it if we feel like it makes sense for this PR.  Or I am happy to merge this as it is and add that new method later.  What do you think?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
gpolumbo-broad: Indeed, just changed the return type for `onGroupUpdate`

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
gpolumbo-broad: Yeah, it's a little odd.  The proxy group is added to the All Users group on Google when the user is created.  Then the user itself is not added to the All Users group in the Sam db until this step.  It's weird, I don't know why it's done this way.  I'm not going to say that we should not fix it now, but I think this PR is in a state where it at least replicates the behavior that was done before (right or wrong) but with hopefully more understandable code.  Fixing the All Users group stuff and CloudExtensions is something I'd really really like to do, but I think that's for another PR.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
Ghost-in-a-Jar: lets follow up on it later this pr already has enough in it

Ghost-in-a-Jar (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
tlangs: Could you make a backlog ticket to track that?

gpolumbo-broad:
jenkins retest

tlangs (approved):

PR #1016, opened 2023-03-07
ID-361 Add sentry release and breadcrumbs.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-361

Adds a release tag and breadcrumbs for sentry which will help diagnose issues.

Breadcrumbs are logs sampled from around the exception, and release tags help filter issues.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -9,7 +9,16 @@ import com.typesafe.scalalogging.LazyLogging
 import io.sentry.{Sentry, SentryOptions}
 import org.broadinstitute.dsde.workbench.dataaccess.PubSubNotificationDAO
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.{Json, Pem}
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -45,10 +45,13 @@ import scala.util.control.NonFatal
 
 object Boot extends IOApp with LazyLogging {
   val sentryDsn: Option[String] = sys.env.get("SENTRY_DSN")
+  val version: Option[String] = Option(getClass.getPackage.getImplementationVersion)
```
gpolumbo-broad, line 49: Quick googling shows this returning a value set in the manifest of the jar.  I'm assuming that for us, this value is a commit hash?  I have no idea.  Did you test how this gets set and what types of values are there?  FWIW, as part of IndependiSam, we want to cut over to using semantic versioning, in which case you may be able to just read an ENV variable to get the version string.  This is trivial enough that we can probably keep it as is, we just should remember to come back and change this when we start semantic versioning.  Perhaps drop a comment here pointing at https://broadworkbench.atlassian.net/browse/ID-485 and make a comment in the ticket to come fix this line.

gpolumbo-broad (approved):

Shakespeared (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
Ghost-in-a-Jar: I added a helpful comment thanks for pointing this out

PR #1023, opened 2023-03-16
ID-498 filter sentry issues

Ghost-in-a-Jar:
This should filter out the vast majority of the issues we def dont care about.
Ticket: https://broadworkbench.atlassian.net/browse/ID-498

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -6,19 +6,9 @@ import cats.data.NonEmptyList
 import cats.effect._
 import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
-import io.sentry.{Sentry, SentryOptions}
 import org.broadinstitute.dsde.workbench.dataaccess.PubSubNotificationDAO
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.{Json, Pem}
-import org.broadinstitute.dsde.workbench.google.{
-  GoogleDirectoryDAO,
-  GoogleKmsInterpreter,
-  GoogleKmsService,
-  HttpGoogleDirectoryDAO,
-  HttpGoogleIamDAO,
-  HttpGoogleProjectDAO,
-  HttpGooglePubSubDAO,
-  HttpGoogleStorageDAO
-}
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -44,18 +44,6 @@ import scala.jdk.CollectionConverters._
...
-  val version: Option[String] = Option(getClass.getPackage.getImplementationVersion)
-  private def initSentry(): Unit = sentryDsn.fold(logger.warn("No SENTRY_DSN found, not initializing Sentry.")) { dsn =>
-    val options = new SentryOptions()
-    options.setDsn(dsn)
-    options.setEnvironment(sys.env.getOrElse("SENTRY_ENVIRONMENT", "unknown"))
-    options.setRelease(version.getOrElse("unknown"))
-
-    Sentry.init(options)
-    logger.info("Sentry initialized")
-  }
```
Ghost-in-a-Jar, line 58: there is a call to this initSentry fn below in boot, just moving it to a companion obj

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/Sentry.scala
```diff
@@ -0,0 +1,36 @@
...
+  def initSentry(): Unit = sentryDsn.fold(logger.warn("No SENTRY_DSN found, not initializing Sentry.")) { dsn =>
+    val options = new SentryOptions()
+    options.setDsn(dsn)
+    options.setEnvironment(sys.env.getOrElse("SENTRY_ENVIRONMENT", "unknown"))
+    options.setRelease(version.getOrElse("unknown"))
+
+    options.setBeforeSend { (event: SentryEvent, hint: Hint) =>
+      event.getThrowable match {
+        case workbenchException: WorkbenchExceptionWithErrorReport if workbenchException.errorReport.statusCode.map(statusCodesToSkip.contains) => null
+        case _ => event
```
Ghost-in-a-Jar, line 29: Can someone double check my logic here

Ghost-in-a-Jar: if null is returned the event isnt sent to sentry btw

Shakespeared (approved):
lgtm

src/main/scala/org/broadinstitute/dsde/workbench/sam/util/Sentry.scala
gpolumbo-broad: Logic makes sense.  I'm less sure about the blanket `statusCodesToSkip`, but this looks like an easy enough thing to tune too.  And I have no idea what are the RIGHT things we should and should not be sending to Sentry yet, so just adding the mechanism here is good.

gpolumbo-broad (approved):

PR #1041, opened 2023-04-05
ID-451 Sam Status should differentiate critical and non-critical systems

gpolumbo-broad:
Ticket: https://broadworkbench.atlassian.net/browse/ID-451 reverting a change we made for https://broadworkbench.atlassian.net/browse/PROD-791 to override the Google Groups status check.

What:

Changes how we calculate the Status for Sam.  There is now a differentiation between "critical" and "non-critical" systems.  The Database is the only critical system and if it is reporting "not OK", then Sam should report "not OK".  If, however, any of the other non-critical Google Services are not reporting "OK", then Sam will still be "OK", but should be considered "degraded".  Sentry should receive an alert if Sam's status is degraded or worse.  

Why:

 We want Sam to be more resilient to Google outages, but we also need to know when it is having problems.

How:

Followed inspiration from Orchestration and Rawls and how they can report "OK" even when some subsystems are having problems.  Rewrote the tests to be more thorough/complete

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -370,7 +370,7 @@ object Boot extends IOApp with LazyLogging {
     val tosService = new TosService(directoryDAO, config.termsOfServiceConfig)
     val userService = new UserService(directoryDAO, cloudExtensionsInitializer.cloudExtensions, config.blockedEmailDomains, tosService)
     val statusService =
-      new StatusService(directoryDAO, cloudExtensionsInitializer.cloudExtensions, DbReference(config.samDatabaseConfig.samRead.dbName, implicitly), 10 seconds)
+      new StatusService(directoryDAO, cloudExtensionsInitializer.cloudExtensions, 10 seconds, 1 minute)
```
Ghost-in-a-Jar, line 367: Do we maybe want to make this something configurable via env vars? just a thought

also i love 'implicitly' as one of the former argument names 😆

Ghost-in-a-Jar (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
gpolumbo-broad: Not a bad idea, but I'm not going to bother for now.  Maybe this will bight us later, but I also don't see it needing to change regularly or vary between environments or anything like that, so I think having it static for now is OK

gpolumbo-broad:
jenkins retest this again for me buddy

tlangs (approved):

gpolumbo-broad:
jenkins retest

gpolumbo-broad:
jenkins retest please

gpolumbo-broad:
jenkins retest this and hopefully we can get by the maven error

gpolumbo-broad:
jenkins retest

gpolumbo-broad:
jenkins retest

gpolumbo-broad:
1 Integration Test flaked out in the swatomation run on Jenkins.  The tests are so flakey lately, I'm taking this as a win and merging.  This PR is all about Sam's status endpoint and all the other tests in the swat run were fine so I think we are OK.

PR #1042, opened 2023-04-05
ID-181 and ID-228 Database resiliency to CloudSQL maintenance

gpolumbo-broad:
Ticket: 
https://broadworkbench.atlassian.net/browse/ID-181
https://broadworkbench.atlassian.net/browse/ID-228

What:

This should merge into https://github.com/broadinstitute/sam/pull/1041

Adds a simple read query to the `DirectoryDAO.checkStatus` method in an attempt to make sure our connections are actually functional.  

Why:
Sometimes during CloudSQL maintenance, our db connections would get all messed up, but not to the point where Sam would be able to tell and it would restart itself.  Instead, it would do things like sit there with 1 connection in the pool that was _just_ functional enough to report as "valid" even though it was unusable.  For some reason, Sam wouldn't try to refresh the connection pool either.  

How:
This change should ensure that the Sam Status endpoint reports "NOT OK" if it is unable to run a simple query against the DB.  In this case, our k8s infrastructure should notice that Sam is unhealthy and attempt to restart it, which is precisely what an on-call engineer would be asked to do in the past when we got paged because Sam was not able to query the DB.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/DirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/DirectoryDAO.scala
@@ -12,7 +12,7 @@ import java.util.Date
 /** Created by dvoet on 5/26/17.
   */
 trait DirectoryDAO {
-  def checkStatus(samRequestContext: SamRequestContext): Boolean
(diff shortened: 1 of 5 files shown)


tlangs (approved):
Heck yeah! 🎉

Ghost-in-a-Jar (approved):

PR #1052, opened 2023-04-10
ID-549 Artifactory Proxy Resolver

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-549

What:

Make Sam use Artifactory as a proxy resolver so that we cache packages from Maven just in case they disappear.

Why:

The `sbt-assembly` plugin temporarily disappeared from Maven last week, causing the Sam Build to fail. If we were caching it, the build would not have failed.

How:

Make Sam look at our internal artifactory before anything else.

---

**PR checklist**

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -10,6 +10,10 @@ import sbtassembly.AssemblyPlugin.autoImport._
 object Settings {
   lazy val artifactory = "https://artifactory.broadinstitute.org/artifactory/"
 
+  val proxyResolvers = List(
+    "internal-maven-proxy" at artifactory + "maven-central"
+  )
+
   lazy val commonResolvers = List(
     "artifactory-releases" at artifactory + "libs-release",
     "artifactory-snapshots" at artifactory + "libs-snapshot"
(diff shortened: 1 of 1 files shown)


tlangs:
jenkins retest

Ghost-in-a-Jar (approved):

tlangs:
jenkins retest

project/Settings.scala
```diff
@@ -65,7 +69,7 @@ object Settings {
     commonBuildSettings ++ commonAssemblySettings ++ commonTestSettings ++ List(
       organization := "org.broadinstitute.dsde.workbench",
       scalaVersion := "2.13.10",
-      resolvers ++= commonResolvers,
+      resolvers := proxyResolvers ++: resolvers.value ++: commonResolvers,
```
gpolumbo-broad, line 72: Not sure where resolvers is being defined, and I'm not sure what the `:=` operator does.  Looking at the scaldoc I couldn't find one on List.  So as long as the operators are right here

gpolumbo-broad (approved):

project/Settings.scala
Ghost-in-a-Jar: it is an sbt specific dsl thing, looks right

Ghost-in-a-Jar (approved):

project/Settings.scala
tlangs: Its what's in Rawls, so I'm assuming its correct 🤷‍♂️

PR #1055, opened 2023-04-13
ID-343 fixing column defaults

gpolumbo-broad:
Ticket: https://broadworkbench.atlassian.net/browse/ID-343

What:

When we added the liquibase migrations before, we forgot to accommodate for existing records.  This change creates new columns with `default` values specified for all the records in the database.  Then it immediately drops the `default` value attribute on those columns so that if users try to insert a new record without specifying a value for these columns they will get an error, which is the behavior we want. 

Why:

Sometimes, in databases, they have data.  

How:

See the "why" - I'm not retyping it, but you can re-read it :) 

---

**PR checklist**

### DIFF
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20230328_add_date_columns_to_user.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20230328_add_date_columns_to_user.xml
@@ -6,19 +6,32 @@
   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
 
   <changeSet logicalFilePath="dummy" author="kfeldman" id="add_date_columns_to_user_table">
+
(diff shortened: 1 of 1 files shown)


Ghost-in-a-Jar (approved):

tlangs (commented):
With the default values, do we still need to be setting new records to epoch in the sam app code?

gpolumbo-broad:
@tlangs - no, the idea is that for all existing records in the db when this goes out, their `created_at` and `updated_at` dates will be the epoch time because we don't know the correct dates for these users.  Going forward though, I like the idea of these values enforcing non-nullness so that a record is never created without these values being set.

tlangs:
But, the default values will also be there for new records that are created without values. So, Sam doesn't need to do it itself. We just need to create new records without specifying values for the columns that the DB provides default values for.

gpolumbo-broad:
No, see the `dropDefaultValue` directives.  I tested it out locally and checked on the database with and without that directive to make sure it was correctly removing the default value attribute set on the column

gpolumbo-broad:
No, see the `dropDefaultValue` directives.  I tested it out locally and checked on the database with and without that directive to make sure it was correctly removing the default value attribute set on the column

tlangs (approved):

gpolumbo-broad:
swatomation tests on Jenkins have been extremely flaky lately.  Only one test failed in this run, it is unrelated and I'm making the call that this Sam fix is safe to merge.

PR #1062, opened 2023-04-21
QA-2062: Investigate Sam user state management to engooden pact verification tests

ichengchang:
Ticket: [QA-2062](https://broadworkbench.atlassian.net/browse/QA-2062)

**Investigate Sam provider state management. Engooden pact verification tests**

What:

There were numerous discussions (within `dsp-contract-testing`) on making the existing Sam provider tests more robust for consumers such as

- BPM
- Leo

Here's a summary of those discussions.

1. There are [several ways of handling auth](https://docs.pact.io/provider/handling_auth). The approach we liked best was to stub out any auth requirements in the provider, with the rationale that those pieces are best tested elsewhere. 

2. For a lot of states, it's possible to set up things you need directly in the provider. For instance, if you're testing a response to a query for user info, you can first make a call to create the user. This simplifies the maintenance on the provider side, and seems more meaningful (and more clear) than pre-baked state.

3. Agreed we are not gonna be setting up a Sam user directly through a login credential. 

The auth approach could be fulfilled with a request filter (Sam side) to intercept an access token and then handle it with some stub out auth logic. 

Depending on auth outcome, make a call to create Sam user from state params associated with a given state, e.g. `user exists` and query for user status info (`user exists`).

Instead of creating a specific user in the cloud db, a simpler approach will be to mock and stub the directory dao and return generic user status info that can be verified with matchers.  This PR illustrates some of these practices.

**Why?**

Making Sam provider verifier tests more robust to support more general consumer tests.

Use mocks and stubs to isolate services that we do not have contract with.

**How?**

Bullet point #1 is addressed by a request filter in Sam provider. The request filter detects a http header `Authorization: Bearer accessToken`. The token needs to be in a valid format.

Bullet point #2. Here we use a stateful mock directory dao and some sophisticated stubbing to mock out the behavior of certain state such as `user exists`.

---

**PR checklist**

[QA-2062]: https://broadworkbench.atlassian.net/browse/QA-2062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
+++ b/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
@@ -57,8 +57,8 @@ class SamProviderSpec
     val policyEvaluatorService = TestPolicyEvaluatorServiceBuilder(directoryDAO, policyDAOOpt = Some(policyDAO)).build
 
     val googleExt = mock[GoogleExtensions]
-    // val mockResourceService = mock[ResourceService]
-    val resourceService: ResourceService = TestResourceServiceBuilder(policyEvaluatorService, policyDAO, directoryDAO, cloudExtensions).withWorkspaceResourceType().build
+    val resourceService: ResourceService =
+      TestResourceServiceBuilder(policyEvaluatorService, policyDAO, directoryDAO, cloudExtensions).withWorkspaceResourceType().build
     val mockManagedGroupService = mock[ManagedGroupService]
     val tosService = MockTosServiceBuilder().withAllAccepted().build
     val azureService = mock[AzureService]
@@ -78,11 +78,6 @@ class SamProviderSpec
         )
       )
     }
-
-    //val fakeWorkspaceResourceType = ResourceType(SamResourceTypes.workspaceName, Set.empty, Set.empty, ResourceRoleName("workspace"))
-    //when {
-    //  mockResourceService.getResourceType(any[ResourceTypeName])
-    //} thenReturn IO.pure(Option(fakeWorkspaceResourceType))
 
     MockSamDependencies(
       resourceService,
--- a/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
+++ b/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
@@ -42,7 +42,7 @@ class SamProviderSpec
   var activeSamUserSubjectId: Option[String] = None
   var activeSamUserEmail: Option[String] = None
   val allUsersGroup: BasicWorkbenchGroup = BasicWorkbenchGroup(CloudExtensions.allUsersGroupName, Set(), WorkbenchEmail("all_users@fake.com"))
-  val defaultSamUser: SamUser = Generator.genWorkbenchUserBoth.sample.get
+  val defaultSamUser: SamUser = Generator.genWorkbenchUserBoth.sample.get.copy(enabled=true)
   val newSamUser: SamUser = Generator.genWorkbenchUserBoth.sample.get
 
   def genSamDependencies: MockSamDependencies = {
--- a/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
+++ b/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
@@ -5,7 +5,7 @@ import akka.http.scaladsl.testkit.ScalatestRouteTest
 import cats.effect.IO
 import cats.effect.unsafe.implicits.global
 import com.typesafe.scalalogging.LazyLogging
-import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchGroup, WorkbenchGroupIdentity, WorkbenchUserId}
+import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchGroup, WorkbenchGroupIdentity, WorkbenchGroupName, WorkbenchUserId}
 import org.broadinstitute.dsde.workbench.oauth2.mock.FakeOpenIDConnectConfiguration
 import org.broadinstitute.dsde.workbench.sam.Generator.genNonPetEmail
 import org.broadinstitute.dsde.workbench.sam.MockTestSupport.genSamRoutes
@@ -44,7 +44,7 @@ class SamProviderSpec
     with LazyLogging
     with MockitoSugar {
 
-  val defaultSamUser: SamUser = Generator.genWorkbenchUserBoth.sample.get
+  val defaultSamUser: SamUser = Generator.genWorkbenchUserBoth.sample.get.copy(enabled=true)
   val allUsersGroup: BasicWorkbenchGroup = BasicWorkbenchGroup(CloudExtensions.allUsersGroupName, Set(defaultSamUser.id), WorkbenchEmail("all_users@fake.com"))
   val defaultResourceTypeActionPatterns = Set(
     SamResourceActionPatterns.alterPolicies,
@@ -100,6 +100,7 @@ class SamProviderSpec
   def genSamDependencies: MockSamDependencies = {
     val userService: UserService = TestUserServiceBuilder()
       .withAllUsersGroup(allUsersGroup)
+      .withWorkbenchGroup(accessPolicy)
       .withEnabledUser(defaultSamUser)
       .withAllUsersHavingAcceptedTos()
       .build
@@ -115,10 +116,14 @@ class SamProviderSpec
     ).thenAnswer((i: InvocationOnMock) => {
       val resourceTypeName = i.getArgument[ResourceTypeName](0)
       val workbenchUserId = i.getArgument[WorkbenchUserId](1)
+      val samRequestContext = i.getArgument[SamRequestContext](2)
       IO {
-        val forEachPolicy = policies.collect {
+        val group = directoryDAO.loadGroup(WorkbenchGroupName(accessPolicy.id.toString), samRequestContext).unsafeRunSync().get
+        print(group)
+        val policies2: Map[WorkbenchGroupIdentity, WorkbenchGroup] = Map(group.id -> group)
+        val forEachPolicy = policies2.collect {
           case (fqPolicyId@FullyQualifiedPolicyId(FullyQualifiedResourceId(`resourceTypeName`, _), _), accessPolicy: AccessPolicy)
-            if accessPolicy.members.contains(workbenchUserId) || accessPolicy.public =>
+                if accessPolicy.members.contains(workbenchUserId) || accessPolicy.public =>
             if (accessPolicy.public) {
               ResourceIdWithRolesAndActions(fqPolicyId.resource.resourceId, RolesAndActions.empty, RolesAndActions.empty, RolesAndActions.fromPolicy(accessPolicy))
             } else {
--- a/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
+++ b/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
@@ -18,7 +18,6 @@ import org.broadinstitute.dsde.workbench.sam.{Generator, MockSamDependencies, Mo
 import org.broadinstitute.dsde.workbench.util.health.{StatusCheckResponse, SubsystemStatus, Subsystems}
 import org.http4s.headers.Authorization
 import org.http4s.{AuthScheme, Credentials}
-import org.mockito.DefaultAnswers
 import org.mockito.invocation.InvocationOnMock
 import org.mockito.scalatest.MockitoSugar
 import org.scalatest.BeforeAndAfterAll
@@ -47,7 +46,17 @@ class SamProviderSpec extends AnyFlatSpec with ScalatestRouteTest with MockTestS
     val mockManagedGroupService = mock[ManagedGroupService]
     val tosService = mock[TosService]
     val azureService = mock[AzureService]
-    val userService: UserService = mock[UserService](DefaultAnswers.ReturnsDeepStubs)
+    val userService: UserService = new UserService(directoryDAO, cloudExtensions, Seq(), defaultTosService)
+    //when {
+    //  userService.directoryDAO
+    //} thenReturn {
+    //  directoryDAO
+    //}
+    //when {
+    //  userService.cloudExtensions
+    //} thenReturn {
+    //  cloudExtensions
+    //}
     when(
       userService.getUserStatusInfo(any[SamUser], any[SamRequestContext])
     ).thenAnswer((i: InvocationOnMock) =>  {
--- a/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
+++ b/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
@@ -63,7 +63,6 @@ class SamProviderSpec
     // Resource service and states for consumer verification
     val resourceService: ResourceService =
       TestResourceServiceBuilder(policyEvaluatorService, policyDAO, directoryDAO, cloudExtensions)
-        .withDummySamUser(defaultSamUser)
         .withRandomDefaultResourceType()
         .withRandomOtherResourceType()
         .withRandomWorkspaceResourceType()
@@ -162,7 +161,7 @@ class SamProviderSpec
       case Some((_, value)) =>
         parseAuth(value)
       case None =>
-        println("no auth header found")
+        logger.debug("no auth header found")
         NoOpFilter
     }
 
@@ -171,16 +170,21 @@ class SamProviderSpec
       .parse(auth)
       .map {
         case Authorization(Credentials.Token(AuthScheme.Bearer, token)) =>
-          println(s"Captured token $token")
+          var proxyToken = token
+          // intended to match a certain class of user token
           token match {
             case "accessToken" =>
-              println("do bearer 'accessToken'")
+              logger.debug("do bearer 'accessToken'")
+              // e.g. proxy token that impersonates a regular user
+              proxyToken = "user" + token
             case _ =>
-              println("do other bearer token")
+              logger.debug("do other bearer token")
+              // e.g. proxy token that impersonates a super user
+              proxyToken = "su" + token
           }
-          SetHeaders("Authorization" -> s"Bearer $token")
+          SetHeaders("Authorization" -> s"Bearer $proxyToken")
         case _ =>
-          println("do other AuthScheme")
+          logger.debug("do other AuthScheme")
           NoOpFilter
       }
       .getOrElse(NoOpFilter)
@@ -204,9 +208,9 @@ class SamProviderSpec
     .withStateManagementFunction(
       StateManagementFunction {
         case ProviderState("user exists", params) =>
-          println("user exists")
+          logger.debug("user exists")
         case _ =>
-          println("other state")
+          logger.debug("other state")
       }
         .withBeforeEach(() => ())
     )
--- a/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
+++ b/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
@@ -58,7 +58,7 @@ class SamProviderSpec
     val policyDAO = StatefulMockAccessPolicyDaoBuilder()
       .withRandomAccessPolicy(SamResourceTypes.workspaceName, Set(defaultSamUser.id))
       .build
-    val policyEvaluatorService = TestPolicyEvaluatorServiceBuilder(directoryDAO, policyDAOOpt = Some(policyDAO)).build
+    val policyEvaluatorService = TestPolicyEvaluatorServiceBuilder(directoryDAO, policyDAO).build
 
     // Resource service and states for consumer verification
     val resourceService: ResourceService =
--- a/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
+++ b/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
@@ -39,10 +39,12 @@ class SamProviderSpec
     with LazyLogging
     with MockitoSugar {
 
+  // The default login user
   val defaultSamUser: SamUser = Generator.genWorkbenchUserBoth.sample.get.copy(enabled = true)
   val allUsersGroup: BasicWorkbenchGroup = BasicWorkbenchGroup(CloudExtensions.allUsersGroupName, Set(defaultSamUser.id), WorkbenchEmail("all_users@fake.com"))
 
   def genSamDependencies: MockSamDependencies = {
+    // Minimally viable Sam service and states for consumer verification
     val userService: UserService = TestUserServiceBuilder()
       .withAllUsersGroup(allUsersGroup)
       .withEnabledUser(defaultSamUser)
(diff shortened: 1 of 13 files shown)


ichengchang:
jenkins retest

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
```diff
@@ -138,19 +137,62 @@ class SamProviderSpec extends AnyFlatSpec with ScalatestRouteTest with MockTestS
...
     case Some(s) if !s.isBlank() => consumerVersionSelectors = consumerVersionSelectors.branch(s, consumerName)
-    case _ => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased
+    case _ => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.mainBranch
   }
 
-  override def provider: ProviderInfoBuilder = ProviderInfoBuilder(
+  // If the auth header in the request is "correct", we can replace it with an auth header that will actually work with our API,
+  // else we leave it as is to be rejected. This request filter is currently not used as the Sam provider
+  // session defaults to defaultSamUser
+  def requestFilter: ProviderRequest => ProviderRequestFilter = req =>
```
gpolumbo-broad: Please refactor this method and extract variables and or smaller functions.  Nest `match` statements, just like nested `if`s or `for` loops increase the cyclomatic complexity and make this harder to understand.  In some cases one nested `match` may make sense, be more readable, or be less confusing.  But I recommend starting off by trying to extract methods until there are no nested `match`es.

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
```diff
@@ -138,19 +137,62 @@ class SamProviderSpec extends AnyFlatSpec with ScalatestRouteTest with MockTestS
...
   consumerBranch match {
     case Some(s) if !s.isBlank() => consumerVersionSelectors = consumerVersionSelectors.branch(s, consumerName)
-    case _ => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased
+    case _ => consumerVersionSelectors = consumerVersionSelectors.deployedOrReleased.mainBranch
   }
 
-  override def provider: ProviderInfoBuilder = ProviderInfoBuilder(
+  // If the auth header in the request is "correct", we can replace it with an auth header that will actually work with our API,
+  // else we leave it as is to be rejected. This request filter is currently not used as the Sam provider
+  // session defaults to defaultSamUser
```
gpolumbo-broad: I'm not totally clear on what this comment is trying to say.  If the filter isn't used, then why is it here?  Will it be used at some point?  Under what conditions?

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/StatefulMockAccessPolicyDaoBuilder.scala
```diff
@@ -0,0 +1,102 @@
...
+      .when(mockedAccessPolicyDAO)
+      .loadPolicy(ArgumentMatchers.eq(policy.id), any[SamRequestContext])
+
+    policy.members.foreach { m =>
+      lenient()
+        .doAnswer { (i: InvocationOnMock) =>
+          val resourceTypeName = i.getArgument[ResourceTypeName](0)
+          val workbenchUserId = i.getArgument[WorkbenchUserId](1)
+          println("policies.size: " + policies.size)
+          IO {
```
gpolumbo-broad: Cyclomatic complexity again.  This is pretty highly nested and a little difficult to follow through the various paths.  Please extract variables and methods.

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/StatefulMockAccessPolicyDaoBuilder.scala
```diff
@@ -0,0 +1,102 @@
...
+        .when(mockedAccessPolicyDAO)
+        .listUserResourcesWithRolesAndActions(
+          ArgumentMatchers.eq(policy.id.resource.resourceTypeName),
+          ArgumentMatchers.eq(WorkbenchUserId(m.toString)),
+          any[SamRequestContext]
+        )
+    }
+  }
+
+  def withAccessPolicy(resourceTypeName: ResourceTypeName, members: Set[WorkbenchSubject]): StatefulMockAccessPolicyDaoBuilder = {
```
gpolumbo-broad: nit: rename this method to something like, `withRandomAccessPolicy`.  Named like this, I would expect this method to take an `AccessPolicy` as an argument.  Instead it is doing some random magic here, so I think naming it as such is a good idea.

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/StatefulMockAccessPolicyDaoBuilder.scala
```diff
@@ -0,0 +1,102 @@
...
+import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
+import org.mockito.ArgumentMatchers
+import org.mockito.Mockito.{RETURNS_SMART_NULLS, lenient}
+import org.mockito.invocation.InvocationOnMock
+import org.mockito.scalatest.MockitoSugar
+
+import scala.collection.concurrent.TrieMap
+import scala.collection.mutable
+
+case class StatefulMockAccessPolicyDaoBuilder() extends MockitoSugar {
```
gpolumbo-broad, line 19: Just looking at the PR summary, seeing this builder made my spidey sense tingle.  We should try to do everything we can to _not_ have to keep track of state in mocks.  I know I did the same thing with some of my earlier test refactorings as a result of how the production code was designed an written.  So you may be in the same bind here, but I encourage you to look hard at this code to see if there is any way you can achieve the same or a similar goal without the complex state management.  If needing to refactor some of the production code is needed to make this cleaner, that is perfectly acceptable.  Just call it out as to why you had to do it.  Of course, this might cascade into a much larger refactor, in which case maybe we just stick with the stateful mock for now.  If we end up keeping this thing around, I request that you please add a comment on this class explaining why it needs to be stateful and what exactly it is about the design of the production code that is forcing us to have to track state in here.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestPolicyEvaluatorServiceBuilder.scala
```diff
@@ -0,0 +1,62 @@
...
+import org.broadinstitute.dsde.workbench.sam.dataAccess.{AccessPolicyDAO, DirectoryDAO, StatefulMockAccessPolicyDaoBuilder}
+import org.broadinstitute.dsde.workbench.sam.model._
+import org.mockito.scalatest.MockitoSugar
+
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext
+
+case class TestPolicyEvaluatorServiceBuilder(
+    directoryDAO: DirectoryDAO,
+    policyDAOOpt: Option[AccessPolicyDAO] = None,
```
gpolumbo-broad:
```suggestion
    maybePolicyDao: Option[AccessPolicyDAO] = None,
```
there is a bit of a convention for `Option`s to be called `maybe…`

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
```diff
@@ -0,0 +1,58 @@
...
+    Set(ResourceAction("alter_policies"), ResourceAction("delete"), ResourceAction("read_policies"), ResourceAction("view"), ResourceAction("non_owner_action"))
+  private val defaultResourceTypeActionPatterns = Set(
+    SamResourceActionPatterns.alterPolicies,
+    SamResourceActionPatterns.delete,
+    SamResourceActionPatterns.readPolicies,
+    ResourceActionPattern("view", "", false),
+    ResourceActionPattern("non_owner_action", "", false)
+  )
+
+  def withWorkspaceResourceType(): TestResourceServiceBuilder = {
```
gpolumbo-broad: Again, I'd rather see this called `withRandomWorkspaceResourceType` or something like that.  Furthermore, I might suggest that this class not be responsible for constructing the `ResourceType` at all.  I think that this method would be more useful in the future if `withWorkspaceResourceType` just took a `ResourceType` argument rather than constructing it in this class.

gpolumbo-broad (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/StatefulMockAccessPolicyDaoBuilder.scala
ichengchang: I refactored the mock and removed external states. This turns out to be sufficient for stubbing the behavior we need for `listUserResourcesWithRolesAndActions`.  I will leave it as this until we have another concrete contract test case that really requires us to think of a refactoring of production code.

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
ichengchang: Refactored code to reduce complexity.

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
ichengchang: In the case when consumer interaction includes an auth header, the filter is where the header will be intercepted for custom processing and/or validation.

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/StatefulMockAccessPolicyDaoBuilder.scala
ichengchang: Reduced complexity by using `AnyOfMatcher` and wrapping some code in `constructResourceIdWithRolesAndActions` function.

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/StatefulMockAccessPolicyDaoBuilder.scala
ichengchang: Renamed to `withRandomAccessPolicy`. If we need a new method `withAccessPolicy(policy: AcessPolicy)` in the future, it can be included here as well.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestPolicyEvaluatorServiceBuilder.scala
ichengchang: Removed the `Option` type, not needed anyway.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
ichengchang: Renamed to `withRandomWorkspaceResourceType`. If we need a new method `withWorkspaceResourceType` in the future, it can be added here as well.

ichengchang:
jenkins retest

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
```diff
@@ -140,34 +145,46 @@ class SamProviderSpec
...
         println("no auth header found")
         NoOpFilter
     }
 
+  private def parseAuth(auth: String): ProviderRequestFilter =
+    Authorization
+      .parse(auth)
+      .map {
+        case Authorization(Credentials.Token(AuthScheme.Bearer, token)) =>
+          println(s"Captured token ${token}")
```
gpolumbo-broad: Probably want to use logger instead of `println`

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
```diff
@@ -140,34 +145,46 @@ class SamProviderSpec
...
+          token match {
```
gpolumbo-broad, line 176: Do we need this `match` statement?  In other words, what is the value-add of printing/logging which format of bearer token we got?  I'm thinking it doesn't matter for the tests that might be using this logic.  I'm sure it was helpful when developing, but for most tests, I don't think they will care which type of token it was, they just want to know if the person is authenticated or not.  If they do care what kind of token it was, well, that sounds like it should be the SUT, not something that gets printed here as a side-effect. 

Also gets rid of a nested conditional :) which I love!

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestPolicyEvaluatorServiceBuilder.scala
```diff
@@ -43,19 +41,13 @@ case class TestPolicyEvaluatorServiceBuilder(
...
-    val policyDAO = policyDAOOpt match {
-      case Some(dao) => dao
-      case None => buildAccessPolicyDao()
-    }
-
+  def build: PolicyEvaluatorService =
     new PolicyEvaluatorService(emailDomain, Map(workspaceResourceType.name -> workspaceResourceType), policyDAO, directoryDAO)
-  }
 
   private def buildAccessPolicyDao(): AccessPolicyDAO = {
```
gpolumbo-broad: According to IntelliJ and my command+f in the Sam code, this method is not used, so I think it should be deleted.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
```diff
@@ -7,30 +7,64 @@ import org.broadinstitute.dsde.workbench.sam.dataAccess.{AccessPolicyDAO, Direct
...
+  private[service] val defaultResourceTypeActionPatterns = Set(
     SamResourceActionPatterns.alterPolicies,
     SamResourceActionPatterns.delete,
     SamResourceActionPatterns.readPolicies,
     ResourceActionPattern("view", "", false),
     ResourceActionPattern("non_owner_action", "", false)
   )
 
-  def withWorkspaceResourceType(): TestResourceServiceBuilder = {
+  def withRandomDefaultResourceType(): TestResourceServiceBuilder = {
```
gpolumbo-broad: Ah, now that I think about these more, I think the better way to do this would be to add a `genResourceType` method in https://github.com/broadinstitute/sam/blob/develop/src/test/scala/Generator.scala.  Then you don't need to have `withRandomDefaultResourceType`, you can just have `withResourceType` and pass it a randomly generated `ResourceType` from `genResourceType`.  Hope that makes sense.  Hopefully that reduces the amount of setup you need to do above in this class too.

gpolumbo-broad: Additionally, if we do it like this, I'm not sure why we need `withRandomOtherResourceType` or `withRandomWorkspaceResourceType`.  If we make the above suggested change, then we can have just 1 `withResourceType` method, and all the "default" or "workspace" or "random" type definition stuff can be handled in the test or in the `Generator`.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
```diff
@@ -45,14 +79,32 @@ case class TestResourceServiceBuilder(
     this
   }
 
+  def withDummySamUser(samUser: SamUser): TestResourceServiceBuilder = {
```
gpolumbo-broad: I see this gets used in 1 test, but `maybeDummySamUser` isn't used anywhere.  Can we delete it?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
```diff
@@ -45,14 +79,32 @@ case class TestResourceServiceBuilder(
...
+      case Some(rt) =>
+        resourceTypes += rt.name -> rt
+      case _ => ()
+    }
+
+    maybeOtherResourceType match {
+      case Some(rt) =>
+        resourceTypes += rt.name -> rt
+      case _ => ()
+    }
```
gpolumbo-broad: Again, if we make the change to just have `withResourceType`, this method becomes much simpler.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
```diff
@@ -7,30 +7,64 @@ import org.broadinstitute.dsde.workbench.sam.dataAccess.{AccessPolicyDAO, Direct
...
 
+import java.util.UUID
 import scala.collection.concurrent.TrieMap
 import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 
 case class TestResourceServiceBuilder(
     policyEvaluatorService: PolicyEvaluatorService,
-    accessPolicyDAO: AccessPolicyDAO,
+    policyDAO: AccessPolicyDAO,
```
gpolumbo-broad: fwiw, I prefer `accessPolicyDAO`

gpolumbo-broad (changes requested):
Ivan, please don't take the volume and type of comments on this PR as criticism that you did anything wrong or incorrectly.  On the the contrary, I'm stoked that you are using the new Builder pattern for the tests!  This is awesome.  It's precisely because this stuff is new and you're the first external contributor to add these that I'm being a little more opinionated about how we are doing things so that we try to establish the right patterns for setting this stuff up.  So thank you!  And I appreciated you working with me and iterating on this stuff.

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
ichengchang: Yup, should have replaced `println` by `logger`. All set now.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestPolicyEvaluatorServiceBuilder.scala
ichengchang: This method has been removed.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
ichengchang: Added a couple of new generators to simplify `TestResourceServiceBuilder` class.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
ichengchang: Refactored code to use a single `withResourceTypes` method to set up initial resource types.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
ichengchang: Renamed `policyDAO` to `accessPolicyDAO`.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
ichengchang: Removed `maybeDummyUser`.

pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
ichengchang: I added a little bit of code and comments to convey the intention of the `match` statement, hopefully it makes more sense now.

ichengchang:
scala tests retest

ichengchang:
jenkins retest

ichengchang:
jenkins retest

src/test/scala/org/broadinstitute/dsde/workbench/sam/matchers/AnyOfMatcher.scala
```diff
@@ -3,5 +3,25 @@ package org.broadinstitute.dsde.workbench.sam.matchers
...
+  /** Use in conjunction with argThat.
+    *
+    * Example usage:
+    * ```
+    * import org.mockito.Mockito._
+    * import org.mockito.ArgumentMatchers._
+    *
+    * val mockObj = mock(classOf[MyClass])
+    *
+    * when(mockObj.myMethod(argThat(new AnyOfMatcher(Seq("foo", "bar", "baz")))))
```
gpolumbo-broad: I _think_ if we add an object to the bottom of this file like:

```
object MatchesOneOf {
  def apply[T](possibleValues: Iterable[T]): AnyOfMatcher[T] = new AnyOfMatcher[T](possibleValues)
}
```

Then we can have tests that read slightly clearer like:  

```
when(mockObj.myMethod(argThat(MatchesOneOf(Seq("foo", "bar", "baz")))))
    .thenReturn("matched") mockObj.myMethod("bar")
verify(mockObj).myMethod(argThat(MatchesOneOf(Seq("foo", "bar", "baz"))))
```
It's just a little extra piece to maintain but I think it makes it read ever so slightly easier which should make the tests easier to read and maintain.  

There might be some other ways we could fiddle with this too in order to make it a little easier to read, but you'd have to fiddle around with it to see what you can make work.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
```diff
@@ -2,109 +2,31 @@ package org.broadinstitute.dsde.workbench.sam.service
...
-        Set(
-          ResourceRole(ResourceRoleName("owner"), defaultResourceTypeActions - ResourceAction("non_owner_action")),
-          ResourceRole(ResourceRoleName("other"), Set(ResourceAction("view"), ResourceAction("non_owner_action")))
-        ),
-        ResourceRoleName("owner")
-      )
-    )
+  def withResourceTypes(): TestResourceServiceBuilder = {
+    Set(genResourceType.sample.get, genWorkspaceResourceType.sample.get)
+      .foreach(rt => resourceTypes += rt.name -> rt)
```
gpolumbo-broad: I would prefer to see this construction of an arbitrary `Set` as part of the test.  Something like, 

```
"myTest" should "do the right thing" in {
    val resourceTypes = Set(genResourceType.sample.get, genWorkspaceResourceType.sample.get)
    val testResourceServiceBuilder = new TestResourceServiceBuilder()
        .withResourceTypes(resourceTypes)
        .build()
    … 
}
```
In this way, the builders are not responsible for creating any of the concrete objects that we care about in the test.  All of those concrete objects that the tests rely on are declared explicitly in the tests, making them easier to read and maintain.  

I guess you could put this another way and say it is a code smell if `org.broadinstitute.dsde.workbench.sam.Generator` is in your list of imports in one of these `Builder` classes.

gpolumbo-broad (changes requested):
This really great Ivan, a lot cleaner, just a couple more suggestions :) As we clean up the bigger parts, some of the more subtle things that we couldn't see before become apparent.  I think we're close though!

src/test/scala/org/broadinstitute/dsde/workbench/sam/matchers/AnyOfMatcher.scala
ichengchang: Renamed `AnyOfMatcher` class to `MatchesOneOf`. Added companion object `MatchesOneOf`.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/TestResourceServiceBuilder.scala
ichengchang: Refactored `withResourceTypes` to accept a collection of resource types as input.

ichengchang:
jenkins retest

gpolumbo-broad (approved):
👍 Thank you for being patient and iterating on this with me, I think this is really great.

ichengchang:
@gpolumbo-broad Thank you for all the feedbacks. I agreed with you that it looks much cleaner and hopefully this sets the stage for a pattern that is easy for developers to follow when writing contract tests in general.

ichengchang:
@Ghost-in-a-Jar Could you please give one more thumb for this PR to merge?

Ghost-in-a-Jar (approved):

PR #1065, opened 2023-04-27
ID-586 Sam signs GS URLs

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-586

What:

Sam can now return signed URLs for GS Objects! 

Why:

Martha is getting retired, and Sam already handles service account keys. It makes sense for Sam to do this, and to not pass keys over the network to other systems.

How:

The first step was to modify [workbench-libs](https://github.com/broadinstitute/workbench-libs/pull/1336) to create signed URLs for gs blobs. Then, I wired that into Sam and wrote tests.

---

**PR checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ sh docker/run-proxy.sh start
 ```shell
 sbt run
 ```
-Navigate to [Sam's (Proxy's) Swagger page](http://localhost:50443/)
+Navigate to [Sam's (Proxy's) Swagger page](https://local.broadinstitute.org:50443/#/)
 
 
 #### Debugging in IntelliJ
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1096,6 +1096,39 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
       x-codegen-request-body-name: scopes
+  /api/google/v1/user/petServiceAccount/{project}/signedUrlForBlob:
+    post:
+      tags:
+        - Google
+      summary: gets a signed URL for the given blob, signed by the Pet Service account of the calling user
+      operationId: getSignedUrlForBlob
+      parameters:
+        - name: project
+          in: path
+          description: Google project of the pet
+          required: true
+          schema:
+            type: string
+      requestBody:
+        description: bucketName and blobName of the object to get a signed URL for
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/SignedUrlRequest'
+        required: true
+      responses:
+        200:
+          description: signed URL for the blob, signed by the Pet Service Account key
+          content:
+            application/json:
+              schema:
+                type: string
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
   /api/google/v1/user/proxyGroup/{email}:
     get:
       tags:
@@ -2954,6 +2987,18 @@ components:
         - https://www.googleapis.com/auth/userinfo.profile
       items:
         type: string
+    SignedUrlRequest:
+      type: object
+      required:
+        - bucketName
+        - blobName
+      properties:
+        bucketName:
+          type: string
+          description: bucket of the blob
+        blobName:
+          type: string
+          description: path to the blob in the bucket
     CreateResourceRequest:
       required:
         - policies
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1100,7 +1100,9 @@ paths:
     post:
       tags:
         - Google
-      summary: gets a signed URL for the given blob, signed by the Pet Service account of the calling user
+      summary: Gets a signed URL for the given blob, signed by the Pet Service account of the calling user. 
+        The signed URL is active for 1 hour and scoped to the permissions of the signing Pet Service Account.
+        Sam will provide a signed URL for any object path, even if that object does not exist.
       operationId: getSignedUrlForBlob
       parameters:
         - name: project
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -14,7 +14,9 @@ import com.google.rpc.Code
 import com.typesafe.scalalogging.LazyLogging
 import net.logstash.logback.argument.StructuredArguments
 import org.broadinstitute.dsde.workbench.dataaccess.NotificationDAO
+import org.broadinstitute.dsde.workbench.google.GooglePubSubDAO.MessageRequest
 import org.broadinstitute.dsde.workbench.google.{GoogleDirectoryDAO, GoogleIamDAO, GoogleKmsService, GoogleProjectDAO, GooglePubSubDAO, GoogleStorageDAO}
+import org.broadinstitute.dsde.workbench.google2.{GcsBlobName, GoogleStorageService}
 import org.broadinstitute.dsde.workbench.model.Notifications.Notification
 import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport.WorkbenchGroupNameFormat
 import org.broadinstitute.dsde.workbench.model._
@@ -32,6 +34,7 @@ import org.broadinstitute.dsde.workbench.util.{FutureSupport, Retry}
 import spray.json._
 
 import java.io.ByteArrayInputStream
+import java.net.URL
 import java.util.Date
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
@@ -56,6 +59,7 @@ class GoogleExtensions(
     val googleKeyCache: GoogleKeyCache,
     val notificationDAO: NotificationDAO,
     val googleKms: GoogleKmsService[IO],
+    val googleStorageService: GoogleStorageService[IO],
     val googleServicesConfig: GoogleServicesConfig,
     val petServiceAccountConfig: PetServiceAccountConfig,
     val resourceTypes: Map[ResourceTypeName, ResourceType],
@@ -174,7 +178,7 @@ class GoogleExtensions(
   // The handler for the subscription will ultimately call GoogleExtensions.synchronizeGroupMembers, which will
   // do all the heavy lifting of creating the Google Group and adding members.
   override def publishGroup(id: WorkbenchGroupName): Future[Unit] =
-    googleGroupSyncPubSubDAO.publishMessages(googleServicesConfig.groupSyncPubSubConfig.topic, Seq(id.toJson.compactPrint))
+    googleGroupSyncPubSubDAO.publishMessages(googleServicesConfig.groupSyncPubSubConfig.topic, Seq(MessageRequest(id.toJson.compactPrint)))
 
   /*
     - managed groups and access policies are both "groups"
@@ -223,7 +227,7 @@ class GoogleExtensions(
       }
 
       // publish all the messages
-      _ <- IO.fromFuture(IO(publishMessages(messages.flatten)))
+      _ <- IO.fromFuture(IO(publishMessages(messages.flatten.map(MessageRequest(_)))))
 
       end <- clock.monotonic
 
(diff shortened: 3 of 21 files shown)


Ghost-in-a-Jar (approved):
might be worth testing out the case where the user doesnt have perms on the gs object, and potentially handling that error

tlangs:
@Ghost-in-a-Jar I'm going to deploy to a BEE and see how it fails and then implement that

tlangs:
jenkins retest

tlangs:
jenkins retest

tlangs:
@Ghost-in-a-Jar turns out google will sign any url, no matter if the object exists or not. I don't think its Sam's job to check for the existence of the object. Its only job is to use the Pet Service Account to sign the URL given.

tlangs:
jenkins retest

tlangs:
jenkins retest

tlangs:
jenkins retest

tlangs:
jenkins retest

Ghost-in-a-Jar (approved):

README.md
```diff
@@ -207,7 +207,7 @@ sh docker/run-proxy.sh start
 ```shell
 sbt run
 ```
-Navigate to [Sam's (Proxy's) Swagger page](http://localhost:50443/)
+Navigate to [Sam's (Proxy's) Swagger page](https://local.broadinstitute.org:50443/#/)
```
gpolumbo-broad, line 210: If we make this change, we also need to add the step here or point to the step that instructs the user on how to modify their `/etc/hosts` file

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -174,7 +178,7 @@ class GoogleExtensions(
   // The handler for the subscription will ultimately call GoogleExtensions.synchronizeGroupMembers, which will
   // do all the heavy lifting of creating the Google Group and adding members.
   override def publishGroup(id: WorkbenchGroupName): Future[Unit] =
-    googleGroupSyncPubSubDAO.publishMessages(googleServicesConfig.groupSyncPubSubConfig.topic, Seq(id.toJson.compactPrint))
+    googleGroupSyncPubSubDAO.publishMessages(googleServicesConfig.groupSyncPubSubConfig.topic, Seq(MessageRequest(id.toJson.compactPrint)))
```
gpolumbo-broad, line 181: Just wondering why this change is in here?  Was this one of the things screwing up tests?  Or maybe something we had to change as a result of a `workbench-libs` version upgrade?  Is this something we should be doing everywhere that we publish messages to a pub-sub queue?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -637,6 +641,14 @@ class GoogleExtensions(
     )
   }
 
+  def getSignedUrl(samUser: SamUser, project: GoogleProject, bucket: GcsBucketName, name: GcsBlobName, samRequestContext: SamRequestContext): IO[URL] =
+    for {
+      petServiceAccount <- createUserPetServiceAccount(samUser, project, samRequestContext)
```
gpolumbo-broad, line 646: 😠 I am not a fan of `createUserPetServiceAccount` both in name and implementation…but that's well out of scope for this ticket.  Just calling it out as naughty and good refactoring fodder.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1096,6 +1096,39 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
       x-codegen-request-body-name: scopes
+  /api/google/v1/user/petServiceAccount/{project}/signedUrlForBlob:
+    post:
+      tags:
+        - Google
+      summary: gets a signed URL for the given blob, signed by the Pet Service account of the calling user
```
gpolumbo-broad: What are credential scopes for the signed URL?  https://cloud.google.com/storage/docs/authentication/signatures#credential-scope

The above link includes an earliest available date, but I know you can also put expiration dates on the signed URL.  Not sure what other limitations can be put on a signed URL and what other defaults might be applied to the signed url.  Since we are not allowing users to specify these parameters on their own, we should do our best to describe the limitations and restrictions or whatever on the signed url.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -637,6 +641,14 @@ class GoogleExtensions(
...
+      petKey <- googleKeyCache.getKey(petServiceAccount)
+      serviceAccountCredentials = ServiceAccountCredentials.fromStream(new ByteArrayInputStream(petKey.getBytes()))
+      url <- googleStorageService.getSignedBlobUrl(bucket, name, serviceAccountCredentials).compile.lastOrError
```
gpolumbo-broad, line 649: Note in this documentation that using this method, the signed url is good for a maximum of 12 hours.  https://cloud.google.com/storage/docs/access-control/signed-urls#signing-iam

Is that going to be a problem?  If Cromwell calls out to Sam in the middle of a run to generate the signed URL, we might be fine, if a job might be running for days or weeks, if we only pass in the signed url (and not the localized file) we might have problems.  I don't really know how cromwell does its thing with regard to this stuff though.  So maybe 12 hours is fine, just want to make sure.  We should also note the 12 hour maximum life span in the Swagger documentation.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesV1Spec.scala
```diff
@@ -246,4 +246,21 @@ class GoogleExtensionRoutesV1Spec extends GoogleExtensionRoutesSpecHelper with S
       status shouldEqual StatusCodes.Forbidden
     }
   }
+
+  "POST /api/google/v1/user/petServiceAccount/{project}/signedUrlForBlob" should "200 with a signed url" in {
+    val (_, samRoutes, projectName) = setupSignedUrlTest()
```
gpolumbo-broad, line 251: In the spirit of test engoodening, I would love to simplify the setup here.  For example, I would consider mocking `googleExtensions.getSignedUrl` directly to return either a valid response or an exception or `Failure` or whatever corresponding to whatever that method does when it fails.  We don't need to test the functionality of `googleExtensions.getSignedUrl` in the route spec, we just need to test that the routing works properly, returns 200, 404, 4XX, 5XX, whatever.  Then we test `googleExtensions.getSignedUrl` in the `GoogleExtensionsSpec`.  

Since `setupSignedUrlTest` is also being added in this test, I would like to ask you to try writing these tests and just mocking out `googleExtensions.getSignedUrl`.  If we can avoid introducing the big complicated setup, I'd prefer to do that.  If it doesn't work, I'd be interested to see what is messing us up (in the spirit of keeping track of the design/implementation subtleties that are forcing us to write more complicated tests than we want)

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
gpolumbo-broad, line 1: 🤮 OMG - this file is SMELLY!  Just the number of lines it and the `GoogleExtensions` constructor repeated 8 billion times with the gigantic parameter list… eww eww eww eww

I think it would be a good idea to do something to get us started with "better" testing here.  Maybe create a new `GoogleExtensionSpecNewAndImproved` or something that does all of this stuff… better.  All it needs to have in it for now is the 1 new test you are adding, just something to get us started on writing all these tests better so that they're more clear, more specific, and overall easier to understand and maintain.

For example, we could make something like a `MockGoogleExtensionBuilder` (see `MockDirectoryDaoBuilder`) so we don't have all those repetitive `null`s and yuck.  The idea would be, yeah, it's a little extra work for this PR, but _somebody_ has to be first.  Then the next person who has to do something in _this_ file, we point them to do it in the new file.  Every time we do that, it will get a little easier and a little better. Bit by bit, I think that's the way we'll need to tackle this.

src/test/scala/org/broadinstitute/dsde/workbench/sam/mock/RealKeyMockGoogleIamDAO.scala
gpolumbo-broad, line 1: I'm wondering if we need this class?  From my quick search through the code, it looks to me like you're just mocking out calls to create a key anyway.  In general, if we can avoid mocking out implementation like this, the better.  If there are ways the code is designed that is making the mocking of this challenging, we should consider changing those designs.

gpolumbo-broad (changes requested):

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
gpolumbo-broad: Just to be clear, it's not YOUR code I'm complaining about.  It's the overall mess that this test suite has become over the years.  :)

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
tlangs: This is a result of the upgrade to workbench-libs

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
tlangs: This isn't being used by Cromwell. It's being used by the UI to generate download links. Also, these links are good for 1 hour, not 12, which is what Martha was doing.

README.md
tlangs: That's already covered in the `firecloud-develop setup` portion of the README

src/main/resources/swagger/api-docs.yaml
tlangs: The scopes are the same as those of the credentials passed to Sam from the UI. Its not doing anything different from Martha here.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesV1Spec.scala
tlangs: I'm hesitant to pour effort into improving these tests. The long-ish term for Sam is to pull this functionality into a separate service, and we're trying to get DrsHub done asap so we can hand it off. I'm not sure its worth the delay to rewrite the tests.

src/test/scala/org/broadinstitute/dsde/workbench/sam/mock/RealKeyMockGoogleIamDAO.scala
tlangs: We need this because workbench-libs requires a real-looking key even in its mocks.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
tlangs: As above, I'm really hesitant to pour a lot of effort into refactoring these tests, since this functionality will be ripped out of Sam and put into a new service. I'd much rather direct effort at getting DRSHub off our hands so that we can get to actually yanking this functionality out of Sam sooner.

tlangs:
jenkins retest

gpolumbo-broad:
@tlangs:

> google will sign any url

I know this is documented in the google stuff, but I also think we should document our Swagger endpoints much better.  I think noting this and any other "settings" or "scopes" on the signed url endpoint is useful so that users know what to expect from our endpoint without also needing to go read through all the google documentation.  Furthermore, it means that when someone asks us a question about how this stuff behaves, we should be able to easily find the answer and not need to go read through all of our code and google documentation to take a stab at it.  If it's in the Swagger docs, they shouldn't even need to ask us in the first place.  If it's expressed via a unit test, we know, "ah, we pass through any url, even if it doesn't exist to google and have them sign it.  That was our intended behavior."

gpolumbo-broad (approved):

tlangs:
@gpolumbo-broad I added some API documentation.

tlangs:
jenkins retest

tlangs:
jenkins retest

PR #1076, opened 2023-05-16
PF-2712: Patch additional groups in an auth domain

mikenev:
Ticket: PF-2712

What: Adds a PATCH endpoint to add groups to an auth domain on a resource.

Why: For the idea of data collections - if we want to import from a data collection into a workspace then the protections set on that data collection should carry over/apply to the workspace as well. 

How: The PATCH endpoint will take in a list of groups for the auth domain. We'll do a diff on that list with the existing groups and add in the missing groups to the auth domain. We don't currently support removing as that has some other considerations.

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1675,11 +1675,11 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
-    post:
+    patch:
       tags:
         - Resources
       summary: Update the groups in the Auth Domain for a resource
-      operationId: postAuthDomainV2
+      operationId: patchAuthDomainV2
       parameters:
         - name: resourceTypeName
           in: path
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1675,6 +1675,41 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+    put:
+      tags:
+        - Resources
+      summary: Update the groups in the Auth Domain for a resource
+      operationId: putAuthDomainV2
+      parameters:
+        - name: resourceTypeName
+          in: path
+          description: Type of resource
+          required: true
+          schema:
+            type: string
+        - name: resourceId
+          in: path
+          description: Id of resource
+          required: true
+          schema:
+            type: string
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              type: array
+              items:
+                type: string
+      responses:
+        204:
+          description: Successfully updated domains
+          content: { }
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
   /api/resources/v2/{resourceTypeName}/{resourceId}/allUsers:
     get:
       tags:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1675,11 +1675,11 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
-    put:
+    post:
       tags:
         - Resources
       summary: Update the groups in the Auth Domain for a resource
-      operationId: putAuthDomainV2
+      operationId: postAuthDomainV2
       parameters:
         - name: resourceTypeName
           in: path
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
@@ -366,7 +366,7 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
 
   def putResourceAuthDomain(resource: FullyQualifiedResourceId, samUser: SamUser, samRequestContext: SamRequestContext): server.Route =
     put {
-      requireAction(resource, SamResourceActions.alterPolicies, samUser.id, samRequestContext) {
+      requireAction(resource, SamResourceActions.readAuthDomain, samUser.id, samRequestContext) {
         entity(as[Set[WorkbenchGroupName]]) { authDomains =>
           complete(resourceService.setResourceAuthDomain(resource, authDomains, samRequestContext).map { response =>
             StatusCodes.OK -> response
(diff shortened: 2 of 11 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1643,6 +1643,41 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+    put:
```
tlangs: This should probably be a PATCH instead of PUT.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -363,6 +364,17 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
       }
     }
 
+  def putResourceAuthDomain(resource: FullyQualifiedResourceId, samUser: SamUser, samRequestContext: SamRequestContext): server.Route =
+    put {
+      requireAction(resource, SamResourceActions.readAuthDomain, samUser.id, samRequestContext) {
```
tlangs: The caller needs more than to be able to read the auth domain on the workspace. There's no `write_auth_domain` action on auth domains, so one will need to be created.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -589,6 +589,18 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
       }
     }
 
+  override def setResourceAuthDomain(
+      resource: FullyQualifiedResourceId,
+      authDomains: Set[WorkbenchGroupName],
+      samRequestContext: SamRequestContext
+  ): IO[Unit] = {
+    serializableWriteTransaction("removeAuthDomainFromResource", samRequestContext) { implicit session =>
```
tlangs: this transaction seems incorrectly named.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
mikenev: This is one of the open questions - what permission level should we require? Do we need something new or can it be something like the resource owner can perform this?

src/main/resources/swagger/api-docs.yaml
dvoet: I think this implementation operates more like a POST than a PUT. I expect PUT to be create or overwrite. This implementation is add only.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -271,6 +271,16 @@ class ResourceService(
         }
       )
 
+  def setResourceAuthDomain(
```
dvoet: This function needs to resynchronize any google groups. Probably need to call `cloudExtensions.onGroupUpdate` with the list of policies for the resource.

dvoet: This function should validate the auth domains probably using `validateAuthDomain` function

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
mikenev: Yeah, makes sense.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -271,6 +271,21 @@ class ResourceService(
...
+      resource: FullyQualifiedResourceId,
+      authDomains: Set[WorkbenchGroupName],
+      userId: WorkbenchUserId,
+      samRequestContext: SamRequestContext
+  ): IO[Set[WorkbenchGroupName]] =
+    for {
+      resourceType <- getResourceType(resource.resourceTypeName)
+      policies <- listResourcePolicies(resource, samRequestContext)
+      _ <- validateAuthDomain(resourceType.get, authDomains, userId, samRequestContext)
+      _ <- cloudExtensions.onGroupUpdate(policies.map(p => FullyQualifiedPolicyId(resource, p.policyName)), samRequestContext)
```
dvoet, line 285: `onGroupUpdate` should come after `accessPolicyDAO.setResourceAuthDomain` to avoid a race condition where the message is handled before the database is updated

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -271,6 +271,21 @@ class ResourceService(
...
+  def setResourceAuthDomain(
+      resource: FullyQualifiedResourceId,
+      authDomains: Set[WorkbenchGroupName],
+      userId: WorkbenchUserId,
+      samRequestContext: SamRequestContext
+  ): IO[Set[WorkbenchGroupName]] =
+    for {
+      resourceType <- getResourceType(resource.resourceTypeName)
+      policies <- listResourcePolicies(resource, samRequestContext)
+      _ <- validateAuthDomain(resourceType.get, authDomains, userId, samRequestContext)
```
dvoet, line 282: validate before listing policies

dvoet (commented):
I think the functionality looks right. I did not really look at the tests. I will defer to the team for actual thumbs.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
mikenev: Thanks Doug, I went back and forth on the order there. This makes sense.

src/main/resources/swagger/api-docs.yaml
tlangs: This feels more like a PATCH to me than a POST. We're updating the value, not creating it. Well, it may be being created if going from 0 -> 1 groups, but more than that is an update instead of a creation.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -589,6 +589,19 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
       }
     }
 
+  override def setResourceAuthDomain(
```
tlangs: very minor. Could this be renamed to `addResourceAuthDomain`? `set` implies it might overwrite current auth domains

Ghost-in-a-Jar (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
mikenev: Yeah, that makes sense

tlangs (approved):

dvoet (approved):

PR #1100, opened 2023-06-12
ID-626 put update user endpoint

sjkobori:
Draft Notes:
  I will be adding the following to this ticket:
- ~~Send back a User object as a response with the updated user (updating the corresponding tests)~~
- ~~Change this to be a PATCH request since we are not providing a complete user object to replace, just partial information~~
- ~~Update all tests in AdminRoutesSpec to pass using the new builder framework~~
- ~~Add update tests corresponding to all response codes sent back by this endpoint~~

Ticket: https://broadworkbench.atlassian.net/browse/ID-626

What:

Adding an endpoint to update the user more generically (just email for now).

Why:

This begins our effort to consolidate our admin endpoints so that we can conform to a standard CRUD infrastructure. Currently, we have several endpoints which update parts of a user but especially if we add more functionality (updating email address etc.) our current architecture will not scale well.

How:

This PR adds a new admin route to update a user by invoking an update user call through our user service. This service access our DAO to update the email address for that user. This will provide a clear avenue to allow any sort of user update being supported by this new endpoint.

I would firstly look at the updated swagger doc (api-docs.yaml) for the new PUT /user/{userId} endpoint. Then the corresponding route in AdminRoutes. Follow that down to UserService and then PostgresDirectoryDAO.

In AdminUserRoutesSpec @gpolumbo-broad and I worked on updating the tests to follow a new builder pattern which abstracts out the construction of a mock DB. That will need to be finished in order to get the other broken tests to pass in that file.

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminRoutes.scala
```diff
@@ -57,6 +57,15 @@ trait AdminRoutes extends SecurityDirectives with SamRequestContextDirectives wi
...
                   }
+                } ~
+                patch {
+                  entity(as[AdminUpdateUserRequest]) { request =>
+                    complete {
+                      userService
+                        .updateUserCrud(WorkbenchUserId(userId), request, samRequestContext)
+                        .map(user => (if (user.isDefined) OK else NotFound) -> user)
+                    }
+                  }
```
Ghost-in-a-Jar, line 68: we give akkahttp a lot of shit but this is very readable imo

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -115,6 +115,28 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
+          // validate all fields to be updated
+          var errorReports = Seq[ErrorReport]()
+          request.email.foreach(email => errorReports = errorReports ++ validateEmail(email, blockedEmailDomains))
+          if (errorReports.nonEmpty) {
+            IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "invalid user update", errorReports)))
+          } else { // apply all updates
+            var updatedUser = user
+            request.email.foreach { email =>
+              directoryDAO.updateUserEmail(userId, email, samRequestContext)
+              updatedUser = user.copy(email = email)
```
Ghost-in-a-Jar, line 131: This feels a little smelly, I think id prefer to do a select for the new user state from the db just in case there are two concurrent requests coming in. This endpoint is for admins only and shouldn't see a lot of traffic so i think two separate db calls and transactions is fine.

Ghost-in-a-Jar (changes requested):
See comment on adding a select statement after the update. Unless we can return the new user state from the update query itself but I am not sure that's possible.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
sjkobori: Yeah I was sort of going back and forth about this. What do you think about using the `RETURNING` parameter in the SQL statement (https://www.postgresql.org/docs/9.5/sql-update.html). That way we can get the record directly from the db without creating a new call?

sjkobori: I am noticing that we do use this same pattern (using the request object to construct the response directly) when we do a createUser, but I do think there is a better solution. I think it might be a good ticket to open if I can find a more general example of returning the status of DB calls. Currently, most of our DAO methods return a Unit which I do not really like the design of so I think this is worth looking into

sjkobori: @Ghost-in-a-Jar I think this may be a larger can of worms than I had initially thought. I think for the time being I can load the user again after updating (two calls). I would like to come back and update this but I think that might be its own ticket to change it so all db calls send back their updated object rather than just a Unit. Out of these two options, which do you think it preferable?

Shakespeared (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
Ghost-in-a-Jar: @sjkobori Okay yeah this is a can of worms agreed lets create a ticket.

Ghost-in-a-Jar (approved):

PR #1113, opened 2023-06-22
ID-663 Admin delete resource endpoint. 

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-663

What:

Adds resource deletion endpoint for admins.

Why:

We need an admin endpoint we can use to delete resources for users without having to be invited to their workspace.

How:

We took the logic from the resource deletion endpoint and added one for admins (requiring admin privs on the incoming request.

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminRoutes.scala
```diff
@@ -109,46 +109,59 @@ trait AdminRoutes extends SecurityDirectives with SamRequestContextDirectives wi
...
+                          }
+                        }
                       }
-                    }
                   }
+                }
               }
-            }
           }
-      }
```
Ghost-in-a-Jar, line 161: There was just a `PathPrefix("resources" / Segment / Segment / "policies")` here before and now we have a new `PathPrefix("resources" / Segment / Segment)` delete endpoint. The old one is now nested in the `PathPrefix("resources" / Segment / Segment)` block.

PR #1131, opened 2023-07-14
ID 651: Allow Resource creation from policy identifiers without policy emails

sjkobori:
Ticket: [ID-651](https://broadworkbench.atlassian.net/browse/ID-651)

~~Putting this here for now just to get visibility. Currently having trouble creating a test without reinventing the testing framework being used.~~ Solved during mobbing by moving the tests to ResourceServiceSpec, where they should have been anyway.

What:

This PR will allow users of our service to create a resource with a "billing-owner" policy included without specifying that policy's email and can instead use the following triple: `(resourceTypeName, policyName, resourceId)`

Why:

The specification of the email can be cumbersome as that information is not as easily accessible as the mentioned triple. By allowing this way of specifying a policy, our other services (in particular RAWLS) can add billing owner's as members to a resource more easily.

How:

This PR accomplishes this task by using the specified triple to construct the corresponding email behind the scenes to add that policy email to the resource upon creation. While this is not an ideal solution, it requires minimal adjustment of the existing codebase. This logic should be able to be stripped out once we deprecate the exposure of the policy email with our new API endpoints.

---

**PR checklist**

[ID-651]: https://broadworkbench.atlassian.net/browse/ID-651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -278,10 +289,21 @@ consistent "has a" relationship is tracked by this ticket: https://broadworkbenc
     actions: Set[ResourceAction],
     roles: Set[ResourceRoleName],
     descendantPermissions: Option[Set[AccessPolicyDescendantPermissions]] = Option(Set.empty),
-    memberPolicies: Option[Set[PolicyIdentifiers]] = Option(Set.empty)
+    memberPolicies: Option[Set[PolicyInfoResponseBody]] = Option(Set.empty)
 ) {
   def getDescendantPermissions: Set[AccessPolicyDescendantPermissions] = descendantPermissions.getOrElse(Set.empty)
 }
+
+@Lenses final case class AccessPolicyMembershipRequest(
```
tlangs: Did we want to start the namespacing of requests and responses now, or in a follow-on ticket.

sjkobori: I am happy to start with these now, but since we did not decide to version this change, I was not sure how we wanted to do it.

sjkobori: @tlangs and I met on Gather and we decided to make a sam.model.api package and to move the affected request/response objects (in this PR) in there for the time being until we can circle back and do namespacing for our SamModel objects

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -714,22 +719,67 @@ class ResourceService(
...
+    }
+  }
+
+  private def loadPolicyByPolicyIdentifiers(policyIdentifiers: PolicyIdentifiers, samRequestContext: SamRequestContext): IO[Option[AccessPolicy]] =
+    accessPolicyDAO.loadPolicy(
+      FullyQualifiedPolicyId(FullyQualifiedResourceId(policyIdentifiers.resourceTypeName, policyIdentifiers.resourceId), policyIdentifiers.policyName),
+      samRequestContext
+    )
+
+  private def collateMemberAndPolicyEmails(memberEmails: Set[WorkbenchEmail], maybePolicyEmails: Option[IO[Set[WorkbenchEmail]]]): IO[Set[WorkbenchEmail]] =
```
tlangs: I don't think this needs to be its own function. You could just tack on the contents of this function onto the end of `getPolicyEmailsFromAccessPolicyMembership` and get the same result. That way, we're maintaining `IO` as the top-level type, and this function isn't used anywhere else.

tlangs (approved):
Other than the comment, LGTM!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -714,22 +719,64 @@ class ResourceService(
...
-        makeCreatablePolicy(accessPolicyName, accessPolicyMembership, samRequestContext)
+      .traverse { case (accessPolicyName, accessPolicyMembershipRequest) =>
+        // grouping member emails and policy emails together
+        val allEmails = getPolicyEmailsFromAccessPolicyMembership(accessPolicyMembershipRequest, samRequestContext)
+          .map { ioPolicyEmails =>
+            for {
+              policyEmails <- ioPolicyEmails
+            } yield accessPolicyMembershipRequest.memberEmails ++ policyEmails
+          }
+          .getOrElse(IO(accessPolicyMembershipRequest.memberEmails))
```
Ghost-in-a-Jar: Nit: it feels like allEmails could be in the for comp below with the other stuff but not 100% sure

Ghost-in-a-Jar (approved):
small nitpick

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2997,6 +2997,34 @@ components:
...
+            type: string
+        descendantPermissions:
+          type: array
+          items:
+            $ref: '#/components/schemas/AccessPolicyDescendantPermissions'
+      description: ""
+    AccessPolicyMembershipRequest:
       required:
         - actions
         - memberEmails
```
cahrens, line 3030: @sjkobori `memberEmails` shouldn't be marked as required, correct? You have to supply either the `memberEmails` or the `memberPolicies` (I will be supplying only `memberPolicies`).

sjkobori: That should be updated in the swagger doc, I can make a ticket t update that. But it functionally works the way you are describing. Providing either will be sufficient.

sjkobori: If you provide an empty list for `memberEmails` and provide `memberPolicies` the endpoint should work correctly. Currently, the documentation aligns with the functionality, as `memberEmails` is not an optional value. It may make sense to change that going forward, but in general, I want to make a pass on all of the collections we are using across our endpoints. I want make sure that we have a consistent way of handling empty collections before making a change here.

PR #1136, opened 2023-07-18
ID-698-Sam-GET-user-endpoint

sjkobori:
Ticket: https://broadworkbench.atlassian.net/browse/ID-698

What:

This PR adds a GET user admin endpoint so that it is possible to retrieve a user record from sam.

Why:

This will allow easier access to sam user data which previously was accessed though the database directly.

How:

This PR adds a v2 user admin route which returns a user record

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3715,15 +3715,6 @@ components:
           type: string
           description: User's email address
           format: email
-        azureB2CId:
-          type: string
-          description: User's Azure B2C Id
-        enabled:
-          type: boolean
-          description: Whether or not the user is enabled
-        acceptedTosVersion:
-          type: string
-          description: The version of terms of service the user accepted
       description: specification of a UpdateUserRequest
   securitySchemes:
     googleoauth:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -56,7 +56,7 @@ paths:
         content:
           'application/json':
             schema:
-              $ref: '#/components/schemas/User'
+              $ref: '#/components/schemas/UpdateUserRequest'
       responses:
         200:
           description: User was updated successfully
@@ -3621,8 +3621,13 @@ components:
           example: my-managed-resource-group
       description: specifies a request for a pet managed identity
     User:
-      required:
       type: object
+      required:
+        - id
+        - email
+        - enabled
+        - createdAt
+        - updatedAt
       properties:
         id:
           type: string
@@ -3656,6 +3661,26 @@ components:
           format: date-time
           description: User's time of last update
       description: specification of a User
+    UpdateUserRequest:
+      type: object
+      properties:
+        googleSubjectId:
+          type: string
+          description: User's Google subject Id
+        email:
+          type: string
+          description: User's email address
+          format: email
+        azureB2CId:
+          type: string
+          description: User's Azure B2C Id
+        enabled:
+          type: boolean
+          description: Whether or not the user is enabled
+        acceptedTosVersion:
+          type: string
+          description: The version of terms of service the user accepted
+      description: specification of a UpdateUserRequest
   securitySchemes:
     googleoauth:
       type: oauth2
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -72,6 +72,38 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/admin/v2/user/{userId}:
+    get:
+      tags:
+        - Admin
+      summary: Retrieves a user record, by user id
+      operationId: adminGetUser
+      parameters:
+        - name: userId
+          in: path
+          description: User ID of the user to be retrieved
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: User was retrieved successfully
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/User'
+        403:
+          description: You do not have admin privileges
+          content: { }
+        404:
+          description: User not found
+          content: { }
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
   /api/admin/v1/user/email/{email}:
     get:
       tags:
@@ -3555,12 +3587,39 @@ components:
       description: specifies a request for a pet managed identity
     User:
       required:
-        - email
       type: object
       properties:
+        id:
+          type: string
+          description: User's Id
+        googleSubjectId:
+          type: string
+          description: User's Google subject Id
         email:
           type: string
           description: User's email address
+          format: email
+        azureB2CId:
+          type: string
+          description: User's Azure B2C Id
+        enabled:
+          type: boolean
+          description: Whether or not the user is enabled
+        acceptedTosVersion:
+          type: string
+          description: The version of terms of service the user accepted
+        createdAt:
+          type: string
+          format: date-time
+          description: User's time of creation
+        registeredAt:
+          type: string
+          format: date-time
+          description: User's time of registration
+        updatedAt:
+          type: string
+          format: date-time
+          description: User's time of last update
       description: specification of a User
   securitySchemes:
     googleoauth:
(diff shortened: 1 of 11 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/sam/api/AdminUserRoutesSpec.scala
```diff
@@ -325,27 +306,78 @@ class AdminUserRoutesSpec extends AnyFlatSpec with Matchers with ScalatestRouteT
...
+      .withNonAdminUser()
+      .withEnabledUser(defaultUser) // "persisted/enabled" user we will check the status of
       .build
     // Act
     Delete(s"/api/admin/v1/user/$defaultUserId/petServiceAccount/myproject") ~> samRoutes.route ~> check {
       status shouldEqual StatusCodes.Forbidden
     }
   }
+
+  "GET /admin/v2/user/{userId}" should "get the corresponding user object (as an admin)" in {
```
Shakespeared: Nit: maybe reword this slightly to something like "get the corresponding user object (if the requesting user is an admin)", just to make it extra explicit that it's the user calling the endpoint who has to be an admin

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/AdminUserRoutesSpec.scala
```diff
@@ -325,27 +306,78 @@ class AdminUserRoutesSpec extends AnyFlatSpec with Matchers with ScalatestRouteT
...
+    Get(s"/api/admin/v2/user/$defaultUserId") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.OK
+      responseAs[SamUser] shouldEqual defaultUser
+    }
+  }
+
+  it should "not allow access to a user object (as an non admin)" in {
+    // Arrange
+    val samRoutes = new MockSamRoutesBuilder(allUsersGroup)
+      .withEnabledUser(defaultUser) // "persisted/enabled" user we will check the status of
```
Shakespeared: Nit: do you mind putting these in the same order as the previous test case? e.g. always having the user calling the endpoint first and the user record being retrieved second. It just makes it easier to read through the tests quickly

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/AdminUserRoutesSpec.scala
```diff
@@ -325,27 +306,78 @@ class AdminUserRoutesSpec extends AnyFlatSpec with Matchers with ScalatestRouteT
...
+      .withNonAdminUser()
+      .build
+
+    // Act and Assert
+    Get(s"/api/admin/v2/user/$defaultUserId") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.Forbidden
+    }
+  }
+
+  it should "not allow access even if a user object does not exist (as an admin)" in {
```
Shakespeared: Instead of me micromanaging your test names, do you mind taking one last look through and making sure the naming of tests and test setup is consistent? I think it's fine as it is but could be even better :) For instance little things like "return a user object" or something instead of just "allow access," and trying to make the admin/non admin distinction easy to parse at a glance. Overall I think these tests look good though!

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpecs/GetUserSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+          .withEnabledUser(userWithBothIds)
+          .build
+        // Act
+        val response = runAndWait(userService.getUser(userWithBothIds.id, samRequestContext))
+
+        // Assert
+        assert(response.nonEmpty, "Getting a user who exists should return a user object object")
+        assert(
+          response.get equals userWithBothIds,
+          "Getting a user with who exists should return the corresponding user."
```
Shakespeared: I think "user with who exists" is a typo

Shakespeared (approved):
overall looks great! I had a few nitpicks, mostly just to improve some of the test naming to make it super clear and readable.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3633,14 +3665,66 @@ components:
...
+      type: object
+      properties:
+        googleSubjectId:
+          type: string
+          description: User's Google subject Id
+        email:
+          type: string
+          description: User's email address
+          format: email
+        azureB2CId:
```
tlangs: Are we sure we want the B2C ID to be updatable?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3633,14 +3665,66 @@ components:
...
+          type: string
+          description: User's Azure B2C Id
+        enabled:
+          type: boolean
+          description: Whether or not the user is enabled
+        acceptedTosVersion:
```
tlangs: The accepted TOS version should only be updated via the TOS endpoints.

src/main/resources/swagger/api-docs.yaml
sjkobori: As it stands right now, the only field that is updatable in practice is the email. The UpdateUserRequest object does contain all of these fields and I realized that the swagger doc was not in sync with the UpdateUser endpoint's request object. We still need to make decisions on what fields we are allowing to update. I think ultimately that the B2CId might be useful to update without having to hit the db directly. We can restrict updating this field to only superadmins or something. Right now, we are using the sam user directives to manage authorization and the level needed to use this endpoint are admins. Going forward, I think we will have a permissions service that had handle what actions are allowed by a user more granularly (nobody can update a user email except for themselves and users with admin roles)

src/main/resources/swagger/api-docs.yaml
sjkobori: This is another thing I wanted to talk about. I do agree that I think only a user can accept their tos (although I think it might be valuable to edit the version if for some reason there is a data discrepancy), but I think eventually we would want to do away with the specific endpoints for updating one field of a user. For now, I can take this out of the swagger doc, would you also like me to remove it from the UpdateUserRequest object in code (right now it is not doing anything) or just leave that as it is?

tlangs: The Update model should probably only contain update-able fields, since our API definition should reflect the functionality of the endpoint. As for the B2C ID, I we need to do a lot of hard thinking about if that should be update-able. I don't think we've gotten any user requests or support tickets involving updating the B2C ID for a user. We do know that there's a need to update emails and Google Subject IDs, so that's probably what we should focus on and provide for now.

tlangs: But, to another point, a set of CRUD endpoints should probably use the same object for requests and responses. So, if we were to have just a `SamUser` object with all the user fields, and use that as both the request and response, then we could say that the only fields that are update-able are the email and Google Subject ID.

sjkobori: I think that we should not be using the same request and response objects since it makes it difficult to know what is required in a request vs what you know is in the response (that is what made updating the `PolicyIdentifiers `complex in the first place). So for example, in the `GET` user endpoint, we can say that the email is a required field. But in the `PATCH` user endpoint, email is not required. This could be done in the `POST` user endpoint by marking certain required field as read only since you need a complete object to make this request. So I still think that there should be a `UpdateUserRequest` object, but I will update it to only have the fields we currently support updating.

Ghost-in-a-Jar (changes requested):
User update request should only have googleSubjectId, email, enabled, and you should wire those to be actually updated by admins if included in the request.

There are probably additional steps needed when updating the email, besides just updating the db record. (thurloe?)

We may want to consider implementing email in a separate pr.

sjkobori:
> User update request should only have googleSubjectId, email, enabled, and you should wire those to be actually updated by admins if included in the request.
> 
> There are probably additional steps needed when updating the email, besides just updating the db record. (thurloe?)
> 
> We may want to consider implementing email in a separate pr.

I think since now the PATCH user endpoint has been disabled in prod, I would like to keep it that way for this ticket and add all the intended features for that endpoint in ticket: https://broadworkbench.atlassian.net/browse/ID-718 . This way it is not longer blocking people that need the GET user endpoint.

Ghost-in-a-Jar (approved):

PR #1140, opened 2023-07-25
ID-710 Add database check to liveness endpoint.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-710

What:

The current liveness check should cause sam to restart if the db connection is bad.

Why:

On db maintenance windows sam gets stuck in a state where the status endpoint is failing (k8s readiness probe is failing so no connections sent to the pod), but the liveness endpoint is succeeding even though the db connection is bad so the pod doesnt get killed. From the cluster's point of view the pod is finishing startup.

How:

Added a select 1 to the liveness check endpoint.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -6,6 +6,7 @@ import cats.data.NonEmptyList
 import cats.effect._
 import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
+import io.prometheus.client.CollectorRegistry
 import org.broadinstitute.dsde.workbench.dataaccess.PubSubNotificationDAO
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.{Json, Pem}
 import org.broadinstitute.dsde.workbench.google.{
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamRequestContextDirectives.scala
```diff
@@ -18,7 +18,6 @@ import scala.util.control.NonFatal
 /** Created by ajang on 2020-05-28
   */
 trait SamRequestContextDirectives {
-  val userService: UserService
```
Ghost-in-a-Jar, line 21: I don't think this was actually being used for anything

sjkobori (approved):
Looks good to me!

tlangs (approved):
Has this been synthetically tested?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -44,23 +45,28 @@ import scala.jdk.CollectionConverters._
...
+
+  def run(args: List[String]): IO[ExitCode] = {
+    initSentry()
+    implicit val system = ActorSystem("sam")
+
     (startup() *> ExitCode.Success.pure[IO]).recoverWith { case NonFatal(t) =>
       logger.error("sam failed to start, trying again in 5s", t)
+
+      Http().shutdownAllConnectionPools() *> system.terminate()
+      CollectorRegistry.defaultRegistry.clear()
```
Ghost-in-a-Jar, line 60: I noticed when sam restarts locally the liveness server still hangs around and causes confusing errors. Also I noticed similar errors relating to prometheus that went away when I added L57.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -93,9 +98,7 @@ object Boot extends IOApp with LazyLogging {
         loggerIO
           .error(t)("FATAL - failure starting liveness http server")
           .unsafeToFuture()(cats.effect.unsafe.IORuntime.global)
-      }
-
-    loggerIO.info("Liveness server has been started").unsafeRunSync()(cats.effect.unsafe.IORuntime.global)
+      } >> loggerIO.info("Liveness server has been started").unsafeToFuture()(cats.effect.unsafe.IORuntime.global)
```
Ghost-in-a-Jar: This just chains them together so they run in order.

PR #1146, opened 2023-07-28
ID-717 Removed logic in updateUserEmail

sjkobori:
Ticket: https://broadworkbench.atlassian.net/browse/ID-717

What:

This PR takes out the database update of user emails in the updateUserRequest flow

Why:

Currently, we need to do more research into how to update a user's email. We need to also update the user's proxy email. So we are pulling back on the update email until that is also supported.

How:

The database call through the `directoryDao` is replaced with a `Unit`.

---

**PR checklist**


Shakespeared (approved):

Ghost-in-a-Jar (approved):
woopsie

PR #1148, opened 2023-08-03
ID-727 GET users by id, googleSubjectId, or azureB2cId

sjkobori:
Ticket: https://broadworkbench.atlassian.net/browse/ID-727

What:

This PR adds a new endpoint which allows the request of a list of user data which match on either the userId, googleSubjectId, or azureB2CId.

Why:

This will allow Thurloe to communicate with sam so that it can properly identify users by any of their corresponding ids and send an email to their email address.

How:

The new endpoint added can be queried by Thurloe to find the user who has the id Thurloe knows about (could be googleSubjectId or azureB2CId) and return the matching user.

---

**PR checklist**


src/main/resources/swagger/api-docs.yaml
```diff
@@ -124,6 +124,37 @@ paths:
...
+  /api/admin/v2/users:
+    get:
+      tags:
+        - Admin
+      summary: gets the users which match the any of the query params
+      operationId: adminGetUserByQuery
+      parameters:
+        - name: id
+          in: query
+          description: User Id ato find users by
```
Ghost-in-a-Jar: nit

Ghost-in-a-Jar (approved):

Shakespeared (approved):

PR #1159, opened 2023-08-14
ID-741 Reduce logs emitted by Sam

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-741

What:

Sam logs a ton. It should log less.

Why:

Logs cost money

How:
Reduce the logging of duplicate information, and reduce verbosity.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
@@ -63,7 +63,7 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with 
       optionalHeaderValueByName(managedIdentityObjectIdHeader).map(_.map(ManagedIdentityObjectId)))
       .as(OIDCHeaders)
       .map { oidcHeaders =>
-        logger.debug(s"Auth Headers: $oidcHeaders")
+        logger.info(s"Auth Headers: $oidcHeaders")
         oidcHeaders
       }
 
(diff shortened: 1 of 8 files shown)


sjkobori (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -29,7 +29,7 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with
     onSuccess {
       getActiveSamUser(oidcHeaders, userService, tosService, samRequestContext).unsafeToFuture()
     }.tmap { samUser =>
-      logger.info(s"Handling request for active Sam User: $samUser")
+      logger.debug(s"Handling request for active Sam User: $samUser")
```
sjkobori, line 32: I made this change for the `asAdminServiceUser` directive I am adding in ID-727 : https://github.com/broadinstitute/sam/pull/1148 so it is consistent

Shakespeared (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
gpolumbo-broad: One of the things we lost with the switch to JWT was that we don't know which user is making any particular request anymore.  That's why this was added.  I admit that it's gross and not very graceful and is creating a lot of noise, but for troubleshooting, debugging, and production support, being able to get the `SamUser` info for a request is really valuable.  

Maybe the alternative is to just log out the `samRequestContext` (or at least part of it) when we log?  I have no idea what's in there or if that would be a terrible idea though.

gpolumbo-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -49,7 +49,7 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with
     onSuccess {
       getSamUser(oidcHeaders, userService, samRequestContext).unsafeToFuture()
     }.tmap { samUser =>
-      logger.info(s"Handling request for (in)active Sam User: $samUser")
+      logger.debug(s"Handling request for (in)active Sam User: $samUser")
```
rtitle, line 52: I know this is merged already, but can we also drop the log statement on L77?
```
logger.info(s"Auth Headers: $oidcHeaders")
```

PR #1161, opened 2023-08-17
[DC-736] Add snapshot builder config view action

s-rubenstein:
Ticket: [Link to Jira ticket](https://broadworkbench.atlassian.net/browse/DC-736)

What:

We want to start developing the Snapshot Builder config in TDR, but want to gate use of it behind admin access for now

Why:

This gives us the ability to develop the config safely without risking other people thinking it is open for use.

How:

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1200,7 +1200,7 @@ resourceTypes = {
         roleActions = ["read_dataset", "read_data", "read_policies", "link_snapshot"]
       }
       admin = {
-        roleActions = ["share_policy::steward", "read_policies", "alter_policies"]
+        roleActions = ["share_policy::steward", "read_policies", "alter_policies", "view_snapshot_builder_config"]
       }
     }
     reuseIds = true
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1181,8 +1181,8 @@ resourceTypes = {
       view_journal = {
         description = "View dataset journal entries"
       }
-      view_snapshot_builder_config = {
-        description = "allows viewing the snapshot builder config"
+      view_snapshot_builder_settings = {
+        description = "allows viewing the snapshot builder settings"
       }
     }
     ownerRoleName = "steward"
@@ -1200,7 +1200,7 @@ resourceTypes = {
         roleActions = ["read_dataset", "read_data", "read_policies", "link_snapshot"]
       }
       admin = {
-        roleActions = ["share_policy::steward", "read_policies", "alter_policies", "view_snapshot_builder_config"]
+        roleActions = ["share_policy::steward", "read_policies", "alter_policies", "view_snapshot_builder_settings"]
       }
     }
     reuseIds = true
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1181,6 +1181,9 @@ resourceTypes = {
       view_journal = {
         description = "View dataset journal entries"
       }
+      view_snapshot_builder_config = {
+        description = "allows viewing the snapshot builder config"
+      }
     }
     ownerRoleName = "steward"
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -1181,6 +1181,9 @@ resourceTypes = {
       view_journal = {
         description = "View dataset journal entries"
       }
+      view_snapshot_builder_config = {
```
pshapiro4broad: Will this have any effect if it's not part of any role? I would expect it to be added to the list of action for one of the roles below, like `admin`.

pshapiro4broad (approved):
Looks OK but I think it need to be part of a `role` to be usable.

src/main/resources/reference.conf
s-rubenstein: Oh shoot, I thought I had committed that, mb.

src/main/resources/reference.conf
```diff
@@ -1197,7 +1200,7 @@ resourceTypes = {
         roleActions = ["read_dataset", "read_data", "read_policies", "link_snapshot"]
       }
       admin = {
-        roleActions = ["share_policy::steward", "read_policies", "alter_policies"]
+        roleActions = ["share_policy::steward", "read_policies", "alter_policies", "view_snapshot_builder_config"]
```
nmalfroy: Admin should be kept really small (where an admin can up their privs if needed).  Is this a temporary thing while this is in dev.  I think this makes more sense as a steward roleAction, and possible custodian (since it has manage_schema and they seem somewhat related) though fair if we want to leave off custodian

nmalfroy (changes requested):
Wondering about the role to give this to.  Is the idea that we don't want users messing with this?

src/main/resources/reference.conf
s-rubenstein: Are there stewards in Prod who could use this action? Technically this code isn’t gated to dev, so I’m hesitant to grant it to a role that users could have.

pshapiro4broad: I was thinking it would be `admin` until it's out of development. I guess it depends on how much we want to gate access to it until it's ready to use.

We should probably have a doc that lists all APIs and all permissions to make things explicit; but generally I was thinking that a steward would create a config and a researcher (possibly `reader`) would be able to `get` it. One complication is that I think snapshot request creation will need a new role, unless we want to use `discoverer` for it. `reader` won't work for this because a researcher who can create a request can't read the dataset.

pshapiro4broad: We could also make a new role just for this feature until it's ready.

nmalfroy: There are.  It's basically the owner of a dataset.  I think while we're building it out, it makes sense to do that but I just want to make sure that we eventually add it to stewards (since they'll be the ones interacting in the long run)

s-rubenstein: Eventually, I think actually this action will belong to Reader (or whatever our new "Aggregate data reader" role is) - but steward for now seems okay. I can move it.

nmalfroy: Oh right. derp, this is the view config, not the change config.  I think it's ok to keep it as admin but either is cool

nmalfroy (approved):
lgtm!

src/main/resources/reference.conf
s-rubenstein: If you're okay keeping it as admin, that's my preference because it keeps it more locked down for now!

nmalfroy: @pshapiro4broad , I'd keep the roles pretty simple so that we don't need to change anything on the TDR side when things are ready to go

nmalfroy: Totally.  I think it's cool to leave as is

PR #1163, opened 2023-08-21
ID-757 remove google group syncing testing from integration tests.

jyang-broad:
Ticket: [<Link to Jira ticket>](https://broadworkbench.atlassian.net/browse/ID-757)

What:

Removes an integration test that appears to test the addition and removal of admins from a group.

Why:

This test alone is 50% of the entire suit's time cost and has unclear goals. Is it testing that behaves as expected, or is it testing that google does that Sam thinks it does. (see Jira ticket for more info)

How:

We deleted the entire test.
---

**PR checklist**


Ghost-in-a-Jar (approved):

tlangs (commented):
My only concern is that this covers some sort of functionality not covered by unit tests. Is there a unit test that tests this without the group sync? I definitely don't like that we're actually talking to google in this test. Maybe it could be converted to a test that moves out the google apis if there's not already a unit test.

Ghost-in-a-Jar:
@tlangs yeah we could mock out the actual calls to google and keep the tests, i dont think it would be too much extra work

Shakespeared (approved):

PR #1176, opened 2023-09-07
ID-739 Actually Fix Invited User Registration Dates

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-739

What:

The query to set the `registered_at` for invited users didn't actually set the `registered_at` value for invited users

Why:

The subquery that sets the `registered_at` value relied on both the GoogleSubjectID and Azure B2C ID being null. However, when registering an invited user, separate calls were made to set the GoogleSubjectID and Azure B2C ID. The first call set the `registered_at` value, but because the user now had a Google Subject ID or Azure B2C ID, the subquery returned no results, so the second query overwrote the `registered_at` date set by the first query.

How:

To fix this, I tried to just simplify the query. However, I realized that the query that updates a login provider ID _should not be updating the `registered_at` column_. This is because there may be reasons to set the GoogleSubjectID/AzureB2CID post registration (its happened before), and having methods that updates the `registered_at` value along with those IDs is _very_ likely to have unintended side effects. So, I've broken out the updating of `registered_at` into its own method, and refactored the code around that.

I also deleted a bunch of old code.

---

**PR checklist**


sjkobori (approved):
lgtm!

Shakespeared (approved):

PR #1178, opened 2023-09-08
[DC-754] Add update_snapshot_builder_settings action

s-rubenstein:
Ticket: https://broadworkbench.atlassian.net/browse/DC-754

What:

Adds an action to update snapshot builder configuration.

Why:

How:

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1184,6 +1184,9 @@ resourceTypes = {
       view_snapshot_builder_settings = {
         description = "allows viewing the snapshot builder settings"
       }
+      update_snapshot_builder_settings = {
+        description = "allows updating the snapshot builder settings"
+      }
     }
     ownerRoleName = "steward"
     roles = {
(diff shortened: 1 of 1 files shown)


nmalfroy (approved):
lgtm!

pshapiro4broad (approved):
This seems OK. I was thinking an existing action might make sense like `edit_dataset` but I don't see that action used in TDR (not sure why) and it doesn't hurt to have a finer grained action right now. If we want to consolidate later on we can do it then.

nmalfroy:
> This seems OK. I was thinking an existing action might make sense like `edit_dataset` but I don't see that action used in TDR (not sure why) and it doesn't hurt to have a finer grained action right now. If we want to consolidate later on we can do it then.

Actually, that's a good point, the `manage_schema` action had that in mind.  The problem is that it's not limited to admins.  Eventually, we may want to get rid of this action and just use `manage_schema` for updating the config (since it's really just dataset metadata)

s-rubenstein:
Yeah, I think the other question for the Snapshot Builder is if we would want to open it to everyone - I think eventually yes, but probably not right now.

PR #1183, opened 2023-09-13
[WOR-1247] Add own action to billing project owner role

marctalbott:
Ticket: [WOR-1247](https://broadworkbench.atlassian.net/browse/WOR-1247)

What:
* Add `own` action to `owner` role on `billing-project` resource

Why:
* We need an action to check that a user owns a billing project to determine if they can migrate workspace buckets in the billing project. There is not currently an `own` action on the `billing-project` resource like there is for other resource types (e.g. workspace). Rather than add a specific `migrate_workspace_buckets` action to the `billing-project` resource type that will be obsolete after the migration is finished, I opted to add the more general `own` action.

How:
* Updates `reference.conf` to add new `own` action and includes that action in the `owner` role list of actions.

---

**PR checklist**

[WOR-1247]: https://broadworkbench.atlassian.net/browse/WOR-1247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -705,11 +705,14 @@ resourceTypes = {
       list_children = {
         description = "list children, e.g. google-project"
       }
+      own = {
+        description = "perform owner actions on this billing project"
+      }
     }
     ownerRoleName = "owner"
     roles = {
       owner = {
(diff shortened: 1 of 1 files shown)


aherbst-broad (approved):
TIL about `own` actions, thanks

cahrens:
I don't quite understand this. We already have the ability to ask what billing projects a user has access to, and what the user's role is on those billing projects. Why is it not sufficient to see if the user has the OWNER role on the billing project?

marctalbott:
> I don't quite understand this. We already have the ability to ask what billing projects a user has access to, and what the user's role is on those billing projects. Why is it not sufficient to see if the user has the OWNER role on the billing project?

Upstream services are not supposed to make authz decisions based on the name of the role to avoid misunderstandings between what an upstream service thinks a role should be able to do and what Sam actually says that role can do. Authz decisions need to be made by checking if a user has a particular action on a resource. A general `own` action sort of skirts around this, but given it's an established pattern in Sam I felt comfortable adding it to the `billing-project` resource for the purpose of bucket migrations

cahrens (approved):

PR #1193, opened 2023-09-27
ID-822 New endpoint to sign Requester Pays urls

tlangs:
Ticket :https://broadworkbench.atlassian.net/browse/ID-822

What:

Introduce a new endpoint that allows the caller to get a signed URL while also specifying the requester pays google project.

Why:

We need to be able to specify which project to bill when signing a url for an object in a requester-pays bucket.

How:

Introduced a new endpoint instead of changing the existing one as to not break clients.

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1029,6 +1029,35 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/google/v1/user/signedUrlForBlob:
+    post:
+      tags:
+        - Google
+      summary: Gets a signed URL for the given blob, signed by the arbitrary Pet Service account of the calling user.
+        The signed URL is active for 1 hour and scoped to the permissions of the signing Pet Service Account.
+        An optional requesterPaysProject can be provided to specify which Google Project is billed.
+        Sam will provide a signed URL for any object path, even if that object does not exist.
+      operationId: getRequesterPaysSignedUrlForBlob
+      requestBody:
+        description: bucketName and blobName of the object to get a signed URL for
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/RequesterPaysSignedUrlRequest'
+        required: true
+      responses:
+        200:
+          description: signed URL for the blob, signed by the user's arbitrary Pet Service Account key
+          content:
+            application/json:
+              schema:
+                type: string
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
   /api/google/v1/user/petServiceAccount/{project}:
     get:
       tags:
@@ -3191,6 +3220,26 @@ components:
           type: boolean
           description: Use the pet service account project as the user project in the request.
           default: true
+    RequesterPaysSignedUrlRequest:
+      type: object
+      required:
+        - bucketName
+        - blobName
+      properties:
+        bucketName:
+          type: string
+          description: bucket of the blob
+        blobName:
+          type: string
+          description: path to the blob in the bucket
+        duration:
+          type: number
+          description: Optional validity duration of the link in minutes. Defaults to 1 hour.
+          default: 60
+        requesterPaysProject:
+          type: string
+          description: Optional Google Project to use for billing.
+          default: undefined
     CreateResourceRequest:
       required:
         - policies
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1033,13 +1033,13 @@ paths:
     post:
       tags:
         - Google
-      summary: Gets a signed URL for the given blob, signed by the arbitrary Pet Service account of the calling user.
-        The signed URL is active for 1 hour and scoped to the permissions of the signing Pet Service Account.
+      summary: Gets a signed URL for the given blob.
+        The signed URL is active for 1 hour and scoped to the permissions of the user.
         An optional requesterPaysProject can be provided to specify which Google Project is billed.
         Sam will provide a signed URL for any object path, even if that object does not exist.
       operationId: getRequesterPaysSignedUrlForBlob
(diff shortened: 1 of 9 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -3191,6 +3220,25 @@ components:
           type: boolean
           description: Use the pet service account project as the user project in the request.
           default: true
+    RequesterPaysSignedUrlRequest:
+      type: object
+      required:
+        - bucketName
+        - blobName
+      properties:
+        bucketName:
```
nmalfroy: I know this is copying existing form but I always found it weird that this was two variables rather than a gs path

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -645,6 +646,22 @@ class GoogleExtensions(
...
+      samUser: SamUser,
+      bucket: GcsBucketName,
+      name: GcsBlobName,
+      duration: Option[Long],
+      requesterPaysProject: Option[GoogleProject],
+      samRequestContext: SamRequestContext
+  ): IO[URL] = {
+    val urlParamsMap: Map[String, String] = requesterPaysProject.map(p => Map(userProjectQueryParam -> p.value)).getOrElse(Map.empty)
+    for {
+      petKey <- IO.fromFuture(IO(getArbitraryPetServiceAccountKey(samUser, samRequestContext)))
```
nmalfroy, line 673: Why do this and not mimic the current behavior of specifying a service account in the URL?

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesV1Spec.scala
```diff
@@ -293,4 +293,33 @@ class GoogleExtensionRoutesV1Spec extends GoogleExtensionRoutesSpecHelper with S
       status shouldEqual StatusCodes.NotFound
     }
   }
+  "POST /api/google/v1/user/signedUrlForBlob" should "200 with a signed url" in {
```
nmalfroy, line 296: If you include the PSA to use, this could (should?) be:
```suggestion
  "POST /api/google/v2/user/petServiceAccount/{project}/signedUrlForBlob" should "200 with a signed url" in {
```

nmalfroy: I just worry this is going to get confusing

nmalfroy (commented):
I was wondering why the switch to an arbitrary PSA?

tlangs:
Regarding the arbitrary PSA, its because I'm guessing that most people who are getting Signed URLs are mostly Readers on workspaces. Project-specific PSAs are really a thing for compute VMs, and Readers don't usually have PSAs created for them. So, instead of proliferating PSAs for every workspace users are downloading a file from, why not just use their Arbitrary PSA?

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesV1Spec.scala
tlangs: I want to get rid of the ability to choose which PSA to use. That might be a "controversial" move, but I think its a better design to just use the Arbitrary PSA

src/main/resources/swagger/api-docs.yaml
tlangs: I'm down to make it just take in a GS Path. I'll do that!

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesV1Spec.scala
nmalfroy: I'm not sure from a permission standpoint.  If you're signing a URL from a given project, it seems wrong to me that the service account would be from some random other project (and it makes the call not deterministic, which makes me nervous).  I get having an option do do this arbitrarily but it seems like we should support both...

nmalfroy (approved):
Ok, thanks for confirming that it still works with arbitrary PSAs

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesV1Spec.scala
Shakespeared: Is there somewhere that we could at least put an explanation of why it's an "arbitrary" PSA? To me that is a code smell. If I came in here to update this code and saw that usage without a comment or doc I'd be concerned about why it's doing that and what implications it might have

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -659,22 +676,33 @@ class GoogleExtensions(
...
-          bucket,
-          name,
-          serviceAccountCredentials,
-          expirationTime = timeInMinutes,
-          expirationTimeUnit = TimeUnit.MINUTES,
-          queryParams = queryParams
-        )
-        .compile
-        .lastOrError
+      url <- getSignedUrl(samUser, bucket, name, duration, urlParamsMap, serviceAccountCredentials)
```
Shakespeared, line 693: does the duration make it possible to set the expiration longer than an hour? If so is that a security issue?

tlangs: Google has a max time of 7 days https://cloud.google.com/storage/docs/access-control/signed-urls

Shakespeared (approved):

tlangs:
@Shakespeared you make a good point about the documentation. If the API doesn't let the caller specify a service account to be used, then lets just not say anything about service accounts! I've updated the API documentation.

PR #1197, opened 2023-09-29
[DR-3244] Add support for auth domains for snapshot readers

nmalfroy:
Ticket: https://broadworkbench.atlassian.net/browse/DR-3244

What:

Note: I didn't include discover_data in the list since presumably, we want the data to be discoverable
by users that aren't in the auth domain
(even if they don't have access)

discover_data doesn't give access to read the file or tabular data in (e.g. data plane data)

Why:

This will allow TDR to set auth domain groups on snapshots

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1224,6 +1224,7 @@ resourceTypes = {
       }
       read_data = {
         description = "Read data from the datasnapshot"
+        authDomainConstrainable = true
       }
       discover_data = {
         description = "Discover data about the datasnapshot"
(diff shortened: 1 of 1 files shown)


nmalfroy:
Not really sure why the pact test is failing...I kinda doubt it's related

dvoet (approved):

src/main/resources/reference.conf
```diff
@@ -1281,6 +1283,7 @@ resourceTypes = {
       }
     }
     reuseIds = true
```
tlangs: You _may_ want to set `reuseIds` to false, just so that an ID without an auth domain can't be re-used for something that does require one. Though, I do doubt that it's an issue, since TDR is generating new UUIDs for every new Snapshot.

tlangs (approved):

src/main/resources/reference.conf
nmalfroy: Ah yeah, might as well to be safe

PR #1200, opened 2023-10-04
ID-815 New terms of service tracking table.

Ghost-in-a-Jar:
Ticket: [ID-815](https://broadworkbench.atlassian.net/browse/ID-815)

What:

This pr adds a new terms of service acceptance/rejection tracking table for users. It is an append only table that tracks userid, tos version, action (accept/reject), createdat.

Why:

For auditing purposes we need to be able to look back on when users accepted or rejected different versions of the tos.

How:

I added a new append-only table and moved tos information from the SamUser class to a new one. Included in the migration is a query that grabs all of the current tos acceptance data and puts it in the new table.

---

**PR checklist**

[ID-815]: https://broadworkbench.atlassian.net/browse/ID-815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20230929_add_tos_audit_table.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20230929_add_tos_audit_table.xml
@@ -10,25 +10,25 @@
       <column name="sam_user_id" type="VARCHAR">
         <constraints primaryKey="true" foreignKeyName="FK_TOS_USER_ID" referencedTableName="SAM_USER" referencedColumnNames="id" deleteCascade="true"/>
       </column>
-      <column name="terms_of_service_version" type="VARCHAR(40)">
+      <column name="version" type="VARCHAR(40)">
         <constraints primaryKey="true"/>
       </column>
       <column name="action" type="VARCHAR">
         <constraints primaryKey="true"/>
       </column>
-      <column name="updated_at" type="timestamptz"/>
+      <column name="created_at" type="timestamptz"/>
     </createTable>
 
     <sql stripComments="true">
       INSERT INTO sam_user_terms_of_service
       (sam_user_id,
-       terms_of_service_version,
+       version,
        action,
-       updated_at)
+       created_at)
       SELECT su.id                    as sam_user_id,
-             su.accepted_tos_version  as terms_of_service_version,
+             su.accepted_tos_version  as version,
              'ACCEPT'                 as action,
-       '1970-01-01 00:00:00-00' as updated_at
+       '1970-01-01 00:00:00-00' as created_at
       FROM sam_user su
       WHERE su.accepted_tos_version is not null;
     </sql>
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20230929_add_tos_audit_table.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20230929_add_tos_audit_table.xml
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+  <changeSet logicalFilePath="dummy" author="tgarwood" id="add_tos_audit_table">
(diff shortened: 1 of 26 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/DirectoryDAO.scala
```diff
@@ -74,7 +74,8 @@ trait DirectoryDAO {
   def setGoogleSubjectId(userId: WorkbenchUserId, googleSubjectId: GoogleSubjectId, samRequestContext: SamRequestContext): IO[Unit]
 
   def acceptTermsOfService(userId: WorkbenchUserId, tosVersion: String, samRequestContext: SamRequestContext): IO[Boolean]
-  def rejectTermsOfService(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Boolean]
+  def rejectTermsOfService(userId: WorkbenchUserId, tosVersion: String, samRequestContext: SamRequestContext): IO[Boolean]
+  def getUserTos(userId: WorkbenchUserId, tosVersion: String, samRequestContext: SamRequestContext): IO[Option[SamUserTos]]
```
Ghost-in-a-Jar: This fn is for the tos database lookup for a given user and tos version.

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/TosTable.scala
```diff
@@ -0,0 +1,39 @@
...
+final case class TosRecord(
+    samUserId: WorkbenchUserId,
+    version: String,
+    action: String,
+    createdAt: Instant
+)
+
+object TosTable extends SQLSyntaxSupportWithDefaultSamDB[TosRecord] {
+  val ACCEPT = "ACCEPT"
+  val REJECT = "REJECT"
```
Ghost-in-a-Jar, line 19: Maybe theres a better way/place to put this? Id love to have an enum but that can be a pain in scala.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20230929_add_tos_audit_table.xml
```diff
@@ -0,0 +1,40 @@
...
+      SELECT su.id                    as sam_user_id,
+             su.accepted_tos_version  as version,
+             'ACCEPT'                 as action,
+       '1970-01-01 00:00:00-00' as created_at
+      FROM sam_user su
+      WHERE su.accepted_tos_version is not null;
+    </sql>
+
+      <dropColumn tableName="sam_user">
+      <column name="accepted_tos_version"/>
```
tlangs, line 37: 😍

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/DirectoryDAO.scala
tlangs: Why do we need to query for a given user _and_ version?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -631,25 +629,37 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
+    }
+  }
+
+  override def getUserTos(userId: WorkbenchUserId, tosVersion: String, samRequestContext: SamRequestContext): IO[Option[SamUserTos]] =
+    readOnlyTransaction("getUserTos", samRequestContext) { implicit session =>
+      val tosTable = TosTable.syntax
+      val column = TosTable.column
+
+      val loadUserTosQuery =
+        samsql"""select ${tosTable.resultAll}
```
tlangs, line 656: We should `LIMIT 1` in this query if we're just gonna get the first anyways.

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/TosTable.scala
tlangs: This is probably fine, given scala's lack of enum types.

tlangs (changes requested):
This is awesome, and quite the hefty PR! Great job maintaining backwards compatibility. We'll want to thoroughly test this on Dev before pushing it out to prod.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/DirectoryDAO.scala
Ghost-in-a-Jar: I figured in case we need to roll back a tos version, also i think it cant hurt to be specific here

tlangs: I agree the method is probably useful, but I don't know if this is the right behavior for the current APIs. A `getUserLatestTos` would probably fit better.

If a user hasn't accepted then current ToS, what happens when `getUserTos` is called as a part of the `termsOfServiceDetails` API endpoint? We would want to return the `SamUserTos` details including that they've accepted the previous version, not an `Empty` response.

Shakespeared (approved):

tlangs (approved):

PR #1201, opened 2023-10-10
ID-792 Load Terms of Service docs from remote sources

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-792

What:

Make Sam load Terms of Service docs from a remote source.

Why:

We want to be able to update the Terra Terms of Service without making code changes. 

How:

Sam will now load the Terms of Service and Privacy Policy texts from a remote source (or optionally, its own classpath for testing) when it boots. It'll keep that text in-memory. To update the Terms of Service version, just host the documents at the configured `TOS_BASE_URL` and bump the `TOS_VERSION` in the deployment configs, then re-deploy Sam. No code changes required!

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamUserDirectives.scala
```diff
@@ -51,15 +51,16 @@ trait SamUserDirectives {
 
   def asAdminServiceUser: Directive0
 
+  val oldTermsOfServiceUrl = "app.terra.bio/#terms-of-service"
```
Ghost-in-a-Jar, line 54: maybe we want to make this configurable?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -63,34 +76,46 @@ class TosService(val directoryDao: DirectoryDAO, val tosConfig: TermsOfServiceCo
...
-    *   terms of service text
-    */
-  def getText(file: String, prettyTitle: String): String = {
+trait RemoteDocument {
+  def apply(uri: String)(implicit actorSystem: ActorSystem, executionContext: ExecutionContext): String
+}
+
+object RemoteDocument extends RemoteDocument with LazyLogging {
+  override def apply(uri: String)(implicit actorSystem: ActorSystem, executionContext: ExecutionContext): String =
+    if (uri.startsWith("classpath")) {
```
Ghost-in-a-Jar: i see why you did this but it feels to me like we should just have this RemoteDocument object only support remote stuff, maybe at least add a comment so as to not confuse a future dev

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -63,34 +76,46 @@ class TosService(val directoryDao: DirectoryDAO, val tosConfig: TermsOfServiceCo
...
+      getTextFromWeb(uri)
+    }
+
+  def getTextFromWeb(uri: String)(implicit actorSystem: ActorSystem, executionContext: ExecutionContext): String = {
+    val future = for {
+      response <- Http().singleRequest(Get(uri))
+      text <- Unmarshal(response).to[String]
+    } yield text
+
+    Await.result(future.withInfoLogMessage(s"Retrieved Terms of Service doc from $uri"), Duration.apply(10, TimeUnit.SECONDS))
```
Ghost-in-a-Jar, line 105: nit: parameterize this and default it to 10 there

Ghost-in-a-Jar (approved):
this looks good, i think we will want to manually test this in dev, maybe even testing switching to a new tos version just to be safe?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamUserDirectives.scala
tlangs: I don't think so. We're eventually going to get rid of this, but it still needs to be here for now since the current endpoints require this to be present in the body to accept the ToS.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
tlangs: I'm not sure the use of parameterizing the timeout here. If it can't load in 10 seconds, Sam should fail to start. This code is only run on startup, so there's not really a use for it while Sam is running day-to-day.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
tlangs: I've renamed this to be clearer.

sjkobori (approved):
What is going on with the `TestKit.shutdownActorSysyem`? Otherwise, looks good to me.

tlangs:
@sjkobori its needed to make sure tests' ActorSystems don't interfere with each other or hang around once the test is complete.

PR #1207, opened 2023-10-13
ID-813 Add TOS rolling acceptance window.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-813

What:

When a new Terms of Service is released, Sam will not immediately start telling the UI that users need to accept a new Terms of Service. For some period of time, longer than the longest possible user session (currently 24 hours), Sam will say that a user’s Terms of Service Acceptance Status is compliant, but the UI will force the user to accept the new Terms of Service on login. This means that users using Terra will not be interrupted as soon as a new Terms of Service is released, but will be required to accept it within a reasonable amount of time.

When the user logs in again, they’ll be forced to accept the new ToS in the UI. CLI users will be able to use Terra until the Acceptance Window expires, same as a Grace Period. But with a grace period, the user can choose to not accept the new ToS, but still use Terra. With the Rolling Acceptance Window, the user MUST accept the Terms of Service before continuing if prompted.

The Grace Period is something we want to get rid of. Right now, AoU wants it so that they can “review” the new ToS, but we’re working on not having that need to happen.

Why:

This pr adds two new config values: `rollingAcceptanceWindowExpiration` (a timestamp), and `rollingAcceptanceWindowPreviousTosVersion` (a string). When we switch to a new TOS version we will need to set the `rollingAcceptanceWindowExpiration` to be at least 24 hours after the switchover, and set `rollingAcceptanceWindowPreviousTosVersion` to be the previous version. If the timestamp is in the future and the user has accepted the version of tos that matches `rollingAcceptanceWindowPreviousTosVersion`, they will be permitted to use the system until the time reaches `rollingAcceptanceWindowExpiration`.

We want to provide a better user experience for when we switch tos versions. This should prevent someone from being kicked out mid session. When their session expires and they have to log in again the ui will force them to accept or decline the tos as usual.

How:

This pr significantly modifies the current TosService logic that decides whether a user can use the system. If they have accepted the previous version (denoted by a config value), and the acceptance window timestamp (another config value) is set to be in the future, then they will be permitted to use the system.

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/config/TermsOfServiceConfig.scala
```diff
@@ -1,12 +1,27 @@
...
+  *   The previous version of the ToS that the user must have accepted in order to be granted access to the system under the rolling window.
   */
 
-case class TermsOfServiceConfig(isTosEnabled: Boolean, isGracePeriodEnabled: Boolean, version: String, baseUrl: String)
+case class TermsOfServiceConfig(
+    isTosEnabled: Boolean,
+    isGracePeriodEnabled: Boolean,
+    version: String,
+    baseUrl: String,
+    rollingAcceptanceWindowExpiration: Instant,
```
tlangs: We might want to make this `Option[Instant]`. Seems like we could skip a bunch of logic if we want to release Sam without a new ToS by setting it to `None`.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/DirectoryDAO.scala
```diff
@@ -76,7 +76,7 @@ trait DirectoryDAO {
   def acceptTermsOfService(userId: WorkbenchUserId, tosVersion: String, samRequestContext: SamRequestContext): IO[Boolean]
   def rejectTermsOfService(userId: WorkbenchUserId, tosVersion: String, samRequestContext: SamRequestContext): IO[Boolean]
   def getUserTos(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Option[SamUserTos]]
-
+  def getUserTos(userId: WorkbenchUserId, tosVersion: String, samRequestContext: SamRequestContext): IO[Option[SamUserTos]]
```
tlangs: rename to `getUserTosVersion`

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/TermsOfServiceConfig.scala
```diff
@@ -1,12 +1,27 @@
...
   * @param version
   *   The latest version of the Terms of Service
-  * @param url
+  * @param baseUrl
   *   The url to the Terra Terms of Service. Used for validation and will be displayed to user in error messages
+  * @param rollingAcceptanceWindowExpiration
+  *   The expiration time for the rolling acceptance window. If the user has not accepted the new ToS by this time, they will be denied access to the system.
+  *   Must be a valid UTC datetime string in ISO 8601 format example: 2007-12-03T10:15:30.00Z
+  *
+  * @param rollingAcceptanceWindowPreviousTosVersion
```
tlangs: I think we can just name this `previousVersion`. It'll be helpful for grace periods, since we could then check to see if the user has accepted the previous version, not just _any_ version. Also, because bootstrapping, we should make this `Option`al. If someone is deploying Sam for the first time, there won't be a previous ToS version!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -50,32 +51,50 @@ class TosService(val directoryDao: DirectoryDAO, val tosConfig: TermsOfServiceCo
...
     val userIsServiceAccount = StandardSamUserDirectives.SAdomain.matches(user.email.value) // Service Account users do not need to accept ToS
-    val userIsPermitted = userTos.exists { tos =>
+
+    if (userIsServiceAccount) {
+      return true
+    }
+    if (userHasRejectedLatestTosVersion(userTos)) {
+      return false
+    }
+    userTos.exists { tos =>
```
tlangs, line 75: I like this short-circuiting. Much easier to read!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -50,32 +51,50 @@ class TosService(val directoryDao: DirectoryDAO, val tosConfig: TermsOfServiceCo
...
       val userHasAcceptedLatestVersion = userHasAcceptedLatestTosVersion(Option(tos))
       val userCanUseSystemUnderGracePeriod = tosConfig.isGracePeriodEnabled && tos.action == TosTable.ACCEPT
       val tosDisabled = !tosConfig.isTosEnabled
```
tlangs: Could probably pull this up to be short-circuited as well.

tlangs (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/TermsOfServiceConfig.scala
```diff
@@ -1,12 +1,27 @@
...
 
+import java.time.Instant
+
 /** Terms of Service configuration.
   * @param isGracePeriodEnabled
   *   Set to true if the grace period for ToS acceptance is active
   * @param version
   *   The latest version of the Terms of Service
-  * @param url
+  * @param baseUrl
```
tlangs, line 10: You may need to rebase. This is old code

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -50,32 +51,56 @@ class TosService(val directoryDao: DirectoryDAO, val tosConfig: TermsOfServiceCo
...
       tos.version.contains(tosConfig.version) && tos.action == TosTable.ACCEPT
     }
+
+  private def userHasRejectedLatestTosVersion(userTos: Option[SamUserTos]): Boolean =
+    userTos.exists { tos =>
+      tos.version.contains(tosConfig.version) && tos.action == TosTable.REJECT
+    }
+
+  private def userHasAcceptedPreviousTosVersion(previousUserTos: Option[SamUserTos]): Boolean =
+    previousUserTos.exists(tos => tos.action == TosTable.ACCEPT)
```
tlangs, line 103: I think the case where there is no previous ToS needs to be included here, and if there isn't a previoius ToS, this function returns `true`

tlangs: Should probably add a test for this as well.

tlangs (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/TermsOfServiceConfig.scala
Ghost-in-a-Jar: I think you just missed this little spot 😅

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
Ghost-in-a-Jar: This actually does return false in that case and there is a test 😄 https://github.com/broadinstitute/sam/pull/1207/files#diff-3a7b95de48385b66cf0a3b94e7e625e53bcce9efc68ce85a67cfe479f2989e1dR153

Ghost-in-a-Jar: (exists returns false if the option is none)

tlangs: Doh. You're right.

tlangs (approved):

sjkobori (approved):
Looks good to me 👍

PR #1208, opened 2023-10-13
ID-690 Get User(s) Endpoint v2

tlangs:
Ticket: [<Link to Jira ticket>](https://broadworkbench.atlassian.net/browse/ID-690)

What:

Some new `users/v2` endpoints to get users, via oneself and via admin powers, all in the same API!

Why:

We want to get rid of some of the separate `/admin` endpoints and gate behavior on standard CRUD endpoints.

How:
- Added a `SamUserResponse` object, which replaces `enabled` with `allowed`.
- Added a `SamUserAllowances` object, which details the reasons a user might be allowed/unallowed.
- Added a `/api/users/v2/self` endpoint. This just gets the SamUserResponse object of the calling user
- Added a `/api/users/v2/{sam_user_id}`. A regular user can only get themselves via their ID. An Admin user can get any SamUserResponse object.
- Added a `/api/users/v2/self/allowed` endpoint to allow a user to interrogate _why_ they might be allowed/unallowed
- Added a `/api/users/v2/{sam_user_id}/allowed` endpoint to allow an admin to interrogate _why_ another user might be allowed/unallowed
- I've also refactored the model code a little bit.

---

**PR checklist**


src/main/resources/swagger/api-docs.yaml
```diff
@@ -3799,6 +3847,44 @@ components:
...
       description: specification of a User
+    SamUserResponse:
+      type: object
+      required:
+        - id
+        - email
+        - enabled
+        - createdAt
+        - updatedAt
+      properties:
```
sjkobori: I think enabled did not get updated to allowed here

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
```diff
@@ -0,0 +1,120 @@
...
+  def userRoutesV2(samUser: SamUser, samRequestContext: SamRequestContext): Route =
+    pathPrefix("users") {
+      pathPrefix("v2") {
+        pathPrefix("self") {
+          // api/users/v2/self
+          pathEndOrSingleSlash {
+            getSamUserResponse(samUser, samRequestContext)
+          } ~
+            // api/users/v2/self/allowed
+            pathPrefix("allowed") {
```
sjkobori: I appreciate these comments, akka formatting is not easy to read...

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
```diff
@@ -1,115 +1,156 @@
...
+  val defaultUser: SamUser = Generator.genWorkbenchUserGoogle.sample.get
+  val otherUser: SamUser = Generator.genWorkbenchUserGoogle.sample.get
+  val adminGroupEmail: WorkbenchEmail = Generator.genFirecloudEmail.sample.get
+  val allUsersGroup: BasicWorkbenchGroup = BasicWorkbenchGroup(CloudExtensions.allUsersGroupName, Set(), WorkbenchEmail("all_users@fake.com"))
+
+  "GET /api/users/v2/self" should "get the user object of the requesting user" in {
+    // Arrange
+    val samRoutes = new MockSamRoutesBuilder(allUsersGroup)
+      .withEnabledUser(defaultUser) // "persisted/enabled" user we will check the status of
+      .withAllowedUser(defaultUser) // "allowed" user we will check the status of
```
sjkobori, line 28: Since these tests are on the API routing layer, they should not have a distinction between and allowed and enabled user. I think we should defer to using `withAllowedUser`

sjkobori: If not possible otherwise, this may be a place where we need to mock out the only routes for backwards compatibility.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
```diff
@@ -1,115 +1,156 @@
...
-    }
-    val (user, samDep, routes) = createTestUser(tosAccepted = true)
-    Get("/register/user/v2/self/info") ~> routes.route ~> check {
+  "GET /api/users/v2/{sam_user_id}" should "return the regular user if they're getting themselves" in {
+    // Arrange
+    val samRoutes = new MockSamRoutesBuilder(allUsersGroup)
+      .withEnabledUsers(Seq(defaultUser, otherUser))
+      .withAllowedUsers(Seq(defaultUser, otherUser))
+      .callAsNonAdminUser()
+      .build
```
sjkobori, line 45: I think now that we have logic dependant on if the calling user's idenitity is an admin or not, we are going to need to pass args into `callAsNonAdminUser` since it would not be true for any arbitrary non admin, it is only true of the nonAdmin who is calling is in fact the user that is being gotten

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/OldUserRoutes.scala
```diff
@@ -151,33 +150,4 @@ trait UserRoutes extends SamUserDirectives with SamRequestContextDirectives {
         }
       }
     }
-
-  def apiUserRoutes(samUser: SamUser, samRequestContext: SamRequestContext): server.Route = pathPrefix("users") {
```
Ghost-in-a-Jar, line 155: are we just totally removing these? should there be a corresponding removal in swagger?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1.scala
```diff
@@ -0,0 +1,45 @@
...
+      } ~
+        pathPrefix("invite") {
+          post {
+            path(Segment) { inviteeEmail =>
+              complete {
+                userService
+                  .inviteUser(WorkbenchEmail(inviteeEmail.trim), samRequestContext)
+                  .map(userStatus => StatusCodes.Created -> userStatus)
+              }
+            }
```
Ghost-in-a-Jar, line 39: nvm i see they just moved here

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
tlangs: I wasn't sure what to do here. I think it is valuable, because a user being enabled doesn't necessarily mean they can use Terra. I think there's value in being able to test and respond accordingly. I do agree that this is clunky, having to specify _both_ things. Maybe all usages of `withEnabledUser(s)` should be migrated to a new `withReadyUser(s)` call, which would do both enabling and allowing, leaving the other two methods there for testing the specifics if need be.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
tlangs: Since we're testing the logic based on the calling user, not the requested user, I think the functionality is still valid. It is kinda un-intuitive though, making _all_ calls to the routes be as an Admin user. I think it could be refactored, but I don't think its necessary to do in this PR

src/main/resources/swagger/api-docs.yaml
tlangs: Done!

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
sjkobori: I think in general, the api mocking builder should only care about mocking things from the api layer's perspective. So the distinction between allowed and enabled isnt a detail which the api is concerned about, so I dont think the mock should have any logic on it in this layer. In the service layer, it is important to make that distinction, but I dont believe so in the api layer.

Ghost-in-a-Jar (approved):
This looks good to me, but just in case we should do some manual testing in dev/staging before the prod deploy of this stuff. We def don't want non-admin users to be able to grab each-other's info.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
sjkobori: I guess what I am trying to get as is: how can we know if the calling non admin user cannot make requests for any other enabled/allowed user? It seems like there would not be any distinction in this test

tlangs:
@Ghost-in-a-Jar Absolutely. I'll test this on a BEE before merging.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
tlangs: Good point. I'll change things.

tlangs: Actually, I'm going to take it back. As I was rewriting, I realized that the difference between `enabled` and `allowed` _is_ significant to the api layer. They mean different things. `allowed` for the v2 endpoints, means the user is able to use the system, which means they're both `enabled` in the db _AND_ have accepted the Terms of Service. These things are spelled out in the API responses, so we need to be able to test them. The test code as written doesn't actually mandate that the test "know" that `allowed` means `enabled` and ToS accepted, but the responses should look right.

tlangs:
@Ghost-in-a-Jar 
```
# Get another user as an admin
$ curl -X 'GET'   'https://sam.tlangs-user-endpoints.bee.envs-terra.bio/api/users/v2/26976461068667bfb9222' \
   -H 'accept: application/json' -H "Authorization: Bearer $(gcloud auth print-access-token tlangs@test.firecloud.org)"
> {"allowed":true,"azureB2CId":"d7d55ae5-07c7-41c0-8349-79305b3fbc52","createdAt":"2023-10-18T16:21:46.873051Z","email":"tlangs@test.firecloud.org","googleSubjectId":"112547541637891094041","id":"26976461068667bfb9222","registeredAt":"2023-10-18T16:21:46.872268Z","updatedAt":"2023-10-18T16:21:49.929332Z"}

# Get another user as a non-admin
$ curl -X 'GET'   'https://sam.tlangs-user-endpoints.bee.envs-terra.bio/api/users/v2/2697646592417bef9cd21'   -H 'accept: application/json' -H "Authorization: Bearer $(gcloud auth print-access-token tlangs@test.firecloud.org)"
> {"causes":[],"message":"You must be an admin.","source":"sam","stackTrace":[],"statusCode":404}
```

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
tlangs: Added the `samUser` arg to `callAs(Non)AdminUser` methods

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/OldUserRoutes.scala
```diff
@@ -33,7 +32,7 @@ trait UserRoutes extends SamUserDirectives with SamRequestContextDirectives {
     })
   }
 
-  def userRoutes(samRequestContext: SamRequestContext): server.Route =
+  def oldUserRoutes(samRequestContext: SamRequestContext): server.Route =
```
gpolumbo-broad, line 35: We should mark this as deprecated

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV1.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.broadinstitute.dsde.workbench.model._
+import org.broadinstitute.dsde.workbench.sam.model.api.SamJsonSupport._
+import org.broadinstitute.dsde.workbench.sam.model.api.SamUser
+import org.broadinstitute.dsde.workbench.sam.service.UserService
+import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
+
+trait UserRoutesV1 extends SamUserDirectives with SamRequestContextDirectives {
+  val userService: UserService
+
+  def userRoutesV1(samUser: SamUser, samRequestContext: SamRequestContext): server.Route = pathPrefix("users") {
```
gpolumbo-broad, line 16: Should have happened before, but we should mark this as deprecated

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
```diff
@@ -0,0 +1,126 @@
...
+  // These routes are wrapped in `withUserAllowInactive` because a user should be able to get info about themselves
+  // Routes that need the user to be active should be wrapped in a directive, such as `withActiveUser`, to ensure
+  // that the user is allowed to use the system.
+  def userRoutesV2(samRequestContextWithoutUser: SamRequestContext): Route =
+    withUserAllowInactive(samRequestContextWithoutUser) { samUser: SamUser =>
+      val samRequestContext = samRequestContextWithoutUser.copy(samUser = Some(samUser))
+      pathPrefix("users") {
+        pathPrefix("v2") {
+          pathPrefix("self") {
+            // api/users/v2/self
```
gpolumbo-broad, line 41: Hahaha ❤️ I was literally just doing this 5 minutes ago in my own PR so I could understand what the hell the routes are.  This syntax is brutal.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/SamUserAllowances.scala
```diff
@@ -0,0 +1,20 @@
...
+import spray.json.DefaultJsonProtocol._
+import spray.json.RootJsonFormat
+
+object SamUserAllowances {
+  implicit val SamUserAllowedResponseFormat: RootJsonFormat[SamUserAllowances] = jsonFormat2(SamUserAllowances.apply)
+
+  def apply(allowed: Boolean, enabled: Boolean, termsOfService: Boolean): SamUserAllowances =
+    SamUserAllowances(allowed, Map("enabled" -> enabled, "termsOfService" -> termsOfService))
+}
+final case class SamUserAllowances(
```
gpolumbo-broad, line 12: or maybe `SamUserAllowance`?

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
gpolumbo-broad: @tlangs you're 100% right that needing to construct the routes _WITH_ the one and only 1 user that we will be calling those routes with is indeed gross.  It's a function of how the routes are written and where we do the permissions checks (pro-tip:  don't sprinkle business logic like authorization checks in your routing logic).  This is definitely a refactoring we should do at some point, and for now, the `callAsFooUser` was the best solution I could come up with that read somewhat nicely and got us around the yucky.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
gpolumbo-broad: It looks like this has been modified, but IIRC I was trying to have these things call each other so that if I wanted to say `.withEnabledUser` it would take care of creating the user automatically.  So carrying that logic forward, I would expect `.withAllowedUser` to make the user enabled as well, since for a user to be Allowed it must also be enabled.

gpolumbo-broad (approved):

sjkobori (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/OldUserRoutes.scala
tlangs: I think we should only mark as deprecated once all the functionality is implemented in the V2 routes

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
tlangs: Its sooooo bad

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/SamUserAllowances.scala
tlangs: I like it being plural, since it details out all the reasons someone might not be allowed to use the system.

PR #1211, opened 2023-10-16
[DR-3245] Include auth domain actions in TDR snapshot roles

samanehsan:
**Ticket:** https://broadworkbench.atlassian.net/browse/DR-3245

**What:** Enable TDR users to read and/or update the auth domain of a snapshot depending on their role.

**Why:** TDR is adding the ability for users to add an auth domain to their snapshot (which will get registered with a `group-constraint` policy in the Terra Policy Service) in order to enforce who is allowed to access the data when it gets exported to a Terra workspace.

**How**: Updated reference.conf `datasnapshot` config

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1263,23 +1263,29 @@ resourceTypes = {
       view_journal = {
         description = "View datasnapshot journal entries"
       }
+      read_auth_domain = {
+        description = "view the auth domain of the snapshot"
+      }
+      update_auth_domain = {
+        description = "update the groups in the auth domain of the snapshot"
+      }
     }
     ownerRoleName = "steward"
     roles = {
       steward = {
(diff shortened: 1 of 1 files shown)


Ghost-in-a-Jar (approved):

src/main/resources/reference.conf
```diff
@@ -1263,23 +1263,29 @@ resourceTypes = {
       view_journal = {
         description = "View datasnapshot journal entries"
       }
+      read_auth_domain = {
+        description = "view the auth domain of the snapshot"
+      }
+      update_auth_domain = {
+        description = "update the groups in the auth domain of the snapshot"
```
s-rubenstein: Is there potentially multiple auth domains for a snapshot or is there only one? If there's only one, I think this comment could be better described as "Update the auth domain of the snapshot" rather than mentioning groups. Just because auth domains are by definition defined by groups.

s-rubenstein (approved):

src/main/resources/reference.conf
samanehsan: Each resource only has one auth domain: https://sam.dsde-dev.broadinstitute.org/#/Resources/getAuthDomainV2

PR #1219, opened 2023-10-20
ID-880 Sam User Attributes

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-880

What:

Sam is the future of user information and preferences. The team has already decided that we want to migrate the data in Thurloe into Sam. 

This will be the start of that process. We’ll create a new table, sam_user_attributes, which will track the user’s marketing comms choice. It will be its own column with a true/false value, meaning there’s no ambiguity to if a user has consented or not. 

Why:

Thurloe has proven to be very difficult to work with. For something as important as GDPR compliance, we need to implement things in a straightforward and developer-friendly fashion as to minimize the risk of being out-of-compliance. 

How:

Implemented the storage of `marketingConsent` from API to DB. Tests included!

---

**PR checklist**

### DIFF
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20231019_sam_user_attributes_table.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20231019_sam_user_attributes_table.xml
@@ -11,7 +11,7 @@
         <constraints primaryKey="true" foreignKeyName="FK_SUA_USER_ID" referencedTableName="SAM_USER" referencedColumnNames="id" deleteCascade="true"/>
       </column>
       <column name="marketing_consent" type="BOOLEAN">
-        <constraints primaryKey="false"/>
+        <constraints primaryKey="false" nullable="false"/>
       </column>
       <column name="updated_at" type="timestamptz"/>
     </createTable>
@@ -25,7 +25,6 @@
              true              as marketing_consent,
              current_timestamp as updated_at
       FROM sam_user su
-      WHERE su.enabled = true;
     </sql>
 
   </changeSet>
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20231019_sam_user_attributes_table.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20231019_sam_user_attributes_table.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+  <changeSet logicalFilePath="dummy" author="tlangs" id="add_sam_user_attributes_table">
+    <createTable tableName="SAM_USER_ATTRIBUTES">
+      <column name="sam_user_id" type="VARCHAR">
+        <constraints primaryKey="true" foreignKeyName="FK_SUA_USER_ID" referencedTableName="SAM_USER" referencedColumnNames="id" deleteCascade="true"/>
+      </column>
+      <column name="marketing_consent" type="BOOLEAN">
+        <constraints primaryKey="false"/>
+      </column>
+      <column name="updated_at" type="timestamptz"/>
+    </createTable>
+
+    <sql stripComments="true">
+      INSERT INTO sam_user_attributes
+      (sam_user_id,
+       marketing_consent,
+       updated_at)
+      SELECT su.id             as sam_user_id,
+             true              as marketing_consent,
+             current_timestamp as updated_at
+      FROM sam_user su
+      WHERE su.enabled = true;
+    </sql>
+
+  </changeSet>
+</databaseChangeLog>
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.workbench.sam.api
 
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+import org.broadinstitute.dsde.workbench.sam.model.api.SamJsonSupport._
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.StatusCodes.{NotFound, OK}
 import akka.http.scaladsl.server.Directives._
@@ -8,7 +9,7 @@ import akka.http.scaladsl.server.{Directive0, ExceptionHandler, Route}
 import cats.effect.IO
 import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.sam.model.api.SamUserResponse._
-import org.broadinstitute.dsde.workbench.sam.model.api.{SamUser, SamUserAttributesRequest, SamUserResponse}
+import org.broadinstitute.dsde.workbench.sam.model.api.{SamUser, SamUserAttributesRequest, SamUserRegistrationRequest, SamUserResponse}
 import org.broadinstitute.dsde.workbench.sam.service.UserService
 import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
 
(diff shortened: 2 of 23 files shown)


tlangs:
I'll test this on a BEE to make sure the db migration works.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20231019_sam_user_attributes_table.xml
```diff
@@ -0,0 +1,31 @@
...
+
+  <changeSet logicalFilePath="dummy" author="tlangs" id="add_sam_user_attributes_table">
+    <createTable tableName="SAM_USER_ATTRIBUTES">
+      <column name="sam_user_id" type="VARCHAR">
+        <constraints primaryKey="true" foreignKeyName="FK_SUA_USER_ID" referencedTableName="SAM_USER" referencedColumnNames="id" deleteCascade="true"/>
+      </column>
+      <column name="marketing_consent" type="BOOLEAN">
+        <constraints primaryKey="false" nullable="false"/>
+      </column>
+      <column name="updated_at" type="timestamptz"/>
```
Ghost-in-a-Jar, line 16: nit: maybe want a created at too? also do we want any auditability of this table like the history of consenting/not consenting to marketing?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
```diff
@@ -123,4 +130,19 @@ trait UserRoutesV2 extends SamUserDirectives with SamRequestContextDirectives {
...
 
+  private def getSamUserAttributes(samUser: SamUser, samRequestContext: SamRequestContext): Route =
+    get {
+      complete {
+        userService.getUserAttributes(samUser.id, samRequestContext).map(response => (if (response.isDefined) OK else NotFound) -> response)
+      }
+    }
+
+  private def patchSamUserAttributes(samUser: SamUser, samRequestContext: SamRequestContext): Route =
+    patch {
```
Ghost-in-a-Jar, line 151: this might be just me but i generally prefer put so that you can keep things idempotent

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
```diff
@@ -154,4 +154,41 @@ class UserRoutesV2Spec extends AnyFlatSpec with Matchers with ScalatestRouteTest
...
+      .withUserAttributes(defaultUser, userAttributes)
+      .callAsNonAdminUser()
+      .build
+
+    // Act and Assert
+    Patch(s"/api/users/v2/self/attributes", userAttributesRequest) ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.OK
+      responseAs[SamUserAttributes] should be(userAttributes)
+    }
+  }
```
Ghost-in-a-Jar, line 230: Should we have one where requesting another user's attributes is denied or is that already covered?

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20231019_sam_user_attributes_table.xml
tlangs: I don't think we need created at, since rows are actually being created when a new user is created. As for auditing, its not in the requirements, so I don't think the slight extra complexity is needed.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
tlangs: This is PATCH because this endpoint will allow the caller to provide values only for the fields they want to update. A PUT would require the caller to specify all fields to store on the server, even if they're not changing.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
tlangs: That's handled in `CreateUserSpec`. Since this is an API layer test, its really only testing to make sure that the ser/der works, and probably shouldn't be testing business logic.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
Ghost-in-a-Jar: Right, but that is sort of nice from our perspective, and there's only one field to really include on this call so far. I think this is fine just wanted to call out puts.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20231019_sam_user_attributes_table.xml
Ghost-in-a-Jar: sounds good!

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
tlangs: I'm hoping that we start getting thurloe data into Sam soon so that this will be fleshed out!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -29,19 +36,28 @@ class UserService(val directoryDAO: DirectoryDAO, val cloudExtensions: CloudExte
...
     openTelemetry.time("api.v1.user.create.time", API_TIMING_DURATION_BUCKET) {
       // Validate the values set on the possible new user, short circuit if there's a problem
       val validationErrors = validateUser(possibleNewUser)
       if (validationErrors.nonEmpty) {
         return IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "invalid user", validationErrors.get)))
       }
+      val registrationRequestErrors = registrationRequest.flatMap(_.validateForNewUser)
+      if (registrationRequestErrors.nonEmpty) {
+        return IO.raiseError(
+          new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "invalid registration request", registrationRequestErrors.get))
```
Ghost-in-a-Jar, line 52: does the ui pr have to go out first for this, so that it has the required fields sent with the registration request?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpecs/CreateUserSpec.scala
```diff
@@ -277,6 +309,20 @@ class CreateUserSpec extends UserServiceTestTraits {
...
+      val cloudExtensions = MockCloudExtensionsBuilder(allUsersGroup).build
+      val userService = new UserService(directoryDAO, cloudExtensions, Seq.empty, defaultTosService)
+
+      // Act
+      val invitedUserStatus = runAndWait(userService.inviteUser(invitedGoogleUser.email, samRequestContext))
+
+      // Assert
+      verify(directoryDAO).setUserAttributes(SamUserAttributes(invitedUserStatus.userSubjectId, marketingConsent = false), samRequestContext)
+    }
+
```
Ghost-in-a-Jar, line 325: Add a test where a user tries to grab another user's attributes?

Ghost-in-a-Jar (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpecs/CreateUserSpec.scala
tlangs: There's no api endpoint to specify a user id of the user to get attributes of, so that's not possible.

Ghost-in-a-Jar: Oh right duh

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
tlangs: This is tied into a new endpoint which isn't being hit yet. If the request object doesn't exist, like it currently doesn't, these errors won't exist.

Ghost-in-a-Jar: Right ok and if the consent boolean isnt there itll default to false is what im reading?

tlangs: yup!

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
sjkobori: I think in the case of a `Not Found` we should be handling that n the api layer since currently that is where the not found logic is. I think in an ideal world we would have some sort of adapter between the routes and the services which would map all the objects returned from the services layer into their corresponding response codes, but currently that logic is scattered between the service and routes layers

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockDirectoryDAO.scala
```diff
@@ -387,4 +371,18 @@ class MockDirectoryDAO(val groups: mutable.Map[WorkbenchGroupIdentity, Workbench
...
+  override def getUserAttributes(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[Option[SamUserAttributes]] =
+    loadUser(userId, samRequestContext).map {
+      case None => None
+      case Some(_) =>
+        userAttributes.get(userId)
+    }
+
+  override def setUserAttributes(samUserAttributes: SamUserAttributes, samRequestContext: SamRequestContext): IO[Unit] =
+    loadUser(samUserAttributes.userId, samRequestContext).map {
+      case None => ()
```
sjkobori: Should we send back a ResponseCode here, instead of a Unit. In the case where you cannot load the user, should we propagate that message back to the caller?

sjkobori (approved):
Left a couple comments, but looks good overall 👍

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/MockDirectoryDAO.scala
tlangs: The db layer shouldn't have any idea about response codes. I'll make it throw an exception instead.

PR #1226, opened 2023-10-25
ID-891 Optimize UserResourcesQuery

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-891

What/why:

This query is accounting for the majority of the cpu load in sam, for some reason postgres is doing a full table scan of `sam_effective_resource_policy` every single time the query is ran, despite there being indexes on the table. This modification forces the join of `sam_effective_resource_policy` to filter down to only resource_ids of the type we are looking for.

How:

This modification to the query should be functionally equivalent, but forces the postgres query planner to do an index scan instead of a full table scan:

before
->  Parallel Seq Scan on sam_effective_resource_policy effpol  (cost=0.00..14559.75 rows=537375 width=24)
after
->  Index Scan using idx_erp_resource_id on sam_effective_resource_policy effpol  (cost=0.43..0.63 rows=6 width=24)

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
@@ -1781,6 +1781,14 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
           from ${ResourceTable as resource}
           join ${ResourceTypeTable as resourceType} on ${resource.resourceTypeId} = ${resourceType.id}
           join ${EffectiveResourcePolicyTable as effPol} on ${resource.id} = ${effPol.resourceId}
+          -- https://broadworkbench.atlassian.net/browse/ID-891
+          -- We had to optimize this query, the following join condition shouldn't make a difference on the results
+          -- but it forces the postgres query planner to do an index scan instead of a full table scan of EffectiveResourcePolicyTable
+            and ${effPol.resourceId} in (
+              select ${resource.id} from ${ResourceTable as resource}
+              join ${ResourceTypeTable as resourceType} on ${resourceType.id} = ${resource.resourceTypeId}
+              where ${resourceType.name} = ${resourceTypeName}
+            )
           join ${PolicyTable as policy} on ${effPol.sourcePolicyId} = ${policy.id}
           $additionalJoins
           where ${resourceType.name} = ${resourceTypeName}


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1781,6 +1781,11 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
           from ${ResourceTable as resource}
           join ${ResourceTypeTable as resourceType} on ${resource.resourceTypeId} = ${resourceType.id}
           join ${EffectiveResourcePolicyTable as effPol} on ${resource.id} = ${effPol.resourceId}
+            and ${effPol.resourceId} in (
+              select ${resource.id} from ${ResourceTable as resource}
+              join ${ResourceTypeTable as resourceType} on ${resourceType.id} = ${resource.resourceTypeId}
+              where ${resourceType.name} = ${resourceTypeName}
```
dvoet, line 1790: if you want to go a little further
```suggestion
              where ${resource.resourceTypeId} = ${resourceTypePKsByName(resourceTypeName)}
```
also could remove the ResourceTypeTable join and replace
```
where ${resourceType.name} = ${resourceTypeName}
```
with
```
where ${resource.resourceTypeId} = ${resourceTypePKsByName(resourceTypeName)}
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1781,6 +1781,11 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
           from ${ResourceTable as resource}
           join ${ResourceTypeTable as resourceType} on ${resource.resourceTypeId} = ${resourceType.id}
           join ${EffectiveResourcePolicyTable as effPol} on ${resource.id} = ${effPol.resourceId}
+            and ${effPol.resourceId} in (
```
dvoet, line 1787: drop a comment about why this is here

dvoet (approved):
made a suggestion for now or for the future

tlangs (commented):
Lets test on Dev and make sure that the query is returning everything and behaving as we expect. Once that's done, to prod we go!

Ghost-in-a-Jar:
@tlangs I ran:
```
(select * from optimized EXCEPT select * from original)
UNION ALL
(select * from original EXCEPT select * from optimized)
```
and that returned nothing so they are returning the same rows

tlangs (approved):

PR #1230, opened 2023-10-26
ID-896 Isolate Misbehaving Query

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-896

What:

Isolate the misbehaving query to the read replica. All other read/writes go to the main db.

Why:

Read replicas have lag. 

How:

Introduce a new `DbReference` to isolate misbehaving queries.

---

**PR checklist**

### DIFF
--- a/src/main/resources/sam.conf
+++ b/src/main/resources/sam.conf
@@ -163,10 +163,10 @@ db {
     password = ${?POSTGRES_PASSWORD}
   }
 
-  sam_bad_query {
-    poolName = "sam_background"
-    poolInitialSize = 5
-    poolMaxSize = 5
+  sam_read_replica {
+    poolName = "sam_read_replica"
+    poolInitialSize = 8
+    poolMaxSize = 8
     poolConnectionTimeoutMillis = 5000
     driver = "org.postgresql.Driver"
     url = ${?POSTGRES_READ_URL}
--- a/src/main/resources/sam.conf
+++ b/src/main/resources/sam.conf
@@ -130,7 +130,7 @@ db {
     poolMaxSize = 8
     poolConnectionTimeoutMillis = 5000
     driver = "org.postgresql.Driver"
-    url = ${?POSTGRES_READ_URL}
+    url = ${?POSTGRES_WRITE_URL}
     user = ${?POSTGRES_USERNAME}
     password = ${?POSTGRES_PASSWORD}
   }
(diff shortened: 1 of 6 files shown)


Shakespeared (approved):

jack-r-warren (approved):
"Looks good to me" - Greg lol

src/main/resources/sam.conf
```diff
@@ -162,6 +162,17 @@ db {
     user = ${?POSTGRES_USERNAME}
     password = ${?POSTGRES_PASSWORD}
   }
+
+  sam_bad_query {
+    poolName = "sam_background"
```
dvoet: different pool name?

src/main/resources/sam.conf
```diff
@@ -162,6 +162,17 @@ db {
...
+    poolInitialSize = 5
+    poolMaxSize = 5
```
dvoet: why not 8?

gpolumbo-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1353,7 +1357,7 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
       userId: WorkbenchUserId,
       samRequestContext: SamRequestContext
   ): IO[Iterable[ResourceIdWithRolesAndActions]] =
-    readOnlyTransaction("listUserResourcesWithRolesAndActions", samRequestContext) { implicit session =>
+    readOnlyTransactionReadReplica("listUserResourcesWithRolesAndActions", samRequestContext) { implicit session =>
```
dvoet, line 1360: do you want to do this for all usages of UserResourcesQuery?

tlangs: Not sure. The query that's been blowing up the DB seems to be the one with `as inherited` in it. Being as surgical as possible seems like the right move here.

src/main/resources/sam.conf
tlangs: because copy-paste errors

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
gpolumbo-broad: I vote that given where we are with this, we isolate this one major offender first.  80/20 rule.  We might be able to make things a lot better at some point, but for now we just need to treat the most severe problem.

PR #1233, opened 2023-10-30
Add to pact readme.

Ghost-in-a-Jar:
Just adding a bit to the readme for pact

---

**PR checklist**

- [ ] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've filled out the [Security Risk Assessment](https://sdarq.dsp-appsec.broadinstitute.org/jira-ticket-risk-assesment) (requires Broad Internal network access) and attached the result to the JIRA ticket

PR #1234, opened 2023-10-30
AJ-1234: Use helmchart name for pact identifier.

jladieu:
This is prefactoring work for [AJ-1234](https://broadworkbench.atlassian.net/browse/AJ-1234)

This pairs with https://github.com/DataBiosphere/terra-workspace-data-service/pull/384, where `wds` also renames its reference from `sam-provider` to just `sam`.

* See [thread](https://broadinstitute.slack.com/archives/C043YJ40719/p1698068076950019) discussing the `-consumer` and `-provider` suffix anti-pattern.
* See [thread](https://broadinstitute.slack.com/archives/C043YJ40719/p1698241095600099) discussing the recommendation to use the helm chart name.
* See [thread](https://broadinstitute.slack.com/archives/C043YJ40719/p1698262189799389) discussing doing this for `sam`.

[AJ-1234]: https://broadworkbench.atlassian.net/browse/AJ-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/verify_consumer_pacts.yml
+++ b/.github/workflows/verify_consumer_pacts.yml
@@ -310,4 +310,4 @@ jobs:
           repo: broadinstitute/terra-github-workflows
           ref: refs/heads/main
           token: ${{ secrets.BROADBOT_TOKEN }} # github token for access to kick off a job in the private repo
-          inputs: '{ "pacticipant": "sam-provider", "version": "${{ needs.verify-consumer-pact.outputs.provider-sha }}" }'
+          inputs: '{ "pacticipant": "sam", "version": "${{ needs.verify-consumer-pact.outputs.provider-sha }}" }'
--- a/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
+++ b/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala
@@ -286,7 +286,7 @@ class SamProviderSpec
       .getOrElse(NoOpFilter)
 
   val provider: ProviderInfoBuilder = ProviderInfoBuilder(
-    name = "sam-provider",
+    name = "sam",
     pactSource = PactSource
       .PactBrokerWithSelectors(
         brokerUrl = pactBrokerUrl


.github/workflows/verify_consumer_pacts.yml
```diff
@@ -307,4 +307,4 @@ jobs:
           repo: broadinstitute/terra-github-workflows
           ref: refs/heads/main
           token: ${{ secrets.BROADBOT_TOKEN }} # github token for access to kick off a job in the private repo
-          inputs: '{ "pacticipant": "sam-provider", "version": "${{ needs.verify-consumer-pact.outputs.provider-sha }}" }'
+          inputs: '{ "pacticipant": "sam", "version": "${{ needs.verify-consumer-pact.outputs.provider-sha }}" }'
```
Ghost-in-a-Jar, line 313: sam also can be a consumer, will this cause any name collisions/issues with that?

Ghost-in-a-Jar (approved):

.github/workflows/verify_consumer_pacts.yml
jladieu: I think ultimately what we want is for sam to self-identify itself as `sam` when acting in both the provider _and_ consumer role.  I think this will let the pact-broker do some dependency graph magic which is currently stymied by having the consumer/provider roles named divergently.

So I think there may be some followup work to rename the consumer side of things as well, but I wanted to try to keep the scope of this change as tight as possible.

Ghost-in-a-Jar: ok nice that makes sense thanks! btw side-note the verify-consumer-pact action was failing on our pr's and it looks like this change fixed it...?

mspector (approved):

.github/workflows/verify_consumer_pacts.yml
jladieu: Oh dang, I didn't realize these were failing previously.

And, I hate to disappoint... but I think that's actually a false 'pass' on this PR...

The first time a pact verification runs for a given provider (in this case, the newly renamed `sam`) it treats the contracts as being in a 'pending' state, and quietly passes the tests (even if they're failing).

You need to go to the [logs](https://github.com/broadinstitute/sam/actions/runs/6694733649/job/18229693055?pr=1234) to see that the tests are still failing with the same looking stack traces as before this PR.

jladieu: Took a quick look at the failure @Ghost-in-a-Jar and I think it's because the `verify_consumer_pacts.yml` needs to somehow run `/bin/bash -c env/test.env` as part of setting up its environment before running sbt test on the `SamProviderSpec`.

I tried a few different incantations (see https://github.com/broadinstitute/sam/pull/1234/commits/5d547ce18d1c5d77097be49553e794a6e73fb359 for latest attempt) but kept hitting permission problems, so I am backing that commit out.

Ghost-in-a-Jar: ok good to know thanks for taking a look!

Ghost-in-a-Jar: @jladieu I managed to fix it with this change https://github.com/broadinstitute/sam/pull/1233/files#diff-88f8d59210f317b7bdd12e6a807c0b84cf473372221acb3ce860f5a0160183b1

jladieu: Perfect, thanks!  I picked up that change and will see how re-running looks.

jladieu: @Ghost-in-a-Jar I think you fixed the issue with permissions, but it looks like maybe another false pass on the PR you linked: If you take a look at: https://github.com/broadinstitute/sam/actions/runs/6711027848/job/18237474362 I'm still seeing some errors (despite the GHA passing).

Example:
```
    1.1) body: $.userEmail Expected 'test@test.com' (String) but received 'tVRgfhGSOLLHNNEcjExKfrcCiMkGNHzDaSDHVkVGwPGSRzTNIwen@gmail.com' (String)

    1.2) body: $.userSubjectId Expected 'testUser' (String) but received '509175583607142998ceb' (String)
```
The bright side is the errors look like they're a lot easier to troubleshoot than the one you fixed since they're concrete pact problems rather than connectivity issues.

mlilynolting (approved):

PR #1236, opened 2023-10-31
ID-897 filter resources endpoint

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-897

`GET /api/resources/v2/{resource_type_name}` is just too darn slow. The DB query to generate the response is one of the most expensive queries in Sam, and it recently caused us to DDoS ourselves when a change was made in Terra-UI that ultimately hit that endpoint millions of times. 

The current endpoint returns all the resources of a resource type that the calling user has access to, including direct and inherited roles and actions. It also returns all public resources of that type. The query backing this runs extraordinarily slowly, often .5-1 seconds, and takes a lot of DB CPU resources.

I’m adding a new endpoint: `GET /api/resources/v3/filter?resourceTypes=resourceType1,resourceType2&policies=policyName1,policyName2&roles=roleName1,roleName2&actions=actionName1,actionName2&includePublic=false`

This “filter” query runs an order-of-magnitude faster through some specific optimizations. The filter parameters are all lists (except includePublic). Results will be the intersection of all filters, excluding empty ones. For example:

* `GET /api/resources/v3/filter?resourceTypes=workspace` would get all the workspaces available to the user, and include all the policies, roles, and actions they have on those workspaces.

* `GET /api/resources/v3/filter?resourceTypes=workspace&roles=reader,writer` would get all workspaces for the user where they are in either the reader OR the writer role. 

The response will look like this:
```
{
	"resources": [
		{
			"resourceType": "workspace",
			"resourceName": "theName"
			"isPublic": Boolean,
			"policies": ["policyName1", "policyName2", ...],
			"roles": ["roleName1", "roleName2", ...],
			"actions": ["actionName1", "actionName2", ...],
		},
		...
	]
}
```
~~In the initial implementation, getting public resources for a user through this query will remain unimplemented, and throw an exception if includePublic=true is set in the request. Some special attention needs to be paid to how we get public resources for a user, since that actually makes up the bulk of the query latency in /api/resource/v2/{resource_type_name}.~~ Ended up implementing it, and its 4x faster than the list resources endpoint

This endpoint is extremely useful for services like Leo, which only want to get the resources for a user for which they have a specific type of access. Currently, Leo is forced to get all resources of a type for a user and filter client-side, but this query will let Leo ask Sam only for what it needs. Efficiencies! 

In initial testing, the DB query supporting the filtering runs 10x faster than the standard list resources query!

---

**PR checklist**


tlangs:
I'm going to change this from v3 to v2 endpoint version

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -28,6 +29,17 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
   val resourceService: ResourceService
   val liquibaseConfig: LiquibaseConfig
 
+  def resourceRoutesV3(samUser: SamUser, samRequestContext: SamRequestContext): server.Route =
+    pathPrefix("resources" / "v3") {
```
tlangs: I need to figure out how to get Akka to recognize this as a separate endpoint rather than an arg to `/api/resources/v2`.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1678,6 +1678,118 @@ class PostgresAccessPolicyDAO(protected val writeDbRef: DbReference, protected v
...
+    val resourceTypeConstraint =
+      if (resourceTypeNames.nonEmpty) samsqls"and ${resource.resourceTypeId} in  (${resourceTypeNames.flatMap(resourceTypePKsByName.get).toSet})"
+      else samsqls""
+    val policyConstraint = if (policies.nonEmpty) samsqls"and ${resourcePolicy.name} in (${policies.toSet})" else samsqls""
+    val roleConstraint = if (roles.nonEmpty) samsqls"and ${resourceRole.role} in (${roles.toSet})" else samsqls""
+    val actionConstraint = if (actions.nonEmpty) samsqls"and ${resourceAction.action} in (${actions.toSet})" else samsqls""
+
+    val policyRoleActionQuery =
+      samsqls"""
+        select distinct ${resource.result.name}, ${resource.result.resourceTypeId}, ${resourcePolicy.result.name}, ${resourceRole.result.role}, ${resourceAction.result.action}, ${resourcePolicy.result.public}
```
Ghost-in-a-Jar: is it possible to do a group by or something instead of a distinct? I try to avoid using distinct if at all possible.

sjkobori (approved):

Ghost-in-a-Jar (approved):

mlilynolting (commented):
this looks great.

PR #1238, opened 2023-11-01
ID-807 Get tos for user v2

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-807

What:

Endpoints for getting a user's ToS status/details for themselves or for another user if they are an admin
---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRoutes.scala
```diff
@@ -54,12 +60,18 @@ trait TermsOfServiceRoutes {
...
+
+  def userTermsOfServiceRoutes(samUser: SamUser, isAdmin: Boolean, samRequestContext: SamRequestContext): server.Route =
+    pathPrefix("termsOfService") {
+      pathPrefix("v1") {
         pathPrefix("user") { // api/termsOfService/v1/user
           pathPrefix("self") { // api/termsOfService/v1/user/self
             pathEndOrSingleSlash {
               get {
-                complete(StatusCodes.NotImplemented)
+                complete(StatusCodes.OK, tosService.getTermsOfServiceDetailsForUser(samUser.id, samUser, isAdmin = false, samRequestContext))
```
tlangs: Should probably be:
```scala
complete {
    tosService.getTermsOfServiceDetailsForUser(samUser.id, samUser, isAdmin = false, samRequestContext).map(_ => StatusCodes.OK -> _)
}
```
just to match the pattern used elsewhere

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRoutes.scala
```diff
@@ -79,15 +91,18 @@ trait TermsOfServiceRoutes {
...
-                complete(StatusCodes.NotImplemented)
-              }
-            } ~
-            pathPrefix("history") { // api/termsOfService/v1/user/{userId}/history
+            validate(samUserIdPattern.matches(userId), "User ID must be alpha numeric") {
+              val requestUserId = WorkbenchUserId(userId)
               pathEndOrSingleSlash {
                 get {
-                  complete(StatusCodes.NotImplemented)
+                  complete(StatusCodes.OK, tosService.getTermsOfServiceDetailsForUser(requestUserId, samUser, isAdmin, samRequestContext))
```
tlangs: Same here

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/MockSamRoutes.scala
```diff
@@ -61,21 +61,24 @@ abstract class MockSamRoutes(
...
-                resourceRoutes(samUser, samRequestContextWithUser) ~
-                  adminRoutes(samUser, samRequestContextWithUser) ~
-                  extensionRoutes(samUser, samRequestContextWithUser) ~
-                  groupRoutes(samUser, samRequestContextWithUser) ~
-                  userRoutesV1(samUser, samRequestContextWithUser) ~
-                  azureRoutes(samUser, samRequestContextWithUser)
-              }
+              publicTermsOfServiceRoutes ~
+                withActiveUser(samRequestContext) { samUser =>
+                  isWorkbenchAdmin(samUser) { isAdmin =>
```
tlangs: We should find another way of doing this. As implemented here, we're checking _every single call_ to see if the user is an admin, even if the route never needs admin permissions. It will introduce an API call to Google for every single request to Sam. Checking for admin permissions should only happen for routes that support admin functionality, to reduce overhead.

tlangs (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -60,9 +58,42 @@ class TosService(val directoryDao: DirectoryDAO, val tosConfig: TermsOfServiceCo
...
+  def getTosDetails(samUser: SamUser, samRequestContext: SamRequestContext): IO[OldTermsOfServiceDetails] =
     directoryDao.getUserTos(samUser.id, samRequestContext).map { tos =>
-      TermsOfServiceDetails(isEnabled = true, tosConfig.isGracePeriodEnabled, tosConfig.version, tos.map(_.version))
+      OldTermsOfServiceDetails(isEnabled = true, tosConfig.isGracePeriodEnabled, tosConfig.version, tos.map(_.version))
+    }
+
+  def getTermsOfServiceDetailsForUser(
+      requestedUserId: WorkbenchUserId,
+      requestingUser: SamUser,
+      isAdmin: Boolean,
```
tlangs: I don't think `isAdmin` should be passed to this function. If it is, then `isWorkbenchAdmin` needs to be run for every single invocation of this method, which is way more than is necessary.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRoutes.scala
```diff
@@ -79,15 +93,22 @@ trait TermsOfServiceRoutes {
...
-                complete(StatusCodes.NotImplemented)
-              }
-            } ~
-            pathPrefix("history") { // api/termsOfService/v1/user/{userId}/history
+            validate(samUserIdPattern.matches(userId), "User ID must be alpha numeric") {
+              val requestUserId = WorkbenchUserId(userId)
               pathEndOrSingleSlash {
-                get {
-                  complete(StatusCodes.NotImplemented)
+                isWorkbenchAdmin(samUser) { isAdmin =>
```
tlangs: This is still calling `isWorkbenchAdmin` way more than is necessary.

tlangs (changes requested):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3255,6 +3255,53 @@ paths:
...
+  /termsOfService/v1/user/{sam_user_id}:
+    get:
+      tags:
+        - TermsOfService
+      summary: gets a user's terms of service status, can only get another user's status if the requester is an admin.
+      operationId: userTermsOfServiceGet
+      parameters:
+        - name: sam_user_id
+          in: path
+          description: the id of the sam user to get terms of service details for
```
sjkobori: I think we should use consistent language about what the object sent back is. I personally prefer status, since I dont think details really means anything here.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
sjkobori: I think in general, if we want to deprecate "admin" routes, we would need to know if the caller is an admin to do our business logic on it. And I do not think it should be done inside the routes layers since I think that should be pretty much absent from logic.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRouteSpec.scala
```diff
@@ -1,51 +1,136 @@
...
 import org.scalatest.concurrent.Eventually.eventually
-import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.funspec.AnyFunSpec
 import org.scalatest.matchers.should.Matchers
-import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 
-class TermsOfServiceRouteSpec extends AnyFlatSpec with Matchers with ScalatestRouteTest with TestSupport {
+class TermsOfServiceRouteSpec extends AnyFunSpec with Matchers with ScalatestRouteTest with TestSupport {
+
+  // Eventually being replaced by REST-ful termsOfService routes
```
sjkobori: Maybe we can remove this comment now?

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRouteSpec.scala
```diff
@@ -1,51 +1,136 @@
...
+      eventually {
+        Get("/tos/text") ~> samRoutes.route ~> check {
+          status shouldEqual StatusCodes.OK
+          responseAs[String].isEmpty shouldBe false
+        }
+      }
+    }
+  }
+
+  // Eventually being replaced by REST-ful termsOfService routes
```
sjkobori: Here too

sjkobori (approved):
Left some comments, but otherwise lgtm!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
tlangs: We need to then push the check for who's an admin or not down to the service layer. We can't have every single call to an endpoint check to see if the user is an admin.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -60,9 +62,51 @@ class TosService(val directoryDao: DirectoryDAO, val tosConfig: TermsOfServiceCo
...
+  def getTosDetails(samUser: SamUser, samRequestContext: SamRequestContext): IO[OldTermsOfServiceDetails] =
     directoryDao.getUserTos(samUser.id, samRequestContext).map { tos =>
-      TermsOfServiceDetails(isEnabled = true, tosConfig.isGracePeriodEnabled, tosConfig.version, tos.map(_.version))
+      OldTermsOfServiceDetails(isEnabled = true, tosConfig.isGracePeriodEnabled, tosConfig.version, tos.map(_.version))
+    }
+
+  def getTermsOfServiceDetailsForUser(
+      requestedUserId: WorkbenchUserId,
+      requestingUser: SamUser,
+      samRequestContext: SamRequestContext
```
tlangs: We should be getting the requesting user from the `samRequestContext` instead of passing it in manually

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -60,9 +62,51 @@ class TosService(val directoryDao: DirectoryDAO, val tosConfig: TermsOfServiceCo
...
+        currentSamUserTos <- loadTosRecordForUser(requestedUserId, Option(tosConfig.version), samRequestContext)
+        previousSamUserTos <- loadTosRecordForUser(requestedUserId, tosConfig.previousVersion, samRequestContext)
+        requestedUser <- loadUser(requestedUserId, samRequestContext)
+      } yield TermsOfServiceDetails(
+        currentSamUserTos.version,
+        currentSamUserTos.createdAt,
+        tosAcceptancePermitsSystemUsage(requestedUser, Option(currentSamUserTos), Option(previousSamUserTos))
+      )
+
+    if (requestedUserId != requestingUser.id) {
```
tlangs: I have a feeling that this is going to be a common pattern. Could we pull it out into some sort of helper function. It would be nice to write something along the lines of
``` 
ensureNecessaryRights(userId, samRequestContext) {
    myLogic...
}
```

tlangs (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRoutes.scala
```diff
@@ -79,15 +93,20 @@ trait TermsOfServiceRoutes {
           } ~
           // The {user_id} route must be last otherwise it will try to parse the other routes incorrectly as user id's
           pathPrefix(Segment) { userId => // api/termsOfService/v1/user/{userId}
-            pathEndOrSingleSlash {
-              get {
-                complete(StatusCodes.NotImplemented)
-              }
-            } ~
-            pathPrefix("history") { // api/termsOfService/v1/user/{userId}/history
+            validate(samUserIdPattern.matches(userId), "User ID must be alpha numeric") {
```
tlangs, line 96: Do we need this check?

tlangs (approved):
Thanks for all the work on this! Just one question.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRoutes.scala
Ghost-in-a-Jar: Greg added it, idk that we NEED it but I think extra validation and user feedback is usually a good thing 🤷

PR #1274, opened 2023-11-28
Fix system allowance for tos reject special case, add /api to tos history endpoint swagger docs.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-811

Fixes a special case for tos where the user has accepted the previous version and is in the acceptance window, but they have rejected the current tos.

I forgot to add /api to the swagger docs for these endpoints.

Also removing duplicate runs of the build ci job

---

**PR checklist**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
     paths-ignore: ['**.md']
   push:
+    branches:
+      - develop
     paths-ignore: ['**.md']
 
 jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -4,6 +4,8 @@ on:
   pull_request:
     paths-ignore: ['**.md']
   push:
+    branches:
+      - develop
```
Ghost-in-a-Jar, line 8: Without this here itll run the build/unit tests twice at the same time for pr pushes, just something i happened to notice.

tlangs (approved):

sjkobori (approved):

PR #1288, opened 2023-12-06
ID-954 Force 60 second timeout on Google Group Sync

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-954

What:

Manually set the request timeout for the group sync endpoint.

Why:

We set the akka server request timeout to 60 seconds, but this doesn't seem to be respected. As a result, we're seeing the group-sync endpoint fail after 20 seconds. 

How:

Wrap it in a directive.

---

**PR checklist**


sjkobori:
What are the practical implications of this timing out? Would this affect any front end behavior?

Shakespeared (approved):

tlangs:
@sjkobori This is actually blocking the creation of TDR datasets in some cases. Sam should allow up to 60 seconds for the request to complete, but in reality, its only allowing 20. Special-casing this fixes the immediate need while we try and find out why Sam isn't respecting the Akka config.

So, it should actually make the user experience better!

sjkobori (approved):

PR #1294, opened 2023-12-11
ID-918 Update OKIO to fix vulnerability.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-918

Okio - Denial of Service

https://defectdojo.dsp-appsec.broadinstitute.org/finding/148482

The following version in use was found to contain a vulnerability: 3.2.0

Upgradingto a non-vulnerable version of this library will fix this finding. The fixed version of this library is: 3.4.0

https://nvd.nist.gov/vuln/detail/CVE-2023-3635


project/Dependencies.scala
```diff
@@ -114,6 +114,8 @@ object Dependencies {
   val slf4jApi: ModuleID = "org.slf4j" % "slf4j-api" % slf4jVersion
   val slf4jSimple: ModuleID = "org.slf4j" % "slf4j-simple" % slf4jVersion
 
+  val okio: ModuleID = "com.squareup.okio" % "okio" % "3.4.0"
```
tlangs: You'll need to exclude `okio` from workbench-libs, since its a transitive dependency of that. Try running `sbt dependencyBrowseTree` to see which wb-libs artifacts need the exclusion

tlangs (changes requested):

project/Dependencies.scala
Ghost-in-a-Jar: Ah yeah I forgot about that thanks.

Shakespeared (approved):

tlangs (approved):

PR #1297, opened 2023-12-12
ID-931, ID-814 Add test for allowing inactive/disallowed users to access self attributes. Also fixing permissions checking for tests.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-931

What:

This is a followup to the previous pr https://github.com/broadinstitute/sam/pull/1295 where I fixed a bug. Inactive/disallowed users should be able to still access basic attributes about themselves so the ui can tell whether to serve up the TOS.

Why:

More test coverage!

How:

I noticed that when I made the test, it succeeded even when the endpoint was still wrapped with `withActiveUser`. I realized that the mock SamRoutes these tests were using from MockSamRoutesBuilder replace `withActiveUser` and `withUserAllowInactive` with a no-op. I replaced with the proper mock calls and then had to fix some other tests that were failing. 

---

**PR checklist**


src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
```diff
@@ -196,6 +196,24 @@ class UserRoutesV2Spec extends AnyFlatSpec with Matchers with ScalatestRouteTest
...
+      .withUserAttributes(defaultUser, userAttributes)
+      .callAsNonAdminUser()
+      .build
+
+    // Act and Assert
+    Get(s"/api/users/v2/self/attributes") ~> samRoutes.route ~> check {
+      status shouldEqual StatusCodes.OK
+      responseAs[SamUserAttributes] should be(userAttributes)
+    }
+  }
```
Ghost-in-a-Jar, line 215: This is the only net new test

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/MockUserServiceBuilder.scala
```diff
@@ -137,6 +145,9 @@ case class MockUserServiceBuilder() extends IdiomaticMockito {
   }
 
   private def makeUsers(samUsers: Iterable[SamUser], mockUserService: UserService): Unit = {
+    mockUserService.getUserFromGoogleSubjectId(any[GoogleSubjectId], any[SamRequestContext]) answers ((googleSubjectId: GoogleSubjectId) =>
```
Ghost-in-a-Jar, line 148: This is necessary to get the StandardSamUserDirectives.getActiveSamUser and StandardSamUserDirectives.getSamUser calls to work

Shakespeared (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/AdminServiceUserRoutesSpec.scala
```diff
@@ -118,6 +118,7 @@ class AdminServiceUserRoutesSpec extends AnyFlatSpec with Matchers with Scalates
     // Arrange
     val samRoutes = new MockSamRoutesBuilder(allUsersGroup)
       .withEnabledUser(defaultUser)
+      .withAllowedUser(defaultUser)
```
Ghost-in-a-Jar, line 121: We have a distinction in sam between a user being 'allowed' to use the system (accepted tos for example) and a user being 'enabled' (a flag on the user database model)

rushtong: This is interesting - I find the phrase `allowed` to be confusing in that an allow state could, in theory, apply to a LOT of different things like `allowed as an admin` or `allowed as a reader` etc. As opposed to something like `withToSAccepted()` or  `withToSRejected()` which is much more specific to the feature you're describing here.

rushtong (approved):
Looks reasonable to me 👍🏽

tlangs (approved):

PR #1322, opened 2024-01-23
ID-1035 Update user patch endpoint to allow nulling of ids

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1035

What:

We want to provide support with an api mechanism to be able to null an azureB2CId or a googleSubjectId on a given user.

Why:

If a user initially registers with google, but then wants to switch to a microsoft login, or vice verse, we need to null either of those ids in sam. If switching from google to microsoft their azureB2CId needs to be nulled in sam's database, if switching from microsoft to google their azureB2CId also needs to be nulled in sam. Otherwise sam will see a conflict in these ids and will throw an error when the user is trying to register with the second identity provider. 

How:

I updated the admin user patch endpoint to include the ability to update a user's googleSubjectId, or azureB2CId. Since there is no differentiation in the scala request parser between including a null value, or not including the value at all in a request, we decided to handle an empty string as a request to null out either of these fields. 

---

**PR checklist**


src/test/scala/org/broadinstitute/dsde/workbench/sam/api/AdminUserRoutesSpec.scala
```diff
@@ -149,6 +149,76 @@ class AdminUserRoutesSpec extends AnyFlatSpec with Matchers with ScalatestRouteT
     }
   }
 
+  "PATCH /admin/v1/user/{userSubjectId}" should "update a user's googleSubjectId" in {
+    // Arrange
+    val samRoutes = new MockSamRoutesBuilder(allUsersGroup)
+      .callAsAdminUser() // enabled "admin" user who is making the http request
+      .withEnabledUser(defaultUser) // "persisted/enabled" user we will check the status of
+      .withAllowedUser(defaultUser)
+      .build
```
Ghost-in-a-Jar, line 76: i am not super happy with how this mock routes builder is setup currently, it feels like we are testing too much of the mock rather than the underlying implementation of the route

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/MockUserServiceBuilder.scala
```diff
@@ -107,7 +107,17 @@ case class MockUserServiceBuilder() extends IdiomaticMockito {
...
+        val newAzureB2CId = if (r.azureB2CId.contains(AzureB2CId(""))) None else if (r.azureB2CId.isDefined) r.azureB2CId else samUser.azureB2CId
+        val newGoogleSubjectId = if (r.googleSubjectId.contains(GoogleSubjectId(""))) None else if (r.googleSubjectId.isDefined) r.googleSubjectId else samUser.googleSubjectId
+        val newUser = samUser.copy(
+          email = r.email.getOrElse(samUser.email),
+          enabled = r.enabled.getOrElse(samUser.enabled),
+          azureB2CId = newAzureB2CId,
+          googleSubjectId = newGoogleSubjectId
+        )
+        IO(Option(newUser))
+      }
```
Ghost-in-a-Jar, line 110: Following up on my comment above I am not a huge fan of having to fake the underlying implementation in these mocks, because we end up testing the mock implementation in the routes test rather than the underlying implementation. Although other test suites should be covering the underlying impl while the routes cover more general request -> response patterns.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -449,18 +449,28 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
                  set (${u.azureB2cId}, ${u.updatedAt}) =
                  ($b2cId,
                    ${Instant.now()}
                  )
-                 where ${u.id} = $userId and (${u.azureB2cId} is null or ${u.azureB2cId} = $b2cId)"""
-          .update()
-          .apply()
+                 where ${u.id} = $userId and ${u.googleSubjectId} is not null"""
+            .update()
+            .apply()
```
Ghost-in-a-Jar: I thought about building these queries dynamically, but I like how clear the if/else logic is this way. Let me know what you think.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -449,18 +449,28 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
     serializableWriteTransaction("setUserAzureB2CId", samRequestContext) { implicit session =>
       val u = UserTable.column
       val results =
-        samsql"""update ${UserTable.table}
+        if (b2cId.value == "")
```
tlangs: In the data-access layer, I think it would be better to accept a real value or `null`. The service layer, or even the API layer, should be the one to translate `""` to `null`.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
tlangs: I think it might actually be better to have an `updateSamUser` method that handles updating both AzureB2CIDs and GoogleSubjectIDs. We can have it dynamically build the query based on the fields present in a `SamUser` input

tlangs (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/SamJsonSupport.scala
```diff
@@ -101,7 +101,7 @@ object SamJsonSupport {
 
   implicit val UserPolicyResponseFormat = jsonFormat5(UserPolicyResponse.apply)
 
-  implicit val UserUpdateRequestFormat = jsonFormat3(AdminUpdateUserRequest.apply)
```
samanehsan, line 104: I'm not too familiar with Scala, but Is there a reason why this is using `jsonFormat4` now? I see some other versions being used in this file so I'm wondering if we should standardize this in a follow-on ticket?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -199,14 +207,35 @@ class UserService(
         // validate all fields to be updated
         var errorReports = Seq[ErrorReport]()
         request.email.foreach(email => errorReports = errorReports ++ validateEmail(email, blockedEmailDomains))
+        request.azureB2CId.foreach(azureB2CId => errorReports = errorReports ++ validateAzureB2CId(azureB2CId))
         if (errorReports.nonEmpty) {
           IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "invalid user update", errorReports)))
+        } else if (request.googleSubjectId.contains(GoogleSubjectId("")) && request.azureB2CId.contains(AzureB2CId(""))) {
+          IO.raiseError(
+            new WorkbenchExceptionWithErrorReport(
+              ErrorReport(StatusCodes.BadRequest, "unable to null both azureB2CId and googleSubjectId in the same request", errorReports)
```
samanehsan: If I accidentally set both the `azureB2CId` and `googleSubjectId` to null in different requests, would everything break? 😅 Should SAM check/prevent this? Or is it okay because I could just call this endpoint again to fix it?

samanehsan: Oh nevermind! I see there's an error that gets thrown when updating the database

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/AdminUserRoutesSpec.scala
samanehsan: Could we pull `samRoutes` out into the class instead of declaring it in each test?

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAOSpec.scala
```diff
@@ -1328,10 +1329,21 @@ class PostgresDirectoryDAOSpec extends RetryableAnyFreeSpec with Matchers with B
...
+        val user = genWorkbenchUserBoth.sample.get
+        dao.createUser(user, samRequestContext).unsafeRunSync()
+
+        dao.loadUser(user.id, samRequestContext).unsafeRunSync().flatMap(_.googleSubjectId) shouldBe user.googleSubjectId
+        dao.setGoogleSubjectId(user.id, newGoogleSubjectId, samRequestContext).unsafeRunSync()
+
+        dao.loadUser(user.id, samRequestContext).unsafeRunSync().flatMap(_.googleSubjectId) shouldBe None
+      }
+
+      "throw an exception when trying to update for a non-existing user" in {
```
samanehsan: Could you add a test for when a user tries to null a `googleSubjectId` for a user that exists but has a null `azureB2cId`? (and vice versa)

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/SamJsonSupport.scala
Ghost-in-a-Jar: yeah good question, it is because this is the function that automatically parses a json blob to the AdminUpdateUserRequest object, I added a new field to AdminUpdateUserRequest so now it is 4 arguments.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/AdminUserRoutesSpec.scala
Ghost-in-a-Jar: When it is declared it sets up the state for the specific test so that would mess things up unfortunately

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -469,6 +469,57 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
+        case UserUpdate(Some(newGoogleSubjectId), Some(newAzureB2CId)) =>
+          samsqls"(${u.googleSubjectId}, ${u.azureB2cId}, ${u.updatedAt})"
+      }
+
+      var updatedUser =
+        samUser.copy(
+          googleSubjectId = userUpdate.newGoogleSubjectId.map(GoogleSubjectId).orElse(samUser.googleSubjectId),
+          azureB2CId = userUpdate.newAzureB2CId.map(AzureB2CId).orElse(samUser.azureB2CId),
+          updatedAt = Instant.now()
+        )
```
Ghost-in-a-Jar, line 492: I tried and couldnt get the update query to return the updated user, so this is the next best thing

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpecs/UpdateUserSpec.scala
```diff
@@ -79,5 +82,146 @@ class UpdateUserSpec extends UserServiceTestTraits {
...
+
+        // Act
+        val error = intercept[WorkbenchExceptionWithErrorReport] {
+          runAndWait(userService.updateUserCrud(googleUser.id, updateRequest, samRequestContext))
+        }
+
+        // Assert
+        assert(
+          error.errorReport.statusCode.value equals StatusCodes.BadRequest,
+          "Nulling a user's azureB2CId when the user has no googleSubjectId should have a Bad Request status"
```
samanehsan: Should this be flipped? (same below):
```suggestion
          "Nulling a user's googleSubjectId when the user has no azureB2CId should have a Bad Request status"
```

samanehsan (approved):
I had a small comment on the test you added, but besides that this looks good! :octocat:

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -72,6 +72,11 @@ object UserStatusDetails {
     tosAccepted: Boolean,
     adminEnabled: Boolean
 )
+
+@Lenses final case class UserUpdate(
```
tlangs: Why make a new class for this instead of just passing the `AdminUpdateUserRequest` into the function?

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/SamModelApi.scala
```diff
@@ -5,6 +5,7 @@ import org.broadinstitute.dsde.workbench.model._
 import org.broadinstitute.dsde.workbench.sam.model._
 import spray.json.{DefaultJsonProtocol, JsObject, JsString, JsValue, RootJsonFormat, deserializationError}
 @Lenses final case class AdminUpdateUserRequest(
+    azureB2CId: Option[AzureB2CId],
```
tlangs: Can we make AdminUserUpdateRequest live in its own file, and then get the `UserUpdateRequestFormat` to be a member of the companion object. I really want to move towards each class being its own file.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -199,15 +207,37 @@ class UserService(
...
         request.email.foreach(email => errorReports = errorReports ++ validateEmail(email, blockedEmailDomains))
+        request.azureB2CId.foreach(azureB2CId => errorReports = errorReports ++ validateAzureB2CId(azureB2CId))
         if (errorReports.nonEmpty) {
           IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "invalid user update", errorReports)))
         } else { // apply all updates
-          var updatedUser = user
-          request.email.foreach { email =>
-            directoryDAO.updateUserEmail(userId, email, samRequestContext)
-            updatedUser = user.copy(email = email)
+          val userUpdate = UserUpdate(request.azureB2CId.map(_.value), request.googleSubjectId.map(_.value))
```
tlangs: How do we feel about moving validation of the object into the object itself? I think it would make sense for an object to know whether or not its valid. That way, we can simplify the service code to just be `updateModel.isValid()`.

tlangs (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
Ghost-in-a-Jar: This was more justified before I did a recent refactor, ill just stick to using AdminUpdateUserRequest 👍

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/AdminUpdateUserRequest.scala
```diff
@@ -0,0 +1,38 @@
...
+) {
+  def isValid(user: SamUser): Seq[ErrorReport] = {
+    import org.broadinstitute.dsde.workbench.google.errorReportSource
+
+    var errorReports = Seq[ErrorReport]()
+    if (azureB2CId.contains(AzureB2CId("null")) && user.googleSubjectId.isEmpty)
+      errorReports = errorReports ++ Seq(ErrorReport(StatusCodes.BadRequest, "Unable to null azureB2CId when the user's googleSubjectId is already null"))
+    else if (googleSubjectId.contains(GoogleSubjectId("null")) && user.azureB2CId.isEmpty)
+      errorReports = errorReports ++ Seq(ErrorReport(StatusCodes.BadRequest, "Unable to null googleSubjectId when the user's azureB2CId is already null"))
+    errorReports
```
tlangs, line 41: Since we don't support updating email or enabled here yet, can we add a validator to reject the request if the user tries to update those?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -199,15 +207,21 @@ class UserService(
...
-            updatedUser = user.copy(email = email)
-          }
-          IO(Some(updatedUser))
+          val updatedUser = if (request.azureB2CId.isDefined || request.googleSubjectId.isDefined) {
+            directoryDAO.updateUser(user, request, samRequestContext)
+          } else IO(Option(user))
+
+          if (request.email.isDefined) {
+            updatedUser.map { u =>
+              directoryDAO.updateUserEmail(userId, request.email.get, samRequestContext)
```
tlangs: IDK if we should be updating the user email at this point. There's more that needs to happen, like putting the new email into the proxy group, updating the all users group, updating other systems, etc.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/AdminUpdateUserRequest.scala
tlangs: Ya know, lets just take out `email` and `enabled`

tlangs (approved):

PR #1326, opened 2024-01-29
ID-1046 Migrate listResourcesAndPoliciesV2 to new query

tlangs:
Ticket: <Link to Jira ticket>

What:
Instead of making everyone else migrate to a new `listResourcesV2` endpoint, why not just change the query powering the `listResourcesAndPoliciesV2` under the hood?

Why:

Lessen the load on everyone else.

How:

Adapt the db response to the response type of `listResourcesAndPoliciesV2`

---

**PR checklist**


Ghost-in-a-Jar (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1686,13 +1690,86 @@ class PostgresAccessPolicyDAO(
         .toSet
     }
   }
-  override def filterResources(
-      samUser: SamUser,
+
+  private val publicResourcesCache: java.util.Map[ResourceTypeName, Seq[FilterResourcesResult]] =
+    Collections.synchronizedMap(new PassiveExpiringMap(1, TimeUnit.HOURS))
```
Ghost-in-a-Jar, line 1695: 1 hour seems like a while, am i correct in thinking that if someone changes a public resource itll take an hour to propagate that change? Do you see that causing any problems?

tlangs: I don't think it would be a problem. Publicizing workspaces is a manual process undertaken by Support, so we can just instruct them that it can take up to an hour for a public workspace to appear.

Ghost-in-a-Jar (approved):

samanehsan (approved):
This looks good to me!

PR #1348, opened 2024-02-13
ID-1086 Strongly Type User Allowances

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1086

What:

Strongly type the `getUserAllowances` response in the api client.

Also fixing a vulnerability in the transitive `json` dependency of the sam client.

Why:

Makes it easier for consumers of the API to use the response

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/SamUserAllowances.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/SamUserAllowances.scala
@@ -2,19 +2,26 @@ package org.broadinstitute.dsde.workbench.sam.model.api
 
 import spray.json.DefaultJsonProtocol._
 import spray.json.RootJsonFormat
-
 object SamUserAllowances {
-  implicit val SamUserAllowedResponseFormat: RootJsonFormat[SamUserAllowances] = jsonFormat2(SamUserAllowances.apply)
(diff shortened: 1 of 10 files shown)


Ghost-in-a-Jar (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/SamUserAllowances.scala
```diff
@@ -2,19 +2,26 @@ package org.broadinstitute.dsde.workbench.sam.model.api
...
 
-  def apply(allowed: Boolean, enabled: Boolean, termsOfService: Boolean): SamUserAllowances =
-    SamUserAllowances(allowed, Map("enabled" -> enabled, "termsOfService" -> termsOfService))
+  def apply(enabled: Boolean, termsOfService: Boolean): SamUserAllowances =
+    SamUserAllowances(enabled && termsOfService, SamUserAllowancesDetails(enabled, termsOfService))
 }
 final case class SamUserAllowances(
     allowed: Boolean,
-    details: Map[String, Boolean]
+    details: SamUserAllowancesDetails
```
samanehsan, line 13: Why is `allowed` not needed any more? Should it be removed from here too?

tlangs: `allowed` is a convenience parameter that just ANDs the different things that could deny a user access. Really useful for consumers, as they don't have to parse the details to know whether or not a user can use Terra

samanehsan (approved):

PR #1350, opened 2024-02-14
ID-1086 Fix JSON dependency vulnerability

tlangs:
Ticket:  https://broadworkbench.atlassian.net/browse/ID-1086

Looks like this _actually_ does what I intended to do, which was to make sure the Sam Client doesn't ship with a vulnerable transitive dependency

---

**PR checklist**

### DIFF
--- a/codegen_java/templates/libraries/okhttp-gson/build.sbt.mustache
+++ b/codegen_java/templates/libraries/okhttp-gson/build.sbt.mustache
@@ -41,6 +41,5 @@ lazy val root = (project in file(".")).
       "com.novocode" % "junit-interface" % "0.10" % "test",
       "org.mockito" % "mockito-core" % "3.12.4" % "test",
       "org.json" % "json" % "20240205"
-    ),
-    dependencyOverrides += "org.json" % "json" % "20240205") ++ publishSettings:_*
+    )) ++ publishSettings:_*
   )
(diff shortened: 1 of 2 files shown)


codegen_java/templates/libraries/okhttp-gson/build.sbt.mustache
```diff
@@ -39,7 +39,8 @@ lazy val root = (project in file(".")).
       "com.google.code.findbugs" % "jsr305" % "3.0.2" % "compile",
       "org.junit.jupiter" % "junit-jupiter-api" % "5.9.1" % "test",
       "com.novocode" % "junit-interface" % "0.10" % "test",
-      "org.mockito" % "mockito-core" % "3.12.4" % "test"
+      "org.mockito" % "mockito-core" % "3.12.4" % "test",
+      "org.json" % "json" % "20240205"
     ),
     dependencyOverrides += "org.json" % "json" % "20240205") ++ publishSettings:_*
```
dvoet: do you still need the dependencyOverrides?

tlangs: Lemme try without it.

tlangs: Looks like we don't. I'm wondering why `dependencyOverrides` didn't work, but I'm not so interested that I'm going to investigate it...

dvoet (approved):

codegen_java/templates/libraries/okhttp-gson/build.sbt.mustache
dvoet: it is probably an override for the local build but doesn't affect the pom, but I am just guessing

okotsopoulos (approved):

PR #1386, opened 2024-03-25
[ID-367] make registered_at non null and update null values to 1970

Shakespeared:
Ticket: [<Link to Jira ticket>](https://broadworkbench.atlassian.net/browse/ID-367)

What:

In using the registered_at date from Sam to send to appcues for the NPS survey timing, we discovered that some registered_at dates are null from when we migrated the dates from Rex to Sam. This PR will backfill all registered_at dates that are currently null and set them to a default value of 1970. This way we can still see which users we don't have accurate registration dates on, but don't have to handle null values. 

---

**PR checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,8 @@ You only have to do this once.
 ```shell
 ./render_config.sh
 ```
+#### Get on the Broad VPN
+You need to be on the non-split VPN to perform the next steps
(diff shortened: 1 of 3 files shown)


Shakespeared:
decided not to do this

PR #1396, opened 2024-04-10
[DC-830] Add snapshot builder permissions

s-rubenstein:
Ticket: https://broadworkbench.atlassian.net/browse/DC-830

  \<Don't forget to include the ticket number in the PR title!\>

What:

This adds the permissions for the snapshot builder to the Sam config. While here, I also decided to use the new Sam _includedRoles_ feature that I noticed, which I think cleans up a lot of the role definitions. I only did it for snapshots because this PR doesn't touch datasets.

What this PR does not do:
1. This PR does not include descendant roles for the dataset<->snapshot relationship - that requires additional discussion
2. This PR does not remove the existing dataset permissions that we have given admins and are currently using to check permissions. I want to remove those later to avoid breaking our dev auth checks

Why:

The cleanup is useful to simplify our role definitions, and help document what is shared across roles. 
The addition of SnapshotBuilderRequests as separate objects allow us to more carefully manage the request objects.

How:

  \<For your reviewers' sake, please describe in ~1 paragraph how this PR accomplishes its goal.\>

  \<If the PR is big, please indicate where a reviewer should start reading it (i.e. which file or function).\>

---

**PR checklist**

- [ ] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've filled out the [Security Risk Assessment](https://sdarq.dsp-appsec.broadinstitute.org/jira-ticket-risk-assesment) (requires Broad Internal network access) and attached the result to the JIRA ticket

PR #1401, opened 2024-04-17
[dc-830] add auth roles for snapshot builder

s-rubenstein:
Ticket: <Link to Jira ticket>

What:

Readd https://github.com/broadinstitute/sam/pull/1396, but waiting until TDR has been changed to accept the new roles and actions.

Why:

How:

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1340,16 +1340,16 @@ resourceTypes = {
   snapshot-builder-request = {
     actionPatterns = {
       get = {
-        description = "Get or list the snapshot builder request
+        description = "Get or list the snapshot builder request"
       }
       update = {
-        description = "Get or list the snapshot builder request
+        description = "Update the snapshot builder request"
       }
       delete = {
-        description = "Get or list the snapshot builder request
+        description = "Delete the snapshot builder request"
       }
       approve = {
-        description = "Approve or reject the snapshot builder request
+        description = "Approve or reject the snapshot builder request"
       }
     }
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -1312,6 +1340,32 @@ resourceTypes = {
...
+        description = "Delete the snapshot builder request"
+      }
+      approve = {
+        description = "Approve or reject the snapshot builder request"
+      }
+    }
+    ownerRoleName = "owner"
+    roles = {
+      owner = {
+        roleActions = ["get", "update", "delete"]
```
pshapiro4broad, line 1361: I didn't notice it before, but it looks like TDR has a convention of including the resource type in the action name, e.g. `update_snapshot`, `read_dataset`, etc. And there is a new action above, `create_snapshot_request`. Although I also see `delete` and `create_snapshot_request` is a `snapshot` action so it makes sense that it has the type in its name.

Maybe there's nothing to do here but I wonder if we should be trying to make the action names more consistent, and if so, what convention should we use.

pshapiro4broad (approved):

src/main/resources/reference.conf
s-rubenstein: I think we should use the shorter name. It is equivalent to `snapshot.delete` rather than `snapshot.delete_snapshot`. One of those has significant redundant information.

rjohanek (approved):
new roles, actions, and resource type look good.

PR #1404, opened 2024-04-18
[TOAZ-355] [TOAZ-356] Use Managed Identity auth when running Azure Control Plane, support for Service Catalog deployed Azure Managed Apps

aaronegrant:
- Refactor SAM ARM authorization code to support using a Managed Identity app id instead of Service Principal clientid/clientsecret

- Service Catalog deployed app support:
Sam maintains its own list of Terra managed app plans https://github.com/broadinstitute/sam/blob/develop/src/main/resources/sam.conf#L192 and validates a billing profile managed app against this list during billing profile creation. 
Service catalog managed apps do not have a plan defined so we will need change the validation in Sam to accept this.


src/main/resources/sam.conf
```diff
@@ -184,6 +184,10 @@ admin {
 azureServices {
   azureEnabled = ${?AZURE_ENABLED}
   allowManagedIdentityUserCreation = ${?AZURE_ALLOW_MANAGED_IDENTITY_USER_CREATION}
+  azureControlPlaneEnabled = ${?AZURE_CONTROL_PLANE_ENABLED}
```
dvoet: This controls specifically whether to use the service catalog or partner center managed resource groups. The name of this config setting should reflect that.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/CrlService.scala
```diff
@@ -58,16 +59,35 @@ class CrlService(config: AzureServicesConfig, janitorConfig: JanitorConfig) {
   }
 
   def getManagedAppPlans: Seq[ManagedAppPlan] = config.managedAppPlans
+  def getAzureControlPlaneEnabled: Boolean = config.azureControlPlaneEnabled
+  def getAuthorizedUserKey: String = config.authorizedUserKey
+  def getKindServiceCatalog: String = config.kindServiceCatalog
```
dvoet: it seems weird to expose these config settings through `CrlService` when they are used in `AzureService`.  `AzureService` probably should be passed `AzureServicesConfig` directly

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -269,6 +274,33 @@ class AzureService(
       } yield mrg
     }
 
+  /** Validates a managed resource group deployed from Azure Service Catalog. Service Catalog apps do not contain a "plan" or publisher. Algorithm:
+    *   1. Resolve the MRG in Azure 2. Get the managed app id from the MRG 3. Resolve the managed app 4. Validate the app kind is "ServiceCatalog" 5. Validate
+    *      that the caller is on the list of authorized users for the app
+    */
+  private def validateServiceCatalogManagedResourceGroup(
```
rtitle, line 283: There is not much difference between `validateServiceCatalogManagedResourceGroup` and `validateManagedResourceGroup` -- would it make sense to consolidate them?

Looks like the only difference is the service catalog app doesn't have a plan. We could potentially have a single method that looks at `managedApp.kind()` to determine whether it needs to validate the plan.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/CrlService.scala
```diff
@@ -58,16 +59,35 @@ class CrlService(config: AzureServicesConfig, janitorConfig: JanitorConfig) {
...
 
-  private def getCredentialAndProfile(tenantId: TenantId, subscriptionId: SubscriptionId): (ClientSecretCredential, AzureProfile) = {
-    val credential = new ClientSecretCredentialBuilder()
+  private def getCredentialAndProfile(tenantId: TenantId, subscriptionId: SubscriptionId): (TokenCredential, AzureProfile) = {
+
+    val managedIdentityCredential = new ManagedIdentityCredentialBuilder()
+      .clientId(config.managedAppWorkloadClientId)
```
rtitle, line 66: Just confirming, if `config.managedAppWorkloadClientId` is null (which I think it would be in the Broad environment) -- will this safely be ignored (e.g. not throw an exception)?

rtitle (commented):
Left a couple comments above, but overall the approach makes sense to me.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/CrlService.scala
aaronegrant: Correct, there will be a message in the log file about unable to auth using Managed Identity and then fall through to the Service Principal credentials

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/MockCrlService.scala
```diff
@@ -45,10 +45,6 @@ object MockCrlService extends MockitoSugar {
       .when(mockCrlService.buildMsiManager(any[TenantId], any[SubscriptionId]))
       .thenReturn(IO.pure(mockMsi))
 
-    lenient()
-      .when(mockCrlService.getManagedAppPlans)
-      .thenReturn(Seq(managedAppPlan))
-
```
dvoet, line 51: why was this removed? I think this is the source of your unit test errors

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -264,11 +271,38 @@ class AzureService(
         appManager <- crlService.buildApplicationManager(mrgCoords.tenantId, mrgCoords.subscriptionId)
         appsInSubscription <- IO(appManager.applications().list().asScala.toSeq)
         managedApp <- IO.fromOption(appsInSubscription.find(_.managedResourceGroupId() == mrg.id()))(managedAppValidationFailure)
-        plan <- validatePlan(managedApp, crlService.getManagedAppPlans)
+        plan <- validatePlan(managedApp, config.managedAppPlans)
```
dvoet: I think this change is causing test failures. In unit tests, `crlService` is mocked and returns results expected by tests but inconsistent with config

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AppConfig.scala
```diff
@@ -174,6 +174,10 @@ object AppConfig {
         if (azureEnabled) {
           Option(
             AzureServicesConfig(
+              config.getBoolean("azureControlPlaneEnabled"),
```
dvoet: the name `azureControlPlaneEnabled` is no longer correct

src/main/scala/org/broadinstitute/dsde/workbench/sam/config/AzureServicesConfig.scala
```diff
@@ -1,10 +1,44 @@
...
+
+object ManagedAppPlan {
+  implicit val managedAppPlanConfigReader: ValueReader[ManagedAppPlan] = ValueReader.relative { config =>
+    ManagedAppPlan(
+      name = config.getString("name"),
+      publisher = config.getString("publisher"),
+      authorizedUserKey = config.getString("authorizedUserKey")
+    )
+  }
+}
```
dvoet: why do we have these new config readers when they already exist in AppConfig.scala?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -407,7 +407,7 @@ object Boot extends IOApp with LazyLogging {
     val resourceTypeMap = config.resourceTypes.map(rt => rt.name -> rt).toMap
     val policyEvaluatorService = PolicyEvaluatorService(config.emailDomain, resourceTypeMap, accessPolicyDAO, directoryDAO)
     val azureService = config.azureServicesConfig.map { azureConfig =>
-      new AzureService(new CrlService(azureConfig, config.janitorConfig), directoryDAO, azureManagedResourceGroupDAO)
+      new AzureService(config.azureServicesConfig.get, new CrlService(azureConfig, config.janitorConfig), directoryDAO, azureManagedResourceGroupDAO)
```
dvoet:
```suggestion
      new AzureService(azureConfig, new CrlService(azureConfig, config.janitorConfig), directoryDAO, azureManagedResourceGroupDAO)
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -44,7 +46,12 @@ class AzureService(
 
   def createManagedResourceGroup(managedResourceGroup: ManagedResourceGroup, samRequestContext: SamRequestContext): IO[Unit] =
     for {
-      _ <- validateManagedResourceGroup(managedResourceGroup.managedResourceGroupCoordinates, samRequestContext)
+      _ <-
+        if (false) {
```
dvoet: probably don't mean `if (false)`

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -269,6 +276,33 @@ class AzureService(
...
+  ): IO[ResourceGroup] =
+    traceIOWithContext("validateServiceCatalogManagedResourceGroup", samRequestContext) { _ =>
+      for {
+        resourceManager <- crlService.buildResourceManager(mrgCoords.tenantId, mrgCoords.subscriptionId)
+        mrg <- lookupMrg(mrgCoords, resourceManager)
+        appManager <- crlService.buildApplicationManager(mrgCoords.tenantId, mrgCoords.subscriptionId)
+        appsInSubscription <- IO(appManager.applications().list().asScala)
+        managedApp <- IO.fromOption(appsInSubscription.find(_.managedResourceGroupId() == mrg.id()))(managedAppValidationFailure)
+        _ <-
+          if (managedApp.kind() == config.kindServiceCatalog && validateUser) {
```
dvoet: ```managedApp.kind() == config.kindServiceCatalog```
if we are here, isn't is already a service catalog?

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -269,6 +276,33 @@ class AzureService(
...
+            validateAuthorizedAppUser(
+              managedApp,
+              ManagedAppPlan("", "", config.authorizedUserKey),
```
dvoet: perhaps refactor `validateAuthorizedAppUser` to take `authorizedUserKey` so we don't have to construct a dummy `ManagedAppPlan` with empty values

dvoet (dismissed):

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
aaronegrant: Not necessarily, Marketplace apps will come back here as well without this check. The managed appplan check would filter them out however. This seemed like a good check to  align the logic with the name of the function

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -44,7 +46,12 @@ class AzureService(
 
   def createManagedResourceGroup(managedResourceGroup: ManagedResourceGroup, samRequestContext: SamRequestContext): IO[Unit] =
     for {
-      _ <- validateManagedResourceGroup(managedResourceGroup.managedResourceGroupCoordinates, samRequestContext)
+      _ <-
+        if (Option(config).isDefined && config.azureServiceCatalogAppsEnabled) {
```
dvoet: why would `config` be null? We should not be passing around nulls. If it truly is nullable, it should be an `Option`

aaronegrant: Switched AzureServicesConfig to Option in AzureServices. Made the check for the ServiceCatalogApps enabled flag a case statement to handle the 3 possible states of AzureServicesConfig

dvoet (approved):

src/main/resources/sam.conf
```diff
@@ -184,6 +184,10 @@ admin {
 azureServices {
   azureEnabled = ${?AZURE_ENABLED}
   allowManagedIdentityUserCreation = ${?AZURE_ALLOW_MANAGED_IDENTITY_USER_CREATION}
+  azureServiceCatalogAppsEnabled = ${?AZURE_SERVICE_CATALOG_APPS_ENABLED}
+  authorizedUserKey = "authorizedTerraUser";
+  kindServiceCatalog = "ServiceCatalog";
```
Shakespeared: nit: the naming of `kindServiceCatalog` and `managedApp.kind` is a bit opaque to me. I understand from context but to me it's hard to interpret. Maybe something like `serviceCatalogAppType` and `managedApp.type` would be a little clearer?

Shakespeared (changes requested):
Not super familiar with this azure functionality but changes seem reasonable to me, requested minor naming changes for clarity

src/main/resources/sam.conf
aaronegrant: @Shakespeared agreed it's not the most useful name. managedApp.kind() is an azure api to get the value of the 'kind' property and will return either 'ServiceCatalog' or 'Marketplace' depending on where the managed app was deployed from. I'll rename the config option to help clarify that we are actually looking at a type of managed app, but we are stuck with the Azure kind() call

Shakespeared: Thanks for clarifying, that makes sense!

Shakespeared (approved):

PR #1406, opened 2024-04-18
ID-1224 Azure Action Managed Identities

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1224

What:

Introducing Action Managed Identities for Azure Resources! This will power Terra's access to data and other resources through Private Endpoints. The first use-case for this is pulling private docker images from Azure Container Registry.

Why:

For the ease of IT Admin of customers, Terra will represent itself with an Action Managed Identity when pulling things through a private endpoint. This means IT Admins can grant a single identity access to the private resource, and Terra will manage user access to that identity. 

An Action Managed Identity is tied to an Action on a Resource. This means that if a user has, for example, the `read` action on resource `A`, they will also have access to the `read-A` Action Managed Identity. 

---

**PR checklist**

### DIFF
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240417_action_managed_identities.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240417_action_managed_identities.xml
@@ -14,14 +14,8 @@
             <column name="resource_action_id" type="BIGINT">
                 <constraints nullable="false" primaryKey="true" foreignKeyName="FK_SAMI_ACTION" referencedTableName="SAM_RESOURCE_ACTION" referencedColumnNames="id"/>
             </column>
-            <column name="tenant_id" type="VARCHAR(40)">
-                <constraints nullable="false" primaryKey="true"/>
-            </column>
-            <column name="subscription_id" type="VARCHAR(40)">
-                <constraints nullable="false" primaryKey="true"/>
-            </column>
-            <column name="managed_resource_group_name" type="VARCHAR">
-                <constraints nullable="false" primaryKey="true"/>
+            <column name="managed_resource_group_id" type="BIGINT">
+                <constraints nullable="false" primaryKey="true" foreignKeyName="FK_SAMI_MRG" referencedTableName="SAM_AZURE_MANAGED_RESOURCE_GROUP" referencedColumnNames="id" deleteCascade="true"/>
             </column>
             <column name="object_id" type="VARCHAR">
                 <constraints nullable="false"/>
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240417_action_managed_identities.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240417_action_managed_identities.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="tlangs" id="action_managed_identities">
+        <createTable tableName="SAM_ACTION_MANAGED_IDENTITY">
+            <column name="resource_id" type="BIGINT">
+                <constraints nullable="false" primaryKey="true" foreignKeyName="FK_SAMI_RESOURCE" referencedTableName="SAM_RESOURCE" referencedColumnNames="id" deleteCascade="true"/>
+            </column>
+
+            <column name="resource_action_id" type="BIGINT">
+                <constraints nullable="false" primaryKey="true" foreignKeyName="FK_SAMI_ACTION" referencedTableName="SAM_RESOURCE_ACTION" referencedColumnNames="id"/>
+            </column>
+            <column name="tenant_id" type="VARCHAR(40)">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="subscription_id" type="VARCHAR(40)">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="managed_resource_group_name" type="VARCHAR">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="object_id" type="VARCHAR">
+                <constraints nullable="false"/>
+            </column>
+            <column name="display_name" type="VARCHAR">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+</databaseChangeLog>
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1559,11 +1559,73 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
-  /api/azure/v1/actionManagedIdentity/{resourceTypeName}/{resourceId}/{action}:
+  /api/azure/v1/actionManagedIdentity/{billingProfileId}/{resourceTypeName}/{resourceId}/{action}:
     post:
       tags:
         - Azure
-      summary: gets or creates an action managed identity that the calling user has access to
+      summary: creates an action managed identity that the calling user has access to
+      operationId: createActionManagedIdentity
+      parameters:
+        - name: billingProfileId
+          in: path
+          description: Billing Profile if the Managed Resource Group to create the Action Managed Identity in
+          required: true
+          schema:
+            type: string
+        - name: resourceTypeName
+          in: path
+          description: Type of resource
+          required: true
+          schema:
+            type: string
+        - name: resourceId
+          in: path
+          description: Id of resource
+          required: true
+          schema:
+            type: string
+        - name: action
+          in: path
+          description: Action to create the managed identity for
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: Action managed identity already exists
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ActionManagedIdentityResponse'
+        201:
+          description: Successfully created the action managed identity
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ActionManagedIdentityResponse'
+        400:
+          description: Invalid or incomplete request body
+          content: { }
+        403:
+          description: Caller does not have the action on the resource
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        404:
+          description: Resource does not exist
+          content: { }
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+  /api/azure/v1/actionManagedIdentity/{resourceTypeName}/{resourceId}/{action}:
+    get:
+      tags:
+        - Azure
+      summary: gets an action managed identity that the calling user has access to
       operationId: getActionManagedIdentity
       parameters:
         - name: resourceTypeName
@@ -1584,20 +1646,13 @@ paths:
           required: true
           schema:
             type: string
-      requestBody:
-        description: The details of the managed resource group in which to create the action managed identity
-        content:
-          'application/json':
-            schema:
-              $ref: '#/components/schemas/GetOrCreateManagedIdentityRequest'
-        required: true
       responses:
         200:
           description: Successfully retrieved the action managed identity without creating it
           content:
             application/json:
               schema:
-                type: string
+                $ref: '#/components/schemas/ActionManagedIdentityResponse'
         201:
           description: Successfully created the action managed identity
           content:
@@ -4294,6 +4349,30 @@ components:
           description: the name of the Azure managed resource group
           example: my-managed-resource-group
       description: specifies a request for a managed identity
+    ActionManagedIdentityId:
+      type: object
+      required:
+        - resourceId
+        - action
+        - billingProfileId
+      properties:
+        resourceId:
+          $ref: '#/components/schemas/FullyQualifiedResourceId'
+        action:
+          type: string
+        billingProfileId:
+          type: string
+    ActionManagedIdentityResponse:
+      type: object
+      properties:
+        id:
+          $ref: '#/components/schemas/ActionManagedIdentityId'
+        objectId:
+          type: string
+        displayName:
+          type: string
+        managedResourceGroupCoordinates:
+          $ref: '#/components/schemas/ManagedResourceGroupCoordinates'
     User:
       type: object
       required:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1479,7 +1479,7 @@ paths:
         content:
           'application/json':
             schema:
-              $ref: '#/components/schemas/GetOrCreatePetManagedIdentityRequest'
+              $ref: '#/components/schemas/GetOrCreateManagedIdentityRequest'
         required: true
       responses:
         200:
@@ -1528,7 +1528,7 @@ paths:
         content:
           'application/json':
             schema:
-              $ref: '#/components/schemas/GetOrCreatePetManagedIdentityRequest'
+              $ref: '#/components/schemas/GetOrCreateManagedIdentityRequest'
         required: true
       responses:
         200:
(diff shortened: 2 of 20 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1559,6 +1559,70 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/azure/v1/actionManagedIdentity/{resourceTypeName}/{resourceId}/{action}:
+    post:
+      tags:
+        - Azure
+      summary: gets or creates an action managed identity that the calling user has access to
```
dvoet: I wonder if we want get-or-create semantics or do we want separate api calls? The initial use case looks like create is a separate step from get. I also wonder if there will be later cases where we want to get the identity only if it exists.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -126,6 +133,70 @@ class AzureService(crlService: CrlService, directoryDAO: DirectoryDAO, azureMana
...
+  def getOrCreateActionManagedIdentity(
+      resource: FullyQualifiedResourceId,
+      resourceAction: ResourceAction,
+      mrgCoordinates: ManagedResourceGroupCoordinates,
+      samUser: SamUser,
+      samRequestContext: SamRequestContext
+  ): IO[(ActionManagedIdentity, Boolean)] = {
+    val id = ActionManagedIdentityId(resource, resourceAction, mrgCoordinates)
+    for {
+      hasPermission <- policyEvaluatorService.hasPermission(resource, resourceAction, samUser.id, samRequestContext)
```
dvoet: Usually permission checks happen in routes, see the `requireOneOfAction` directive. The reasons being that a) there is a common place for permission checks to happen, b) the service code is more reusable and c) common handling of 404 vs 403 responses

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -126,6 +133,70 @@ class AzureService(crlService: CrlService, directoryDAO: DirectoryDAO, azureMana
...
+        // pet exists in Sam DB - return it
+        case Some(p) => IO.pure((p, false))
+        // pet does not exist in Sam DB - create it
+        case None => createActionManagedIdentity(id, samRequestContext)
+      }
+    } yield ami
+
+  }
+
+  private def createActionManagedIdentity(
```
dvoet, line 163: Is there anything in call tree that validates that the MRG is trusted like by calling `validateManagedResourceGroup` or ensuring that the MRG coordinates exist in the db?

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutes.scala
```diff
@@ -57,6 +57,36 @@ trait AzureRoutes extends SecurityDirectives with LazyLogging with SamRequestCon
...
+                    throw new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"action $action not found"))
+                  }
+                  pathEndOrSingleSlash {
+                    postWithTelemetry(
+                      samRequestContext,
+                      "resourceType" -> resource.resourceTypeName,
+                      "resource" -> resource.resourceId,
+                      "action" -> resourceAction
+                    ) {
+                      entity(as[ManagedResourceGroupCoordinates]) { mrgCoordinates =>
```
dvoet: Should the input be a billing profile id instead of full coordinates? Sam should already have the coordinates in the DB. Then we can be sure the MRG is kosher.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240417_action_managed_identities.xml
```diff
@@ -0,0 +1,35 @@
...
+            </column>
+            <column name="tenant_id" type="VARCHAR(40)">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="subscription_id" type="VARCHAR(40)">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="managed_resource_group_name" type="VARCHAR">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
```
dvoet: instead of storing the MRG coordinates, should this be a FK to SAM_AZURE_MANAGED_RESOURCE_GROUP?

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
tlangs: I've recently been questioning this approach. I'm not sure which is better: doing a permission check at the "point of ingress" or the "point of action". 

Doing it at the point of ingress means that the service code can be simpler, but it opens up possibilities for developer error. Because the check and the action are not co-located, it becomes easy to call the service code from another place in the codebase, maybe a place where the developer didn't or forgot to implement a permissions check. 

By co-locating the permissions check with the action in the service code, it takes that burden off the developer. A developer no longer needs to carry that "context" as they develop: they can be sure that the service code won't let the user do something they're not allowed to do. However, it does make the service code a bit more complicated. 

I'm not sure which approach is better. Maybe neither is. Interested in hearing your thoughts.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -126,6 +138,93 @@ class AzureService(crlService: CrlService, directoryDAO: DirectoryDAO, azureMana
...
+      mrgCoordinates <- mrgOpt match {
+        case Some(mrg) => IO.pure(mrg.managedResourceGroupCoordinates)
+        case None =>
+          IO.raiseError(
+            new WorkbenchExceptionWithErrorReport(
+              ErrorReport(StatusCodes.NotFound, s"Managed Resource Group with Billing Profile ID [${id.billingProfileId}] does not exist")
+            )
+          )
+      }
+      _ <- validateManagedResourceGroup(mrgCoordinates, samRequestContext)
```
Ghost-in-a-Jar: Make sure that this actually prevents resources in azure from being created, note concurrency in for yields

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1559,6 +1559,125 @@ paths:
...
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/azure/v1/actionManagedIdentity/{billingProfileId}/{resourceTypeName}/{resourceId}/{action}:
+    post:
+      tags:
+        - Azure
+      summary: creates an action managed identity that the calling user has access to
+      operationId: createActionManagedIdentity
+      parameters:
+        - name: billingProfileId
```
rtitle, line 1569: Just noticing: the [getPetManagedIdentity](https://sam.dsde-prod.broadinstitute.org/#/Azure/getPetManagedIdentity) endpoint takes the MRG coordinates in the POST body, and looks up the billing profile ID from the Sam DB. This endpoint does the opposite: it takes the billing profile ID in the path and looks up the MRG coordinates from the Sam DB.

I guess either style works because Sam stores a mapping between billing profile id <> MRG coordinates. Do we want to be consistent?

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240417_action_managed_identities.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="tlangs" id="action_managed_identities">
+        <createTable tableName="SAM_ACTION_MANAGED_IDENTITY">
```
rtitle, line 9: Should this table have any unique constraints? Like should an `action_managed_identity` be unique by (resource, action, mrg)?

Also -- I know we're starting with org-wide identities for now. If we discover we want a per-workspace or per-user/workspace action managed identity in the future, how extensible is this? (Not saying we necessarily need to engineer for this now, just curious.)

tlangs: The primary key being the `resource_id`, `resource_action_id`, and `azure_managed_resource_group_id` will make sure they're unique.

Action Managed Identities can be applied to any action on any resource, so they're very extensible in that way. We don't have the capacity to do per-user Action Managed Identities, but that also kinda flies in the face of what Action Managed Identities are meant to be used for. Namely, they're a way to abstract over Terra users and represent the Terra platform's access to something.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -126,6 +136,79 @@ class AzureService(crlService: CrlService, directoryDAO: DirectoryDAO, azureMana
...
+      mrgOpt <- azureManagedResourceGroupDAO.getManagedResourceGroupByBillingProfileId(id.billingProfileId, samRequestContext)
+      mrg <- IO.fromOption(mrgOpt)(
+        new WorkbenchExceptionWithErrorReport(
+          ErrorReport(StatusCodes.NotFound, s"Managed Resource Group with Billing Profile ID [${id.billingProfileId}] does not exist")
+        )
+      )
+      mrgCoordinates = mrg.managedResourceGroupCoordinates
+      // mapping the result of the validate call to ensure that validation happens before anything is created in Azure
+      validatedMrgCoordinates <- validateManagedResourceGroup(mrgCoordinates, samRequestContext).map(_ => mrg.managedResourceGroupCoordinates)
+      msiManager <- crlService.buildMsiManager(validatedMrgCoordinates.tenantId, validatedMrgCoordinates.subscriptionId)
```
rtitle, line 177: Just mentioning that `crlService.buildMsiManager` should be correctly Janitorized in BEEs/test envs. So action managed identities should get cleaned up in integration tests.

src/main/resources/swagger/api-docs.yaml
tlangs: We want to be consistent, and I actually want to migrate Pet Azure Managed Identities to just use the Billing Profile ID in the path. But, that would be a separate PR.

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
tlangs: Sweet!

src/main/resources/swagger/api-docs.yaml
rtitle: Sounds good!

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
tlangs: I moved everything to the routes layer

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
tlangs: Done

src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutes.scala
tlangs: Done

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceSpec.scala
```diff
@@ -107,13 +139,219 @@ class AzureServiceSpec(_system: ActorSystem) extends TestKit(_system) with AnyFl
...
+    azureService.deleteActionManagedIdentity(actionManagedIdentityId, samRequestContext).unsafeRunSync()
+
+    // action managed identity should not exist in postgres
+    directoryDAO.loadActionManagedIdentity(actionManagedIdentityId, samRequestContext).unsafeRunSync() shouldBe None
+
+    // managed identity should not exist in Azure
+    msiManager.identities().listByResourceGroup(managedResourceGroupName.value).asScala.toList.exists { i =>
+      i.name() === azureService.toManagedIdentityNameFromAmiId(actionManagedIdentityId)
+    } shouldBe false
+  }
```
Ghost-in-a-Jar, line 278: This delete test could maybe just be merged with the above creation test? feels like there is a lot of overlap. Up to you though.

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceUnitSpec.scala
```diff
@@ -0,0 +1,225 @@
...
+        val testActionManagedIdentity = ActionManagedIdentity(testActionManagedIdentityId, testObjectId, testDisplayName, testMrgCoordinates)
+
+        when(mockDirectoryDAO.loadActionManagedIdentity(testActionManagedIdentityId, samRequestContext)).thenReturn(IO.pure(Option(testActionManagedIdentity)))
+        when(mockCrlService.buildMsiManager(testMrgCoordinates.tenantId, testMrgCoordinates.subscriptionId)).thenReturn(IO.pure(mockMsiManager))
+        when(mockMsiManager.identities()).thenReturn(mockIdentities)
+        doNothing.when(mockIdentities).deleteById(testObjectId.value)
+        when(mockDirectoryDAO.deleteActionManagedIdentity(testActionManagedIdentityId, samRequestContext)).thenReturn(IO.pure(()))
+
+        // Act & Assert
+        azureService.deleteActionManagedIdentity(testActionManagedIdentityId, samRequestContext).unsafeRunSync()
```
Ghost-in-a-Jar, line 171: no assertions? i guess if it completes without error we are good?

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceUnitSpec.scala
```diff
@@ -0,0 +1,225 @@
...
+        doNothing.when(mockIdentities).deleteById(testObjectId.value)
+        doNothing.when(mockIdentities).deleteById(testObjectId2.value)
+        when(mockDirectoryDAO.loadActionManagedIdentity(testActionManagedIdentityId, samRequestContext)).thenReturn(IO.pure(Some(testActionManagedIdentity)))
+        when(mockDirectoryDAO.loadActionManagedIdentity(testActionManagedIdentityId2, samRequestContext)).thenReturn(IO.pure(Some(testActionManagedIdentity2)))
+        when(mockDirectoryDAO.deleteActionManagedIdentity(testActionManagedIdentityId, samRequestContext)).thenReturn(IO.pure(()))
+        when(mockDirectoryDAO.deleteActionManagedIdentity(testActionManagedIdentityId2, samRequestContext)).thenReturn(IO.pure(()))
+        when(mockAzureManagedResourceGroupDAO.deleteManagedResourceGroup(testBillingProfileId, samRequestContext)).thenReturn(IO.pure(1))
+
+        // Act & Assert
+        azureService.deleteManagedResourceGroup(testBillingProfileId, samRequestContext).unsafeRunSync()
```
Ghost-in-a-Jar, line 221: same comment as above test, maybe worth spying on something and asserting a method was called?

Ghost-in-a-Jar: maybe assert that delete action managed identities was called?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -137,6 +158,10 @@ class ResourceServiceSpec
   private val constrainableService =
     new ResourceService(constrainableResourceTypes, constrainablePolicyEvaluatorService, policyDAO, dirDAO, NoExtensions, emailDomain, Set.empty)
 
+  val mockAzureService = mock[AzureService]
+  private val azuredService =
```
Ghost-in-a-Jar: is this a typo should this be azureService? Also isnt there a AzureService class separate from the ResourceService class? Should this val be 'resourceService' instead?

tlangs: This is a `ResourceService` with azure functionality enabled. The`ResourceService` in this test file that doesn't have azure functionality enabled is just called `service`

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceUnitSpec.scala
tlangs: Yup! If one of the mocked functions wasn't called, ScalaTest will complain

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
```diff
@@ -1779,6 +1804,54 @@ class ResourceServiceSpec
...
+        ),
+        ManagedIdentityObjectId(UUID.randomUUID().toString),
+        ManagedIdentityDisplayName(s"${resource.resourceId.value}-${action.value}"),
+        managedResourceGroupCoordinates
+      )
+      runAndWait(dirDAO.createActionManagedIdentity(ami, samRequestContext))
+    }
+    when(mockAzureService.deleteActionManagedIdentity(any[ActionManagedIdentityId], any[SamRequestContext])).thenReturn(IO.unit)
+
+    assert(dirDAO.getAllActionManagedIdentitiesForResource(resource, samRequestContext).unsafeRunSync().nonEmpty)
```
Ghost-in-a-Jar, line 1841: I am confused shouldnt this return something at this line and then after deleting the resource not return anything?

Ghost-in-a-Jar (approved):
Left a couple comments/questions on the tests but overall looks great!

Shakespeared (approved):

rtitle (commented):
I just want to make sure I understand the data flow for these endpoints (is this captured somewhere?). Specifically, where in the flow is this new Sam endpoint being called?

1. User runs a workflow with a private ACR image
2. TES calls Sam `get/createActionManagedIdentity` with the user's credential to obtain an `ACR-pull` action managed identity (how does TES determine that this is a private ACR image?)
3. TES calls WSM `createBatchPool` specifying both the pet and the action managed identities
4. WSM creates the batch pool as normal, assigning both identities
5. TES Runner (which runs on the batch node) uses the assigned action managed identity to pull the image down (it does not call Sam at this point)

Is that about right? cc @jgainerdewar as well.

rtitle:
We talked faces! I'm good :)

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureServiceUnitSpec.scala
tlangs: That's done on lines 216 and 217. If those mocks aren't called, the test will fail.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
tlangs: Ah, this passes because the test runs against a real database, but a fake azure service. I've re-arranged the lines to that it makes a little more sense.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceServiceSpec.scala
tlangs: renamed for clarity.

PR #1411, opened 2024-04-25
ID-614 Ensure Email Uniqueness for Managed Groups

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-614

Sam wasn't checking for email uniqueness across all subject types when creating a managed group. This meant that someone could register a managed group that was already a registered user, and that would cause problems down the line. This change makes group creation check across all subject types, preventing this situation.

This change also adds an additional check to user invitation. Recently, a user invited a group to a workspace before the group was created in Sam. This meant Sam created a new user with the email address of the group. This should not be possible. So, I've made it so that Sam disallows inviting users with the same domain that it uses for groups.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1604,37 +1604,6 @@ resourceTypes = {
     reuseIds = true
   }
 
-  private_azure_container_registry = {
-    actionPatterns = {
-      delete = {
-        description = "Delete this private acr"
-      }
-      read_policies = {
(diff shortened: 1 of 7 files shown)


Ghost-in-a-Jar (approved):

samanehsan (approved):
Lgtm! Thanks for the walk-through 🚶‍♀️

PR #1414, opened 2024-04-30
ID-1174 Add combined user state api call to reduce needed ui api calls

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1174

What:

Adds a combined user state api call in order to reduce the amount of api calls needed from the ui to sam in order to load user state. 

Why:

Right now, Terra-UI makes[ 5 calls at once to Sam to get user information](https://github.com/DataBiosphere/terra-ui/blob/HEAD/src/auth/auth.ts#L479). This led to a recent incident where we added a call to Sam from Terra-UI on login, but forgot to tell Sam to ignore the ToS acceptance for that call.

In general, Terra-UI is very chatty with Terra APIs. This leads to an increased risk of call failure. This new endpoint will make the UI less error-prone in its communication with Sam, as well as simplify the login code in the UI.

How:

I added a new api endpoint that returns the user's samUser, allowances, attributes, terms of service details, and enterprise features.

These are all to replace the multiple api calls in the ui [here](https://github.com/DataBiosphere/terra-ui/blob/HEAD/src/auth/auth.ts#L496) with a single call.

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
```diff
@@ -164,6 +182,31 @@ trait UserRoutesV2 extends SamUserDirectives with SamRequestContextDirectives {
...
+          enterpriseFeatures <- resourceService
+            .listResourcesFlat(
+              samUser.id,
+              Set(ResourceTypeName("enterprise-feature")),
+              Set.empty,
+              Set(ResourceRoleName("user")),
+              Set.empty,
+              includePublic = false,
+              samRequestContext
+            )
```
Ghost-in-a-Jar, line 202: I matched this with [the call that the ui is doing](https://github.com/DataBiosphere/terra-ui/blob/c7623a8574750ca830bd8d48e7fa0c2a415a938a/src/libs/ajax/User.ts#L198)

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/MockSamRoutesBuilder.scala
```diff
@@ -27,6 +27,8 @@ class MockSamRoutesBuilder(allUsersGroup: WorkbenchGroup)(implicit system: Actor
   private val userServiceBuilder: MockUserServiceBuilder = MockUserServiceBuilder()
   private val mockTosServiceBuilder = MockTosServiceBuilder()
 
+  val mockResourceService = mock[ResourceService]
```
Ghost-in-a-Jar, line 30: I didnt want to have to add a whole MockResourceServiceBuilder thing just for this so I made this mock available for mocking in the test i wrote

Shakespeared (approved):

samanehsan (approved):

PR #1420, opened 2024-05-02
ID-1263 Correct Error Deleting Action Managed Identities

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1263

When deleting action managed identities, the query didn't use the composite `sam_resource_action.action` `sam_resource_action.resourceTypeId` key when getting Action Managed Identities to delete. This meant that the subquery returned resource action id's for any action name, even if two different actions on different resource types had the same name. This caused a sql error where a subquery returned more than one result. No bueno! 

Corrected the bug. This is brand new functionality, so bugs like this are bound to happen!

---

**PR checklist**


rtitle (approved):
Thanks for noticing and fixing!

samanehsan (approved):

PR #1425, opened 2024-05-03
ID-1268 Private Endpoint Access Resource Type

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1268

We don't know what we want yet, so lets make anything possible.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1666,22 +1666,22 @@ resourceTypes = {
     reuseIds = true
   }
 
-  azure_private_endpoint_access = {
+  azure_managed_identity = {
     actionPatterns = {
       delete = {
-        description = "Delete this private endpoint access"
+        description = "Delete this azure managed identity"
       }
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -1665,6 +1665,37 @@ resourceTypes = {
     allowLeaving = false
     reuseIds = true
   }
+
+  azure_private_endpoint_access = {
```
dvoet: `azure_managed_identity`?

tlangs: Are we representing an identity, or an endpoint, or the thing on the other side of the endpoint?

rtitle: +1 `azure_managed_identity`, this is really modeling access to a managed identity (we have yet to model the private endpoint)

dvoet (approved):

rtitle (approved):

PR #1410, opened 2024-04-25
TOAZ-351 openIdConnect in api-docs.yaml

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/TOAZ-351

update to latest version of workbench-libs oauth2 and adopt open api securitySchemes:
```
    oidc:
      type: openIdConnect
      openIdConnectUrl: OPEN_ID_CONNECT_URL
      x-tokenName: id_token
```

---

**PR checklist**

- [ ] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've filled out the [Security Risk Assessment](https://sdarq.dsp-appsec.broadinstitute.org/jira-ticket-risk-assesment) (requires Broad Internal network access) and attached the result to the JIRA ticket

PR #1430, opened 2024-05-07
[ID-1266] Fix missing auth methods in Sam generated client

rtitle:
**Ticket:**

https://broadworkbench.atlassian.net/browse/ID-1266

**What:**

The latest Sam generated client has no way to authenticate! The `setAccessToken` method we were previously using throws an exception. After this PR, clients will need to use `setBearerToken` instead of `setAccessToken`.

**Why:**

I noticed this when [upgrading WSM](https://github.com/DataBiosphere/terra-workspace-manager/pull/1724) to the latest Sam client version. I think this broke after we recently [updated](https://github.com/broadinstitute/sam/pull/1410) the OpenAPI security schemes for Sam (and other services). `openapi-generator` [does not support](https://openapi-generator.tech/docs/generators/java#security-feature) the `openIdConnect` security scheme, which is what we now use.

**How:**

1. Add a new `bearerAuth` scheme to the Sam OpenAPI docs. The Sam swagger page now has this option:

image

2. Update `openapi-generator-cli` to the latest version

3. Added an `ApiClient.mustache` file based off https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache. The only change is to add a dummy record to the `authentications` Map for the `oidc` security scheme. `openapi-generator` unfortunately validates that every security scheme has an entry in the Map.

I don't love checking in mustache code to Sam, but don't see a better way to work around lack of `openIdConnect` support.

---

**PR checklist**


codegen_java/templates/ApiClient.mustache
```diff
@@ -0,0 +1,1924 @@
...
+        initHttpClient();
+
+        // Setup authentications (key: authentication name, value: authentication).{{#authMethods}}{{#isBasic}}{{#isBasicBasic}}
+        authentications.put("{{name}}", new HttpBasicAuth());{{/isBasicBasic}}{{#isBasicBearer}}
+        authentications.put("{{name}}", new HttpBearerAuth("{{scheme}}"));{{/isBasicBearer}}{{/isBasic}}{{#isApiKey}}
+        authentications.put("{{name}}", new ApiKeyAuth({{#isKeyInHeader}}"header"{{/isKeyInHeader}}{{#isKeyInQuery}}"query"{{/isKeyInQuery}}{{#isKeyInCookie}}"cookie"{{/isKeyInCookie}}, "{{keyParamName}}"));{{/isApiKey}}{{#isOAuth}}
+        authentications.put("{{name}}", new OAuth());{{/isOAuth}}{{/authMethods}}{{#withAWSV4Signature}}
+        authentications.put("AWS4Auth", new AWS4Auth());{{/withAWSV4Signature}}
+        
+        /*
```
rtitle, line 147: I tried to indicate where the Terra changes are with this comment

scripts/gen_java_client.sh
```diff
@@ -1,5 +1,5 @@
 set -e
 
-docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli:v7.3.0 generate -i /local/src/main/resources/swagger/api-docs.yaml -g java -o /local/codegen_java --api-package org.broadinstitute.dsde.workbench.client.sam.api --model-package org.broadinstitute.dsde.workbench.client.sam.model --template-dir /local/codegen_java/templates --library okhttp-gson --additional-properties useJakartaEe=true,disallowAdditionalPropertiesIfNotPresent=false
+docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli:v7.5.0 generate -i /local/src/main/resources/swagger/api-docs.yaml -g java -o /local/codegen_java --api-package org.broadinstitute.dsde.workbench.client.sam.api --model-package org.broadinstitute.dsde.workbench.client.sam.model --template-dir /local/codegen_java/templates --library okhttp-gson --additional-properties useJakartaEe=true,disallowAdditionalPropertiesIfNotPresent=false
```
rtitle, line 3: Updated `openapi-generator-cli` version

src/main/resources/swagger/api-docs.yaml
```diff
@@ -4717,4 +4718,7 @@ components:
     oidc:
       type: openIdConnect
       openIdConnectUrl: OPEN_ID_CONNECT_URL
-      x-tokenName: id_token
\ No newline at end of file
+      x-tokenName: id_token
+    bearerAuth:
```
dvoet, line 4722: do you need `bearerAuth` given you are adding oidc as a `HttpBearerAuth`?

rtitle: `bearerAuth` adds `ApiClient.setBearerToken()`

Without it I can still do:
```
    ((HttpBearerAuth) apiClient.getAuthentications().get("oidc")).setBearerToken(token);
```
but that feels a little unwieldy.

rtitle: I guess another idea is I could mustache `setAccessToken` to set the bearer token for `oidc`. But I kinda want to limit how much custom code we add…

dvoet: why doesn't `apiClient.setBearerToken(token);` work?

rtitle: We connected in slack. `apiClient.setBearerToken` is conditionally only added when `bearerAuth` is included in the YAML.

We think it's reasonable to support both oidc (for interactive flows) and bearer (for programmatic access) security schemes.

tlangs (approved):

Ghost-in-a-Jar (approved):

codegen_java/templates/ApiClient.mustache
```diff
@@ -0,0 +1,1921 @@
...
+        throw new RuntimeException("No API key authentication configured!");
+    }
+
+    /**
+     * Helper method to set access token for the first OAuth2 authentication.
+     *
+     * @param accessToken Access token
+     */
+    public void setAccessToken(String accessToken) {
+        /*
```
rtitle, line 632: The other Terra addition

rtitle:
I tested this on a BEE and it is working with WSM 🎉 Things are green so I'm going to go ahead and auto-merge.

PR #1440, opened 2024-05-22
ID-1282 Finalize Azure Private Resource Types

tlangs:
Ticket: [<Link to Jira ticket>](https://broadworkbench.atlassian.net/browse/ID-1282)

What:

We've done a lot of technical feasibility work, and settled on what resource types need to exist for Private Azure Container Registries and Storage Accounts. This gets rid of previous resource types and actions, which will need to be manually removed from production.

Since resource types are upserted on boot, current resources should continue to work. We should migrate away from them and delete the resource types that are no longer in use.

Why:

Representing resources types as their logical Terra functional pieces gives us a lot of flexibility around access control and the evolution of features. A raw `azure_managed_identity` doesn't let Sam do what it does best: manage users' access to things. By representing an `azure_private_storage_account` as it's own resource, we can use Sam to manage which users can `read` and which user's can `write`. The Action Managed Identities are just an implementation detail of how the underlying cloud resource is accessed.

How:

Update `reference.conf`

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1655,9 +1655,6 @@ resourceTypes = {
       write = {
         description = "write to the private azure storage account"
       }
-      identify = {
-        description = "DEPRECATED - use the identity to access the storage account"
-      }
       "share_policy::owner" = {
         description = "change the membership of the owner policy for this private azure storage account"
       }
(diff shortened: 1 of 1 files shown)


rtitle (approved):

jgainerdewar (approved):

Ghost-in-a-Jar (approved):

src/main/resources/reference.conf
```diff
@@ -1649,54 +1649,32 @@ resourceTypes = {
...
+        description = "change the membership of the writer policy for this private azure storage account"
       }
     }
-    ownerRoleName = "admin"
+    ownerRoleName = "owner"
     roles = {
-      admin = {
-        roleActions = ["delete", "read_policies", "use", "share_policy::admin", "share_policy::user", "identify"]
+      owner = {
+        roleActions = ["delete", "read_policies", "share_policy::owner", "share_policy::reader", "share_policy::writer", "read", "write"]
```
Ghost-in-a-Jar, line 1671: do they need pull_image?

tlangs: I think it makes sense that an `owner` has all the permissions of other roles.

PR #1451, opened 2024-06-04
[TOAZ-355] [TOAZ-356] Use Managed Identity auth when running Azure Control Plane, support for Service Catalog deployed Azure Managed Apps 

aaronegrant:
Refactor SAM ARM authorization code to support using a Managed Identity app id instead of Service Principal clientid/clientsecret/tenantid if optional config 'AZURE_MANAGED_APP_WORKLOAD_CLIENT_ID' is set

Service Catalog deployed app support, enabled with 'AZURE_SERVICE_CATALOG_ENABLED' config

Sam maintains its own list of Terra managed app plans https://github.com/broadinstitute/sam/blob/develop/src/main/resources/sam.conf#L192 and validates a billing profile managed app against this list during billing profile creation.
Service catalog managed apps do not have a plan defined so we will need change the validation in Sam to accept this.


src/main/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureService.scala
```diff
@@ -228,31 +228,41 @@ class AzureService(
   /** Resolves a managed resource group in Azure and returns the terra.billingProfileId tag value. This is used for access control checks during route handling.
     */
   def getBillingProfileId(request: GetOrCreatePetManagedIdentityRequest, samRequestContext: SamRequestContext): IO[Option[BillingProfileId]] =
-    // get the billing profile id from the database
-    // if not there, for backwards compatibility, get the billing profile id from a tag on the Azure resource
```
dvoet, line 232: this backwards compatibility is no longer needed

src/test/scala/org/broadinstitute/dsde/workbench/sam/azure/AzureRoutesSpec.scala
```diff
@@ -139,26 +119,6 @@ class AzureRoutesSpec extends AnyFlatSpec with Matchers with ScalatestRouteTest
     }
   }
 
-  it should "successfully create a pet managed identity for a user using MRG Azure tag for backwards compatibility" in {
```
dvoet, line 142: backwards compatibility removed

dvoet (approved):

Shakespeared (approved):

PR #1457, opened 2024-06-12
ID-1306 CBAS Submission, Filtering by Resource Parent, and Getting Resource Creator

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1306

To help our Workflows friends, we need to add a `cbas-submissions` resource type, with `owner` and `reader` roles. This will allow CBAS to not need to store Sam User IDs, and instead let Sam do what Sam does best! 

We also need to augment the `listResourcesV2` endpoint so that it can take in a `parentResourceId` and constrain its results to resources with only that `parentResourceId`. 

I've added an optional query parameter to the `listResourcesV2` endpoint, which takes in a list of resource parents to filter on. The resource parents are formatted as `resourceTypeName:resourceId`, since we need both to be able to do a proper DB join. The `:` character is not allowed in resource names, so its an OK character to use as a separator. 

The response from `listResourcesV2` has not changed, as doing so might break any consumers of the endpoint. If there are no consumers that would break, it may be worth it to include this parent information in the response. 

I've added a new endpoint: `getResourceCreator`. If a user has the `read_creator` action, they'll be able to get the email address of the creator of the resource. Its odd that Sam wasn't storing this before!

---

**PR checklist**

### DIFF
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240613_resource_created_by.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240613_resource_created_by.xml
@@ -8,9 +8,10 @@
     <changeSet logicalFilePath="dummy" author="tlangs" id="resource_created_by">
         <addColumn tableName="sam_resource">
             <column name="created_by" type="VARCHAR">
-                <constraints nullable="true" foreignKeyName="FK_SR_CREATOR" referencedTableName="SAM_USER" referencedColumnNames="id"/>
+                <constraints nullable="true"/>
             </column>
         </addColumn>
+        <addForeignKeyConstraint baseTableName="sam_resource" baseColumnNames="created_by"  constraintName="FK_SR_CREATOR" referencedTableName="SAM_USER" referencedColumnNames="id" onDelete="SET NULL" />
     </changeSet>
 
 </databaseChangeLog>
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240613_resource_created_by.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240613_resource_created_by.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="tlangs" id="resource_created_by">
+        <addColumn tableName="sam_resource">
+            <column name="created_by" type="VARCHAR">
+                <constraints nullable="true" foreignKeyName="FK_SR_CREATOR" referencedTableName="SAM_USER" referencedColumnNames="id"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
+</databaseChangeLog>
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -172,6 +172,7 @@ resourceTypes = {
           wds-instance = ["owner"]
           kubernetes-app = ["manager"]
           kubernetes-app-shared = ["owner", "user"]
+          cbas-submission = ["reader"]
         }
       }
       application = {
@@ -188,6 +189,7 @@ resourceTypes = {
           google-project = ["pet-creator"]
           wds-instance = ["writer"]
           kubernetes-app-shared = ["user"]
+          cbas-submission = ["reader"]
         }
       }
       reader = {
@@ -198,6 +200,7 @@ resourceTypes = {
           google-project = ["pet-creator"]
           wds-instance = ["reader"]
           kubernetes-app-shared = ["user"]
+          cbas-submission = ["reader"]
         }
       }
       discoverer = {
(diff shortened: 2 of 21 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ResourceRoutes.scala
```diff
@@ -605,38 +605,65 @@ trait ResourceRoutes extends SamUserDirectives with SecurityDirectives with SamM
...
+          actions: Option[String],
+          parentResources: Option[String],
+          includePublic: Boolean,
+          format: String
+      ) =>
+        val parsedResourceTypes = resourceTypes.map(_.split(",").map(ResourceTypeName(_)).toSet).getOrElse(Set.empty)
+        val parsedPolicies = policies.map(_.split(",").map(AccessPolicyName(_)).toSet).getOrElse(Set.empty)
+        val parsedRoles = roles.map(_.split(",").map(ResourceRoleName(_)).toSet).getOrElse(Set.empty)
+        val parsedActions = actions.map(_.split(",").map(ResourceAction(_)).toSet).getOrElse(Set.empty)
+        val parsedParentResourceIds = parentResources
```
tlangs, line 635: a little cleanup here to keep our code DRY

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
```diff
@@ -197,6 +197,7 @@ trait UserRoutesV2 extends SamUserDirectives with SamRequestContextDirectives {
               Set.empty,
               Set(ResourceRoleName("user")),
               Set.empty,
+              Set.empty,
```
tlangs, line 200: Listing all user resources doesn't need to filter on parent, so an empty set is provided here.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1816,10 +1824,15 @@ class PostgresAccessPolicyDAO(
     val notNullConstraintRoleAction =
       samsqls"and not (${resourceRole.role} is null and ${resourceAction.action} is null)"
     val notNullConstraintPolicyAction = samsqls"and not (${resourceAction.action} is null)"
+    val resourceParentsConstraint =
+      if (parentResourceIds.nonEmpty) samsqls"and ${parentResourceIds
+          .map(parentResourceId => samsqls"(${resourceParent.resourceTypeId} = ${resourceTypePKsByName(parentResourceId.resourceTypeName)} and ${resourceParent.name} = ${parentResourceId.resourceId})")
+          .reduce((acc, clause) => samsqls"$acc or $clause")}"
+      else samsqls""
```
tlangs, line 1850: A series of `or` clauses is needed here because we're joining on the fully-qualified resource id, which is actually 2 values.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240613_resource_created_by.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="tlangs" id="resource_created_by">
+        <addColumn tableName="sam_resource">
+            <column name="created_by" type="VARCHAR">
```
tlangs, line 10: Adding the column `created_by` isn't just useful for submissions. Its really odd that Sam doesn't know what user created a resource!

src/main/resources/reference.conf
```diff
@@ -1674,6 +1677,32 @@ resourceTypes = {
...
+        roleActions = ["abort", "create_with_parent"]
+        includedRoles = ["reader"]
+      }
+      reader = {
+        roleActions = ["read", "read_creator"]
+      }
+    }
+    allowLeaving = false
+    reuseIds = false
+  }
```
tlangs: This will allow CBAS to no longer need to store Sam User IDs!

src/main/resources/reference.conf
```diff
@@ -1674,6 +1677,35 @@ resourceTypes = {
...
+      abort = {
+        description = "Abort this cbas-submission"
+      }
+      read = {
+        description = "read information about the submission"
+      }
+      create_with_parent = {
+        description = "Enables creating the request object with a parent"
+      }
+      read_creator = {
```
tlangs, line 1692: This resource action will allow those who have it to be able to get the resource creator's email. This is similar to `read_policies`, where users are allowed to get the emails of people in policies on a resource.

src/main/resources/reference.conf
```diff
@@ -1674,6 +1677,35 @@ resourceTypes = {
     allowLeaving = false
     reuseIds = true
   }
+
+  cbas-submission = {
```
tlangs, line 1681: This will allow CBAS to no longer need to store Sam User IDs!

src/main/resources/reference.conf
```diff
@@ -1674,6 +1677,35 @@ resourceTypes = {
...
+      read_creator = {
+        description = "Enables reading the creator of the resource"
+      }
+    }
+    ownerRoleName = "owner"
+    roles = {
+      owner = {
+        roleActions = ["abort", "create_with_parent"]
+        includedRoles = ["reader"]
+      }
```
cjllanwarne, line 1701: How hard is it to change these after creation if we wanted to (say) have a "canceller" role separate from the owner/creator?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2976,6 +2985,44 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+  /api/resources/v2/{resourceTypeName}/{resourceId}/creator:
```
cjllanwarne, line 2988: This is fine but it's probably more helpful to have a way to batch the lookup of `resource => createdBy`. Either here or in the `listResources` response perhaps?

PR #1465, opened 2024-06-26
ID-1319 Fix missing terms of service error.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1319

What:

This pr fixes a bug that was introduced to the UI in a previous change we made. The bug was introduced in this PR where we swapped out various calls that the ui does to sam on login, to a single call. The issue is that previously the terms of service call was swallowing 404 errors when the user has no tos acceptance history. Now instead of throwing a 404 we will just return a successful response that reflects that the user hasn't accepted any tos previously.

Example response shape instead of 404:
{
  latestAcceptedVersion: None,
  acceptedOn: None,
  permitsSystemUsage: false,
  isCurrentVersion: false,
}

https://github.com/DataBiosphere/terra-ui/pull/4844/files

How:

Instead of returning a 404 when there is no tos history for the user we return a successful response that reflects the tos state of the user. There should not be any change in the UI required to fix the bug after this PR is merged. 

Start in [TosService.scala](https://github.com/broadinstitute/sam/pull/1465/files#diff-0a231acaf05c514471578d80e97b2138fb42aae179a932068cc550e0fa2f78d0)

---

**PR checklist**


tlangs:
Are we violating CRUD principals here? If someone hasn't accepted any ToS, shouldn't their status be not found?

Shakespeared (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -2078,6 +2078,9 @@ class GoogleExtensionSpec(_system: ActorSystem)
 
     googleExtensions.fireAndForgetNotifications(messages)
 
+    // fireAndForgetNotifications is asynchronous, so we need to wait for the future to complete
+    Thread.sleep(500)
```
Ghost-in-a-Jar, line 2082: This test failed in gha so I added this to reduce flakeyness

Ghost-in-a-Jar: I reran and it passed

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
```diff
@@ -189,7 +189,17 @@ trait UserRoutesV2 extends SamUserDirectives with SamRequestContextDirectives {
         for {
           allowances <- userService.getUserAllowances(samUser, samRequestContext)
           maybeAttributes <- userService.getUserAttributes(samUser.id, samRequestContext)
-          termsOfServiceDetails <- tosService.getTermsOfServiceDetailsForUser(samUser.id, samRequestContext)
+          termsOfServiceDetails <- tosService.getTermsOfServiceDetailsForUser(samUser.id, samRequestContext).recoverWith {
+            case e: WorkbenchExceptionWithErrorReport if e.errorReport.statusCode == Option(StatusCodes.NotFound) =>
```
tlangs: Could we make `getTermsOfServiceDetailsForUser` return an `Option`? I'd love to not rely on exceptions for expected control flow. We should make the ToS Routes translate a `None` to a 404.

tlangs (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/TosService.scala
```diff
@@ -96,28 +96,17 @@ class TosService(
...
-  private def getLatestTermsOfServiceAcceptance(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[SamUserTos] = for {
-    maybeTermsOfServiceRecord <- directoryDao.getUserTermsOfService(userId, samRequestContext, Option(TosTable.ACCEPT))
-    latestUserTermsOfService <- maybeTermsOfServiceRecord
-      .map(IO.pure)
-      .getOrElse(
-        IO.raiseError(
-          new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Could not find current terms of service for user:${userId}"))
-        )
-      )
-  } yield latestUserTermsOfService
```
Ghost-in-a-Jar, line 119: This fn is unnecessary now

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRoutes.scala
```diff
@@ -120,7 +121,10 @@ trait TermsOfServiceRoutes extends SamUserDirectives with SamRequestContextDirec
                 pathEndOrSingleSlash {
                   getWithTelemetry(samRequestContext, userIdParam(requestUserId)) {
                     complete {
-                      tosService.getTermsOfServiceDetailsForUser(requestUserId, samRequestContext)
+                      tosService.getTermsOfServiceDetailsForUser(requestUserId, samRequestContext).map {
+                        case response @ TermsOfServiceDetails(None, None, false, false) => StatusCodes.NotFound -> response
+                        case response => StatusCodes.OK -> response
```
tlangs: Sorry for being a curmudgeon. Can `getTermsOfServiceDetailsForUser` itself return an `Optional`? The "Combined" response should modify the `Optional.Empty` to be what it expects.

tlangs (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRoutes.scala
Ghost-in-a-Jar: ok yeah that sounds good no worries

tlangs (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/TermsOfServiceRouteSpec.scala
```diff
@@ -221,8 +222,12 @@ class TermsOfServiceRouteSpec extends AnyFunSpec with Matchers with ScalatestRou
...
         .withDisabledUser(Generator.genWorkbenchUserGoogle.sample.get)
 
-      Get("/api/termsOfService/v1/user/self/history") ~> mockSamRoutesBuilder.build.route ~> check {
-        status shouldEqual StatusCodes.NotFound
+      Get("/api/termsOfService/v1/user/self/history") ~> Route.seal(mockSamRoutesBuilder.build.route) ~> check {
+        status shouldEqual StatusCodes.OK
+        withClue(s"${responseAs[String]} is not parsable as an instance of `TermsOfServiceHistory`.") {
+          val response = responseAs[TermsOfServiceHistory]
+          response.history shouldBe empty
+        }
```
Ghost-in-a-Jar, line 230: I noticed this test didnt actually reflect the behavior of the getTermsOfServiceHistoryForUser fn

PR #1466, opened 2024-06-26
[TOAZ-363] Allow UAMI accounts to be added to service admin whitelist

bennettn4:
Ticket: (https://broadworkbench.atlassian.net/browse/TOAZ-363)

What:

UAMI accounts do not have OIDC_email field populated through accesstoken, need to be dynamically built from b2cid

Why:

Needed if Thurloe is running as a user assigned managed identity on azure and needs service account admin whitelist

How:

Additional check in service account admin whitelist based UAMI email

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -49,7 +49,10 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with
...
     Directives
       .mapInnerRoute { r =>
-        if (!adminConfig.serviceAccountAdmins.contains(oidcHeaders.email)) {
+        if (
+          !adminConfig.serviceAccountAdmins.contains(oidcHeaders.email) &&
+          !adminConfig.serviceAccountAdmins.contains(WorkbenchEmail(s"${oidcHeaders.externalId.map(_.value).toOption.orNull}@uami.terra.bio"))
+        ) {
           reject(AuthorizationFailedRejection)
         } else {
           logger.info(s"Handling request for service admin account: ${oidcHeaders.email}")
```
dvoet: I think this would be easier to follow in the affirmative, without the nots
```
        if (
          adminConfig.serviceAccountAdmins.contains(oidcHeaders.email) ||
          adminConfig.serviceAccountAdmins.contains(WorkbenchEmail(s"${oidcHeaders.externalId.map(_.value).toOption.orNull}@uami.terra.bio"))
        ) {
          // do the logic on starting on line 58
        } else {
          reject(AuthorizationFailedRejection)
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
```diff
@@ -49,7 +49,10 @@ trait StandardSamUserDirectives extends SamUserDirectives with LazyLogging with
   def asAdminServiceUser: Directive0 = requireOidcHeaders.flatMap { oidcHeaders =>
     Directives
       .mapInnerRoute { r =>
-        if (!adminConfig.serviceAccountAdmins.contains(oidcHeaders.email)) {
+        if (
+          !adminConfig.serviceAccountAdmins.contains(oidcHeaders.email) &&
+          !adminConfig.serviceAccountAdmins.contains(WorkbenchEmail(s"${oidcHeaders.externalId.map(_.value).toOption.orNull}@uami.terra.bio"))
```
dvoet: It bothers me that if the external id is absent, this ends up checking `null@uami.terra.bio`. I know that that is not real but it also should never be treated as valid. We also should only check for this kind of email if `oidcHeaders.email` is empty.
I suggest using a function like
```
  private def isAdminUami(oidcHeaders: OIDCHeaders) = {
    val maybeUamiEmail = oidcHeaders.externalId.map(b2cId => WorkbenchEmail(s"${b2cId.value}@uami.terra.bio")).toOption
    oidcHeaders.email.value.isEmpty && maybeUamiEmail.isDefined && adminConfig.serviceAccountAdmins.contains(maybeUamiEmail.get)
  }
```

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
bennettn4: Inverted boolean logic for readability in [ea35970](https://github.com/broadinstitute/sam/pull/1466/commits/ea359701ed0994c4ab517a75dafe378ab117b05c)

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala
bennettn4: Added suggested helper function in [ea35970](https://github.com/broadinstitute/sam/pull/1466/commits/ea359701ed0994c4ab517a75dafe378ab117b05c)

dvoet (approved):

PR #1470, opened 2024-06-29
PROD-972 ID-1323 Eagerly Remove Google Group Members

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/PROD-972
https://broadworkbench.atlassian.net/browse/ID-1323

What:

When we hit the 24-hour quota for the Google Directory API, Google Group synchronization grinds to a halt. Any membership removals that should happen fail to do so until the 24-hour quota rolls over. This can lead us to fail to adhere to SLAs for access revocation.

To mitigate this, I'm making removals from policy groups immediately hit Google to remove the member. Since we're limited on the group membership read quota, this operation should succeed even if that quota is reached. 

To make this change as narrow as possible, messages will still be published to the Pub/Sub topic. It will be a no-op when Sam gets the group membership, since the user will already have been removed.
---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMessageReceiver.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMessageReceiver.scala
@@ -67,6 +67,8 @@ class GoogleGroupSyncMessageReceiver(groupSynchronizer: GoogleGroupSynchronizer)
     import WorkbenchIdentityJsonSupport._
     import org.broadinstitute.dsde.workbench.sam.google.SamGoogleModelJsonSupport._
 
+    val foo = report.map { case (group, syncReportItems) => (group, syncReportItems.groupBy(_.operation)) }
+
     if (errorReports.isEmpty) {
       logger.info(s"synchronized google group", StructuredArguments.raw("syncDetail", report.toJson.compactPrint))
       consumer.ack()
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMessageReceiver.scala
```diff
@@ -67,6 +67,8 @@ class GoogleGroupSyncMessageReceiver(groupSynchronizer: GoogleGroupSynchronizer)
     import WorkbenchIdentityJsonSupport._
     import org.broadinstitute.dsde.workbench.sam.google.SamGoogleModelJsonSupport._
 
+    val foo = report.map { case (group, syncReportItems) => (group, syncReportItems.groupBy(_.operation)) }
```
Shakespeared: is this `foo` for testing?

Shakespeared (approved):
seems reasonable, left one question

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMessageReceiver.scala
tlangs: yup! Good catch!

bstitson-broad (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSyncMessageReceiver.scala
tlangs: I'm honestly surprised the compiler didn't yell at me for an unused variable.

PR #1472, opened 2024-07-01
PROD-972 ID-1324 Deduplicate group synchronization calls to google.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1324

What:

We make a LOT of calls to the google groups api, a lot of which are duplicated and unnecessary. This PR aims to reduce the number of calls to google group apis during the group syncing 

Why:

We are hitting the quota for listing group memberships every day and ended up with a prod incident over the weekend. 

How:

The plan is to record the last synchronized version and current version of group, only synchronize group if group version is greater than last synchronized version.

Pseudocode:

```
Pull message from pubsub, check current version of group against last synchronized version
if equal:
	do nothing
if the group version is greater than last synchronized version
	sychronize the group

	update last sycnronized version to the version of the group when the synchronization started
```

In DB migration set initial values of current group version to 1 and last synchronized version to 0

Testing/monitoring notes:
* can test manually by pushing our own messages into the queue
* add unit tests in GoogleExtensionSpec
* add some telemetry around no-ops and list group membership 
---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/DirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/DirectoryDAO.scala
@@ -45,7 +45,9 @@ trait DirectoryDAO {
 
   def isGroupMember(groupId: WorkbenchGroupIdentity, member: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Boolean]
 
-  def updateSynchronizedDate(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Unit]
+  def updateSynchronizedDateAndVersion(group: WorkbenchGroup, samRequestContext: SamRequestContext): IO[Unit]
+
+  def updateGroupUpdatedDateAndVersionWithSession(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Unit]
 
   def getSynchronizedDate(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Option[Date]]
 
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
@@ -85,7 +85,7 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val 
         val r = ResourceTable.syntax("r")
         val rt = ResourceTypeTable.syntax("rt")
 
-        samsql"""select ${g.result.email}, ${gm.result.memberUserId}, ${sg.result.name}, ${p.result.name}, ${r.result.name}, ${rt.result.name}
+        samsql"""select ${g.result.email}, ${gm.result.memberUserId}, ${sg.result.name}, ${p.result.name}, ${r.result.name}, ${rt.result.name}, ${g.result.version}, ${g.result.lastSynchronizedVersion}
                   from ${GroupTable as g}
                   left join ${GroupMemberTable as gm} on ${g.id} = ${gm.groupId}
                   left join ${GroupTable as sg} on ${gm.memberGroupId} = ${sg.id}
@@ -100,7 +100,9 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val 
               rs.stringOpt(sg.resultName.name).map(WorkbenchGroupName),
               rs.stringOpt(p.resultName.name).map(AccessPolicyName(_)),
               rs.stringOpt(r.resultName.name).map(ResourceId(_)),
-              rs.stringOpt(rt.resultName.name).map(ResourceTypeName(_))
+              rs.stringOpt(rt.resultName.name).map(ResourceTypeName(_)),
+              rs.get[Integer](g.resultName.version),
+              rs.get[Option[Integer]](g.resultName.lastSynchronizedVersion)
             )
           }
           .list()
@@ -112,13 +114,16 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val 
       } else {
         val email = results.head._1
         val members: Set[WorkbenchSubject] = results.collect {
-          case (_, Some(userId), None, None, None, None) => userId
-          case (_, None, Some(subGroupName), None, None, None) => subGroupName
-          case (_, None, Some(_), Some(policyName), Some(resourceName), Some(resourceTypeName)) =>
+          case (_, Some(userId), None, None, None, None, _, _) => userId
+          case (_, None, Some(subGroupName), None, None, None, _, _) => subGroupName
+          case (_, None, Some(_), Some(policyName), Some(resourceName), Some(resourceTypeName), _, _) =>
             FullyQualifiedPolicyId(FullyQualifiedResourceId(resourceTypeName, resourceName), policyName)
         }.toSet
 
-        Option(BasicWorkbenchGroup(groupName, members, email))
+        val version = results.head._7
+        val lastSynchronized = results.head._8
+
+        Option(BasicWorkbenchGroup(groupName, members, email, version, lastSynchronized))
       }
 
   override def loadGroupEmail(groupName: WorkbenchGroupName, samRequestContext: SamRequestContext): IO[Option[WorkbenchEmail]] =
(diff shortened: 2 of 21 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -169,21 +174,29 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
-  override def updateSynchronizedDate(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Unit] =
-    serializableWriteTransaction("updateSynchronizedDate", samRequestContext) { implicit session =>
+  override def updateSynchronizedDateAndVersion(group: WorkbenchGroup, samRequestContext: SamRequestContext): IO[Unit] =
+    serializableWriteTransaction("updateSynchronizedDateAndVersion", samRequestContext) { implicit session =>
       val g = GroupTable.column
-      samsql"update ${GroupTable.table} set ${g.synchronizedDate} = ${Instant.now()} where ${g.id} = (${workbenchGroupIdentityToGroupPK(groupId)})"
+      samsql"""update ${GroupTable.table}
+              set ${g.synchronizedDate} = ${Instant.now()},
+                  ${g.lastSynchronizedVersion} = ${group.version}
+              where ${g.id} = (${workbenchGroupIdentityToGroupPK(group.id)})"""
```
dvoet: I think here you need another condition to avoid downgrading the version
`and (COALESCE(${g.lastSynchronizedVersion}, 0) < ${group.version}`

I think this situation would be rare and require an order of events like:
1. thread A picks up a message and loads group with version 2
2. group is updated to version 3
3. thread B picks up later message and loads group with version 3
4. thread B completes sync and sets latest sync version to 3
5. thread A wakes up and completes sync and sets latest sync version to 2
I think this requires a long nap by thread A, but stranger things have happened.

dvoet:
I think you also need to update the group version somewhere in or before [GoogleExtension.onGroupUpdate](https://github.com/broadinstitute/sam/blob/0b43abd3df9e2f74e36d0ebee26f3a1f846cccb7/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala#L210). This is the logic that figures out all the policies related to an auth domain that must be synced. The related policies need to have their versions bumped so that the GoogleGroupSynchronizer doesn't ignore them.

dvoet:
also need to update the group version in [overwritePolicyMembersInternal](https://github.com/broadinstitute/sam/blob/0b43abd3df9e2f74e36d0ebee26f3a1f846cccb7/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala#L992)

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
Ghost-in-a-Jar: Ah right I forgot about that case

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -192,7 +192,12 @@ class GoogleExtensionSpec(_system: ActorSystem)
 
       added.foreach(email => verify(mockGoogleDirectoryDAO).addMemberToGroup(target.email, WorkbenchEmail(email.value.toLowerCase)))
       removed.foreach(email => verify(mockGoogleDirectoryDAO).removeMemberFromGroup(target.email, WorkbenchEmail(email.value.toLowerCase)))
-      verify(mockDirectoryDAO).updateSynchronizedDate(target.id, samRequestContext)
+
+      target match {
+        case _: BasicWorkbenchGroup => verify(mockDirectoryDAO).updateSynchronizedDateAndVersion(target, samRequestContext)
+        case p: AccessPolicy =>
+          verify(mockDirectoryDAO).updateSynchronizedDateAndVersion(p.copy(members = p.members + CloudExtensions.allUsersGroupName), samRequestContext)
+      }
```
Ghost-in-a-Jar, line 200: This started failing because updateSynchronizedDateAndVersion was being called for the policy with the addition of the all users group in the policy. Still trying to get my head around exactly why. A lot is mocked out here and this test is really only focused around group emails so I think this change is fine.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -298,6 +287,7 @@ class ResourceService(
         } else IO.unit
       policies <- listResourcePolicies(resource, samRequestContext)
       _ <- accessPolicyDAO.addResourceAuthDomain(resource, authDomains, samRequestContext)
+      _ <- policies.traverse(p => directoryDAO.updateGroupUpdatedDateAndVersionWithSession(FullyQualifiedPolicyId(resource, p.policyName), samRequestContext))
```
Ghost-in-a-Jar, line 290: This updates all of the policies when an auth domain is added to a resource

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/DirectoryDAO.scala
```diff
@@ -45,7 +45,9 @@ trait DirectoryDAO {
 
   def isGroupMember(groupId: WorkbenchGroupIdentity, member: WorkbenchSubject, samRequestContext: SamRequestContext): IO[Boolean]
 
-  def updateSynchronizedDate(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Unit]
+  def updateSynchronizedDateAndVersion(group: WorkbenchGroup, samRequestContext: SamRequestContext): IO[Unit]
```
Ghost-in-a-Jar, line 48: We want to pass in the group here so that we can use the version of the group that is held in memory when the sync started.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/DirectoryDAO.scala
```diff
@@ -45,7 +45,9 @@ trait DirectoryDAO {
...
+
+  def updateGroupUpdatedDateAndVersionWithSession(groupId: WorkbenchGroupIdentity, samRequestContext: SamRequestContext): IO[Unit]
```
Ghost-in-a-Jar, line 50: This is needed to update the policies in resource service and it just creates a write db session and then calls updateGroupUpdatedDateAndVersion

Ghost-in-a-Jar: updateGroupUpdatedDateAndVersion is also only on PostgresGroupDAO which ResourceService doesnt have access to

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -231,10 +233,18 @@ consistent "has a" relationship is tracked by this ticket: https://broadworkbenc
...
+    lastSynchronizedVersion: Option[Int] = None
 )
 
-@Lenses final case class BasicWorkbenchGroup(id: WorkbenchGroupName, members: Set[WorkbenchSubject], email: WorkbenchEmail) extends WorkbenchGroup
+@Lenses final case class BasicWorkbenchGroup(
+    id: WorkbenchGroupName,
+    members: Set[WorkbenchSubject],
+    email: WorkbenchEmail,
+    version: Int = 1,
+    lastSynchronizedVersion: Option[Int] = None
```
Ghost-in-a-Jar, line 246: Defaulting these values (including in the other models) was to avoid having to change BasicWorkbenchGroup in a ton of places. With the proper tests in place I think this is ok.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleGroupSynchronizer.scala
```diff
@@ -51,7 +54,15 @@ class GoogleGroupSynchronizer(
...
       for {
-        group <- loadSamGroup(groupId, samRequestContext)
+        group: WorkbenchGroup <- loadSamGroup(groupId, samRequestContext)
+        // If group.version > group.lastSynchronizedVersion, then the group needs to be synchronized
+        // Else Noop
+        _ <-
+          if (group.version > group.lastSynchronizedVersion.getOrElse(0)) {
+            IO.unit
+          } else {
+            IO.raiseError(new GroupAlreadySynchronized)
```
dvoet, line 64: I don't really like handling this with an exception since this is not really an exceptional situation, we just want an early escape.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
dvoet: probably because the policy is set to public

dvoet (approved):
1 comment but don't hold this up based on it

samanehsan (approved):
Thanks @Ghost-in-a-Jar! 🎉

PR #1475, opened 2024-07-02
PROD-972 Limit google group syncs

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/PROD-972

What:

This change is based on the observation that if a user is not a member of a policy then the policy does not need to be synchronized when the associated auth domain is changed. If the are not on the policy and they are removed from the auth domain, they should not have been in policy google group in the first place. If they are not on the policy and are added to the auth domain, they should not be added to the policy google group.

Why:

reduce the number of google group sync messages

How:

whenever group membership is changed (really policy membership), capture the membership additions and removals and limit the policies synchronized to only those that contain that member, directly or indirectly.

---

**PR checklist**


samanehsan (approved):
Thanks @dvoet! The tests really helped clarify the intended behavior here 💯

Ghost-in-a-Jar (approved):

PR #1476, opened 2024-07-03
[DC-1142] Add delete permission to snapshot stewards on requests

s-rubenstein:
Ticket: https://broadworkbench.atlassian.net/browse/DC-1142

What:
This adds delete to the approver role, so that snapshot stewards have the ability to delete snapshot access requests. 

Why:

This allows cleanup of requests to unblock deletion of snapshots. Mostly important for tests and local development.

How:

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1361,7 +1361,7 @@ resourceTypes = {
         roleActions = ["get", "update", "delete", "create_with_parent"]
       }
       approver = {
-        roleActions = ["get", "approve"]
+        roleActions = ["get", "approve", "delete"]
       }
     }
     reuseIds = false


pshapiro4broad (approved):

rjohanek (commented):
Why do we want the approver to be able to delete snapshot requests? I know it is so that the underlying snapshot can be deleted, but what is the case that the underlying snapshot needs to be deleted? Is this just for us/ease of testing, if so does this have any repercussions that we need to consider? Like if approvers are deleting requests instead of rejecting them, it leaves little information for researchers. Is that something we need to worry about?

s-rubenstein:
@rjohanek I think most of the time we won't want them deleting requests, and we probably don't really want to encourage deleting of requests in general. However, for the time being we need to enable test cleanup, and this is the best way to do that. I think the risk of approvers deleting requests is low (if they do, it may be confusing to researchers, but it is functionally equivalent to rejecting.)

What we really want here is a cascade functionality for deleting Sam resources, but that would be a much greater change/feature for Sam, so we're punting on that.

rjohanek (approved):

PR #1480, opened 2024-07-08
ID-1327 revert synchronous group removals and refactor

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1327

Reverting the synchronous group sync for policy removals. This was a band-aid, but has the potential to create long request latencies. Now that google group synchronization is fixed, its no longer needed.

Also did a little refactor of the sync de-duplication. Instead of relying on an exception for control flow, we now rely on an `Option`. This is best-practice, since using exceptions for control flow incurs a significant performance penalty in JVM languages.

---

**PR checklist**

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
@@ -307,7 +307,9 @@ class GoogleExtensionRoutesSpec extends GoogleExtensionRoutesSpecHelper with Sca
     // A duplicate call should return OK and an empty response
     Post(s"/api/google/resource/${resourceType.name}/foo/${resourceType.ownerRoleName.value}/sync") ~> routes.route ~> check {
       status shouldEqual StatusCodes.OK
-      responseAs[Map[WorkbenchEmail, Seq[SyncReportItem]]] shouldBe empty
+      val response = responseAs[Map[WorkbenchEmail, Seq[SyncReportItem]]]
+      response.keys should not be empty
+      response.values.flatten shouldBe empty
     }
   }
 
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
@@ -304,11 +304,14 @@ class GoogleExtensionRoutesSpec extends GoogleExtensionRoutesSpecHelper with Sca
       responseAs[Map[WorkbenchEmail, Seq[SyncReportItem]]] should not be empty
     }
 
-    // A duplicate call should return OK and an empty response
+    // A duplicate call should return OK and the email with no sync report items
     Post(s"/api/google/resource/${resourceType.name}/foo/${resourceType.ownerRoleName.value}/sync") ~> routes.route ~> check {
       status shouldEqual StatusCodes.OK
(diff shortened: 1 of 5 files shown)


Ghost-in-a-Jar (approved):

dvoet (approved):

Ghost-in-a-Jar (changes requested):
can you retain the shape of the response to what it was before

tlangs:
@Ghost-in-a-Jar done

Ghost-in-a-Jar (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
```diff
@@ -307,7 +307,9 @@ class GoogleExtensionRoutesSpec extends GoogleExtensionRoutesSpecHelper with Sca
     // A duplicate call should return OK and an empty response
```
okotsopoulos: This comment looks to be out of date given these changes.

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
```diff
@@ -307,7 +307,9 @@ class GoogleExtensionRoutesSpec extends GoogleExtensionRoutesSpecHelper with Sca
...
     Post(s"/api/google/resource/${resourceType.name}/foo/${resourceType.ownerRoleName.value}/sync") ~> routes.route ~> check {
       status shouldEqual StatusCodes.OK
-      responseAs[Map[WorkbenchEmail, Seq[SyncReportItem]]] shouldBe empty
+      val response = responseAs[Map[WorkbenchEmail, Seq[SyncReportItem]]]
+      response.keys should not be empty
```
okotsopoulos: (Ditto for the existing assertion on the initial sync call.)

Can we make this assertion more meaningful, rather than only checking emptiness?  E.g. is there something we know about the expected `WorkbenchEmail` that we can assert?

okotsopoulos (commented):
Thanks!  I left a few suggestions for test improvement.

>  Now that google group synchronization is fixed, its no longer needed.

I'm probably out of the loop here, is there somewhere I can go to learn more about how Google Group synchronization was fixed (ex. an incident review page)?

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
Ghost-in-a-Jar: Yeah look at what I did in my other pr (that ill ultimately close after this pr merged)
https://github.com/broadinstitute/sam/pull/1479/files

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
tlangs: Thanks! Changed to reflect ✨reality✨

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutesSpec.scala
tlangs: Asserted that the response contains an email starting with `policy-`. The part after `policy-` could be anything, but is likely a UUID, but I didn't think that was too important to check. The API should just be expecting _an_ email, not a specific email, since there's no way for the caller to know the email before the call!

okotsopoulos (approved):

PR #1482, opened 2024-07-08
ID-1328 system resource access policy creation on boot

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1328

What:

create resource access policies from config at creation tome

Why:

Terra requires some system access policies for it's services, e.g. rawls, leo, etc. To date, there have only been a couple and they have been trivial. We expect more and more complicated policies soon. Better to manage them in code.

How:

A new config stanza called `resourceAccessPolicies` is read at boot time and fed into existing code to overwrite policies. The config has the same structure as the api to create policies.

---

**PR checklist**

### DIFF
--- a/src/main/resources/sam.conf
+++ b/src/main/resources/sam.conf
@@ -243,3 +243,52 @@ janitor {
   trackResourceTopicId = ${?JANITOR_TRACK_RESOURCE_TOPIC_ID}
 }
 
+terra.support.emails = [
+  // dynamically configuring lists is hard, so 10 slots are provided for support emails
+  // add more if more than 10 are ever needed
+  ${?TERRA_SUPPORT_EMAIL_0}
+  ${?TERRA_SUPPORT_EMAIL_1}
+  ${?TERRA_SUPPORT_EMAIL_2}
+  ${?TERRA_SUPPORT_EMAIL_3}
+  ${?TERRA_SUPPORT_EMAIL_4}
+  ${?TERRA_SUPPORT_EMAIL_5}
+  ${?TERRA_SUPPORT_EMAIL_6}
+  ${?TERRA_SUPPORT_EMAIL_7}
+  ${?TERRA_SUPPORT_EMAIL_8}
+  ${?TERRA_SUPPORT_EMAIL_9}
+]
+
+resourceAccessPolicies {
+  resource_type_admin {
+    workspace {
+      support {
+        memberEmails = ${terra.support.emails}
+        roles = ["support"]
+      }
+    }
+    managed-group {
+      support {
+        memberEmails = ${terra.support.emails}
+        roles = ["support"]
+      }
+    }
+    billing-project {
+      support {
+        memberEmails = ${terra.support.emails}
+        roles = ["support"]
+      }
+    }
+    dataset {
+      support {
+        memberEmails = ${terra.support.emails}
+        roles = ["support"]
+      }
+    }
+    datasnapshot {
+      support {
+        memberEmails = ${terra.support.emails}
+        roles = ["support"]
+      }
+    }
+  }
+}
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
@@ -44,6 +44,7 @@ import org.broadinstitute.dsde.workbench.sam.db.DbReference
 import org.broadinstitute.dsde.workbench.sam.google._
 import org.broadinstitute.dsde.workbench.sam.model._
 import org.broadinstitute.dsde.workbench.sam.service._
+import org.broadinstitute.dsde.workbench.sam.util.SamRequestContext
 import org.broadinstitute.dsde.workbench.sam.util.Sentry.initSentry
 import org.broadinstitute.dsde.workbench.util.DelegatePool
 import org.typelevel.log4cats.StructuredLogger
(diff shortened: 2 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -80,6 +81,20 @@ object Boot extends IOApp with LazyLogging {
 
         _ <- dependencies.policyEvaluatorService.initPolicy()
 
+        // make sure all users referenced by resourceAccessPolicies exist
+        _ <- appConfig.resourceAccessPolicies.flatMap { case (_, policy) => policy.memberEmails }.toList.traverse { email =>
+          dependencies.samApplication.userService.inviteUser(email, SamRequestContext()).attempt
```
tlangs, line 86: Are we sure we want to silently do this? Since this is on-boot, I'm wondering if it would be better to hard-fail if the users don't exist, with some descriptive message saying there's a configuration error.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
```diff
@@ -80,6 +81,20 @@ object Boot extends IOApp with LazyLogging {
...
+        }
+
+        // create resourceAccessPolicies but don't prevent startup if any fail
+        policyTrials <- dependencies.samApplication.resourceService.upsertResourceAccessPolicies(appConfig.resourceAccessPolicies)
+        _ = policyTrials.map {
+          case Left(t) =>
+            logger.error("NON-FATAL - failure creating configured policy on startup, continuing", t)
```
tlangs: Why is an error non-fatal? If Sam can't upsert the policies, then I would expect startup to fail, because Sam won't be behaving how I expect.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -83,6 +83,42 @@ class ResourceService(
...
+          if (resourceType.name.isResourceTypeAdmin) {
+            overwriteAdminPolicy(
+              resourceType,
+              fullyQualifiedPolicyId.accessPolicyName,
+              fullyQualifiedPolicyId.resource,
+              accessPolicyMembershipRequest,
+              samRequestContext
+            )
+          } else {
+            overwritePolicy(
```
tlangs: Do we expect to be setting up non resource_type_admin policies with this functionality?

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: In a BEE, the user's probably won't exist because our services register their service account users on startup. Any new live environment will also have this issue. The accounts exist in our live environments because they were registered by the same process many moons ago.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: no current expectation, but it seemed just as easy to support it as to throw an exception

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
dvoet: It seemed drastic to have sam not start up because of this. I suppose with k8s there would not be down time, but it would still be disruptive to have to deal with it. I was not thinking about it as Sam not behaving as expected. I was thinking that Sam does not really care what the policies are. Your call, I am happy to go either way and your team supports it.

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
tlangs: Ah, good point about BEEs!

src/main/scala/org/broadinstitute/dsde/workbench/sam/Boot.scala
tlangs: I think I'd rather have it be disruptive. A bee would not start up if a policy can't be upserted, and we require a successful `create-bee` action to succeed before merging. I agree Sam shouldn't care what the policies are, but if a policy defined in `reference.conf` can't be upserted, I would consider that an error state.

src/main/resources/sam.conf
```diff
@@ -243,3 +243,52 @@ janitor {
   trackResourceTopicId = ${?JANITOR_TRACK_RESOURCE_TOPIC_ID}
 }
 
+terra.support.emails = [
```
dvoet, line 246: this new config should replace [this script](https://github.com/broadinstitute/dsp-scripts/blob/master/terra-support-sam-policies/terra-support-sam-policies.py)

src/main/resources/sam.conf
```diff
@@ -243,3 +243,52 @@ janitor {
...
+  ${?TERRA_SUPPORT_EMAIL_4}
+  ${?TERRA_SUPPORT_EMAIL_5}
+  ${?TERRA_SUPPORT_EMAIL_6}
+  ${?TERRA_SUPPORT_EMAIL_7}
+  ${?TERRA_SUPPORT_EMAIL_8}
+  ${?TERRA_SUPPORT_EMAIL_9}
+]
+
+resourceAccessPolicies {
+  resource_type_admin {
```
rtitle, line 262: What's an example for doing this for services (Leo, Rawls)? Wondering if our existing roles will work, or if it would be better to create some new ones for service management of resources..

dvoet: my intent was to create new roles with hand picked actions. The workspace application role is probably ok for Leo but every other resource type probably needs other roles.

src/main/resources/sam.conf
dvoet: see also https://github.com/broadinstitute/terra-helmfile/pull/5706

tlangs (approved):

samanehsan (approved):

PR #1492, opened 2024-07-15
ID-1333 Add group version increment to onPolicyUpdate

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1333

What:

This adds group version incrementation to onPolicyUpdate which should cover our bases as far as policy member updates are concerned. It was a mistake to not add it here in the first place and instead try to put it a level up, we missed some areas where memberships are updated. 

We also had to make sure group versions are updated for nested policy groups constrained by an auth domain

Why:

Users removed from auth domains are not losing access to google resources.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
@@ -717,6 +717,10 @@ class ResourceService(
       changeEvents = createAccessChangeEvents(policyId.resource, originalPolicies, updatedPolicies)
       _ <- AuditLogger.logAuditEventIO(samRequestContext, changeEvents.toSeq: _*)
 
+      _ <- directoryDAO.updateGroupUpdatedDateAndVersionWithSession(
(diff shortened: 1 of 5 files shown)


tlangs (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -885,10 +889,6 @@ class ResourceService(
           for {
             originalPolicies <- accessPolicyDAO.listAccessPolicies(policyId.resource, samRequestContext)
             policyChanged <- accessPolicyDAO.setPolicyIsPublic(policyId, public, samRequestContext)
-            _ <- directoryDAO.updateGroupUpdatedDateAndVersionWithSession(
-              FullyQualifiedPolicyId(policyId.resource, policyId.accessPolicyName),
-              samRequestContext
-            )
```
Ghost-in-a-Jar, line 891: Note that the next thing that happens is onPolicyUpdateIfChanged which calls onPolicyUpdate if the policy actually changed.

PR #1493, opened 2024-07-16
[DDO-3704] Don't try to create pets in inactive projects

rtitle:
Ticket: https://broadworkbench.atlassian.net/browse/DDO-3704

What:

Throw a sensible error and don't retry if a GCP pet is requested for an inactive project.

Why:

Janitor deletes GCP projects for workspaces in BEE environments to save costs. Unfortunately project-less workspaces are undeletable in Terra UI, in part because the Sam `getPetServiceAccount` retries for a long time and eventually returns 500/502. This PR makes Sam `getPetServiceAccount` APIs handle this situation more gracefully.

How:

This PR changes `getPetServiceAccount` to fail fast with 400 for an inactive project, instead of retrying and eventually returning 500/502.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -356,8 +356,8 @@ class GoogleExtensions(
         // SA does not exist in google, create it and add it to the proxy group
         case None =>
           for {
-            _ <- assertProjectIsActive(project)
             _ <- assertProjectInTerraOrg(project)
+            _ <- assertProjectIsActive(project)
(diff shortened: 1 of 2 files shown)


tlangs (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -356,6 +356,7 @@ class GoogleExtensions(
         // SA does not exist in google, create it and add it to the proxy group
         case None =>
           for {
+            _ <- assertProjectIsActive(project)
```
dvoet: should the call to `assertProjectInTerraOrg` come first?

rtitle: Good call, it makes sense to check if it's a Terra project first

Ghost-in-a-Jar (approved):

PR #1498, opened 2024-07-22
ID-1300 Admin Get Users by ID

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1300

In order for Orchestration to modify group membership using Sam User ID from ECM, it needs to be able to add users from the response to Sam Groups. Since Sam Group Endpoints only deal in email addresses, we need a way to get emails from Sam User IDs.

This is an admin-only functionality, as Sam User IDs should not be discoverable by other users. IE: A user can know their own Sam User ID, but should not be able to find out the Sam User IDs of others. Because of this, only authorized Terra Services will be able to call this API endpoint. In this case, its Thurloe and Orch.

Another idea is to make a V2 Sam Group API, which deals only in Sam User IDs, but that’s a much larger change and not necessary to accomplish the goal of deprecating Shibboleth and Thurloe.

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -70,7 +70,7 @@ paths:
     get:
       tags:
         - Admin
-      summary: Retrieves a user record, by user id
+      summary: Retrieves a user record, by user id. Limited to 1000 user IDs.
       operationId: adminGetUser
       parameters:
         - name: userId
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -189,9 +189,18 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/User'
+        400:
+          description: Request invalid. Request body must be a list of less than 1000 Sam User IDs.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
         403:
           description: You do not have service admin privileges
-          content: { }
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
         500:
           description: Internal Server Error
           content:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -168,6 +168,36 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+    post:
+      tags:
+        - Admin
+      summary: Gets a list of users for a list of Sam User IDs
+      operationId: adminGetUsersByIDs
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                type: string
+      responses:
+        200:
+          description: list of matching users
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/User'
+        403:
+          description: You do not have service admin privileges
+          content: { }
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
   /api/admin/v1/user/{userId}/disable:
     put:
       tags:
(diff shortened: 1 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ServiceAdminRoutes.scala
```diff
@@ -53,6 +54,13 @@ trait ServiceAdminRoutes extends SecurityDirectives with SamRequestContextDirect
               .map(users => (if (users.nonEmpty) OK else NotFound) -> users)
           }
         }
+      } ~
+      postWithTelemetry(samRequestContext) {
```
Ghost-in-a-Jar, line 61: shouldnt this be get?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ServiceAdminRoutes.scala
```diff
@@ -53,6 +54,13 @@ trait ServiceAdminRoutes extends SecurityDirectives with SamRequestContextDirect
...
+        entity(as[Seq[WorkbenchUserId]]) { samUserIds =>
+          complete {
+            userService.getUsersByIds(samUserIds, samRequestContext)
+          }
+        }
```
Ghost-in-a-Jar, line 72: Should we have a limit on how many you can get at one time?

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpecs/GetUsersByIdsSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+      val userService = TestUserServiceBuilder()
+        .withAllUsersGroup(allUsersGroup)
+        .build
+      // Act
+      val response = runAndWait(userService.getUser(userWithBothIds.id, samRequestContext))
+
+      // Assert
+      assert(response.isEmpty, "Getting a nonexistent user should not find a user")
+    }
+  }
```
Ghost-in-a-Jar, line 48: should we have any test cases that assert the admin only functionality or is that properly covered by other tests?

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/ServiceAdminRoutes.scala
tlangs: I'm thinking POST, since the number of IDs could get pretty unwieldy in a url.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpecs/GetUsersByIdsSpec.scala
tlangs: I believe this is covered by other tests, but I'll check.

tlangs: Found the test, but it was just named slightly wrong. It tests that only authorized service accounts can call the API

Ghost-in-a-Jar (approved):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -168,6 +168,36 @@ paths:
...
+          description: list of matching users
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/User'
+        403:
+          description: You do not have service admin privileges
+          content: { }
```
samanehsan: Why isn't this `#/components/schemas/ErrorReport` like the 500 error?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -168,6 +168,36 @@ paths:
...
+      responses:
+        200:
+          description: list of matching users
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/User'
+        403:
```
samanehsan, line 198: I think you also need a block for a `400` response here (if someone requests > 1000 users)

samanehsan (approved):

PR #1520, opened 2024-08-16
ID-1372 descendants inherit auth domains

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1372

What:

propagate auth domain to descendant resources when
1. resources are created with a parent that has an auth domain
2. resource parent is set to a resource with an auth domain
3. auth domain groups are added to a resource with progeny

db script to propagate auth domains to all existing descendants of resources with auth domains

Why:

see design doc linked from ticket

How:

brute force

---

**PR checklist**


mlilynolting:
Does this require a migration of existing descendants of resources with auth domains?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -916,32 +971,12 @@ class ResourceService(
   def getResourceParent(resourceId: FullyQualifiedResourceId, samRequestContext: SamRequestContext): IO[Option[FullyQualifiedResourceId]] =
     accessPolicyDAO.getResourceParent(resourceId, samRequestContext)
 
-  /** In this iteration of hierarchical resources, we do not allow child resources to be in an auth domain because it would introduce additional complications
-    * when keeping Sam policies with their Google Groups. For more details, see
-    * https://docs.google.com/document/d/10qGxsV9BeM6-N_Zk27_JIayE509B8LUQBGiGrqB0taY/edit#heading=h.dxz6xjtnz9la
```
mlilynolting, line 921: Is this still a concern?

dvoet: It is no longer a concern. I am not sure why I though it was in the first place. I was probably thinking about how we might sync inherited policies to google, but thinking about it now, those policies don't actually exist so there is no way to sync them and there is no reason to since the membership will always be the same as the source policy. So if you need to sync a group in order to add it google IAM somewhere, just use the source policy.

LizBaldo (commented):
The inheritance logic makes sense to me. One question that I have though (and can be outside the scope of this PR) is wether or not actions on the auth-domain constrained resource are also constrained 🤔 (for instance for an auth-domain constrained workspace, the child google project, and grandchild runtime should be auth-domain constrained as well, but what about things like the connect action on the runtime itself?)

dvoet:
> The inheritance logic makes sense to me. One question that I have though (and can be outside the scope of this PR) is wether or not actions on the auth-domain constrained resource are also constrained 🤔 (for instance for an auth-domain constrained workspace, the child google project, and grandchild runtime should be auth-domain constrained as well, but what about things like the connect action on the runtime itself?)

That is beyond the scope of this PR. But when the time comes, [something like this](https://github.com/broadinstitute/sam/blob/692d12bb32e905275986b905bcd8df7f4519dbef/src/main/resources/reference.conf#L93) for the `connect` action will be required. That is step 6 of the solution in the auth domain section of the design doc.

LizBaldo (approved):
LGTM, but I would feel better if somebody who has more experience in Sam than me to have a look as well :)

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240820_descendant_auth_domains.xml
```diff
@@ -0,0 +1,27 @@
...
+                union
+                    select childResource.id, ancestorResource.group_id
+                    from SAM_RESOURCE childResource
+                    join ancestor_resource ancestorResource on ancestorResource.resource_id = childResource.resource_parent_id
+                )
+
+            insert into sam_resource_auth_domain(resource_id, group_id)
+            select ar.resource_id, ar.group_id
+            from ancestor_resource ar
+            on conflict do nothing
```
Ghost-in-a-Jar, line 23: this is elegant, nice

Ghost-in-a-Jar (approved):
whether or not we want this is more of a product question, but the implementation + tests look good. I would be really careful with the migration ofc because it seems like it could take a while in prod.

PR #1523, opened 2024-08-23
[ID-1351] store and retrieve user attributes with Sam

Shakespeared:
Ticket: [<Link to Jira ticket>](https://broadworkbench.atlassian.net/browse/ID-1351)

What:

We are migrating the Sam User Attributes out of Thurloe and into Sam. User attributes are "profile" information coming from the UI, either from new user signup or when a user updates their personal info through the UI. 

This PR will store all new user attributes or updated user attributes  in the Sam DB, and when a user updates their attributes that update will be updated or added in the Sam DB as well. This does NOT sever the connection with Thurloe, as we need to backfill user info before we do that. 

Why:

The Thurloe KV store is bad and gross and we want this info in Sam so it will be a) consistent and b) easily retrievable. 

How:

I updated the existing `samUserAttributes` and `samUserAttributesRequest` objects to include these additional values, and updated the corresponding `userAttributes` API endpoints to return the new fields. 

This PR includes a bunch of test updates, but most of them are essentially just adding in these new fields to existing tests because the idea of `userAttributes` already exists and is tested. 

---

**PR checklist**


src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240806_add_profile_info_table.xml
```diff
@@ -0,0 +1,84 @@
...
+        </column>
+    </addColumn>
+    <addColumn tableName="SAM_USER_ATTRIBUTES">
+        <column name="created_at" type="timestamptz">
+            <constraints nullable="true"/>
+        </column>
+    </addColumn>
+    </changeSet>
+    <changeSet logicalFilePath="dummy" author="kfeldman" id="move_updated_at_column_to_end_of_table">
+<!--        renaming updated_at so I can recreate it as a new column, then transfer existing data -->
```
Shakespeared, line 75: I would love thoughts on this liquibase change... I wanted to re-order the columns so `updatedAt` is the last one. Unfortunately the column reordering commands don't seem to work in postgres, so this was the best way I found to do it. I don't love having to drop the column but in testing this did seem to work.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -1355,11 +1343,57 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
+        ) values (
+        ${userAttributes.userId},
+        ${userAttributes.marketingConsent},
+        ${userAttributes.firstName},
+        ${userAttributes.lastName},
+        ${userAttributes.organization},
+        ${userAttributes.contactEmail},
+        ${userAttributes.title},
+        ${userAttributes.department},
+        ARRAY[${userAttributes.interestInTerra.getOrElse(Array.empty[String])}]::text[],
```
Shakespeared, line 1384: I had to do a lot of experimenting to get the types correct, this finally didn't throw any type errors

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/tables/UserAttributesTable.scala
```diff
@@ -4,13 +4,25 @@ import org.broadinstitute.dsde.workbench.model._
...
+    contactEmail: Option[String],
+    title: Option[String],
+    department: Option[String],
+    interestInTerra: Option[List[String]],
+    programLocationCity: Option[String],
+    programLocationState: Option[String],
+    programLocationCountry: Option[String],
+    researchArea: Option[List[String]],
+    additionalAttributes: Option[String],
+    createdAt: Option[Instant],
```
Shakespeared, line 24: I did make both of these timestamp fields optional to account for users who already have a record in the DB without a `createdAt` date, and to allow for (slightly) easier testing

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240806_add_profile_info_table.xml
```diff
@@ -0,0 +1,84 @@
...
+        <renameColumn tableName="SAM_USER_ATTRIBUTES" oldColumnName="updated_at" newColumnName="old_updated_at"/>
+        <addColumn tableName="SAM_USER_ATTRIBUTES">
+            <column name="updated_at" type="timestamptz" valueComputed="old_updated_at">
```
samanehsan, line 78: If the value is computed with `old_updated_at`, will the values stay after that column is dropped? I'm not familiar with the `valueComputed` tag!

samanehsan (approved):
Lgtm! Just had a question about the liquibase migration 🤔

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240806_add_profile_info_table.xml
Shakespeared: yes it will persist those values, I just did some manual testing to confirm. It should just be seeding the db with the values from `old_updated_at` but not creating a linkage between those columns, so when we drop `old_updated_at` the new `updated_at` should persist.

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20240806_add_profile_info_table.xml
```diff
@@ -0,0 +1,84 @@
...
+        </column>
+    </addColumn>
+    <addColumn tableName="SAM_USER_ATTRIBUTES">
+        <column name="additional_attributes" type="JSONB">
+            <constraints nullable="true"/>
+        </column>
+    </addColumn>
+    <addColumn tableName="SAM_USER_ATTRIBUTES">
+        <column name="created_at" type="timestamptz">
+            <constraints nullable="true"/>
```
Ghost-in-a-Jar, line 70: Maybe this shouldnt be nullable?

Ghost-in-a-Jar: Nevermind, just saw your comment about why this is nullable

Ghost-in-a-Jar (approved):

PR #1525, opened 2024-08-26
[IA-4295] Add API to get a pet service account token for a user

rtitle:
Ticket: https://broadworkbench.atlassian.net/browse/IA-4295

What:

Sam has APIs to get pet keys and pet tokens as users. However the "admin" pet routes (the ones that require the  `cloud-extension/google` permission and take an email) only support pet keys. This PR adds admin pet routes to get pet tokens.

Why:

Leonardo relies on the admin pet routes to get pet tokens for users. Today Leonardo is requesting keys from Sam and using those to generate tokens. However Leonardo is doing some refactoring of its auth layer (including switching to use the Sam client), and it would be nice to remove the key logic and just fetch tokens from Sam directly. 

How:

Updated swagger docs, `GoogleExtensionRoutes`, and `GoogleExtensions` to implement the new APIs.

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1082,7 +1082,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
   /api/google/v1/petServiceAccount/{project}/{email}/token:
-    get:
+    post:
       tags:
         - Google
       summary: gets a token for the user's pet service account, get_pet_private_key
@@ -1177,10 +1177,11 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
   /api/google/v1/petServiceAccount/{email}/token:
-    get:
+    post:
       tags:
         - Google
-      summary: gets a token for the user's arbitrary pet service account, get_pet_private_key action on cloud-extension/google required
+      summary: gets a token for the user's arbitrary pet service account, get_pet_private_key
+        action on cloud-extension/google required
       operationId: getUserArbitraryPetServiceAccountToken
       parameters:
         - name: email
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1034,7 +1034,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
-  /api/google/v1/petServiceAccount/{project}/{email}:
+  /api/google/v1/petServiceAccount/{project}/{email}/key:
     get:
       tags:
         - Google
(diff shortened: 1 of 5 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1034,7 +1034,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
-  /api/google/v1/petServiceAccount/{project}/{email}:
+  /api/google/v1/petServiceAccount/{project}/{email}/key:
```
rtitle: I noticed this API was inconsistent: it returns a key, but all the other APIs that return keys are suffixed with `/key`. So I updated the swagger definition, but made the server code handle both patterns (with `/key` and without). I think this should be safe, but let me know if you disagree.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -455,30 +455,55 @@ class GoogleExtensions(
       key <- googleKeyCache.getKey(pet)
     } yield key
 
-  def getPetServiceAccountToken(user: SamUser, project: GoogleProject, scopes: Set[String], samRequestContext: SamRequestContext): Future[String] =
```
rtitle, line 458: I was bothered that some of the `getPet*` methods returned `IO` and others returned `Future`. So I refactored slightly to make them all return `IO` (some of the dependent methods are still in terms of `Future`).

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -455,30 +455,55 @@ class GoogleExtensions(
...
-    getPetServiceAccountKey(user, project, samRequestContext).unsafeToFuture().flatMap { key =>
-      getAccessTokenUsingJson(key, scopes)
+  def getPetServiceAccountToken(user: SamUser, project: GoogleProject, scopes: Set[String], samRequestContext: SamRequestContext): IO[String] =
+    getPetServiceAccountKey(user, project, samRequestContext).flatMap { key =>
+      IO.fromFuture(IO(getAccessTokenUsingJson(key, scopes)))
     }
 
+  def getPetServiceAccountToken(
```
rtitle, line 463: The general pattern is we overload these methods: one with a `SamUser` and one with a `WorkbenchEmail`. So now we have 8 total combinations:
```
getPetServiceAccountKey(user|email)
getPetServiceAccountToken(user|email)
getArbitraryServiceAccountKey(user|email)
getArbitraryServiceAccountToken(user|email)
```

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/NewGoogleExtensionsSpec.scala
```diff
@@ -206,4 +210,168 @@ class NewGoogleExtensionsSpec(_system: ActorSystem)
       }
     }
   }
+  "GoogleExtensions: Pet Service Accounts" - {
```
rtitle, line 213: Added some pet tests to `NewGoogleExtensionsSpec` - this test is nice because it's easier to mock behavior thatn `GoogleExtensionsSpec`.

src/main/resources/swagger/api-docs.yaml
dvoet: this will change the generated client, I would leave it alone

rtitle: ok, removed

samanehsan (approved):

Ghost-in-a-Jar (approved):

PR #1528, opened 2024-08-28
ID-1369 Admin endpoint repair access

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1369

What:

We recently saw a case where a user is missing their proxy group, this new endpoint should create the proxy group if it doesnt exist and then send off synchronization messages to synchronize google with sam’s database.

Why:

If a user doesn't have a proxy group they cant use the platform at all. 

How:

I added a new admin only endpoint that does the following:
1. Creates the user's proxy group
2. Adds the user to the proxy group and all users group
3. Sends off a group synchronization message for any groups they are a member of

I went a little light on the unit tests because this is an admin endpoint and not an application feature, and I am going to thoroughly test it out in dev before using it in prod. 

---

**PR checklist**


src/test/scala/org/broadinstitute/dsde/workbench/sam/service/MockUserService.scala
```diff
@@ -123,9 +123,6 @@ class MockUserService(
   def setGoogleSubjectId(userId: WorkbenchUserId, googleSubjectId: GoogleSubjectId, samRequestContext: SamRequestContext): IO[Unit] =
     directoryDAO.setGoogleSubjectId(userId, googleSubjectId, samRequestContext)
 
-  def listUserDirectMemberships(userId: WorkbenchUserId, samRequestContext: SamRequestContext): IO[LazyList[WorkbenchGroupIdentity]] =
-    directoryDAO.listUserDirectMemberships(userId, samRequestContext)
```
Ghost-in-a-Jar, line 127: This actually wasn't being used anywhere

dvoet (approved):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -98,6 +98,34 @@ paths:
...
+          description: User ID of the user to have their cloud access repaired
+          required: true
+          schema:
+            type: string
+      responses:
+        204:
+          description: User access was repaired successfully (as long as the group sync messages succeed)
+        403:
+          description: You do not have admin privileges
+          content: { }
```
samanehsan, line 119: Should the content here (and with the 404) be the same as the 500 error?
```
content:
  application/json:
    schema:
```

samanehsan (approved):
Lgtm! :octocat: I just had a question about the api swagger definition

src/main/resources/swagger/api-docs.yaml
Ghost-in-a-Jar: I think this is unnecessary for 403/404, the reason the error report is there for 500 is because that is an unexpected error where it could be helpful to include additional information. For 404/403 it is obvious what the problem is.

PR #1529, opened 2024-08-28
ID-1369 Repair cloud access followup: fix swagger

Ghost-in-a-Jar:
Changing get -> put in swagger and add cloudExtensions.onUserEnable call to the new endpoint. This is what actually adds stuff to the user's proxy group after it is created.


Shakespeared (approved):

samanehsan (approved):

ungwudik:
@Ghost-in-a-Jar Double-checking that the two checks that failed here were not blockers?

PR #1535, opened 2024-09-06
[IA-5060] Inherit manager roles to notebook-cluster and persistent-disk

rtitle:
Ticket: https://broadworkbench.atlassian.net/browse/IA-5060

What:

Phase 1 of the [Simplify Leo Access Control](https://broadworkbench.atlassian.net/browse/IA-5059) epic.

This PR: 
* Allows `notebook-cluster` and `perstistent-disk` resource types to inherit from `google-project` or workspace.
* Makes `notebook-cluster/connect` and `persistent-disk/attach` actions auth domain constrainable.

Why:

We are trying to clean up/simplify Leonardo access control logic. Part of this is standardizing on consistent Sam resource types and using inheritance to propagate permissions. This PR sets up the Sam model; subsequent Leonardo changes will be made to migrate to it.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -175,6 +175,8 @@ resourceTypes = {
           wds-instance = ["owner"]
           kubernetes-app = ["manager"]
           kubernetes-app-shared = ["owner", "user"]
+          notebook-cluster = ["manager"]
+          persistent-disk = ["manager"]
         }
       }
       application = {
@@ -610,17 +612,8 @@ resourceTypes = {
       read_policies = {
         description = "list all policies and policy details for this google-project"
       }
-      list_notebook_cluster = {
-        description = "list all notebook clusters in this google-project"
-      }
       launch_notebook_cluster = {
         description = "launch a new notebook cluster in this google-project"
-      }
-      delete_notebook_cluster = {
-        description = ""
-      }
-      stop_start_notebook_cluster = {
-        description = "stop and start notebook clusters in this google-project"
       }
       list_persistent_disk = {
         description = "list all persistent disks in this google-project"
@@ -656,10 +649,12 @@ resourceTypes = {
     ownerRoleName = "owner"
     roles = {
       owner = {
-        roleActions = ["read_policies", "list_notebook_cluster", "delete_notebook_cluster", "list_persistent_disk", "delete_persistent_disk", "delete", "get_parent", "set_parent", "list_children"]
+        roleActions = ["read_policies", "delete", "get_parent", "set_parent", "list_children"]
         includedRoles = ["notebook-user", "pet-creator"]
         descendantRoles = {
           kubernetes-app = ["manager"]
+          notebook-cluster = ["manager"]
+          persistent-disk = ["manager"]
         }
       }
       notebook-user = {
@@ -790,6 +785,7 @@ resourceTypes = {
       }
       connect = {
         description = "connect to the Jupyter notebook running on the notebook cluster"
+        authDomainConstrainable = true
       }
       delete = {
         description = "delete the notebook cluster"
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -803,22 +814,30 @@ resourceTypes = {
...
+      }
     }
     reuseIds = false
+    authDomainConstrainable = true
   }
   persistent-disk = {
     actionPatterns = {
       read = {
-        description = "read metadata and contents of persistent disk"
+        description = "read metadata of persistent disk"
```
rtitle, line 835: `read` is just used for listing disks; `attach` is needed to actually view the contents of a disk

src/main/resources/reference.conf
```diff
@@ -790,6 +800,7 @@ resourceTypes = {
       }
       connect = {
         description = "connect to the Jupyter notebook running on the notebook cluster"
+        authDomainConstrainable = true
```
LizBaldo, line 803: I am a bit confused about how granular the authDomainConstrainable is here. I thought that every resources / actions would inherit that configuration from the workspace after Doug's PR

rtitle: My understanding was you still have to annotate which actions on the child are authDomainConstrainable (like `connect`). @dvoet can you confirm?

src/main/resources/reference.conf
```diff
@@ -803,22 +814,30 @@ resourceTypes = {
...
     roles = {
       creator = {
         roleActions = ["status", "connect", "delete", "read_policies", "stop_start", "modify"]
       }
+      manager = {
+        roleActions = ["status", "delete", "read_policies"]
+      }
     }
     reuseIds = false
+    authDomainConstrainable = true
```
rtitle: I blindly followed this pattern from other resources -- does `authDomainConstrainable = true` need to be specified here at the resource level, in addition to the action level?

rtitle:
VERY draft Leo PR to start creating Sam resources with parents (phase 2): https://github.com/DataBiosphere/leonardo/pull/4781/

I'll try to test these together next week.

src/main/resources/reference.conf
dvoet: @rtitle is correct. The auth domain is inherited but which actions have the auth domain applied is controlled by the `authDomainConstrainable` setting.

src/main/resources/reference.conf
dvoet: hmmm... interesting. Looking at the code, [only the action level setting is used](https://github.com/broadinstitute/sam/blob/1cf4c65577fc7d321fb68bda447e08e059e3f5c5/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala#L142).

dvoet (approved):

src/main/resources/reference.conf
rtitle: Interesting… I'll just set it at the action level then I think.

rtitle:
@LizBaldo requested review on this PR. I tested on a BEE and verified that all the resources (runtime, disk, app) continue to work normally.

LizBaldo (approved):

tlangs (approved):

ungwudik:
@rtitle or anyone here - Is the failure in `orch-swat-test-job (dev, qa)` a known issue or flakiness?

rtitle:
@ungwudik I noticed that, the Orch test failed with:
```
[info] - should link an eRA Commons account with access to none of the supported closed-access datasets *** FAILED *** (4 seconds, 225 milliseconds)
[info]   org.broadinstitute.dsde.workbench.service.RestException: {"causes":[],"message":"<!DOCTYPE html>\n<html lang=\"en\">\n<head>\n<meta charset=\"utf-8\">\n<title>Error</title>\n</head>\n<body>\n<pre>Cannot GET /dev/public-key.pem</pre>\n</body>\n</html>\n","source":"shibboleth","stackTrace":[],"statusCode":404}
```
Looks like the Orch swat test failed to get a dev shibboleth endpoint. This PR didn't touch anything related to orch/shibboleth. I'm not sure if this is a known dev infra issue? (cc @tlangs)

PR #1543, opened 2024-09-18
[IA-5062] Add resource access policies for Leo resource migration

rtitle:
Ticket: https://broadworkbench.atlassian.net/browse/IA-5062

[Design doc](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/3364683793/IA-5062+Migrate+existing+notebook-cluster+and+persistent-disk+resources+to+have+parents)

What:

This PR configures resource access policies in Sam conf to add a `leo_migration` policy containing configurable privileged emails.

Why:

To simplify Leo access control logic, we are running a migration script to set parents for historical `notebook-cluster` and `persistent-disk` Sam resources without a parent. We plan to run the migration as a privileged `@firecloud.org` user, which will need administrative permissions on those Sam resource types.

How:

This approach leverages Sam resource access policies and inheritance of `resource_type_admin` permissions.

---

**PR checklist**

- [ ] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've filled out the [Security Risk Assessment](https://sdarq.dsp-appsec.broadinstitute.org/jira-ticket-risk-assesment) (requires Broad Internal network access) and attached the result to the JIRA ticket

PR #1546, opened 2024-09-23
IA-5062 fix: switch to using descendent roles instead of actions

rtitle:
Ticket: https://broadworkbench.atlassian.net/browse/IA-5062

What:

https://github.com/broadinstitute/sam/pull/1543 added the `leo_migration` resource access policies to Sam. We think it was generating a lot of inserts by using descendent actions instead of roles, which was causing issues in Sam dev. Switching to roles to be more performant.

Why:

How:

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -251,7 +251,7 @@ resourceTypes = {
         ]
       }
       # See https://broadworkbench.atlassian.net/browse/IA-5062
-      # This policy can be removed after the migration is complete.
+      # This role can be removed after the migration is complete.
       leo_migration = {
         roleActions = ["read_policies", "add_child"]
       }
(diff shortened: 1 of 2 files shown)


src/main/resources/reference.conf
```diff
@@ -250,6 +250,11 @@ resourceTypes = {
           "read"
         ]
       }
+      # See https://broadworkbench.atlassian.net/browse/IA-5062
+      # This role can be removed after the migration is complete.
+      leo_migration = {
```
tlangs: can this be renamed `leo_migrator` to match the other role names?

tlangs (changes requested):

dvoet (approved):

src/main/resources/reference.conf
rtitle: yes, good call

rtitle: Fixed

tlangs (approved):

PR #1564, opened 2024-10-15
[CORE-58] Cascade policy deletion 

calypsomatic:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-58

This PR now cascade deletes policies when deleting a resource.  Instead of throwing an error that some policy members are in use like `deleteResource` previous did, it simply removes the problematic rows from the table.  

I've demonstrated on my BEE that with this code, when a workspace has policies attached to it due to an imported TDR snapshot, it can be successfully deleted without any errors or orphaned policy references.

Why:

Users were unable to delete any workspace that had had a TDR snapshot with a policy imported into it, as those policies had entries in Sam's database that were not being deleted.

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1006,6 +1006,34 @@ class PostgresAccessPolicyDAO(
...
+      val deleteFlatQuery = samsql"""delete from ${GroupMemberFlatTable as gmf} where ${gmf.memberGroupId} in
+                         (select distinct ${gmf.result.memberGroupId}
+                          from ${GroupMemberFlatTable as gmf}
+                          join ${PolicyTable as p} on ${gmf.memberGroupId} = ${p.groupId}
+                          where ${p.resourceId} = (${loadResourcePKSubQuery(resourceId)}))
+                     """
+      logger.info(s"deleteFlatQuery: ${deleteFlatQuery.statement}")
+      deleteFlatQuery.update().apply()
+    }
+  }
```
davidangb: is the `select distinct` necessary for both of these queries? Does a simple `select` return the same result? (I don't know the answer, this is an honest question)

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -445,6 +445,23 @@ class ResourceService(
...
+      _ <- cloudDeletePolicies(resource, samRequestContext)
+      _ <- deleteActionManagedIdentitiesForResource(resource, samRequestContext)
+
+      // leave a tomb stone if the resource type does not allow reuse
+      leaveTombStone = !resourceTypes(resource.resourceTypeName).reuseIds
+      _ <- accessPolicyDAO.deleteResource(resource, leaveTombStone, samRequestContext)
+
+      _ <- AuditLogger.logAuditEventIO(samRequestContext, ResourceEvent(ResourceDeleted, resource))
+    } yield ()
+
```
davidangb: there's a lot of copy-paste here between `deleteResource` and `deleteResourceCascade`. Can these be consolidated? Looks like the only difference is the invocation of `checkNoPoliciesInUse(resource, samRequestContext)` vs `accessPolicyDAO.removePolicyGroupsInUse(resource, samRequestContext)` in the second step of the for comprehension.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2123,6 +2123,39 @@ paths:
...
+        400:
+          description: Cannot delete a resource with children. Delete the children first then try again.
+          content: { }
+        403:
+          description: You do not have permission to perform this action on the resource or permissions on the resource's parent
+          content: { }
+        404:
+          description: Resource type does not exist or you are not a member of any
+            policy on the resource
+          content: { }
```
davidangb: I know you expressed wanting to make this a separate API, but looking at it I am wondering again if it can be a flag/option on the existing delete-resource API

davidangb (commented):
Some comments inline about separate APIs/functions vs adding to the existing ones; otherwise, looks like a good approach to me. I'd love a knowledgable Sam person to review the SQL queries but it seems like a good high-level approach.

Can you write tests for this?

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
calypsomatic: Yep!  This was all just to prove the concept, definitely planning on cleaning it up.

src/main/resources/swagger/api-docs.yaml
calypsomatic: It can, that would just cascade changes throughout the codebase.  I *think* if we make the flag optional it won't require updating everyone that calls the API but I'm not positive.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
dvoet: a `using` clause may be better, [see this example](https://github.com/broadinstitute/sam/blob/ed9641ef91be396708174c0c95345e0f2ac03f9f/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/EffectivePolicyMutationStatements.scala#L31)

dvoet:
I think this is fine and should just happen all the time. No new api, not even an option on the existing api. Just do it.

One other change that needs to happen is that any groups that are modified need to be synchronized by calling `cloudExtensions.onGroupUpdate`.

So, I would update `ResourceService.deleteResource` directly to first figure out any groups that are changing and sync them instead of calling `checkNoPoliciesInUse`. Then I would modify `accessPolicyDAO.deleteResource` to include your 2 new delete statements in the same transaction as the other deletes.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -569,6 +569,28 @@ class PostgresAccessPolicyDAO(
...
+      logger.warn(s"deleteQuery: ${deleteQuery.statement}")
+      deleteQuery.update().apply()
+
+      val deleteFlatQuery = samsql"""delete from ${GroupMemberFlatTable as gmf}
+                 using ${PolicyTable as p}
+                 where ${gmf.memberGroupId} = ${p.groupId}
+                 and ${p.resourceId} = (${loadResourcePKSubQuery(resource)})
+                     """
+      logger.info(s"deleteFlatQuery: ${deleteFlatQuery.statement}")
+      deleteFlatQuery.update().apply()
```
dvoet: I would pull this into another function like the ones below for readability

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -441,21 +440,11 @@ class ResourceService(
       // leave a tomb stone if the resource type does not allow reuse
       leaveTombStone = !resourceTypes(resource.resourceTypeName).reuseIds
       _ <- accessPolicyDAO.deleteResource(resource, leaveTombStone, samRequestContext)
+      _ <- cloudSyncPolicies(resource, samRequestContext)
```
dvoet: The resource was deleted in the line above, the database calls in `cloudSyncPolicies` won't find anything. I think you need to get the list of groups that will be changed by having the resource removed before deleting the resource then sync them afterwards.

And something else I think I missed: `PostgresGroupDAO.updateGroupUpdatedDateAndVersion` needs to be called for each group updated.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -569,6 +569,28 @@ class PostgresAccessPolicyDAO(
...
+      val gmf = GroupMemberFlatTable.syntax("gmf")
+
+      // Remove policy members from group tables to prevent FK Violations
+      val deleteQuery =
+        samsql"""delete from ${GroupMemberTable as gm}
+                 using ${PolicyTable as p}
+                 where ${gm.memberGroupId} = ${p.groupId}
+                 and ${p.resourceId} = (${loadResourcePKSubQuery(resource)})
+                     """
+      logger.warn(s"deleteQuery: ${deleteQuery.statement}")
```
dvoet: remove logging

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -569,6 +569,28 @@ class PostgresAccessPolicyDAO(
...
+      deleteQuery.update().apply()
+
+      val deleteFlatQuery = samsql"""delete from ${GroupMemberFlatTable as gmf}
+                 using ${PolicyTable as p}
+                 where ${gmf.memberGroupId} = ${p.groupId}
+                 and ${p.resourceId} = (${loadResourcePKSubQuery(resource)})
+                     """
+      logger.info(s"deleteFlatQuery: ${deleteFlatQuery.statement}")
```
dvoet: remove logging

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
calypsomatic: Just to make sure I understand what you're asking, I should do something like:
```
affectedPolicies <- accessPolicyDAO.findAffectedPolicies(resource, samRequestContext) //New method
_ <- accessPolicyDAO.deleteResource(resource, leaveTombStone, samRequestContext)
_ <- cloudSyncPolicies(affectedPolicies, samRequestContext) //Update this method to take in policies instead of finding them
_ <- updateGroupUpdatedDateAndVersion(affectedPolicies, samRequestContext)
```
Is that correct?

dvoet: yes but `updateGroupUpdatedDateAndVersion` should come before `cloudSyncPolicies`. The sync process keeps track of the last synchronized version and won't do anything if that version is the same as the current version. So the order of operations should be change the policy/group in the sam db, bump the version, cloud sync.

And here is yet another idea (sorry if I am jerking you around): instead of your new database queries to bulk delete entries, you can list the policies that need to be updated along with which policy is being removed. For each of those pairs (policy changing, policy removed), call `removeSubjectFromPolicy` (in the same class, `ResourceService`). That takes care of the database calls, syncing, everything. The code would look something like 
```
affectedPolicies <- accessPolicyDAO.findAffectedPolicies(resource, samRequestContext) //New method
_ <- affectedPolicies.traverse { case (policyToUpdate, policyToRemove) => removeSubjectFromPolicy(policyToUpdate, policyToRemove, samRequestContext) }
```
where `findAffectedPolicies` (or maybe there is a better name) returns a `IO[List[(FullyQualifiedPolicyId, FullyQualifiedPolicyId)]]`

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -951,6 +952,81 @@ class PostgresAccessPolicyDAO(
...
+  """
+      query.foreach { rs =>
+        println(s"pg.name: ${rs.string(pg.resultName.name)}, rt.name: ${rs.string(rt.resultName.name)}, r.name: ${rs.string(r.resultName.name)}, g.name: ${rs
+            .string(p.resultName.name)}")
+      }
+      query
+        .map { rs =>
+          val resourceTypeName1 = rs.get[ResourceTypeName](rt.resultName.name)
+          val resourceId1 = rs.get[ResourceId](r.resultName.name)
+          val accessPolicyName1 = rs.get[AccessPolicyName](pg.resultName.name)
```
dvoet: pg is an alias to the group table so its name will not be an access policy name

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -951,6 +952,81 @@ class PostgresAccessPolicyDAO(
...
+      resourceId: FullyQualifiedResourceId,
+      samRequestContext: SamRequestContext
+  ): IO[List[(FullyQualifiedPolicyId, FullyQualifiedPolicyId)]] =
+    readOnlyTransaction("findAffectedPolicyGroups", samRequestContext) { implicit session =>
+      val g = GroupTable.syntax("g")
+      val pg = GroupTable.syntax("pg")
+      val gm = GroupMemberTable.syntax("gm")
+      val p = PolicyTable.syntax("p")
+      val rt = ResourceTypeTable.syntax("rt")
+      val r = ResourceTable.syntax("r")
```
dvoet: I suggest naming these something more readable and include whether they are representing rows from the container or member policy. Same with fullyQualifiedPolicyId1 vs fullyQualifiedPolicyId2. One is the container and one is the member. Naming them better will help keep things straight.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -432,30 +432,21 @@ class ResourceService(
...
 
       // remove from cloud first so a failure there does not leave sam in a bad state
       _ <- cloudDeletePolicies(resource, samRequestContext)
       _ <- deleteActionManagedIdentitiesForResource(resource, samRequestContext)
 
       // leave a tomb stone if the resource type does not allow reuse
       leaveTombStone = !resourceTypes(resource.resourceTypeName).reuseIds
+      affectedPolicies <- accessPolicyDAO.findAffectedPolicyGroups(resource, samRequestContext) // New method
+      _ <- affectedPolicies.traverse { case (policyToUpdate, policyToRemove) => removeSubjectFromPolicy(policyToUpdate, policyToRemove, samRequestContext) }
+      _ <- cloudSyncPolicies(affectedPolicies, samRequestContext) // TODO: use this on affected policies
```
davidangb: if you are using `removeSubjectFromPolicy`, do you still need the call to `cloudSyncPolicies`? Does `removeSubjectFromPolicy` handle the syncing internally?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/AccessPolicyDAO.scala
```diff
@@ -130,7 +130,12 @@ trait AccessPolicyDAO {
       samRequestContext: SamRequestContext
   ): IO[Seq[FilterResourcesResult]]
 
-  def checkPolicyGroupsInUse(resourceId: FullyQualifiedResourceId, samRequestContext: SamRequestContext): IO[List[Map[String, String]]]
+  // TODO: better name?
+  def findAffectedPolicyGroups(
+      resourceId: FullyQualifiedResourceId,
+      samRequestContext: SamRequestContext
+  ): IO[List[(FullyQualifiedPolicyId, FullyQualifiedPolicyId)]]
+
```
davidangb, line 137: findPolicyGroupMemberhips? findPolicyGroupParticipation? maybe findPolicyGroupsInUse to reflect the method it is supplanting?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -569,6 +569,7 @@ class PostgresAccessPolicyDAO(
 
   override def deleteResource(resource: FullyQualifiedResourceId, leaveTombStone: Boolean, samRequestContext: SamRequestContext): IO[Unit] =
     serializableWriteTransaction("deleteResource", samRequestContext) { implicit session =>
+      deletePolicyMembersFromGroups(resource)
```
davidangb: is this still necessary here, given that `ResourceService` calls `removeSubjectFromPolicy`?

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAOSpec.scala
```diff
@@ -3553,6 +3553,61 @@ class PostgresAccessPolicyDAOSpec extends AnyFreeSpec with Matchers with BeforeA
...
+        dao.createResource(childResource, samRequestContext).unsafeRunSync()
+
+        // Add child policy to parent policy
+        dirDao.addGroupMember(parentPolicy.id, childPolicy.id, samRequestContext).unsafeRunSync()
+
+        val policyGroups = dao.findAffectedPolicyGroups(childResourceFullyQualifiedId, samRequestContext).unsafeRunSync()
+
+        policyGroups should not be empty
+        policyGroups.head._1 shouldEqual parentPolicy.id
+        policyGroups.head._2 shouldEqual childPolicy.id
```
davidangb: I think you could also write this as:
```scala
policyGroups should contain theSameElementsAs List(parentPolicy.id, childPolicy.id)
```

optional; pick whatever approach is more consistent with Sam code. I do see `theSameElementsAs` used elsewhere in Sam.

dvoet: need an extra set of parens in there
`policyGroups should contain theSameElementsAs List((parentPolicy.id, childPolicy.id))`
and that 1 line can replace your 3

src/test/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAOSpec.scala
```diff
@@ -3553,6 +3553,61 @@ class PostgresAccessPolicyDAOSpec extends AnyFreeSpec with Matchers with BeforeA
...
+        )
+
+        val parentResource =
+          Resource(parentResourceFullyQualifiedId.resourceTypeName, parentResourceFullyQualifiedId.resourceId, Set.empty, Set(parentPolicy))
+        val childResource =
+          Resource(
+            childResourceFullyQualifiedId.resourceTypeName,
+            childResourceFullyQualifiedId.resourceId,
+            Set.empty,
+            Set(childPolicy)
```
dvoet: I would add 2 more policies to `childResource`, one that is not a member of a `parentResource` policy and on that is a member of a different policy. Maybe even add another parent resource.

dvoet (approved):
I think this looks right. Approved assuming the cleanup David pointed and my test comments.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
calypsomatic: Not sure, actually.  Maybe @dvoet knows?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
calypsomatic: I guess not!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
dvoet: you don't need to call `cloudSyncPolicies`

davidangb (approved):
thanks for sticking with this one!

PR #1577, opened 2024-11-02
CORE-137 Optimize filter resources

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-137

What:

The query implemented in `PostgresAccessPolicyDAO.filterPrivateResources` has been highlighted by query insights as far and away the biggest load by time on the database. It is called on average 1.1 million times per day! It is the query behind the [api to list user resources](https://sam.dsde-prod.broadinstitute.org/#/Resources/listResourcesAndPoliciesV2). Around the beginning of October, we also noticed in dev that this query was [causing stability](https://broadinstitute.slack.com/archives/C03GMG4DUSE/p1728403148967779) issues because the AoU user had access to too many google-project resources. Those were cleaned up but the vulnerability remains.

How:

There were 2 main observations:
1. including actions associated to roles for each resource in the query results multiplied the number of rows returned, for the AoU case there were millions. But the role to action mapping is static from Sam config and is already loaded in memory. Moving all this extra data around is wasteful and is probably leading to excess garbage collection.
2. removing the joins to figure out auth domain information sped up the AoU queries. Prior versions of this query loaded auth domains separately especially because not all resource types are affected by them. Revert to that.

Overall, the workhorse query is much smaller in terms of joins.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
@@ -1780,9 +1780,8 @@ from ${GroupMemberTable as groupMemberTable}
     }
   }
 
-  private val publicResourcesCache: ConcurrentMap[ResourceTypeName, Seq[FilterResourcesResult]] = {
+  private val publicResourcesCache: ConcurrentMap[ResourceTypeName, Seq[FilterResourcesResult]] =
     Caffeine.newBuilder().expireAfterWrite(1, TimeUnit.HOURS).build[ResourceTypeName, Seq[FilterResourcesResult]]().asMap()
-  }
 
   private def getPublicResourcesOfType(resourceTypeName: ResourceTypeName, samRequestContext: SamRequestContext): IO[Seq[FilterResourcesResult]] = {
     val resourcePolicy = PolicyTable.syntax("resourcePolicy")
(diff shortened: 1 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1779,8 +1777,8 @@ from ${GroupMemberTable as groupMemberTable}
     }
   }
 
-  private val publicResourcesCache: java.util.Map[ResourceTypeName, Seq[FilterResourcesResult]] =
-    Collections.synchronizedMap(new PassiveExpiringMap(1, TimeUnit.HOURS))
+  private val publicResourcesCache: ConcurrentMap[ResourceTypeName, Seq[FilterResourcesResult]] =
+    Caffeine.newBuilder().expireAfterWrite(1, TimeUnit.HOURS).build[ResourceTypeName, Seq[FilterResourcesResult]]().asMap()
```
dvoet, line 1781: I switched to a concurrent hashmap implementation because synchronized maps do not allow concurrent reads. Even if the reads are fast, synchronized can lead to jam ups and interference between different resource types.

davidangb:
> including actions associated to roles for each resource in the query results multiplied the number of rows …. But the role to action mapping is static from Sam config and is already loaded in memory.

If the role:action association exists both as static config and in the db, does this change lead to a possible inconsistency?

1. A user is granted a role on some resource; the role's actions are persisted to the db for this user
2. A DSP engineer modifies the role:action mapping in the static config
3. This list-resources API now returns the updated actions for the role, but some other APIs still rely on the db for actions and return a different result

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -116,7 +116,10 @@ class ResourceService(
                 samRequestContext
               )
             }
-        } yield upsertedPolicy
+        } yield {
+          logger.info(s"Upserted policy $fullyQualifiedPolicyId")
```
davidangb, line 120: is this intentional or a leftover from development?

davidangb (approved):
This looks ok to me. I have to admit I am too unfamiliar with it to be very confident. I get the high-level design change and that sounds good.

Does the contract test failure indicate a real change in functionality?

dvoet:
> If the role:action association exists both as static config and in the db, does this change lead to a possible inconsistency?
> 
> 1. A user is granted a role on some resource; the role's actions are persisted to the db for this user
> 2. A DSP engineer modifies the role:action mapping in the static config
> 3. This list-resources API now returns the updated actions for the role, but some other APIs still rely on the db for actions and return a different result

No, step 1 in your list is not how it works. The role is assigned to the user, not the role's actions. The role:action association stored in the database is only update on startup to match what is in config.

davidangb:
> No, step 1 in your list is not how it works. The role is assigned to the user, not the role's actions. The role:action association stored in the database is only update on startup to match what is in config.

Got it - thank you!

marctalbott (approved):

PR #1604, opened 2024-12-05
CORE-210 add can_access action

dvoet:
https://github.com/DataBiosphere/leonardo/pull/4808 must merge first

Ticket: https://broadworkbench.atlassian.net/browse/CORE-210

When a user loses access to a workspace, they need to lose all access to Leonardo compute resources as well. We will do this with a new concept in Sam called a prerequisite action. A prerequisite action is an action on a resource that is required but not sufficient to perform any other action on the same resource. In the case of Leo resources (notebook clusters, persistent disks and kubernetes apps), this action, `can_access`, is passed down from the containing workspace through the `can-compute` role which includes the `accessor` Leo role. It is also passed down through the workspace `owner` role which includes the `manager` role on the Leo resource. The net effect is that in order to do anything on these resources, the caller must be a writer with compute access or an owner of a workspace. If the caller loses that access, the Leo resources in the workspace continue to exist but are inaccessible by the caller. If the caller regains access, they regain access to the Leo resources as well.

---

**PR checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,15 @@ The “owner” role of a resource generally will include delete action and acti
 
 Resource types define the set of available actions for all resources of that type. It also defines a set of roles and their associated actions. Roles are useful because it can be cumbersome to deal with granular actions and as a point of extensibility (when new actions are added to resource types, they can be added to roles as well, effectively adding the action to all resources with that role). Creating and maintaining resource types is achieved through [configuration](src/main/resources/reference.conf).
 
+#### Resource Type Configuration
+Configuration options for resource types are defined in the `resourceTypes` section of the configuration file. Each resource type has the following fields:
+* `actionPatterns` - a set of regex patterns that are used to document and validate the actions available for resources of this type
+* `roles` - a set of roles that are available for resources of this type. Roles are a collection of actions. Both roles and actions can be assigned to resource policies, but it is highly recommended to use roles because they are easier to change and affect all resources with that role as opposed to updating policies to add new actions.
+* `ownerRoleName` - the name of the role that is considered the "owner" role for all resources of this type. All resources must have a policy with this role or have a parent.
+* `reuseIds` - whether to allow reusing ids when creating resources of this type. This is important to prevent when using auth domains because users should not be able to delete then recreate a resource in Sam omitting the auth domain. Should be false when using UUIDs for Sam resource ids. Default is false.
+* `allowLeaving` - whether to allow users to leave resources of this type, otherwise an owner must remove them. Default is false.
+* `prerequisiteAction` - an optional action that must be granted before a user can perform any other actions on a resource of this type. Useful for resources that require some kind of access to a parent resource before accessing a child.
+
 ### Public Policies
 There are some cases where it is desirable to grant actions or roles to all authenticated users. For example, granting read-only access to public workspaces. In this case a policy can be created that has the appropriate actions or roles and set to public. Resources with public policies show up when listing resources for a user. For this reason it is not always desirable to allow everyone to make public policies. Again, the example is public workspaces. Public workspaces show up for everyone and should be curated.
 
@@ -62,7 +71,7 @@ Group - Create, delete, read, list, add/remove users and groups. Nested groups a
 
 ### Built In Actions
 * read_policies - may read all policies of a resource 
-* alter_policies - may change any policy of a resource
+* alter_policies - may add or change any policy of a resource, use sparingly, prefer share_policy below for more control over policy structure
 * delete - may delete a resource
 * share_policy::{policy name} - may add/remove members to/from specified policy of a resource
 * read_policy::{policy name} - may read specified policy of a resource
--- a/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20241205_prereq_action_column.xml
+++ b/src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20241205_prereq_action_column.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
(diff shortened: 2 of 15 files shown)


src/main/resources/reference.conf
```diff
@@ -826,11 +831,17 @@ resourceTypes = {
...
+      # the creator role is assigned directly to the notebook-cluster resource and does not include can_access
+      # but can_access is required to perform any action so must be granted via the parent resource using
+      # either the manager or accessor role
       creator = {
         roleActions = ["status", "connect", "delete", "read_policies", "stop_start", "modify", "set_parent", "get_parent"]
       }
       manager = {
-        roleActions = ["status", "delete", "read_policies"]
+        roleActions = ["status", "delete", "read_policies", "can_access"]
+      }
```
davidangb, line 843: I see you added `descendantRoles` for `accessor` but not `manager` (for both notebook-cluster and kubernetes-app). How is `manager` assigned to users? Is that assigned directly somewhere by Leo?

davidangb (approved):
:+1: with one question inline

src/main/resources/reference.conf
dvoet: manager is already a descendant role from workspace owner

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20241205_prereq_action_column.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog logicalFilePath="dummy"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
+
+    <changeSet logicalFilePath="dummy" author="dvoet" id="rt_prereq_action_column">
+        <addColumn tableName="sam_resource_type">
+            <column name="prerequisite_action" type="VARCHAR">
```
marctalbott, line 10: Should this be a FK? I suppose the owner role isn't either so seems fine.

marctalbott (approved):

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20241205_prereq_action_column.xml
dvoet: making it a FK creates some circular headaches when creating resource types: insert the resource type first, then actions, then update the resource type. Not worth it IMO.

src/main/scala/org/broadinstitute/dsde/workbench/sam/db/SamTypeBinders.scala
```diff
@@ -17,123 +17,123 @@ import scalikejdbc.TypeBinder
...
-    def apply(rs: ResultSet, index: Int): ResourceActionPK = ResourceActionPK(rs.getLong(index))
+    def apply(rs: ResultSet, label: String): ResourceActionPK = nullSafe(rs.getLong(label), ResourceActionPK)
+    def apply(rs: ResultSet, index: Int): ResourceActionPK = nullSafe(rs.getLong(index), ResourceActionPK)
   }
 
   implicit val resourceActionNameTypeBinder: TypeBinder[ResourceAction] = new TypeBinder[ResourceAction] {
-    def apply(rs: ResultSet, label: String): ResourceAction = ResourceAction(rs.getString(label))
-    def apply(rs: ResultSet, index: Int): ResourceAction = ResourceAction(rs.getString(index))
+    def apply(rs: ResultSet, label: String): ResourceAction = nullSafe(rs.getString(label), ResourceAction)
+    def apply(rs: ResultSet, index: Int): ResourceAction = nullSafe(rs.getString(index), ResourceAction)
```
dvoet, line 66: The changes in this file were sparked because `prerequisiteAction` is optional and the prior code resulted in these functions returning `ResourceAction(null)` which is not valid. I fixed all the other functions in this file as well just in case.

src/main/resources/reference.conf
```diff
@@ -862,17 +878,27 @@ resourceTypes = {
...
+        roleActions = ["delete", "read", "read_policies", "can_access"]
+      }
+      accessor = {
+        roleActions = ["can_access"]
       }
     }
     reuseIds = false
+    prerequisiteAction = "can_access"
   }
   kubernetes-app = {
```
davidangb, line 897: does `kubernetes-app` need `prerequisiteAction = "can_access"` too?

davidangb (approved):
one question about `kubernetes-app` config inline.

Is there any call for `prerequisiteAction` to be an array of actions instead of just a single action? That's probably too complicated, but if we want multiple we should do it now so we don't have to change the models later.

proactive :+1:

src/main/resources/reference.conf
dvoet: I assume you mean `kubernetes-app-shared` and if so, no because _all_ permissions are already inherited. The other resources behave differently because there is a `creator` role that is assigned directly to a single user.

dvoet:
> Is there any call for prerequisiteAction to be an array

I thought about this and could not decide if it should be implemented as AND (must have all prereqs) or OR (any prereq will suffice). So I kept is simple :)

I don't think updating the models later will be much of a problem

src/main/resources/reference.conf
dvoet: actually, you were right, I missed it

README.md
```diff
@@ -50,6 +50,15 @@ The “owner” role of a resource generally will include delete action and acti
 
 Resource types define the set of available actions for all resources of that type. It also defines a set of roles and their associated actions. Roles are useful because it can be cumbersome to deal with granular actions and as a point of extensibility (when new actions are added to resource types, they can be added to roles as well, effectively adding the action to all resources with that role). Creating and maintaining resource types is achieved through [configuration](src/main/resources/reference.conf).
 
+#### Resource Type Configuration
```
LizBaldo, line 53: ❤️

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20241205_prereq_action_column.xml
LizBaldo, line 10: Just curious as to why the `prerequisite_action` is a VARCHAR instead of an ENUM?

LizBaldo (approved):

src/main/resources/org/broadinstitute/dsde/sam/liquibase/changesets/20241205_prereq_action_column.xml
dvoet: the value is not built-in, it comes from configuration

PR #1613, opened 2024-12-20
CORE-242 bulk membership update

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-242

In addition to add the new api, refactored how some other calls validate (see comments in PR).

---

**PR checklist**

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1975,6 +1975,49 @@ paths:
                 oneOf:
                   - $ref: '#/components/schemas/FilteredResourcesFlatResponse'
                   - $ref: '#/components/schemas/FilteredResourcesHierarchicalResponse'
+  /api/resources/v2/bulkMembershipUpdate:
+    post:
+      tags:
+        - Resources
+      summary: Bulk update the membership for resources
+      description: |
+        This endpoint is used to add and remove members of multiple policies across multiple resource in one api call.
+        Each policy is updated in order in its own transaction. If a policy does not exist on a resource, 
+        processing stops with an error but all updates up to that point are committed. Missing member emails or policies fail fast. Additions are processed
+        before removals so if a subject is in both the add and remove lists for a policy, the ultimate result is removal.
+        This call is idempotent, if members being added already exist or members being removed do not exist, nothing will happen.
+      operationId: bulkMembershipUpdateV2
+      requestBody:
+        description: The details of the membership updates
+        content:
+          'application/json':
+            schema:
+              type: array
+              items:
+                $ref: '#/components/schemas/BulkMembershipUpdateRequestV2'
+        required: true
+      responses:
+        200:
+          description: Successfully updated membership
+          content: { }
+        400:
+          description: members or policies are not found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        403:
+          description: Caller does not have permission to update the membership of one or more of the policies or one or more resources do not exist
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
   /api/resources/v2/{resourceTypeName}:
     get:
       tags:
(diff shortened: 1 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresAccessPolicyDAO.scala
```diff
@@ -1038,7 +1039,6 @@ from ${GroupMemberTable as groupMemberTable}
     }
     removeAllGroupMembers(groupId)
     insertGroupMembers(groupId, memberList)
-    updateGroupUpdatedDateAndVersion(id)
```
dvoet, line 1041: this is an internal function, calling `updateGroupUpdatedDateAndVersion` is redundant

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -778,10 +837,6 @@ class ResourceService(
       changeEvents = createAccessChangeEvents(policyId.resource, originalPolicies, updatedPolicies)
       _ <- AuditLogger.logAuditEventIO(samRequestContext, changeEvents.toSeq: _*)
 
-      _ <- directoryDAO.updateGroupUpdatedDateAndVersionWithSession(
-        FullyQualifiedPolicyId(policyId.resource, policyId.accessPolicyName),
-        samRequestContext
-      )
```
dvoet, line 784: I think this was here as a failsafe but was double doing work. I made sure it is correctly called in the DAO. Maybe a follow on would be to figure out a trigger even though I despise triggers.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -851,51 +906,14 @@ class ResourceService(
   ): IO[Set[ValidatableAccessPolicy]] =
     policies.toList
       .traverse { case (accessPolicyName, accessPolicyMembershipRequest) =>
-        for {
-          // grouping member emails and policy emails together
```
dvoet, line 855: It looks like this code was getting the email address for each policy and adding them to `memberEmails` of the policy. The problem is that policies that do not exist are just ignored. There is also no reason to load a policy by its identifiers, get its email, then use the email to load its identifiers. I changes all this to validate policies (so I could use that for the new api) then use the policy identifiers directly.

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/ResourceService.scala
```diff
@@ -215,7 +255,9 @@ class ResourceService(
   private def constructAccessPolicy(resourceType: ResourceType, resourceId: ResourceId, validatableAccessPolicy: ValidatableAccessPolicy, public: Boolean) =
     AccessPolicy(
       FullyQualifiedPolicyId(FullyQualifiedResourceId(resourceType.name, resourceId), validatableAccessPolicy.policyName),
-      validatableAccessPolicy.emailsToSubjects.values.flatten.toSet,
+      validatableAccessPolicy.emailsToSubjects.values.flatten.toSet ++ validatableAccessPolicy.memberPolicies
+        .getOrElse(Set.empty)
+        .map(_.toFullyQualifiedPolicyId),
```
dvoet, line 276: See [comment below](https://github.com/broadinstitute/sam/pull/1613/files#r1894413066). `memberPolicies` are no longer included in `emailsToSubjects` so add them here when constructing the policy

calypsomatic (approved):
One small nit: the ticket specifies that the API should be a PATCH instead of a POST.  Does it matter?

dvoet:
> One small nit: the ticket specifies that the API should be a PATCH instead of a POST. Does it matter?

I changed the ticket ;)

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2001,6 +2001,49 @@ paths:
...
+  /api/resources/v2/bulkMembershipUpdate:
+    post:
+      tags:
+        - Resources
+      summary: Bulk update the membership for resources
+      description: |
+        This endpoint is used to add and remove members of multiple policies across multiple resource in one api call.
+        Each policy is updated in order in its own transaction. If a policy does not exist on a resource, 
+        processing stops with an error but all updates up to that point are committed. Missing member ids, emails or policies fail fast. Additions are processed
+        before removals so if a subject is in both the add and remove lists for a policy, the ultimate result is removal.
```
samanehsan, line 2013: What about checking that the add and remove lists are different and throwing an error if there's any overlap?

samanehsan (approved):

PR #1617, opened 2025-01-03
CORE-232: user supportSummary API

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-232

## What:

New admin API for information about a user, including direct, indirect, and unsynchronized group membership counts.

This is implemented by adding the group membership counts to the existing logic for the `getSamUserCombinedState` API, and then creating a new admin API which reuses that logic but allows an admin to execute that logic for any user.

## Review Notes:
1. This PR requires super-admin to call the new supportSummary API. This is temporary. To keep PRs small, I hope to merge the PR as-is - including the restrictive permissions - and then merge #1620, and then finally create another PR that opens up the new supportSummary API to the support team by checking for resource_type_admin permissions instead of super-admin permissions.
2. In my testing while running a local Sam against a local Postgres, I always have 1 unsynchronized group: all_users. Will this be true in prod as well? If so, we probably want to add a special case for that group to avoid counting it.

## Example
This new API returns a payload that looks like:

```json
{
  "additionalDetails": {
    "enterpriseFeatures": {
      "format": "flat",
      "resources": []
    }
  },
  "allowances": {
    "allowed": true,
    "details": {
      "enabled": true,
      "termsOfService": true
    }
  },
  "attributes": {
    "marketingConsent": true,
    "userId": "2736276474111094565b5"
  },
  "favoriteResources": [],
  "groupMembershipCounts": {
    "directSynchronized": 2,
    "totalSynchronized": 4,
    "unsynchronized": 1
  },
  "samUser": {
    "azureB2CId": "9ec01dd3-ddec-4273-8e34-39fdd2a480ab",
    "createdAt": "2025-01-07T19:01:14.166491Z",
    "email": "davidan.dev@gmail.com",
    "enabled": true,
    "googleSubjectId": "111137720113699110757",
    "id": "2736276474111094565b5",
    "registeredAt": "2025-01-07T19:01:14.164172Z",
    "updatedAt": "2025-01-07T19:01:18.917368Z"
  },
  "termsOfServiceDetails": {
    "acceptedOn": "2025-01-07T19:01:18.889191Z",
    "isCurrentVersion": true,
    "latestAcceptedVersion": "1",
    "permitsSystemUsage": true
  }
}
```

---

**PR checklist**


src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminRoutes.scala
```diff
@@ -284,4 +296,33 @@ trait AdminRoutes extends SecurityDirectives with SamRequestContextDirectives wi
...
+      indirectGroupMemberships <- userService.countIndirectGroupMemberships(samUser, samRequestContext)
+    } yield SamUserSupportSummaryResponse(
+      samUser,
+      allowances,
+      maybeAttributes,
+      termsOfServiceDetails.getOrElse(TermsOfServiceDetails(None, None, permitsSystemUsage = false, isCurrentVersion = false)),
+      Map("enterpriseFeatures" -> enterpriseFeatures.toJson),
+      directGroupMembershipCount = directGroupMemberships,
+      indirectGroupMembershipCount = indirectGroupMemberships
+    )
```
davidangb: This follows the pattern set by [getSamUserCombinedState in UserRoutesV2](https://github.com/broadinstitute/sam/blob/2fbfc2201f19f3da56a2b6114e85b55021d184d0/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala#L203-L230), in which business logic that requires multiple services is implemented in the route class. Is there a better place to put this logic?

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -528,6 +528,20 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
+      val loadUserQuery = samsql"""select ${userTable.resultAll}
+                                    from ${UserTable as userTable}
+                                    where ${userTable.email} = ${email}"""
+      loadUserQuery
+        .map(UserTable(userTable))
+        .single()
+        .apply()
+        .map(UserTable.unmarshalUserRecord)
+    }
+
```
davidangb, line 544: modeled after `loadUserByAzureB2CId`

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -718,6 +732,28 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
+    readOnlyTransaction("countDirectGroupMemberships", samRequestContext) { implicit session =>
+      val query = samsql"""select count(distinct g.id) indirectMembershipCount
+                          from sam_group g
+                          join sam_group_member_flat gmf on g.id = gmf.group_id
+                          where g.synchronized_date is not null
+                          and gmf.member_user_id = ${samUser.id}"""
+
+      query.map(rs => rs.int(1)).single().apply().getOrElse(0)
+    }
+
```
davidangb, line 756: I based these SQL queries off previous Slack discussion. I believe they are correct … but would love confirmation/oversight!

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -505,6 +508,12 @@ class UserService(
       case None => IO.raiseError(new WorkbenchExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"User $workbenchUserId not found")))
     }
   }
+
+  def countDirectGroupMemberships(samUser: SamUser, samRequestContext: SamRequestContext): IO[Int] =
+    directoryDAO.countDirectGroupMemberships(samUser, samRequestContext)
+
+  def countIndirectGroupMemberships(samUser: SamUser, samRequestContext: SamRequestContext): IO[Int] =
+    directoryDAO.countIndirectGroupMemberships(samUser, samRequestContext)
```
davidangb: I did not create unit tests for the three new methods here in UserService, since they don't add any logic above/beyond the PostgresDirectoryDAO methods.

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -718,6 +732,28 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
+                          from sam_group g
+                          join sam_group_member gm on g.id = gm.group_id
+                          where g.synchronized_date is not null
+                          and gm.member_user_id = ${samUser.id}"""
+
+      query.map(rs => rs.int(1)).single().apply().getOrElse(0)
+    }
+
+  override def countIndirectGroupMemberships(samUser: SamUser, samRequestContext: SamRequestContext): IO[Int] =
+    readOnlyTransaction("countDirectGroupMemberships", samRequestContext) { implicit session =>
```
marctalbott: nit: typo
```suggestion
    readOnlyTransaction("countIndirectGroupMemberships", samRequestContext) { implicit session =>
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
marctalbott: They look correct to me but I haven't done any manual verification

src/main/resources/swagger/api-docs.yaml
```diff
@@ -146,6 +146,32 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/UserStatus'
+  /api/admin/v1/user/email/{email}/supportSummary:
```
marctalbott, line 149: I generally favor the usage of Sam user ids over emails when identifying users, but I expect emails will usually be more convenient for callers of this API. It might be helpful to expose this information via both email and user id in this case.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminRoutes.scala
marctalbott: I don't love having the business logic in the route class, but alternatives are probably not worth the effort and the pattern already exists in Sam so seems fine to me.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/SamUserSupportSummaryResponse.scala
```diff
@@ -0,0 +1,21 @@
...
+  implicit val SamUserSupportSummaryResponseFormat: RootJsonFormat[SamUserSupportSummaryResponse] = jsonFormat7(SamUserSupportSummaryResponse.apply)
+}
+final case class SamUserSupportSummaryResponse(
+    samUser: SamUser,
+    allowances: SamUserAllowances,
+    attributes: Option[SamUserAttributes],
+    termsOfServiceDetails: TermsOfServiceDetails,
+    additionalDetails: Map[String, JsValue],
+    directGroupMembershipCount: Int,
+    indirectGroupMembershipCount: Int // TODO: for clarity, should this be "directAndIndirectGroup..."?
```
marctalbott: Yeah, I might think to add the two counts together to get total otherwise. Maybe `totalGroupMembershipCount`?

marctalbott: I also wonder if they should indicate that they are counts of synchronized groups and not total group membership in Sam

marctalbott (commented):
My understanding is that we can't expand access to the Sam super admin group to everyone who will need to call this API, so creating a `user` `resource_type_admin` resource seems reasonable to me even though it's definitely a departure from other `resource_type_admin` resources

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminRoutes.scala
dvoet: I would put this in `userService` (which already knows about `tosService`).

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -718,6 +732,28 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
     }
   }
 
+  override def countDirectGroupMemberships(samUser: SamUser, samRequestContext: SamRequestContext): IO[Int] =
```
dvoet:
```suggestion
  override def countDirectSynchronizedGroupMemberships(samUser: SamUser, samRequestContext: SamRequestContext): IO[Int] =
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
```diff
@@ -718,6 +732,28 @@ class PostgresDirectoryDAO(protected val writeDbRef: DbReference, protected val
...
+      val query = samsql"""select count(distinct g.id) directMembershipCount
+                          from sam_group g
+                          join sam_group_member gm on g.id = gm.group_id
+                          where g.synchronized_date is not null
+                          and gm.member_user_id = ${samUser.id}"""
+
+      query.map(rs => rs.int(1)).single().apply().getOrElse(0)
+    }
+
+  override def countIndirectGroupMemberships(samUser: SamUser, samRequestContext: SamRequestContext): IO[Int] =
```
dvoet:
```suggestion
  override def countIndirectSynchronizedGroupMemberships(samUser: SamUser, samRequestContext: SamRequestContext): IO[Int] =
```

src/main/scala/org/broadinstitute/dsde/workbench/sam/dataAccess/PostgresDirectoryDAO.scala
dvoet: They also look correct to me. One thing to note is that this does not include public resources but we probably want to deal with that separately.

dvoet (commented):
Sam currently has 2 admin mechanisms, the newer backed by resource type admin and the older backed by the fc-admins google group. It looks like the older is only used to gate access to user specific apis (should verify this). It would be great to consolidate to the newer. I think that would mean creating a `user` resource type admin resource even though `user` is not a resource type. I could see that being a follow on ticket. All of our support staff need to be suitable though, so they may already be members of fc-admins.

dvoet:
> This new API is very similar to the [getSamUserCombinedState](https://sam.dsde-dev.broadinstitute.org/#/Users/getSamUserCombinedState) API. Should they be consolidated?

I think it would be nice if they were the same api under hood. Exposing the counts to the user may be useful

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminRoutes.scala
davidangb: Unfortunately `UserService` doesn't know about `ResourceService`, which we need to get the enterprise features. Whaddya think - is it worth giving `UserService` that knowledge, or do we not actually care about the enterprise features at this point?

samanehsan (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/AdminRoutes.scala
davidangb: Moved to `UserService`. This is kinda hacky in that the method requires the caller to pass in a `ResourceService`. I am hesitant to change the `UserService` constructor just for this one API.

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/api/SamUserSupportSummaryResponse.scala
davidangb: I refactored this; now we have counts for `directSynchronized`, `totalSynchronized`, and `unsynchronized`.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2.scala
```diff
@@ -200,35 +203,6 @@ trait UserRoutesV2 extends SamUserDirectives with SamRequestContextDirectives wi
...
-          samUser,
-          allowances,
-          maybeAttributes,
-          termsOfServiceDetails.getOrElse(TermsOfServiceDetails(None, None, permitsSystemUsage = false, isCurrentVersion = false)),
-          Map("enterpriseFeatures" -> enterpriseFeatures.toJson),
-          favoriteResources
-        )
-      }
-    }
-
```
davidangb, line 231: This logic moved to `UserService` to make it reusable

src/main/resources/swagger/api-docs.yaml
davidangb: Yeah, I am expecting that support will be dealing with email addresses instead of userids - but it would be trivial to add a supportSummary API that accepts an id if we needed one

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
```diff
@@ -505,6 +509,60 @@ class UserService(
...
+      termsOfServiceDetails.getOrElse(TermsOfServiceDetails(None, None, permitsSystemUsage = false, isCurrentVersion = false)),
+      GroupMembershipCounts(
+        directSynchronized = directGroupMemberships,
+        totalSynchronized = indirectGroupMemberships,
+        unsynchronized = unsynchronizedGroupMemberships
+      ),
+      Map("enterpriseFeatures" -> enterpriseFeatures.toJson),
+      favoriteResources
+    )
+
```
davidangb, line 565: I think it's kinda smelly to require `ResourceService` as an argument to this method. But I also didn't want to update all 67 calls to the UserService constructor just to support this one method.

src/test/scala/org/broadinstitute/dsde/workbench/sam/api/UserRoutesV2Spec.scala
```diff
@@ -260,184 +254,6 @@ class UserRoutesV2Spec extends AnyFlatSpec with Matchers with TimeMatchers with
     }
   }
```
davidangb, line 256: Since the `getSamUserCombinedState()` method moved from `UserRoutesV2` to `UserService`, all of the following tests moved from here to the new `SupportSummarySpec`.

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/UserServiceSpecs/SupportSummarySpec.scala
```diff
@@ -0,0 +1,208 @@
+package org.broadinstitute.dsde.workbench.sam.service.UserServiceSpecs
```
davidangb, line 1: These are the tests moved from `UserRoutesV2Spec`, but I also cleaned them up a bunch to avoid code repetition and Sonar warnings

src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala
dvoet: I guess I agree. An alternative would be to create another Service class that depends on both but 🤷 . I leave it up to you.

dvoet (approved):

marctalbott (approved):

PR #1631, opened 2025-01-28
CORE-278: optimize google key cache; more aggressive cleanup

davidangb:
_experimental, do not review yet_

---

At a high level, the logic in the Google key cache is:
1. Look for an active cached key
2. If found: use that key
3. If not found: clean up old keys, create a new key, use the new key

This PR contains improvements to the implementation of this logic:

* When looking for an active cached key, consider all cached keys and not just the most recent one
* When cleaning up old keys, clean up more cases of invalid keys
* When cleaning up old keys, catch errors on key deletion. Previously, an error during key deletion could stop the whole flow and prevent the user from receiving a valid key

---

**PR checklist**

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLockSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLockSpec.scala
@@ -32,7 +32,7 @@ class DistributedLockSpec extends AsyncFlatSpec with Matchers with TestSupport {
     res.attempt.map(r => r.isRight shouldBe true).unsafeToFuture()
   }
 
-  it should "fail if there's same lock has already been set within 30 seconds" in {
+  it should "fail on acquireLock if there's same lock has already been set within 30 seconds" in {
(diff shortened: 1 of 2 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/sam/util/DistributedLockSpec.scala
```diff
@@ -46,6 +46,20 @@ class DistributedLockSpec extends AsyncFlatSpec with Matchers with TestSupport {
...
+    val res = lockResource.use { lock =>
+      for {
+        _ <- lock.withLock(lockDetails).use(_ => IO.pure(()))
+        _ <- IO.sleep(2 seconds)
+        failed <- lock.withLock(lockDetails).use(_ => IO.pure(())).attempt
+      } yield failed.swap.toOption.get.asInstanceOf[FailToObtainLock].getMessage shouldBe s"can't get lock: $lockDetails"
+    }
+    res.unsafeToFuture()
+  }
+
```
davidangb: this unit test fails; I wrote the test to understand what the lock is actually doing

PR #1632, opened 2025-01-29
CORE-278: error logging; flexibility in finding active keys

davidangb:

## What:

This assists in debugging CORE-278 but does not solve that ticket:
* enhance logging when retrieving a pet SA key
* when querying the key cache, use any active/valid key instead of only considering the most recent cache entry

## Why:

Hopefully, getting this logging into prod will make it easier to pinpoint the actual problems in CORE-278.

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
@@ -123,15 +123,20 @@ class GoogleKeyCache(
   private def retrieveActiveKey(pet: PetServiceAccount): IO[(Option[String], List[GcsObjectName], List[ServiceAccountKey])] =
     for {
       (keysFromCache, keysFromIam) <- fetchKeysFromCacheAndIam(pet)
-      maybeMostRecentKey = keysFromCache.sortBy(_.timeCreated.toEpochMilli).lastOption
-      mostRecentKey <- maybeMostRecentKey match {
-        case Some(mostRecentKey) if isKeyActive(mostRecentKey, keysFromIam) =>
+
+      // TODO CORE-278: this could result in log spam
+      _ = if (keysFromIam.length >= 10)
+        logger.warn(s"danger: pet ${pet.serviceAccount.displayName.value} has ${keysFromIam.length} keys (cache has ${keysFromCache.length})")
+
+      maybeActiveKey = keysFromCache.sortBy(_.timeCreated.toEpochMilli).findLast(x => isKeyActive(x, keysFromIam))
+      activeKey <- maybeActiveKey match {
+        case Some(key) =>
           googleStorageAlg
-            .unsafeGetBlobBody(googleServicesConfig.googleKeyCacheConfig.bucketName, GcsBlobName(mostRecentKey.value))
+            .unsafeGetBlobBody(googleServicesConfig.googleKeyCacheConfig.bucketName, GcsBlobName(key.value))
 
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -123,15 +127,20 @@ class GoogleKeyCache(
...
       (keysFromCache, keysFromIam) <- fetchKeysFromCacheAndIam(pet)
-      maybeMostRecentKey = keysFromCache.sortBy(_.timeCreated.toEpochMilli).lastOption
-      mostRecentKey <- maybeMostRecentKey match {
-        case Some(mostRecentKey) if isKeyActive(mostRecentKey, keysFromIam) =>
+
+      // TODO CORE-278: this could result in log spam
+      _ = if (keysFromIam.length >= 10)
+        logger.warn(s"danger: pet ${pet.serviceAccount.displayName.value} has ${keysFromIam.length} keys (cache has ${keysFromCache.length})")
+
+      maybeActiveKey = keysFromCache.sortBy(_.timeCreated.toEpochMilli).findLast(x => isKeyActive(x, keysFromIam))
```
davidangb: in cases where the bucket-based key cache and the actual IAM list of keys are out of sync for whatever reason, this change will be slightly more flexible in finding a usable key. It will only make a difference in cases where the pet has multiple keys already. If there is some kind of race condition or contention where we create multiple keys in rapid succession, this change should have a better chance at finding a usable one.

calypsomatic (approved):

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -123,15 +127,20 @@ class GoogleKeyCache(
...
+      activeKey <- maybeActiveKey match {
+        case Some(key) =>
           googleStorageAlg
-            .unsafeGetBlobBody(googleServicesConfig.googleKeyCacheConfig.bucketName, GcsBlobName(mostRecentKey.value))
+            .unsafeGetBlobBody(googleServicesConfig.googleKeyCacheConfig.bucketName, GcsBlobName(key.value))
 
         case _ => IO.pure(None)
```
dvoet:
```
      maybeActiveKey = keysFromCache.sortBy(_.timeCreated.toEpochMilli).findLast(isKeyActive(_, keysFromIam))
      activeKey <- maybeActiveKey.map { key =>
          googleStorageAlg
            .unsafeGetBlobBody(googleServicesConfig.googleKeyCacheConfig.bucketName, GcsBlobName(key.value))
        }.getOrElse(IO.pure(None))
```

dvoet (approved):
I left 1 style suggestion. I don't think this actually changes any functionality. The PR description says something about returning any "active" key, not just the last one. But in reality there should only be 1 "active" key at any given time. And even if there is more than one then we just get the last one anyway.

davidangb:
> I don't think this actually changes any functionality

@dvoet this shouldn't change any functionality _unless_ the IAM listing is eventually consistent.

request 1: create new key 1, save key 1 to cache
request 2: sees key 1 in cache but not in IAM, considers key 1 inactive. Create new key 2, save key 2 to cache.
request 3: sees key 2 in cache (most recent), but key 2 isn't in IAM. Goes to create new key 3.

The change in this PR would allow request 3 to consider both key 2 and key 1 … hopefully by that time the IAM listing has caught up for key 1 and the request can use it.

PR #1637, opened 2025-02-06
CORE-278: more resilience when retrieving pet keys

davidangb:

This is fairly pervasive rewrite of `GoogleKeyCache`; apologies to reviewers.

In an effort to reduce key thrashing and to make a best effort to provide a usable service account key, we're changing the logic on how keys are retrieved. Much of this is in response to Google Support's information that newly-created keys are not consistently usable for something on the order of "typically 2 minutes, potentially 7 minutes or longer"

Glossing over some details, the logic has changed thusly:

_**previously**_
read keys from the key cache. Find any key in cache which is less than 12 days old and which has a corresponding key in the Google IAM listing. If no key found, create a new one.

_**in this PR**_
read keys from the key cache. Look for, in order of preference:
1. keys which are older than 15 minutes but less than 12 days old
2. keys which are older than 12 days
3. keys which are newer than 15 minutes

there is additional logic in this PR to clean up keys prior to creating a new one, and to avoid creating multiple keys in quick succession.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
@@ -278,7 +278,7 @@ class GoogleKeyCache(
       logger.info(
         s"cleanupKeys: ${pet.id.project.value}-${pet.serviceAccount.subjectId.value} is reaping ${toDelete.size} keys due to quota"
       )
-      toDelete.parTraverse(cachedKey => removeKey(pet, cachedKey.keyId))
+      toDelete.traverse(cachedKey => removeKey(pet, cachedKey.keyId))
     } else {
       IO.unit
     }
@@ -291,7 +291,7 @@ class GoogleKeyCache(
       logger.info(
         s"cleanupKeys: ${pet.id.project.value}-${pet.serviceAccount.subjectId.value} is reaping ${unknownKeyIds.size} orphaned keys"
       )
-      unknownKeyIds.toList.parTraverse(keyId => IO.fromFuture(IO(googleIamDAO.removeServiceAccountKey(pet.id.project, pet.serviceAccount.email, keyId))))
+      unknownKeyIds.toList.traverse(keyId => IO.fromFuture(IO(googleIamDAO.removeServiceAccountKey(pet.id.project, pet.serviceAccount.email, keyId))))
     } else {
       IO.unit
     }
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
@@ -97,21 +97,28 @@ class GoogleKeyCache(
   /** Retrieve a key for this pet, creating keys as necessary.
     */
   override def getKey(pet: PetServiceAccount): IO[String] = {
-    def maybeCreateKey(withinLock: Boolean, fallback: () => IO[String]): IO[String] =
-      for {
-        maybeActiveKey <- findActiveKey(pet, withinLock)
-        activeKey <- maybeActiveKey match {
-          case Some(existingActiveKey) => IO.pure(existingActiveKey)
-          case None => fallback()
-        }
-      } yield activeKey
-
-    def failed(): IO[String] =
-      throw new WorkbenchException("Could not create or retrieve key")
-
     // 5 minute lock timeout chosen to match how long key creation polling can take
+    // TODO CORE-278: when we remove polling from workbench-libs, turn this lock time back down
     val lockDetails = LockDetails(s"${pet.id.project.value}-getKey", pet.serviceAccount.subjectId.value, 5 minutes)
-    maybeCreateKey(withinLock = false, () => distributedLock.withLock(lockDetails).use(_ => maybeCreateKey(withinLock = true, () => failed())))
+
+    for {
+      // try to find a key using read-only logic by specifying withinLock = false
+      maybeActiveKey <- findActiveKey(pet, withinLock = false)
+      activeKey <- maybeActiveKey match {
+        case Some(key) => IO.pure(key)
+        case None =>
+          // obtain a lock, then try again to find a key, allowing key creation by specifying withinLock = true
+          distributedLock
+            .withLock(lockDetails)
+            .use(_ => findActiveKey(pet, withinLock = true).attempt)
+            .map {
+              case Right(Some(key)) => key
+              case Right(None) => throw new WorkbenchException("Could not create or retrieve key: unexpected problem")
+              case Left(t) =>
+                throw new WorkbenchException(s"Could not create or retrieve key: ${t.getMessage}")
+            }
+      }
+    } yield activeKey
   }
 
   /** List all cached keys from the Google bucket for this pet.
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
@@ -166,9 +166,9 @@ class GoogleKeyCache(
       // if not within lock, return None. This will signal callers to obtain a lock and re-call this function.
       None
     } else if (nascentKeys.isEmpty && retiredKeys.nonEmpty) {
-      // if within lock, retired keys exist, and no nascent keys exist: return newest retired key and trigger key creation
-      // TODO CORE-278: trigger new-key creation; can be async
-      newestOf(retiredKeys)
+      // if within lock, retired keys exist, and no nascent keys exist:
+      // return newest retired key and trigger key creation
+      furnishNewKey(pet).map(_ => newestOf(retiredKeys)).unsafeRunSync()
     } else {
       // no keys exist; create one and return it
       Option(furnishNewKey(pet).unsafeRunSync())
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
@@ -26,6 +26,8 @@ import org.broadinstitute.dsde.workbench.sam.model.CachedKey.keyPathPattern
 
 import java.time.Instant
 
+import cats.effect.unsafe.implicits.global
+
 /** Created by mbemis on 1/10/18.
   */
 class GoogleKeyCache(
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -157,19 +226,40 @@ class GoogleKeyCache(
...
           else if (e.getDetails.getCode == StatusCodes.BadRequest.intValue && e.getDetails.getMessage == "Precondition check failed.")
             throw new WorkbenchException("You may have reached the 10 key limit on service accounts.")
           else
             throw new WorkbenchException(s"Error creating key for service account: ${e.getDetails.getCode}: ${e.getDetails.getMessage}")
+        case t: Throwable =>
+          logger.error(
+            s"furnishNewKey: ${pet.id.project.value}-${pet.serviceAccount.subjectId.value} ${t.getMessage}",
+            t
+          )
+          throw new WorkbenchException(s"Error creating key for service account: ${t.getMessage}")
```
davidangb, line 269: this is the antipattern of both logging and throwing. I really wanted to log the stack trace for these errors, but really didn't want to include the stack trace when showing the exception to the user … so I'm logging and throwing.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -92,57 +94,124 @@ class GoogleKeyCache(
...
+      // trigger key creation and then return newest retired key
+      for {
+        _ <- cleanupAndCreateKey(pet, keysInCache)
+        retiredKey <- newestOf(retiredKeys)
+      } yield {
+        logger.warn(
+          s"searchCachedKeys: ${pet.id.project.value}-${pet.serviceAccount.subjectId.value} is using a retired key; no nascent keys exist"
+        )
+        retiredKey
+      }
```
davidangb, line 205: in this case, I'd love to fire-and-forget the new key creation and return the retired key as fast as possible to the user. However, that would mean we release the distributed lock before new key creation finishes, which could lead to multiple keys being created in rapid succession if callers request keys multiple times.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -92,57 +94,124 @@ class GoogleKeyCache(
...
-    def cleanupAndCreateKey(keysFromCache: List[GcsObjectName], keysFromIam: List[ServiceAccountKey]): IO[String] = {
-      logger.info(
-        s"cleanupAndCreateKey: ${pet.id.project.value}-${pet.serviceAccount.subjectId.value} with ${keysFromCache.length} in cache and ${keysFromIam.length} in IAM"
-      )
-      for {
-        _ <- IO.fromFuture(IO(cleanupUnknownKeys(pet, keysFromCache, keysFromIam)))
-        key <- furnishNewKey(pet)
-      } yield key
-    }
+    def failed(): IO[String] =
```
dvoet: I don't think there is a situation where `failed()` is executed. The old structure of the code was
1. try to get a key
2. if no key, create one

But we need the locking thing and thus a double check on #1. I keep finding the structure used hard to follow (and I probably wrote it).

Maybe this is better
```
    val lockDetails = LockDetails(s"${pet.id.project.value}-getKey", pet.serviceAccount.subjectId.value, 5 minutes)
    for {
      maybeActiveKey <- findActiveKey(pet, withinLock = false)
      activeKey <- maybeActiveKey match {
        case Some(existingActiveKey) => IO.pure(existingActiveKey)
        case None =>
          distributedLock
            .withLock(lockDetails)
            .use(_ => findActiveKey(pet, withinLock = true))
            .map(_.getOrElse(throw new WorkbenchException("Could not create or retrieve key")))
      }
    } yield activeKey
```

either that or stop passing `withinLock` to `findActiveKey` and bump that functionality out somewhere

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
dvoet: I don't think it is really worth it. I do think we should remove the polling in workbench libs though.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -178,25 +268,39 @@ class GoogleKeyCache(
...
+    */
+  private def cleanupKeys(pet: PetServiceAccount, cachedKeyObjects: List[CachedKey], serviceAccountKeys: List[ServiceAccountKey]): IO[Unit] = {
+    // if 10 or more keys, delete until we are below 10.
+    val tooManyFunction: IO[_] = if (cachedKeyObjects.size >= 10) {
+      val numToDelete = cachedKeyObjects.size - 9
+      val toDelete = cachedKeyObjects.sortBy(_.timeCreated).takeRight(numToDelete)
+      logger.info(
+        s"cleanupKeys: ${pet.id.project.value}-${pet.serviceAccount.subjectId.value} is reaping ${toDelete.size} keys due to quota"
+      )
+      toDelete.parTraverse(cachedKey => removeKey(pet, cachedKey.keyId))
```
dvoet: `parTraverse` is a bad idea. Google does not alway like concurrent modifications which leads to retry thrashing. Just use `traverse`

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -178,25 +268,39 @@ class GoogleKeyCache(
...
+    val cachedKeyIds: Set[ServiceAccountKeyId] = cachedKeyObjects.map(_.keyId).toSet
+    val unknownKeyIds: Set[ServiceAccountKeyId] = serviceAccountKeys.map(_.id).toSet -- cachedKeyIds
 
-    !keyRetired && keyExistsForSA
-  }
+    val orphansFunction: IO[_] = if (unknownKeyIds.nonEmpty) {
+      logger.info(
+        s"cleanupKeys: ${pet.id.project.value}-${pet.serviceAccount.subjectId.value} is reaping ${unknownKeyIds.size} orphaned keys"
+      )
+      unknownKeyIds.toList.parTraverse(keyId => IO.fromFuture(IO(googleIamDAO.removeServiceAccountKey(pet.id.project, pet.serviceAccount.email, keyId))))
```
dvoet: parTraverse again

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
```diff
@@ -178,25 +268,39 @@ class GoogleKeyCache(
...
+    } else {
+      IO.unit
+    }
 
-  private def cleanupUnknownKeys(pet: PetServiceAccount, cachedKeyObjects: List[GcsObjectName], serviceAccountKeys: List[ServiceAccountKey]): Future[Unit] = {
-    val cachedKeyIds = cachedKeyObjects.map(_.value).collect { case keyPathPattern(_, _, keyId) => ServiceAccountKeyId(keyId) }
-    val unknownKeyIds: Set[ServiceAccountKeyId] = serviceAccountKeys.map(_.id).toSet -- cachedKeyIds.toSet
+    for {
+      _ <- tooManyFunction
+      _ <- orphansFunction
```
dvoet, line 308: I would do orphans first since that may bring it down below 10

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
dvoet: This is only removing keys from google, not the cache. They should be removed from the cache too. Note that the cache has notifications setup on deletion to trigger sam to remove the key. So you could remove the key from the cache an pray that everything is fast enough so that there is space by the time you create a new key. Or you could remove from both.

dvoet (commented):
some comments, I did not look through the tests yet

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
davidangb: hah I found this logic confusing too and rewrote it kinda like your suggestion, but then reverted to not impose too much change. I'll revisit.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
davidangb: This is calling `GoogleKeyCache.removeKey()`, which explicitly removes from cache and from google

dvoet: oh, good, then never mind

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
davidangb: I found this section tricky. First, it counts the number of keys _in cache_ and reduces that below 10. Then, it looks for any keys _in google_ which aren't in the cache and deletes those.

I really wanted to use the count from cache for the first step, since the count from Google is eventually-consistent. I realize the possibility we'd delete more than necessary in corner cases. given this context, do you think this is worth changing?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
davidangb: rewritten, with some extra complications to propagate error messages

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala
dvoet: sound good, no change

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCacheSpec.scala
```diff
@@ -0,0 +1,219 @@
...
+  it should "return None when no keys exist in cache" in {
+    // this case triggers creation of a new key, so it returns None when not within a lock
+    val input: List[CachedKey] = List()
+    val actual = newKeyCache().searchCachedKeys(pet, input, withinLock = false).unsafeRunSync()
+    actual shouldBe None
+  }
+
+  behavior of "GoogleKeyCache.searchCachedKeys(), when within a lock"
+
+  it should "return None when only retired keys exist" in {
```
dvoet: is this test name accurate?

dvoet (approved):
I don't care enough to argue, but I think the `warn` level logs should be `info`

src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCacheSpec.scala
davidangb: no, it sure isn't! fixed.

samanehsan (approved):

PR #1441, opened 2024-05-22
ID-1270 create_with_parent Action

tlangs:
Ticket: https://broadworkbench.atlassian.net/browse/ID-1270

  \<Don't forget to include the ticket number in the PR title!\>

What:

Introduce a `create_with_parent` resource action, which will allow resource creation with a parent resource without letting users re-parent the resource later.

Why:

Needed for future functionality.

How:

Just add the resource action.

---

**PR checklist**

- [ ] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've filled out the [Security Risk Assessment](https://sdarq.dsp-appsec.broadinstitute.org/jira-ticket-risk-assesment) (requires Broad Internal network access) and attached the result to the JIRA ticket

PR #1640, opened 2025-02-11
DT-1202: Allow creating datasnapshot as a child resource of dataset

pshapiro4broad:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/DT-1202

What:

To support inheriting dataset custodian role as steward on a snapshot, we need to allow

- creating the `datasnapshot` with a parent resource (`create_with_parent`)
- adding a child resource to a `dataset` (`add_child`)

Also, we need to specify that `dataset` `custodian` or `steward` are `steward` on `datasnapshot`, if the parent resource is set.

Why:

This is support the goal of avoiding adding dataset custodians to every snapshot, to prevent hitting GCP quota limits.

How:

With this change
- users with `snapshot_creator` or greater role on `dataset` can `add_child` to `dataset`
- users with `custodian` or greater role on `dataset` can `list_children`
- users with `steward` role can `create_with_parent` to `datasnapshot`, and `get_parent`
- `dataset` `custodian` or greater role are `steward` on child `datasnapshot`s.

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1456,14 +1456,14 @@ resourceTypes = {
     ownerRoleName = "steward"
     roles = {
       steward = {
-        roleActions = ["share_policy::steward", "share_policy::custodian", "update_passport_identifier", "view_journal"]
+        roleActions = ["share_policy::steward", "share_policy::custodian", "update_passport_identifier", "view_journal", "create_with_parent"]
         includedRoles = ["custodian"]
         descendantRoles = {
           snapshot-builder-request = ["approver"]
         }
       }
       custodian = {
-        roleActions = ["delete", "edit_datasnapshot", "update_snapshot", "share_policy::reader", "share_policy::aggregate_data_reader", "share_policy::discoverer", "read_policies", "set_public", "update_auth_domain", "lock_resource", "unlock_resource", "get_parent", "create_with_parent"]
+        roleActions = ["delete", "edit_datasnapshot", "update_snapshot", "share_policy::reader", "share_policy::aggregate_data_reader", "share_policy::discoverer", "read_policies", "set_public", "update_auth_domain", "lock_resource", "unlock_resource", "get_parent"]
         includedRoles = ["reader"]
       }
       reader = {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1338,6 +1338,9 @@ resourceTypes = {
       add_child = {
         description = "add a child resource"
       }
+      list_children = {
+        description = "list child resources"
+      }
     }
     ownerRoleName = "steward"
     roles = {
@@ -1356,7 +1359,7 @@ resourceTypes = {
         roleActions = ["read_dataset", "read_data", "ingest_data"]
       }
       snapshot_creator = {
-        roleActions = ["read_dataset", "read_data", "read_policies", "link_snapshot", "add_child"]
+        roleActions = ["read_dataset", "read_data", "read_policies", "link_snapshot", "add_child", "list_children"]
       }
       admin = {
         roleActions = ["share_policy::steward", "read_policies", "alter_policies", "unlock_resource"]
@@ -1444,20 +1447,23 @@ resourceTypes = {
         description = "add a child resource"
       }
       create_with_parent = {
-        description = "Enables creating the request object with a parent"
+        description = "Enables creating a snapshot with a parent dataset"
+      }
+      get_parent = {
+        description = "get parent of snapshot"
       }
     }
     ownerRoleName = "steward"
     roles = {
       steward = {
-        roleActions = ["share_policy::steward", "share_policy::custodian", "update_passport_identifier", "view_journal", "create_with_parent"]
+        roleActions = ["share_policy::steward", "share_policy::custodian", "update_passport_identifier", "view_journal"]
         includedRoles = ["custodian"]
         descendantRoles = {
           snapshot-builder-request = ["approver"]
         }
       }
       custodian = {
-        roleActions = ["delete", "edit_datasnapshot", "update_snapshot", "share_policy::reader", "share_policy::aggregate_data_reader", "share_policy::discoverer", "read_policies", "set_public", "update_auth_domain", "lock_resource", "unlock_resource"]
+        roleActions = ["delete", "edit_datasnapshot", "update_snapshot", "share_policy::reader", "share_policy::aggregate_data_reader", "share_policy::discoverer", "read_policies", "set_public", "update_auth_domain", "lock_resource", "unlock_resource", "get_parent"]
         includedRoles = ["reader"]
       }
       reader = {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1335,6 +1335,9 @@ resourceTypes = {
       "unlock_resource" = {
         description = "Can unlock a resource"
       }
+      add_child = {
+        description = "add a child resource"
+      }
     }
     ownerRoleName = "steward"
     roles = {
@@ -1348,7 +1351,7 @@ resourceTypes = {
         roleActions = ["read_dataset", "read_data", "ingest_data"]
       }
       snapshot_creator = {
-        roleActions = ["read_dataset", "read_data", "read_policies", "link_snapshot"]
+        roleActions = ["read_dataset", "read_data", "read_policies", "link_snapshot", "add_child"]
       }
       admin = {
         roleActions = ["share_policy::steward", "read_policies", "alter_policies", "unlock_resource"]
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -1435,11 +1438,14 @@ resourceTypes = {
       add_child = {
         description = "add a child resource"
       }
+      create_with_parent = {
```
dvoet, line 1449: What is this action for? I think you probably want `set_parent`. https://github.com/broadinstitute/sam/blob/develop/README.md#resource-and-policy-management

pshapiro4broad: I got it from https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala#L34
which is used by [`requireCreateWithOptionalParent`](https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala#L70):
```
        // parents are allowed for a resource type if the owner role contains the SamResourceActions.setParent or SamResourceActions.createWithParent action
        val parentAllowed = resourceType.roles
          .find(_.roleName == resourceType.ownerRoleName)
          .exists(role => role.actions.contains(SamResourceActions.setParent) || role.actions.contains(SamResourceActions.createWithParent))
```

pshapiro4broad: It looks like when this was added in https://github.com/broadinstitute/sam/pull/1441, the README wasn't updated. I can switch to using `set_parent` if that's preferred. Relevant slack thread: https://broadinstitute.slack.com/archives/C03GMG4DUSE/p1715026844061349

src/main/resources/reference.conf
```diff
@@ -1345,12 +1348,15 @@ resourceTypes = {
       custodian = {
         roleActions = ["manage_schema", "create_datasnapshot", "ingest_data", "soft_delete", "hard_delete", "unlink_snapshot", "list_snapshots", "lock_resource", "unlock_resource", "share_policy::custodian", "share_policy::ingester", "share_policy::snapshot_creator"]
         includedRoles = ["snapshot_creator"]
+        descendantRoles = {
```
pshapiro4broad, line 1354: It wasn't obvious to me how this works but I'm hoping that `descendantRoles` is shared along with `roleActions` when `includedRoles` is used.

src/main/resources/reference.conf
dvoet: Interesting. I guess you can keep your `create_with_parent`. If I had been paying attention, I might have suggested that they use `set_parent` but not grant `remove_child` on the parent.

src/main/resources/reference.conf
```diff
@@ -1437,11 +1443,14 @@ resourceTypes = {
...
       }
+      create_with_parent = {
+        description = "Enables creating the request object with a parent"
+      }
     }
     ownerRoleName = "steward"
     roles = {
       steward = {
-        roleActions = ["share_policy::steward", "share_policy::custodian", "update_passport_identifier", "view_journal"]
+        roleActions = ["share_policy::steward", "share_policy::custodian", "update_passport_identifier", "view_journal", "create_with_parent"]
```
dvoet, line 1459: You don't actually need to grant `create_with_parent` to anyone. The action's existence on the resource type is used only as a flag to allow resources to be created with a parent

src/main/resources/reference.conf
dvoet: `included_roles` is like role inheritance. The `custodian` includes everything from the `snapshot_creator` role. `descendantRoles` are those that are applied to children, grandchildren, etc. The `steward` role is granted to to all descendant `datasnapshot`s.

dvoet (approved):
I don't know if you intend to add parents after creating snapshots or not. If you do intend to add parents after the fact, then you will need `set_parent`

src/main/resources/reference.conf
pshapiro4broad: > I might have suggested that they use set_parent but not grant remove_child on the parent.

Well, TDR is the only user of `create_with_parent` so we could stop doing it if you want us to. I don't have a strong opinion one way or the other.

src/main/resources/reference.conf
pshapiro4broad: When I tested this change in a bee, I found that it didn't work until I added `create_with_parent` as an action on the `steward` (owner) role.

I also tried adding `create_with_parent` as an action on `custodian`, which is included by `steward`, but that didn't work.

Looking at the Scala code, in `SecurityDirectives.requireCreateWithOptionalParent()`:
```
        // parents are allowed for a resource type if the owner role contains the SamResourceActions.setParent or SamResourceActions.createWithParent action
        val parentAllowed = resourceType.roles
          .find(_.roleName == resourceType.ownerRoleName)
          .exists(role => role.actions.contains(SamResourceActions.setParent) || role.actions.contains(SamResourceActions.createWithParent))
```

It appears that this is what the code is requiring -- the action must directly be on the owner role.

pshapiro4broad: I'm wondering if the failure of this to work with included roles is a bug. Looking at the code path for the `set_parent` operation (`ResourceRoutes.setResourceParent()`), the code uses `SecurityDirectives.requireAction()` which looks like it does support included roles.

src/main/resources/reference.conf
```diff
@@ -1345,12 +1351,15 @@ resourceTypes = {
...
+        descendantRoles = {
+          datasnapshot = ["steward"]
+        }
       }
       ingester = {
         roleActions = ["read_dataset", "read_data", "ingest_data"]
       }
       snapshot_creator = {
-        roleActions = ["read_dataset", "read_data", "read_policies", "link_snapshot"]
+        roleActions = ["read_dataset", "read_data", "read_policies", "link_snapshot", "add_child", "list_children"]
```
snf2ye: I think this should be at the dataset custodian role instead of snapshot_creator. I think we only want to let dataset custodian and stewards update this inheritSteward role and set children.

snf2ye: Is there an add_child endpoint in Sam? I only see setParent. Do we need to instead add the "set_parent" role to datasnapshots?

snf2ye: Thanks for talking through this with me:
- "add_child" needs to be on the snapshot_creator role so that on snapshot create, the new snapshot can be added as a child to the dataset. We expect for snapshot_creators to perform this option. 
- "add_child" is checked on "create_with_parent" or "set_parent", so we need both. 

However, I'm not sure about leaving "list_children" on the snapshot_creator. This will effectively grant the snapshot_creator with "list_snapshots" permissions, which was previously reserved for dataset stewards and custodians. Can we move "list_children" up to custodian level instead?

src/main/resources/reference.conf
```diff
@@ -1437,18 +1446,24 @@ resourceTypes = {
...
-        roleActions = ["share_policy::steward", "share_policy::custodian", "update_passport_identifier", "view_journal"]
+        roleActions = ["share_policy::steward", "share_policy::custodian", "update_passport_identifier", "view_journal", "create_with_parent"]
         includedRoles = ["custodian"]
         descendantRoles = {
           snapshot-builder-request = ["approver"]
         }
       }
       custodian = {
-        roleActions = ["delete", "edit_datasnapshot", "update_snapshot", "share_policy::reader", "share_policy::aggregate_data_reader", "share_policy::discoverer", "read_policies", "set_public", "update_auth_domain", "lock_resource", "unlock_resource"]
+        roleActions = ["delete", "edit_datasnapshot", "update_snapshot", "share_policy::reader", "share_policy::aggregate_data_reader", "share_policy::discoverer", "read_policies", "set_public", "update_auth_domain", "lock_resource", "unlock_resource", "get_parent"]
```
snf2ye, line 1466: Why do we need "get_parent"?

src/main/resources/reference.conf
pshapiro4broad: > However, I'm not sure about leaving "list_children" on the snapshot_creator. This will effectively grant the snapshot_creator with "list_snapshots" permissions, which was previously reserved for dataset stewards and custodians. Can we move "list_children" up to custodian level instead?

Yes, I think you're right, a snapshot creator shouldn't need this permission and it could let them have more permissions than they had before.  I'll move it to `custodian`.

src/main/resources/reference.conf
pshapiro4broad: I think we'll need it in the UI, to find out of a snapshot has a parent dataset which is providing steward users to the snapshot. The UI can use this to fetch and show the dataset custodian users as "inherited" stewards in the `Roles & Memberships` tab.

snf2ye (approved):

PR #1655, opened 2025-03-24
CORE-402 recurse includedRoles to list actions

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-402

What:

Add a new function to `ResourceType` that will get the actions for a role, recursing through `includedRoles` and use it in places that expect the current list of actions to be complete.

Why:

How:

---

**PR checklist**

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
@@ -77,9 +77,8 @@ trait SecurityDirectives {
       case None => Directives.pass // no parent specified, proceed
       case Some(parent) =>
         // parents are allowed for a resource type if the owner role contains the SamResourceActions.setParent or SamResourceActions.createWithParent action
-        val parentAllowed = resourceType.roles
-          .find(_.roleName == resourceType.ownerRoleName)
-          .exists(role => role.actions.contains(SamResourceActions.setParent) || role.actions.contains(SamResourceActions.createWithParent))
+        val parentAllowedActions = Set(SamResourceActions.setParent, SamResourceActions.createWithParent)
+        val parentAllowed = parentAllowedActions.diff(resourceType.getRoleActions(resourceType.ownerRoleName)).nonEmpty
         if (!parentAllowed) {
           Directives.failWith(
             new WorkbenchExceptionWithErrorReport(
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
@@ -78,7 +78,7 @@ trait SecurityDirectives {
       case Some(parent) =>
         // parents are allowed for a resource type if the owner role contains the SamResourceActions.setParent or SamResourceActions.createWithParent action
         val parentAllowedActions = Set(SamResourceActions.setParent, SamResourceActions.createWithParent)
-        val parentAllowed = parentAllowedActions.diff(resourceType.getRoleActions(resourceType.ownerRoleName)).nonEmpty
+        val parentAllowed = parentAllowedActions.intersect(resourceType.getRoleActions(resourceType.ownerRoleName)).nonEmpty
         if (!parentAllowed) {
           Directives.failWith(
             new WorkbenchExceptionWithErrorReport(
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
```diff
@@ -141,6 +141,29 @@ object UserStatusDetails {
...
+    } else {
+      roles
+        .find(_.roleName == roleName)
+        .map(role =>
+          role.includedRoles.foldLeft(role.actions)((accumulatedActions, includedRole) =>
+            accumulatedActions ++ getRoleActions(includedRole, visitedRoles + roleName)
+          )
+        )
+        .getOrElse(Set.empty)
+    }
```
davidangb, line 166: I _think_ there's a corner case here where the recursion exits early:

Role "one" includes roles "two" and "three"

Role "two" includes roles "A", "B", "C"

Role "three" includes roles "B", "C", "D", "E"

As soon as the recursion hits "B" for the second time, it will exit - potentially never considering "D" or "E" (yeah, it's a Set, it won't process alphabetically, but hopefully this gets the point across)

I also wonder if this corner case is even possible in the current `reference.conf` and how important it is to guard against

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
```diff
@@ -77,9 +77,8 @@ trait SecurityDirectives {
       case None => Directives.pass // no parent specified, proceed
       case Some(parent) =>
         // parents are allowed for a resource type if the owner role contains the SamResourceActions.setParent or SamResourceActions.createWithParent action
-        val parentAllowed = resourceType.roles
-          .find(_.roleName == resourceType.ownerRoleName)
-          .exists(role => role.actions.contains(SamResourceActions.setParent) || role.actions.contains(SamResourceActions.createWithParent))
+        val parentAllowedActions = Set(SamResourceActions.setParent, SamResourceActions.createWithParent)
+        val parentAllowed = parentAllowedActions.diff(resourceType.getRoleActions(resourceType.ownerRoleName)).nonEmpty
```
davidangb: should this be `intersect` instead of `diff`?

src/test/scala/org/broadinstitute/dsde/workbench/sam/model/ResourceTypeSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+      resourceType.getRoleActions(role.roleName) should be(role.actions)
+    }
+  }
+
+  it should "return the correct actions for a resource type with includedRoles" in {
+    val resourceType = Generator.genResourceType.sample.get
+    val testRole = ResourceRoleName("includedRole")
+    val withIncludedRoles = resourceType.copy(
+      roles = resourceType.roles + ResourceRole(testRole, Set.empty, resourceType.roles.map(_.roleName))
+    )
```
davidangb, line 22: if I'm reading this correctly, the includedRoles' actions are an exact duplicate of the non-included roles' actions, so this test doesn't really exercise the recursion. Could you use a different role/actions for the included ones?

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
dvoet: I am not following why there would be an early exit. The `foldLeft` will evaluate all `includedRoles`. Roles "two" and "three" evaluate in separate call trees so in your example "B" is actually visited twice. `visitedRoles` does not prevent multiple visits, just cycles.

Maybe you are thinking that `visitedRoles` has a broader scope than it does. Each call to `getRoleActions` gets its own copy of `visitedRoles`. `visitedRoles + roleName` does not modify `visitedRoles` but creates a new Set with `roleName` added.

src/test/scala/org/broadinstitute/dsde/workbench/sam/model/ResourceTypeSpec.scala
dvoet: This should exercise the recursion. `testRole` is not given any direct actions, it only includes other roles. So any actions it has come from the recursion.

src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SecurityDirectives.scala
dvoet: yup, that's why the tests are failing

src/main/scala/org/broadinstitute/dsde/workbench/sam/model/SamModel.scala
davidangb: yup, I see this now, I was misreading the recursion, all good!

davidangb (approved):

calypsomatic (approved):

PR #1656, opened 2025-03-27
[CORE-346] add read action

kevinmarete:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-346

  \<Don't forget to include the ticket number in the PR title!\>

What:

  \<For your reviewers' sake, please describe in a sentence or two what this PR is accomplishing (usually from the users' perspective, but not necessarily).\>

Why:

  \<For your reviewers' sake, please describe in ~1 paragraph what the value of this PR is to our users or to ourselves.\>

How:

  \<For your reviewers' sake, please describe in ~1 paragraph how this PR accomplishes its goal.\>

  \<If the PR is big, please indicate where a reviewer should start reading it (i.e. which file or function).\>

---

**PR checklist**

- [ ] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've filled out the [Security Risk Assessment](https://sdarq.dsp-appsec.broadinstitute.org/jira-ticket-risk-assesment) (requires Broad Internal network access) and attached the result to the JIRA ticket

PR #1672, opened 2025-04-08
DT-1421: add a new resource type for TDR's google projects

pshapiro4broad:
Ticket: http://broadworkbench.atlassian.net/browse/DT-1421

*What:*

- add a new resource type for TDR's google projects
- add actions and descendantRoles to datasnapshot and dataset.

*Why:*

TDR needs a way to communicate permissions to Rawls, but we don't want to have a circular call chain where Rawls needs to call TDR.

*How:*

Sam is a common service that both TDR and Rawls call into, so we're using Sam to store this information.

Verified using https://github.com/DataBiosphere/jade-data-repo/blob/ps/dt-1421-datarepo-google-project-support-sam-test/src/test/java/bio/terra/integration/SamResourcesTest.java by running the test manually in a BEE

---

**PR checklist**

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1189,7 +1189,7 @@ resourceTypes = {
         description = "set the parent resource of this resource"
       }
       link = {
-        description = "link this resource to a dataset or datasnapshot"
+        description = "link this resource to a billing-project"
       }
       read = {
         description = "read this resource"
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -1174,6 +1174,35 @@ resourceTypes = {
     }
     reuseIds = true
   }
+  datarepo-google-project = {
+    actionPatterns = {
+      read_policies = {
+        description = "list all policies and policy details for this resource"
+      }
+      delete = {
+        description = "delete this resource"
+      }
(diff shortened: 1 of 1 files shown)


src/main/resources/reference.conf
```diff
@@ -1174,6 +1174,35 @@ resourceTypes = {
...
+      }
+      get_parent = {
+        description = "get the parent resource of this resource"
+      }
+      set_parent = {
+        description = "set the parent resource of this resource"
+      }
+      link = {
+        description = "link this resource to a dataset or datasnapshot"
+      }
```
davidangb, line 1199: In #1652, we added a `link` action to `google-project` to control the ability to set/remove billing from the underlying Google project. Is this `link` action suitable for that as well? If not, we'll need a new action for billing.

davidangb (approved):
When this merges and Rawls is updated to use this new resource type, potentially #1656   and #1652 can be reverted.

kevinmarete (approved):

src/main/resources/reference.conf
pshapiro4broad: You're right, the description here is wrong. This _is_ for linking to a billing project, not to a dataset or snapshot.

PR #1676, opened 2025-04-14
[CORE-445] Make hasPermissionOneOf case-insensitive

samanehsan:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-445

What:
In the resources `setPublic` API endpoint, if a user specifies a policy name that is not all lowercase, Sam throws a permissions error because checking the policy is case-sensitive. This is also an issue for other endpoints where the policy is an API param.

How:
Override equals and hashcode of `ResourceAction` to ignore case when comparing two actions.

**PR checklist**

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
@@ -681,32 +681,6 @@ class PolicyEvaluatorServiceSpec extends RetryableAnyFlatSpec with Matchers with
 
     res.unsafeRunSync()
   }
-
-  "hasPermissionOneOf" should "not be case-sensitive" in {
-    val user = genWorkbenchUserBoth.sample.get
-    val resource = genResource.sample.get.copy(resourceTypeName = defaultResourceType.name)
-
-    val samplePolicy = genPolicy.sample.get
-    val policyWithUser = AccessPolicy.members.set(samplePolicy.members + user.id)(samplePolicy)
-    val policyWithAction =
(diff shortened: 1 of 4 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
```diff
@@ -681,6 +681,32 @@ class PolicyEvaluatorServiceSpec extends RetryableAnyFlatSpec with Matchers with
...
+      _ <- policyDAO.createResourceType(managedGroupResourceType, samRequestContext)
+      _ <- dirDAO.createUser(user, samRequestContext)
+      _ <- savePolicyMembers(policy)
+      _ <- policyDAO.createResourceType(defaultResourceType, samRequestContext)
+      _ <- policyDAO.createResource(resource, samRequestContext)
+      _ <- policyDAO.createPolicy(policy, samRequestContext)
+
+      r <- constrainableService.policyEvaluatorService.hasPermissionOneOf(
+        resource.fullyQualifiedId,
+        Set(SamResourceActions.alterPolicies, SamResourceActions.sharePolicy(AccessPolicyName("READER"))),
```
marctalbott: I know this is closer to what we actually do, but we should remove `alterPolicies` from the check here to be sure that the case-insensitive match is working and we aren't passing because the user has `alterPolicies`

marctalbott (approved):

src/test/scala/org/broadinstitute/dsde/workbench/sam/service/PolicyEvaluatorServiceSpec.scala
samanehsan: That's a good point!

dvoet:
I wonder if it would work to override equals and hashcode for ResourceAction

calypsomatic (approved):

dvoet (approved):

PR #1693, opened 2025-06-16
CORE-562 optimize sql for list resources

dvoet:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-562

I asked Claude how to make the sql better and after a couple iterations we came up with eliminating the union in favor of joining both role actions and direct actions at the same time. I hope that the resulting query does about half the work due to dropping the union. The danger is a cartesian product when there are lots of direct actions but that is rare.

Here are some explain analyze plans (query at the bottom of each too). First for dvoet at broad, should be fast:
[old query](https://github.com/user-attachments/files/20763291/dvoet.explain.before.txt)
[new query](https://github.com/user-attachments/files/20763295/dvoet.explain.after.txt)
Looks like a 50% reduction!

Next for the AoU SA, this is the one that causes problems:
[old query](https://github.com/user-attachments/files/20763299/auo.explain.before.txt)
[new query](https://github.com/user-attachments/files/20763301/aou.explain.after.txt)
Note the Seq Scans on `sam_effective_resource_policy`. Overall time looks about the same but cost seems to be halfish.

At the end of the day, I am looking to decrease CPU load and not overall response time so I would like to run a trial of this in prod.


davidangb (approved):
:+1: very interesting. I would have thought that adding more joins would make the query more expensive ... but I see in the query costs that it is going down. I'm curious to see how it performs in production!

samanehsan (approved):

PR #1716, opened 2025-09-04
CORE-681: forget-project API

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-681

Create a new API at `DELETE /api/google/v1/project/{project}` to "forget" a Google project.

"Forgetting" means:
* remove all known pet service accounts in this project from their proxy groups
* delete knowledge of these pet service accounts from the Sam db
* recursively delete the google-project resource and all its children from the Sam db

This does NOT include deleting anything from the cloud (the project still exists and pet service accounts still exist). It is the responsibility of the caller (Rawls) to ensure these are all deleted - typically, by deleting the project itself which will delete everything in it.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -319,6 +319,15 @@ class GoogleExtensions(
       }
     } yield deletedSomething
 
+  def forgetUserPetServiceAccount(userId: WorkbenchUserId, project: GoogleProject, samRequestContext: SamRequestContext): IO[Boolean] =
+    for {
+      maybePet <- directoryDAO.loadPetServiceAccount(PetServiceAccountId(userId, project), samRequestContext)
+      forgotSomething <- maybePet match {
+        case Some(pet) => forgetPetServiceAccount(pet, samRequestContext).map(_ => true)
+        case None => IO.pure(false) // didn't find the pet, nothing to forget
+      }
+    } yield forgotSomething
+
   def createUserPetServiceAccount(user: SamUser, project: GoogleProject, samRequestContext: SamRequestContext): IO[PetServiceAccount] = {
     val (petSaName, petSaDisplayName) = toPetSAFromUser(user)
     // The normal situation is that the pet either exists in both the database and google or neither.
@@ -561,7 +570,6 @@ class GoogleExtensions(
 
   /** Delete Sam's knowledge of this pet service account. Does not make any changes in the cloud.
     */
-  // TODO CORE-681: unit tests
   private def forgetPetServiceAccount(petServiceAccount: PetServiceAccount, samRequestContext: SamRequestContext): IO[Unit] =
     for {
       // disable the pet service account
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
@@ -795,7 +795,7 @@ class GoogleExtensions(
     } yield Map("pets" -> allProjectPets.size, "resources" -> numResourcesDeleted)
   }
 
-  override val allSubSystems: Set[Subsystems.Subsystem] = Set() // Set(Subsystems.GoogleGroups, Subsystems.GooglePubSub, Subsystems.GoogleIam)
+  override val allSubSystems: Set[Subsystems.Subsystem] = Set(Subsystems.GoogleGroups, Subsystems.GooglePubSub, Subsystems.GoogleIam)
 
 }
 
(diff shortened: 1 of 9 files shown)


src/test/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionSpec.scala
```diff
@@ -2155,6 +2206,175 @@ class GoogleExtensionSpec(_system: ActorSystem)
     messageLog should contain theSameElementsAs formattedMessages
   }
 
+  "forgetProject" should "recursively delete child resources" in {
```
davidangb, line 2209: this is a pretty massive unit test, but I really wanted coverage here

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -751,7 +770,33 @@ class GoogleExtensions(
       .lastOrError
   }
 
-  override val allSubSystems: Set[Subsystems.Subsystem] = Set(Subsystems.GoogleGroups, Subsystems.GooglePubSub, Subsystems.GoogleIam)
+  // TODO CORE-681: unit tests
```
calypsomatic: Are there more unit tests to be written in a later PR?

calypsomatic (approved):
I'm guessing the answer is "not really", but is there anything we can do to either enforce or at least strongly signal to callers of this API that they need to delete resources in the cloud?

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
davidangb: oh whoops, I had that in there as a note to self, I meant to remove it when I wrote the tests. Will fix.

davidangb:
> I'm guessing the answer is "not really", but is there anything we can do to either enforce or at least strongly signal to callers of this API that they need to delete resources in the cloud?

@calypsomatic options I thought of, but don't really like:

1. Make this API only callable by the Rawls service account. This delegates enforcement to Rawls which is probably better but means all API calls will be as a superuser.
2. Sam could attempt to retrieve the project from the cloud, and fail the API if the project exists. This could have false positives ... what if Sam just doesn't have the right permission and thinks the project is gone even though it still exists?

Ideas are welcome!!!

calypsomatic:
> > I'm guessing the answer is "not really", but is there anything we can do to either enforce or at least strongly signal to callers of this API that they need to delete resources in the cloud?
> 
> @calypsomatic options I thought of, but don't really like:
> 
> 1. Make this API only callable by the Rawls service account. This delegates enforcement to Rawls which is probably better but means all API calls will be as a superuser.
> 2. Sam could attempt to retrieve the project from the cloud, and fail the API if the project exists. This could have false positives ... what if Sam just doesn't have the right permission and thinks the project is gone even though it still exists?
> 
> Ideas are welcome!!!

Yeah I don't really like option 2.  Right now it looks like it's just requiring the caller to have delete permission, or was there some other check I was missing?  Who do we expect to be making this call normally?  Maybe it's fine as is and we should just have a way to discover and clean up orphaned google resources.

davidangb:
> Who do we expect to be making this call normally?

@calypsomatic to complete this ticket, this API should get called by Rawls' delete-billing-project API. So it'll be triggered by a user who wants to delete a v1 billing project.  Eventually, we could also use it in Rawls' delete-workspace API to make that API a bit more robust and move some logic out of Rawls.

src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensions.scala
```diff
@@ -751,7 +770,32 @@ class GoogleExtensions(
...
+      // forget all pets in this project
+      _ <- allProjectPets.traverse { pet =>
+        forgetPetServiceAccount(pet, samRequestContext)
+      }
+      // recursively delete this google-project resource and its children
+      numResourcesDeleted <- recursiveDeleteResource(projectResourceId, 0)
+    } yield Map("pets" -> allProjectPets.size, "resources" -> numResourcesDeleted)
+  }
+
+  override val allSubSystems: Set[Subsystems.Subsystem] = Set() // Set(Subsystems.GoogleGroups, Subsystems.GooglePubSub, Subsystems.GoogleIam)
```
samanehsan: Did you mean to leave this comment in?

davidangb: aggg no! Will fix, thank you for catching this

davidangb: reverted in [c749bf0](https://github.com/broadinstitute/sam/pull/1716/commits/c749bf051125cba87d8c0376835701b25e3b805d)

samanehsan (approved):

PR #1749, opened 2026-06-29
[CTM-559] Admin endpoint to set allowExternalMembers on existing Google groups

marctalbott:
New groups already get `allowExternalMembers=true` at creation; this adds a one-off, super-admin-triggered migration to flip the same setting on groups created before that.

`PUT /google/v1/groups/allowExternalMembers/migrate/{tier[,tier...]}` runs the given tiers (`proxy` or a resource type name) in order, in the background. Each tier runs idempotently through the coordinated-backoff `GoogleDirectoryDAO`, metered and keyset-paginated a page at a time so a tier never materializes its whole population in memory. The proxy tier also re-adds each enabled user's email to their proxy group. `GET .../migrate/status` returns a per-tier progress report.

- **Resource-type tiers cover both** the policy-backed groups and the aggregate group whose name matches the resource id (e.g. a managed group's own group), so managed groups are fully migrated.
- **Resume is automatic and idempotent.** Progress (state, counts, cursor) is persisted per tier. Re-invoking a tier resumes a crashed/failed run from its last recorded cursor with cumulative counts; a `completed` tier is skipped when other tiers are requested alongside it, and re-runs are safe.
- **Unknown tier selectors are rejected with a 400** (against the configured resource types + `proxy`) rather than silently "completing" zero groups.

Proposed run order: `proxy` → `managed-group` → `datasnapshot`/`dataset`/`spend-profile` → `workspace`/`billing-project`.

Ticket: https://broadworkbench.atlassian.net/browse/CTM-559

### DIFF
--- a/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/MockTestSupport.scala
+++ b/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/MockTestSupport.scala
@@ -25,7 +25,13 @@ import org.broadinstitute.dsde.workbench.sam.config._
 import org.broadinstitute.dsde.workbench.sam.dataAccess._
 import org.broadinstitute.dsde.workbench.sam.db.TestDbReference
 import org.broadinstitute.dsde.workbench.sam.db.tables._
-import org.broadinstitute.dsde.workbench.sam.google.{GoogleExtensionRoutes, GoogleExtensions, GoogleGroupSynchronizer, GoogleKeyCache}
+import org.broadinstitute.dsde.workbench.sam.google.{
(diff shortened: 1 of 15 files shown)


jgainerdewar (approved):
Leaving a rubber-stamp thumb, should be carefully reviewed by others as well.

snf2ye (approved):
Thanks for walking through it with us. Looks like a solid approach. I also ran through a claude PR review and I didn't see any issues of concern. 
I think the next step of testing this out on a small resource on dev sounds good. 

Thank you!

aednichols (approved):
I agree with the high-level concept outlined during the walk through. Did not review line-by-line as I am not a Sam expert.

PR #1753, opened 2026-07-09
[CTM-559] Make external members migration throughput configurable

marctalbott:
Replaces the fixed 100ms per-item throttle and sequential processing with a per-run `queriesPerMinute` rate limiter and bounded-concurrency page processing, so throughput can be tuned for the upcoming Google quota increase. 

Concurrency is controlled via a single `queriesPerMinute` parameter which determines the pace at which we process each group. For example, if `queriesPerMinute` were 10, we would process 5 groups per minute (1 get and 1 patch per group) and we would start processing a group every 12 seconds. As `queriesPerMinute` increases, the time between when we start processing decreases. Overall concurrency is limited in practice by Google's latency. We will only be processing as many groups at once as `queriesPerMinute` allows us to start in the time it takes Google to process a group. There is a maximum bound of 128 to keep Google slowness from causing out of control concurrency.

Ticket: https://broadworkbench.atlassian.net/browse/CTM-559

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
@@ -318,25 +318,30 @@ trait GoogleExtensionRoutes extends ExtensionRoutes with SamUserDirectives with 
               } ~
                 // PUT .../migrate/{tier[,tier...]} -> run the given priority tiers ("proxy" or resource type names) in order, in the background.
                 // Returns immediately; safe to re-run to resume (completed tiers are skipped, a crashed tier resumes from its last cursor).
+                // Optional queriesPerMinute query param sets the Groups Settings API rate limit (the single throughput control) for this run.
                 path(Segment) { tierSelector =>
-                  putWithTelemetry(samRequestContext, "tiers" -> new ValueObject { val value: String = tierSelector }) {
-                    complete {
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleExtensionRoutes.scala
```diff
@@ -318,25 +318,30 @@ trait GoogleExtensionRoutes extends ExtensionRoutes with SamUserDirectives with
...
-                      if (unknown.nonEmpty)
-                        IO.pure[(StatusCode, String)](
-                          StatusCodes.BadRequest -> s"Unknown migration tier(s): ${unknown.mkString(", ")}. Valid tiers: proxy or a resource type name."
-                        )
-                      else
-                        groupExternalMembersMigrator
-                          .migrate(tiers, samRequestContext)
-                          .start
-                          .as(StatusCodes.Accepted -> s"Started allowExternalMembers migration for tiers ${tiers.map(_.value).mkString(", ")}")
+                  parameters("queriesPerMinute".as[Int].optional) { queriesPerMinute =>
```
marctalbott, line 323: This is easier to read with `Hide whitespace` turned on

davidangb (approved):
this is some thorny code! Looks correct to me. I am excited to see how it does in practice.

snf2ye (approved):

aednichols (approved):

PR #917, opened 2022-11-29
ID-299 Upgrade Netty to fix medium vulnerability.

Ghost-in-a-Jar:
Ticket: https://broadworkbench.atlassian.net/browse/ID-299

What:

 Security fix

Why:

security

How:

Explicitly bumping netty to the most recent v4 version

---

**PR checklist**

- [ ] I've followed [the instructions](https://github.com/broadinstitute/sam/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've filled out the [Security Risk Assessment](https://broadinstitute.github.io/dsp-appsec-security-risk-assessment/) and attached the result to the JIRA ticket

PR #1755, opened 2026-07-24
CTM-549: bump netty, bouncycastle, and jackson

davidangb:
Ticket: https://broadworkbench.atlassian.net/browse/CTM-549

Remove `netty-all` as a direct dependency; bump all netty transitive dependencies to `4.2.16.Final` (CTM-549)

Bump bouncycastle transitive dependencies to `1.85`

Bump Jackson 3 (jackson-core, jackson-databind) to `3.2.1` (CTM-571)

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -10,6 +10,7 @@ object Dependencies {
   val scalikejdbcVersion = "3.4.2"
   val postgresDriverVersion = "42.7.8"
   val sentryVersion = "6.15.0"
+  val nettyV = "4.2.16.Final"
 
   val workbenchLibV = "9254061" // If updating this, make sure googleStorageLocal in test dependencies is up-to-date
   val workbenchUtilV = s"0.10-$workbenchLibV"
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -66,7 +67,6 @@ object Dependencies {
   val scalaCheck: ModuleID = "org.scalacheck" %% "scalacheck" % scalaCheckV % "test"
 
   val reactorNetty: ModuleID = "io.projectreactor.netty" % "reactor-netty" % "1.2.13"
-  val nettyAll: ModuleID = "io.netty" % "netty-all" % "4.2.13.Final"
```
davidangb, line 69: netty is pulled in transitively; we don't need to directly specify netty-all. It's possible someone did this previously as a means to override the netty version.

davidangb: Update: this was added way back in #917 as a way to specify the version of netty. We don't need the direct dependency.

jgainerdewar (approved):

LizBaldo (approved):

